PR #113, opened 2017-06-21
fix(oauth2): do not use remote google fonts for the default login page

### LINKED ISSUE #112: [oauth2] do not fetch remote icon fonts for default login page
We must not fetch google fonts for the default login page to avoid http 404 responses for servers with no internet access or servers behind a proxy

tcompiegne:
This closes #112


brasseld:
This wouldn't be required anymore when https://github.com/gravitee-io/graviteeio-access-management/issues/104 is fixed.

PR #122, opened 2017-07-20
Add am-gateway Dockerfile.

fmeriaux:


brasseld:
Included in 1.4.0 with compose

PR #244, opened 2018-06-07
Add the "client_credentials" grant type to the default admin client

pletessier:
See issue https://github.com/gravitee-io/issues/issues/1281


tcompiegne:
Hi @pletessier, don't you think we should also update existing admin client ?

Your code only apply if the admin client doesn't exist.

pletessier:
Hi @tcompiegne ,
From my point of view, setting the grant_type at init is enough. I just need that to be able to create an access_token the first time. If user changes are made after, I suppose that Gravitee shouldn't overwrite them. It would be confusing.

tcompiegne:
All right, according to our contributing guide, can you sign the [CLA](https://download.gravitee.io/cla.pdf), and I will accept your PR.

```
Contributor License Agreement

Please sign our Contributor License Agreement (CLA) and send it to contact@graviteesource.com. For any code changes to be accepted, the CLA must be signed.
```

Thanks.
Regards.

tcompiegne:
👍

Issue #460, opened 2018-11-28
Make nginx sub_filter optional to allow using other reverse proxies like Traefik


pletessier:
In the Nginx config, a sub_filter is applied to serve the ui resources behind the reverse-proxy : 
```
location /am/ui/ {
      proxy_pass http://am-webui/;
      proxy_set_header   Host $host;
      proxy_set_header   X-Real-IP $remote_addr;
      proxy_set_header   X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header   X-Forwarded-Host $server_name;
      sub_filter '<base href="/"' '<base href="/am/ui/"';
      sub_filter_once on;
}
```

This is OK when working with Nginx, but I think that management-ui should be able to serve its resources on a configurable base path. It would allow using another reverse-proxy (like Traefik) which cannot rewrite the HTML content.

PR #461, opened 2018-11-28
Add the possibility to change the ui basepath to avoid using nginx sub_filter

pletessier:
#460

### DIFF
--- a/docker/management-ui/files/entrypoint.sh
+++ b/docker/management-ui/files/entrypoint.sh
@@ -20,6 +20,8 @@ setup() {
     echo "Configure management API and management UI"
     cat /var/www/html/constants.json.template | \
     sed "s#http://localhost:8093/#${MGMT_API_URL}#g;s#http://localhost:4200/#${MGMT_UI_URL}#g" > /var/www/html/constants.json
+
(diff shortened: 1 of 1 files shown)


docker/management-ui/files/entrypoint.sh
```diff
@@ -20,6 +20,8 @@ setup() {
     echo "Configure management API and management UI"
     cat /var/www/html/constants.json.template | \
     sed "s#http://localhost:8093/#${MGMT_API_URL}#g;s#http://localhost:4200/#${MGMT_UI_URL}#g" > /var/www/html/constants.json
+
+    [[ -n $BASEPATH ]] && sed -i "s%<base href=\"/\">%<base href=\"$BASEPATH\">%g" /var/www/html/index.html
```
brasseld: Where / How is defined BASEPATH ?

brasseld:
Hi @pletessier 

Is this PR still relevant?

pletessier:
Hi @brasseld 

I don't need it anymore since I'm using Traefik and "Host" rules rather than "Path".

But I still think it should be configurable, preferably in Angular.

brasseld:
Unfortunately, the base path can not be configured using Angular so we are forced to rely on the RP in front of the UI to "rewrite" this path...

tcompiegne:
A lot of configuration has been changed since this PR. 

To be closed.

PR #558, opened 2019-04-11
Provide default URL and credentials in the docker instructions

paneq:
* Please fix me...


paneq:
It turns out it is not `admin / admin` but `admin / adminadmin` as per https://docs.gravitee.io/am/1.x/am_userguide_authentication.html :) It would be nice to add it here

brasseld:
You can add it by yourself in the doc :)

tcompiegne:
The documentation has been updated since.

PR #606, opened 2019-06-19
Client secret might be skipped if not required by the client

JesusTheHun:
According to https://tools.ietf.org/html/rfc6749#section-4.1.3, the client might not need to be authenticated if client credentials have not been issued. 
A client secret cannot be qualified as a credential if it does not allow the client to authenticate. Therefore if the grant type "client_credentials" is not given to the client, the grant types _password_, _authorization_code_ in the context of a non-private client such as a web app with PKCE flow, the request may skip client authorization.

This PR does not handle the case where an _authorization_code_ grant type is given to a private client and therefore requires a client authentication. 

I suggest we allow the administrator to flag the client as "private" or more specifically a flag to "enforce client authentication" with a default value as true by default.

I may enhance this PR with such flag if a) you would like me to do so b) indicate where I should put my fingers to implement such feature

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/provider/ClientAuthenticationProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/provider/ClientAuthenticationProvider.java
@@ -56,7 +56,10 @@ public class ClientAuthenticationProvider implements AuthProvider {
                 .findByClientId(clientId)
                 .subscribe(
                         client -> {
-                            if (client.getClientSecret().equals(clientSecret)) {
+                            boolean secretMatches = client.getClientSecret().equals(clientSecret);
+                            boolean secretIsRequired = client.isSecretRequired();
+
+                            if (secretMatches || !secretIsRequired) {
                                 authHandler.handle(Future.succeededFuture(new Client(client)));
                             } else {
                                 authHandler.handle(Future.failedFuture(new BadClientCredentialsException()));
--- a/gravitee-am-model/src/main/java/io/gravitee/am/model/Client.java
+++ b/gravitee-am-model/src/main/java/io/gravitee/am/model/Client.java
@@ -679,6 +679,10 @@ public class Client implements Cloneable{
         this.accountSettings = accountSettings;
     }
 
+    public boolean isSecretRequired() {
+        return getAuthorizedGrantTypes().contains(GrantType.CLIENT_CREDENTIALS);
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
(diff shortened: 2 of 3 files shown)


brasseld:
Can you squash the two commits into a single one please

JesusTheHun:
It's done

lusoalex:
> It's done

@JesusTheHun First thanks a lot for your contribution!

My opinion is you are right on the need, Code flow does not always need client_secret, especially for mobile using PKCE.

But I really think that the way to implement it should be based on the application type of the client.
Something like application_type: 
 - "web" ==> Code flow + client_secret required 
 - "mobile" ==> Code flow + client_secret not mandatory + PKCE required 
 - "Single Page App" ==> Implicit + client_secret not required
 - "Server to Server" ==> client_credentials so client_secret required 
 - "trusted" ==> Ressource Owner Password Flow so client_secret required
 - "hybrid" ==> When code & implicit are mixed, client_secret required on the token endpoint

That requires to create one client per use case instead of mixing multiple grants on the same client_id.
I strongly believe this is the best way but is also more complex to implement, and later to upgrade already created clients.

What's your feedback?

JesusTheHun:
I thought about that and somehow agree. My issue with this approach is a) you decide on behalf of the user and b) it requires more changes on the UI side and the API side, since you replace the choice of grant_type by the app type. 

My fallback is what I proposed in my initial post, let the user decide and keep grant_type choices as they are, but allow the user to dis-enforce client_secret requirement. The downside of my approach is you should, according to the spec, auto-enforce client authentication if later on client_credentials is enabled.

What is broadly done by competitors is simply to required client_secret when the type of request is a  _client_credentials_ grant type, regardless of the grant types available. This is not strictly spec compliant but it is actually not a security exposure. Even though we agree it is not a good practice to have a single client with many grant types.

So I guess it's your call !

tcompiegne:
Hello guys and @JesusTheHun 

First of all, thanks for you contribution, as you spoke with @lusoalex we must see further on this topic. 

At GraviteeSource, we have already begun the work on this subject called `Application Management`.

Some couple issues have already been created : 

https://github.com/gravitee-io/issues/issues/1973 (Application Management)
https://github.com/gravitee-io/issues/issues/2090 (OAuth 2.0 Public Clients)

The goal is to create a new `Application Creation Workflow` with an application/client's type as starting point.

image

We also want to create some "expert mode" UI screens to hide advanced (too technical) options that most people do not to use (or do not want to change).

The most difficult part is to be consistent with the OAuth 2.0 and OIDC specs. We will must compose with the following OIDC client parameters and use them to consider all possible scenarios.

- `application_type` (native or web)
- `token_endpoint_auth_method` (client_secret_post, client_secret_basic, client_secret_jwt, private_key_jwt, and none)

We could image the following possible combinations : 

- web + none => SPA
- web + client_secret_basic => Web App
- native + none => Mobile App
- native + secret => Desktop App
- ...

And of course we must ensure the backward compatibility :).

A big topic, but we definitely move forward quickly on this.

JesusTheHun:
Hello Titouan,

I had a hunch when I saw the app type property in the api but not in the UI. It's cool that you are already working on this. If you need some help or user feedback about this don't hesitate to ping me here or on gitter.

tcompiegne:
Closed by https://github.com/gravitee-io/graviteeio-access-management/pull/707

PR #638, opened 2019-08-06
feat(idp): internal refactoring to handle new type of identity provider

tcompiegne:
closes gravitee-io/issues#2185


brasseld:
Closes https://github.com/gravitee-io/issues/issues/2510

PR #670, opened 2019-10-28
Update gravitee.yml

lazki:
Added PATCH because we need it to edit domain configuration. Otherwise you get a «Invalid CORS request» when modifying a domain configuration in the AM UI, on Dynamic Client Registration on the «Client Registration Settings».


tcompiegne:
👍

PR #687, opened 2019-12-04
Always enhance request with users permissions if required

ventuc:
This PR fixes Access Management issue https://github.com/gravitee-io/issues/issues/2958.

Sopes related to user's permissions are always granted when the client is configured with option "Enhance scopes" (that is if `client.isEnhanceScopesWithUserPermissions() == true`).

Related test cases are included in this PR.


tcompiegne:
Hi @ventuc, very nice work, can you just change the base branch to the `2.10.x` ? That way, we can do a 2.10.4 maintenance release.

Regards

ventuc:
Hi @tcompiegne, I've just changed base from `master` to `2.10.x`. Thanks!

tcompiegne:
Hi @ventuc, all tests passed successfully.

Great job, thank you 👏

PR #690, opened 2019-12-07
Fix users search pagination in Management API

ventuc:
This PR fixes issue https://github.com/gravitee-io/issues/issues/2994 by taking into account the page parameter in the users list endpoint of the Management API not only for users listing but also for searching (when parameter `q` is passed).

The actual fix is in the MongoDB query: call to `skip()` was missing in class `MongoDbRepository`. The rest are simply adaptations of methods signatures, in order to pass the `page` parameter from the endpoint to the query.

Added also a test case to check the correctness of paginated results.


ventuc:
Please note that for file UsersResource only line 85 is really changed, the of the file is a before... don't know why git is showing that all the lines have been changed. For my IDE only line 85 is changed.

tcompiegne:
Hi @ventuc ,  nice work, thanks again !

PR #714, opened 2020-02-09
fix:snapshot_dependencies

lusoalex:
After an update of the master, I was unable to build the project due to unresolved dependencies.  
After checking, the root cause was due to snapshot dependencies.  

Instead of asking developers to add them into their maven settings.xml, I think this is more developer-friendly to add the repository directly on your pom.xml as done into this PR.

Regards,
Alexandre.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -239,4 +239,13 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <!-- Allow oss snapshot without requiring the developer to add it into it's own maven settings.xml -->
+    <repositories>
+        <repository>
+            <id>snapshots-repo</id>
+            <name>snapshots repository</name>
+            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
+        </repository>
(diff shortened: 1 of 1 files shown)


brasseld:
Does it make sense to add this into gravitee-parent to make it available for all gravitee modules ?

lusoalex:
> Does it make sense to add this into gravitee-parent to make it available for all gravitee modules ?

Depends if you expect to use snapshots in any submodules. (only not relevant for gravitee-am-ui from my point of view).

You may also reject any external snapshots through a "release" profile when you build a stable release.

FYI, to test/reproduce the unresolved dependencies, I used this :
```bash
rm -rf ~/.m2/repository/io/gravitee/plugin/gravitee-plugin-core/1.12.0-SNAPSHOT/
```

tcompiegne:
closed by https://github.com/gravitee-io/gravitee-parent/commit/55aa3c4518ffdc3eadae0b9a6636c300f1339a90

PR #833, opened 2020-06-01
feat(uma2): add new resource server application type

tcompiegne:
closes gravitee-io/issues#3850

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ResourceServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ResourceServiceImpl.java
@@ -15,11 +15,15 @@
  */
 package io.gravitee.am.service.impl;
 
+import io.gravitee.am.model.Application;
+import io.gravitee.am.model.User;
 import io.gravitee.am.model.common.Page;
 import io.gravitee.am.model.uma.Resource;
 import io.gravitee.am.repository.management.api.ResourceRepository;
+import io.gravitee.am.service.ApplicationService;
 import io.gravitee.am.service.ResourceService;
 import io.gravitee.am.service.ScopeService;
+import io.gravitee.am.service.UserService;
 import io.gravitee.am.service.exception.*;
 import io.gravitee.am.service.model.NewResource;
 import io.reactivex.Completable;
@@ -33,8 +37,7 @@ import org.springframework.stereotype.Component;
 
 import java.net.MalformedURLException;
 import java.net.URI;
-import java.util.Date;
-import java.util.List;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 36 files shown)


gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.ts
```diff
@@ -70,6 +70,10 @@ export class ApplicationOverviewComponent implements OnInit {
     return this.application.type.toLowerCase() === 'service';
   }
 
+  isUmaApp(): boolean {
+    return this.application.type.toLowerCase() === 'resource_server';
```
lusoalex, line 74: The application should be considered as a resource application as soon as it get the `uma_protection` scope.  
By running current uma postman scenario, the current "RS-Client" is not considered as a UMA resource app while it can generate resource set & permissions tickets.  
Should we update the application type automatically after every scopes changes?

gravitee-am-ui/src/app/domain/applications/application/advanced/resources/resources.component.html
```diff
@@ -0,0 +1,67 @@
...
+-->
+<div fxLayout="column">
+  <div class="gv-form-section">
+    <div class="gv-form-section-title">
+      <h5>Resources</h5>
+      <mat-divider></mat-divider>
+    </div>
+    <div>
+      <p>Administrators of this application can list resources registered by resource owners via the UMA 2.0 Resource Set Registration Endpoint.</p>
+      <p>The endpoint defined by this specification at which the resource server registers resource sets it wants the authorization server to know about to apply access policies.</p>
```
lusoalex, line 26:
```suggestion
      <p><i>This endpoint defined by the UMA 2.0 specification is used by the resource server to register resource sets it wants the authorization server to protect and manage access policies.</i></p>
```
Suggestion pour compréhension?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ResourceServiceImpl.java
```diff
@@ -170,4 +218,18 @@ public Completable delete(Resource resource) {
         }
         return Single.just(toValidate);
     }
+
+    private User convert(User user) {
```
lusoalex, line 222: It seems more a kind of clone or "partial clone" than a convert right?
_When I read it, I first thought it was two different kind of User, then as there's no package name in both, I understood that was more a kind of clone..._

gravitee-am-ui/src/app/domain/applications/application/advanced/resources/resources.component.html
tcompiegne: FP

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ResourceServiceImpl.java
tcompiegne: FP

gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.ts
tcompiegne: For now I will leave it as it is.

PR #860, opened 2020-07-02
feat(vhost): allow to expose security domain using virtual hosts

jhaeyaert:
Closes gravitee-io/issues#3199
Closes gravitee-io/issues#3507


tcompiegne:
👍

PR #935, opened 2020-10-09
feat(grant): enhance user on ExtensionGrant to get dynamic scopes

micivray:
In the case of ExtensionGrantGranter.resolveResourceOwner the user's roles permissions are not loaded, whereas we expect them to be added to user's scope.
These permissions/scopes are well loaded by AuthorizationCodeGranter or RefreshTokenGranter for example.

_An other solution that the one proposed could be to modify the AbstractTokenGranter grant method but it didn't seem relevant as the issue is only on extension grant._

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/granter/extensiongrant/ExtensionGrantGranter.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/granter/extensiongrant/ExtensionGrantGranter.java
@@ -20,6 +20,7 @@ import io.gravitee.am.common.oidc.idtoken.Claims;
 import io.gravitee.am.extensiongrant.api.ExtensionGrantProvider;
 import io.gravitee.am.gateway.handler.common.auth.UserAuthenticationManager;
 import io.gravitee.am.gateway.handler.common.auth.idp.IdentityProviderManager;
+import io.gravitee.am.gateway.handler.common.user.UserService;
 import io.gravitee.am.gateway.handler.oauth2.exception.InvalidGrantException;
 import io.gravitee.am.gateway.handler.oauth2.exception.UnauthorizedClientException;
 import io.gravitee.am.gateway.handler.oauth2.service.granter.AbstractTokenGranter;
@@ -55,6 +56,7 @@ public class ExtensionGrantGranter extends AbstractTokenGranter {
     private final ExtensionGrant extensionGrant;
     private final UserAuthenticationManager userAuthenticationManager;
     private final IdentityProviderManager identityProviderManager;
+    private final UserService userService;
     private Date minDate;
 
     public ExtensionGrantGranter(ExtensionGrantProvider extensionGrantProvider,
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/granter/extensiongrant/ExtensionGrantGranter.java
```diff
@@ -124,6 +128,7 @@ public boolean handle(String grantType, Client client) {
                                         user.setRoles(idpUser.getRoles());
                                         return user;
                                     })
+                                    .flatMap(user -> userService.enhance(user).toMaybe())
```
tcompiegne, line 131: Hey @micivray, you cannot use directly the `enhance` method because the current user hold the ID of the IDP user (external_id). You need first to retrieve the internal user from Gravitee, here is a snippet

```
.flatMap(user -> userService.findByDomainAndExternalIdAndSource(domain, user.getId(), idp)
                                 .flatMap(userService.enhance(user).toMaybe())
                                 .defaultEmpty(user))

tcompiegne (commented):
Hi @micivray thanks for the PR, you can check my comments.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/granter/extensiongrant/ExtensionGrantGranter.java
micivray: Ok thank you @tcompiegne. I didn't think that I needed to load any other information as it was working as is (local test with a LDAP IDP) just setting: user id with idpUser id.

In this case do you advice to call userService.findByDomainAndUsernameAndSource if userService.findByDomainAndExternalIdAndSource is not found (empty) ?

bcollard: Hi TItouan!
Here is a little update if this can help understand why @micivray is struggling with the suggestion you made.
Our flow is: 
(end user / RO) --> Client App --> Corporate OIDC server --> get an ID_TOKEN, then used a JWT bearer --> (Gravitee APIM) --> custom API inserting the user identity into an OpenLDAP, within a group carried by one of the JWT Bearer claim --> Gravitee AM --> OpenLDAP (as IdP) to fetch the user's group --> Gravitee AM to map user's group to a role and to dynamically add scopes

We have a more detailed procedure (24 pages) on how-to reproduce the bug detailed in issue https://github.com/gravitee-io/issues/issues/3534, that we can share privately.

So here, as @micivray said, the "user is not saved in our case", as you expect in your suggestion.

What do you think?

micivray: You can see my proposal with the fix below to enhance user roles and permissions without calling the enhance method with the wrong ID.

PR #1029, opened 2021-02-04
fix(ui): fix domain vhost listening host pattern

tcompiegne:
fixes gravitee-io/issues#4958


jhaeyaert:
I added some fixes. @tcompiegne can you check please.

PR #1038, opened 2021-02-07
fix(resetPassword): use one time token for reset password action

tcompiegne:
fixes gravitee-io/issues#4914


jhaeyaert (approved):
Seems good to me. I let you merge yourself and let you check if we should also display the error when accessing to the reset form too (currently the error only appears after posting the form).

PR #1041, opened 2021-02-10
fix(proxy): add property to explicitly enable httpclient proxy

jhaeyaert:
Closes gravitee-io/issues#5048


jhaeyaert:
I converted the PR to Draft because we still need to discuss how we should document the new `httpClient.proxy.enabled` property knowing that we have to apply this fix on 3.0.x, 3.5.x and master

aelamrani:
How do you plan to communicate on this BC?

tcompiegne:
No BC since it has never worked. Communication will be made through our enterprise channel with our customers.

PR #1090, opened 2021-04-02
fix(cockpit): set cockpit status to deleted when handling GoodBye com…

jhaeyaert:
…mand

Fixes gravitee-io/issues#5325

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/GoodbyeCommandHandler.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/GoodbyeCommandHandler.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.am.management.service.impl.commands;
 
+import io.gravitee.am.model.Installation;
 import io.gravitee.am.service.InstallationService;
 import io.gravitee.cockpit.api.command.Command;
 import io.gravitee.cockpit.api.command.CommandHandler;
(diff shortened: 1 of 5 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/GoodbyeCommandHandler.java
```diff
@@ -49,8 +53,11 @@ public GoodbyeCommandHandler(InstallationService installationService) {
 
     @Override
     public Single<GoodbyeReply> handle(GoodbyeCommand command) {
-        return installationService.delete()
-                .andThen(Single.just(new GoodbyeReply(command.getId(), CommandStatus.SUCCEEDED)))
+        return installationService.getOrInitialize()
+                .map(Installation::getAdditionalInformation)
+                .doOnSuccess(additionalInfos -> additionalInfos.put(COCKPIT_INSTALLATION_STATUS, DELETED_STATUS))
+                .flatMap(installationService::setAdditionalInformation)
```
tcompiegne, line 59: Can we do all that in the delete() method of the installationService ? 

Not a huge fan of having a setter method at service level.

tcompiegne (changes requested):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/GoodbyeCommandHandler.java
jhaeyaert: I've added an `addAdditionalInformation` method on `InstallationService` and did the work here. I preferred not to change the `delete` method as the `GoodByeCommand` does not delete the installation anymore but instead just set the cockpit status to `DELETED`.

PR #1106, opened 2021-04-14
fix(memberships): properly check primary owner unicity

tcompiegne:
Fixes gravitee-io/issues#5428


loriepisicchio (approved):
I was not really able to test it as Cockpit now sends only the PrimaryOwner once, but code looks good

PR #1136, opened 2021-04-30
feat(pagination) add pagination feature for oauth scopes + fix paging…

DCourtneyGravitee:
… into querying issues for domains, scopes, and applications closes gravitee-io/issues#5213


leleueri:
Hi,

I checked the PR and I think that some tests are missing on the repositories for the search method. 
I did some tests using MongoDB & PostgreSQL, it seems that for PostgreSQL the search is case sensitive this is not the case with MongoDB.

leleueri (changes requested):
Hi,

I checked the PR and I think that some tests are missing on the repositories for the search method. 
I did some tests using MongoDB & PostgreSQL, it seems that for PostgreSQL the search is case sensitive this is not the case with MongoDB.

PR #1138, opened 2021-04-30
fix(gateway): handle encoded redirectUri

leleueri:
Fixes gravitee-io/issues#5508


leleueri:
Still have so work to do

PR #1154, opened 2021-05-10
return only the requested scopes when the enhance scopes with user permission option is active

phbpx:
Fixes [gravitee-io/issues#3839](https://github.com/gravitee-io/issues/issues/3839)


klinux (approved):

tcompiegne:
Well done guys 👍

PR #1198, opened 2021-06-09
chore(vertx4): move to gravitee-bom and use vertx4.

jhaeyaert:
Related to gravitee-io/issues#5220


leleueri:
Some workflow regarding nightly deployment seems to be missing from the `.circleci/config.yml` file.

PR #1203, opened 2021-06-15
feat: Infobip plugin

droderuan:
Hi! I created this plugin to make it possible to use the 2FA with SMS service from [Infobip](https://www.infobip.com/). The plugin was created using [the Infobip SDK](https://github.com/infobip/infobip-api-java-client) and I used the Twilio plugin as a reference.

### DIFF
--- a/gravitee-am-resource/gravitee-am-resource-infobip/pom.xml
+++ b/gravitee-am-resource/gravitee-am-resource-infobip/pom.xml
@@ -1,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>gravitee-am-resource</artifactId>
+        <groupId>io.gravitee.am.resource</groupId>
+        <version>3.9.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>gravitee-am-resource-infobip</artifactId>
+    <name>Infobip 2FA</name>
+
+    <properties>
+        <maven.compiler.source>16</maven.compiler.source>
+        <maven.compiler.target>16</maven.compiler.target>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>io.gravitee.am.resource</groupId>
+            <artifactId>gravitee-am-resource-api</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.gravitee.am.service</groupId>
+            <artifactId>gravitee-am-service</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Infobip -->
+        <dependency>
+            <groupId>com.infobip</groupId>
+            <artifactId>infobip-api-java-client</artifactId>
+            <version>3.0.1</version>
+        </dependency>
+
+        <!-- rxjava -->
+        <dependency>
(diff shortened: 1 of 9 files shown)


tcompiegne:
Hi @droderuan, amazing work, we will do the review ASAP.

gravitee-am-resource/gravitee-am-resource-infobip/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>gravitee-am-resource-infobip</artifactId>
+    <name>Infobip 2FA</name>
+
+    <properties>
+        <maven.compiler.source>16</maven.compiler.source>
+        <maven.compiler.target>16</maven.compiler.target>
+    </properties>
```
tcompiegne, line 35: Can you remove these properties, java version is managed by our gravitee parent module.

gravitee-am-resource/gravitee-am-resource-infobip/src/main/java/io/gravitee/am/resource/infobip/provider/InfobipResourceProvider.java
```diff
@@ -0,0 +1,120 @@
...
+        String apiKeyPrefix = this.configuration.getApiKeyPrefix();
+        String messageId = this.configuration.getMessageId();
+        String applicationId = this.configuration.getApplicationId();
+
+        ApiClient apiClient = new ApiClient();
+        apiClient.setApiKeyPrefix(apiKeyPrefix);
+        apiClient.setApiKey(apiKey);
+        apiClient.setBasePath(baseUrl);
+
+        this.tfaApi = new TfaApi(apiClient);
```
tcompiegne, line 66: Client creation should be done only one time at start up (afterPropertiesSet() method for example). This way you avoid to initiate the client every request.

gravitee-am-resource/gravitee-am-resource-infobip/src/main/java/io/gravitee/am/resource/infobip/provider/InfobipResourceProvider.java
```diff
@@ -0,0 +1,120 @@
...
+                boolean isSuccessful = sendCodeResponse.getSmsStatus().equals("MESSAGE_SENT");
+
+                if(!isSuccessful) {
+                    emitter.onError(new SendChallengeException("Message not sent"));
+                }
+
+                this.pinId = sendCodeResponse.getPinId();
+
+                LOGGER.debug("Infobip Verification code asked with ID '{}'", sendCodeResponse.getPinId());
+                emitter.onComplete();
```
tcompiegne, line 87: Put this code in a "else" section, otherwise both emitter.onError and  emitter.onComplete(); will be called if there is an error.

gravitee-am-resource/gravitee-am-resource-infobip/src/main/java/io/gravitee/am/resource/infobip/provider/InfobipResourceProvider.java
```diff
@@ -0,0 +1,120 @@
...
+
+                LOGGER.debug(
+                        "Infobip Verification code with ID '{}' verified with status '{}'",
+                        this.pinId,
+                        verified);
+                if (!verified) {
+                    emitter.onError(new InvalidCodeException("Challenger not verified"));
+                }
+
+                emitter.onComplete();
```
tcompiegne, line 113: Put this code in a "else" section, otherwise both emitter.onError and  emitter.onComplete(); will be called if there is an error.

tcompiegne (changes requested):

gravitee-am-resource/gravitee-am-resource-infobip/src/main/java/io/gravitee/am/resource/infobip/provider/InfobipResourceProvider.java
droderuan: Hi @tcompiegne
So, I see on the twilio plugin, they use this:
```java
 @Override
    public ResourceProvider start() throws Exception {
        Twilio.init(configuration.getAccountSid(), configuration.getAuthToken());
        return this;
    }

    @Override
    public ResourceProvider stop() throws Exception {
        Twilio.destroy();
        return this;
    }
```

Do I need to put the client instantiation on the `start` method?

tcompiegne: Yes exactly, and if there is a destroy method for your client, you can call it in the stop() method.

droderuan: I changed it, but just for curiosity: what would be the `afterPropertiesSet()`? I looked into the interfaces of `MFAResourceProvider` and didn't find it. Or is it more like that I'm setting the configurations and then, **after it**, I do:
```java
this.tfaApi = new TfaApi(apiClient);
```

Is just I'm confused if `afterPropertiesSet()` is a method hahahaha

tcompiegne: :) It's just the method of InitializingBean Spring object. Which can also be called in addition of the start method. But start method is fine.

tcompiegne (approved):

PR #1272, opened 2021-07-21
fix(passwordless): make HTTP IdP compatible with passwordless

tcompiegne:
fixes gravitee-io/issues#5889

### DIFF
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
@@ -18,17 +18,17 @@ package io.gravitee.am.identityprovider.http.authentication;
 import io.gravitee.am.common.exception.authentication.AuthenticationException;
 import io.gravitee.am.common.exception.authentication.InternalAuthenticationServiceException;
 import io.gravitee.am.common.oidc.StandardClaims;
-import io.gravitee.am.identityprovider.api.Authentication;
-import io.gravitee.am.identityprovider.api.AuthenticationProvider;
-import io.gravitee.am.identityprovider.api.DefaultUser;
-import io.gravitee.am.identityprovider.api.User;
+import io.gravitee.am.identityprovider.api.*;
 import io.gravitee.am.identityprovider.http.HttpIdentityProviderMapper;
 import io.gravitee.am.identityprovider.http.HttpIdentityProviderResponse;
 import io.gravitee.am.identityprovider.http.HttpIdentityProviderRoleMapper;
 import io.gravitee.am.identityprovider.http.authentication.spring.HttpAuthenticationProviderConfiguration;
+import io.gravitee.am.identityprovider.http.configuration.HttpAuthResourcePathsConfiguration;
 import io.gravitee.am.identityprovider.http.configuration.HttpIdentityProviderConfiguration;
 import io.gravitee.am.identityprovider.http.configuration.HttpResourceConfiguration;
 import io.gravitee.am.identityprovider.http.configuration.HttpResponseErrorCondition;
+import io.gravitee.am.service.exception.AbstractManagementException;
+import io.gravitee.am.service.exception.TechnicalManagementException;
 import io.gravitee.common.http.HttpHeader;
 import io.gravitee.common.http.HttpHeaders;
 import io.gravitee.common.http.MediaType;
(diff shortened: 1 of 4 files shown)


gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
```diff
@@ -182,7 +121,135 @@
...
+        // prepare request
+        final HttpAuthResourcePathsConfiguration authResourceConfiguration = configuration.getAuthenticationResource().getPaths();
+        if (authResourceConfiguration == null) {
+            return Maybe.empty();
+        }
+        if (authResourceConfiguration.getLoadPreAuthUserResource() == null) {
+            return Maybe.empty();
+        }
+
+        final HttpResourceConfiguration readResourceConfiguration = authResourceConfiguration.getLoadPreAuthUserResource();
```
leleueri, line 133: IMHO Before going further, we should validate the configuration. We should at least provide an URI and a HTTP Method to process the request otherwise we should log a WARN and return empty.
Wdyt?

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
```diff
@@ -182,7 +121,135 @@
...
+                }
+            }
+        }
+
+        // if remote API call failed, throw exception
+        if (lastException != null) {
+            throw lastException;
+        }
+        if (responseBody == null) {
+            return null;
```
leleueri, line 249: This method is called twice and after each call an exception is thrown in case of null value. 
Why do we not thrown the  exception here?

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/resources/schemas/schema-form.json
```diff
@@ -102,6 +102,104 @@
               "exception"
             ]
           }
+        },
+        "paths": {
+          "type": "object",
+          "id": "urn:jsonschema:io:gravitee:am:identityprovider:http:HttpAuthResourcePathsConfiguration",
+          "title": "Paths",
```
leleueri, line 109: I don't understand the meaning of "Paths" in this case.
Maybe we could have "loadPreAuthUserResource" at the same level as "authenticationResource" ?

Also I'm wondering if a checkbox to enable or disable the "loadPreAuthUserResource" make sens?

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
```diff
@@ -182,7 +121,135 @@
...
+        }
+        return responseBody.startsWith("[") ?
+                new JsonArray(responseBody).getJsonObject(0).getMap() : new JsonObject(responseBody).getMap();
```
leleueri, line 252: Don't understand the test on the array, before we only process the object. 
In the array case, do we have to take the first entry or do we have to interpret all the entries as an object and merge them into single object?

leleueri (changes requested):

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
tcompiegne: FP

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
tcompiegne: FP

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/resources/schemas/schema-form.json
tcompiegne: It's mainly to have clarity from the UI.

Also I'm wondering if a checkbox to enable or disable the "loadPreAuthUserResource" make sens?

No because it's not like the UserResource where something is loaded into the gateway. This one is dynamically enable or disable with the presence of the value.

gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/authentication/HttpAuthenticationProvider.java
tcompiegne: It's because we can now fetch the user information and sometimes API can return an array of 1 element.

leleueri (approved):

PR #1324, opened 2021-09-02
fix(event): propagate node stop event when shutdown

jhaeyaert:
Closes gravitee-io/issues#5966

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,9 +67,9 @@
 
     <properties>
         <gravitee-bom.version>1.0</gravitee-bom.version>
-        <gravitee-common.version>1.16.2</gravitee-common.version>
+        <gravitee-common.version>1.22.0-SNAPSHOT</gravitee-common.version>
         <gravitee-plugin.version>1.18.0</gravitee-plugin.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -67,9 +67,9 @@
 
     <properties>
         <gravitee-bom.version>1.0</gravitee-bom.version>
-        <gravitee-common.version>1.16.2</gravitee-common.version>
+        <gravitee-common.version>1.22.0-SNAPSHOT</gravitee-common.version>
```
remisultan, line 70: Why are we putting `-SNAPSHOT` versions ?

jhaeyaert: Not sure to understand your question. Just for your information, this PR is part of this issue gravitee-io/issues#5966 which contains multiples PR on different repositories. We put a SNAPSHOT version in order to embed the fix in AM to be able to be properly test it. The SNAPSHOT will be removed once the PRs related to the issue will be merge and the release will be performed.

remisultan: > We put a SNAPSHOT version in order to embed the fix in AM to be able to properly test it.
@jhaeyaert that exactly answers my question. Thank you !

PR #1336, opened 2021-09-08
Create SECURITY.md

tiagoros:
Updating APIM and AM Security policy aligned with our responsible disclosure policy.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,60 @@
+# Security Policy
+## Supported Versions
+Current supported versions for Gravitee Access Management:
+| Version | Supported          |
+| ------- | ------------------ |
+| LTS:    |                    |
+| 3.10.x  | :white_check_mark: |
+| 3.5.x   | :white_check_mark: |
+|         |                    |
+| < 3.10  | :x:                |
+| < 3.5   | :x:                |
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,60 @@
+# Security Policy
+## Supported Versions
+Current supported versions for Gravitee Access Management:
+| Version | Supported          |
+| ------- | ------------------ |
+| LTS:    |                    |
```
remisultan, line 6: Wouldn't it be more interesting to target to the online doc for the supporting versions rather than having to maintain the version directly in the file  🤔 ?
I believe we are maintaining the same way all the products? (Like APIM for instance has the same LTS as AM?)

Having to maintain versions would be prone to forget to modify the file.
Requesting @tcompiegne

tiagoros: Yes, def and I would agree 100% with that @remisultan, but until then can we please put this one in?
@tiffanymarchi can support us on this too and she def should be in the loop.

remisultan: Awesome.
I have created an issue so we don't forget about it: https://github.com/gravitee-io/issues/issues/6156

In the meantime, I'll merge this.

PR #1340, opened 2021-09-10
fix(idp): sanitize credentials for HTTP IdP

tcompiegne:
fixes gravitee-io/issues#6147


gravitee-am-identityprovider/gravitee-am-identityprovider-http/src/main/java/io/gravitee/am/identityprovider/http/utils/SanitizeUtils.java
```diff
@@ -0,0 +1,52 @@
...
+     * @param credentials
+     * @param authenticationHttpHeaders
+     * @return
+     */
+    public static String sanitize(String credentials, List<HttpHeader> authenticationHttpHeaders) {
+        if (authenticationHttpHeaders == null || authenticationHttpHeaders.isEmpty()) {
+            return credentials;
+        }
+
+        if (!authenticationHttpHeaders
```
leleueri, line 41: if the content-type isn't specify, the body mat be a Json (this is the case with the default values, the body is a JSON but no headers are defined.) Do we have to consider JSON as default payload? If any, we have to test the presence of the Content-Type header separately.

PR #1370, opened 2021-09-26
fix(org): only select organization idps during authentication

tcompiegne:
fixes gravitee-io/issues#6181


leleueri:
I push a fix on this PR (dedicated fixup commit) to fix an issue.
The original proposal for this PR prevented an OrganizationUser from authenticating because the GRAVITEE IDP isn't referenced into the list of identities for the organization.

leleueri (commented):
I push a fix on this PR (dedicated fixup commit) to fix an issue.
The original proposal for this PR prevented an OrganizationUser from authenticating because the GRAVITEE IDP isn't referenced into the list of identities for the organization.

PR #1377, opened 2021-09-28
fix: test EnrolledFactorSecurity to avoid NPE and validate the accoun…

leleueri:
…t information (phoneNumbe & email) before factor enrollment

fixes gravitee-io/issues#6258


remisultan (approved):
Retested the whole login flow
LGTM

PR #1441, opened 2021-11-03
fix(oauth): better support of prompt login for external OIDC IdP

tcompiegne:
fixes gravitee-io/issues#6477


remisultan (approved):
LGTM

Just a small warning when backporting

PR #1460, opened 2021-11-22
fix: always return up to date IDP and Email Template in Management se…

leleueri:
…rvice

fixes gravitee-io/issues#6590


leleueri:
Here is a Postman collection with environment description in order to validate the fix.

Without this fix, the user creation may fail (404) or the email template may not be the one we send using the REST API call
With this fix, the user creation succeed and the email received uses the custom template.

[Uploading SUPPORT_2222.postman.zip…]()

tcompiegne (approved):

PR #1469, opened 2021-11-29
feat: separate idp roles from static roles

remisultan:
fixes: https://github.com/gravitee-io/issues/issues/6515

⚠️ Please check the section where the OrganizationUserService updates the roles. There are some questions around it.

https://github.com/gravitee-io/gravitee-access-management/blob/d22277611e29a32bdd8e3dac3851217b219b087c/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/service/impl/AuthenticationServiceImpl.java#L92

Right here whenever the user already exists within the Organization users. If there are no roles within the identityProvider, We skip the role update. This causes for any new users from an idp not to be updated whenever the rule is not present anymore.

If we were to replace the roles, we would have to make some specific checks (is the user not from an internal idp, which does not sound really great)

The solution would be to make a mapper and create also its contrary to roll back users from an initial role. This works because users only have one role in Organization settings

We could also make dynamic roles on Memberships but does that make sense ?

Any other ideas ?

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
@@ -17,13 +17,15 @@ package io.gravitee.am.gateway.handler.scim.mapper;
 
 import io.gravitee.am.common.oidc.StandardClaims;
 import io.gravitee.am.gateway.handler.scim.model.*;
-import io.gravitee.am.gateway.handler.scim.service.impl.UserServiceImpl;
 import io.gravitee.am.identityprovider.api.DefaultUser;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
-import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.Optional.ofNullable;
+import static java.util.stream.Collectors.toList;
 
 /**
  * @author Eric LELEU (eric.leleu at graviteesource.com)
@@ -45,17 +47,17 @@ public class UserMapper {
         Name name = new Name();
         name.setGivenName(user.getFirstName());
         name.setFamilyName(user.getLastName());
-        name.setMiddleName(get(additionalInformation, StandardClaims.MIDDLE_NAME, String.class));
+        name.setMiddleName(get(additionalInformation, StandardClaims.MIDDLE_NAME));
         scimUser.setName(name.isNull() ? null : name);
         scimUser.setDisplayName(user.getDisplayName());
         scimUser.setNickName(user.getNickName());
 
-        scimUser.setProfileUrl(get(additionalInformation, StandardClaims.PROFILE, String.class));
+        scimUser.setProfileUrl(get(additionalInformation, StandardClaims.PROFILE));
         scimUser.setTitle(user.getTitle());
         scimUser.setUserType(user.getType());
         scimUser.setPreferredLanguage(user.getPreferredLanguage());
-        scimUser.setLocale(get(additionalInformation, StandardClaims.LOCALE, String.class));
-        scimUser.setTimezone(get(additionalInformation, StandardClaims.ZONEINFO, String.class));
+        scimUser.setLocale(get(additionalInformation, StandardClaims.LOCALE));
+        scimUser.setTimezone(get(additionalInformation, StandardClaims.ZONEINFO));
         scimUser.setActive(user.isEnabled());
         scimUser.setEmails(toScimAttributes(user.getEmails()));
         // set primary email
@@ -65,7 +67,7 @@ public class UserMapper {
             attribute.setPrimary(true);
             if (scimUser.getEmails() != null) {
                 Optional<Attribute> optional = scimUser.getEmails().stream().filter(attribute1 -> attribute1.getValue().equals(attribute.getValue())).findFirst();
-                if (!optional.isPresent()) {
+                if (optional.isEmpty()) {
                     scimUser.setEmails(Collections.singletonList(attribute));
                 }
             } else {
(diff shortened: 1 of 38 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
```diff
@@ -77,7 +79,7 @@ public static User convert(io.gravitee.am.model.User user, String baseUrl, boole
         scimUser.setPhotos(toScimAttributes(user.getPhotos()));
         scimUser.setAddresses(toScimAddresses(user.getAddresses()));
         scimUser.setEntitlements(user.getEntitlements());
-        scimUser.setRoles(user.getRoles());
+        scimUser.setRoles(getRoleStream(user).flatMap(List::stream).sorted().distinct().collect(toList()));
```
leleueri, line 82: With `getRoleStream` doesn't flatten the two list? 
I would prefer a `getAllRoles` method that includes the processing flow `flatMap(List::stream).sorted().distinct().collect(toList()))` as this method is only used here.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/JWTAuthenticationFilter.java
```diff
@@ -102,26 +103,26 @@ public Authentication attemptAuthentication(HttpServletRequest request, HttpServ
...
         try {
-            JWT payload = jwtParser.parse(authToken);;
+            JWT payload = jwtParser.parse(authToken);
             Map<String, Object> claims = new HashMap<>(payload);
             claims.put(Claims.ip_address, remoteAddress(request));
             claims.put(Claims.user_agent, userAgent(request));
+
             DefaultUser user = new DefaultUser((String) claims.get(StandardClaims.PREFERRED_USERNAME));
             user.setId((String) claims.get(StandardClaims.SUB));
-            user.setRoles(user.getRoles() != null ? user.getRoles() : (List<String>) claims.get(CustomClaims.ROLES));
```
leleueri, line 118: Are you sure that the roles are not used later in AuthenticationServiceImpl.onAuthenticationSuccess ?

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/UserRolesResource.java
```diff
@@ -79,6 +76,7 @@ public void list(
             @PathParam("environmentId") String environmentId,
             @PathParam("domain") String domain,
             @PathParam("user") String user,
+            @QueryParam(value = "dynamic") @DefaultValue(value = "false") boolean dynamic,
```
leleueri, line 79: this new query parameter isn't used in the UI so the roles assigned by the RoleMapper aren't displayed.
IMHO we have to make two calls from the UI to get all roles and disable deletion for dynamic one.
Or maybe we can remove this parameter and always return the both type of roles.
In any cases, we have to return a 400 erro on role deletion if the role is part of the dynamic ones.

wdyt?

leleueri (commented):
with SCIM, if I update a user info on a user description obtained using a get order (user has static and dynamic roles) then the dynamic role is persisted in the static one.

Don't know if it has been discussed with PO team but maybe we could exclude roles that are present into the "dynamic roles" of the user.

wdyt?

Also removing the RoleMapper from Organization IDP don't rollback the user role to the default one. But it is maybe related to your comment...

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/UserRolesResource.java
remisultan: I may go on the first solution and making 2 calls.
Since the separation is made, I'll keep it simple and display them as well

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/JWTAuthenticationFilter.java
remisultan: FP

remisultan:
> with SCIM, if I update a user info on a user description obtained using a get order (user has static and dynamic roles) then the dynamic role is persisted in the static one.
> 
> Don't know if it has been discussed with PO team but maybe we could exclude roles that are present into the "dynamic roles" of the user.
> 
> wdyt?

+1 open to discuss

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/spring/user/SpringDynamicUserRoleRepository.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.stereotype.Repository;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@Repository
+public interface SpringDynamicUserRoleRepository extends RxJava2CrudRepository<JdbcUser.Role, String> {
+    @Query("select * from dynamic_user_roles r where r.user_id = :user")
+    Flowable<JdbcUser.Role> findByUserId(@Param("user") String userId);
```
leleueri, line 32: :thinking: Should it be JdbcUser.DynamicRole ?

leleueri (commented):
LGTM, we just have to discuss the SCIM behavior with @tcompiegne and demonstrate the role screen with him.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/spring/user/SpringDynamicUserRoleRepository.java
remisultan: FP

leleueri (approved):

PR #1473, opened 2021-12-02
feat(oidc): support POST end_session_endpoint (logout)

tcompiegne:
closes gravitee-io/issues#6643


leleueri (changes requested):
I did a quick review of the code base (I will do another hit later) but during my tests I found following issues : 
* when the "single sign out" is enabled, the op_id_token is present into the AM id_token, is it expected ? Until now, we filter thisi attribute
* when the "single sign out" is enabled, If I call the logout endpoint using GET method and the id_token_hint parameter, the External IDP (Azure AD) rises and error. (The logout on azure work if the id_token_hint isn't provided)

leleueri:
as discussed on slack, these two issues are not related to this PR.

leleueri (approved):

PR #1474, opened 2021-12-02
fix(users): add missing last_password_reset field for the JDBC repository

tcompiegne:
fixes gravitee-io/issues#6664


leleueri (approved):
I fixed this PR as the lastPasswordReset wasn't persisted.
I also add a check into the Repository test module

PR #1477, opened 2021-12-03
feat: keep original access/id token from idp

remisultan:
closes: https://github.com/gravitee-io/issues/issues/6549

### DIFF
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/authentication/AbstractOpenIDConnectAuthenticationProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/authentication/AbstractOpenIDConnectAuthenticationProvider.java
@@ -110,7 +110,7 @@ public abstract class AbstractOpenIDConnectAuthenticationProvider extends Abstra
                 builder.addParameter(io.gravitee.am.common.oidc.Parameters.NONCE, SecureRandomString.generate());
             }
             // add state if provided.
-            if(!StringUtils.isEmpty(state)) {
+            if (!StringUtils.isEmpty(state)) {
                 builder.addParameter(Parameters.STATE, state);
             }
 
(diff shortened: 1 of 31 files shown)


leleueri (changes requested):
I would prefer an option in each schema.json of IDPs that are able to provide the ACCESS_TOKEN & ID_TOKEN. In your implementation, you suppose that all external IDP are able to provide this information, but this is not the case for CAS or SAML for example. So for these two types of IDP we will expose a toggle that has no effect. 

We can discuss it if you want.

gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/authentication/AbstractOpenIDConnectAuthenticationProvider.java
```diff
@@ -175,9 +175,15 @@ public Request signInUrl(String redirectUri, String state) {
                     if (httpResponse.statusCode() != 200) {
                         throw new BadCredentialsException(httpResponse.statusMessage());
                     }
-
                     JsonObject response = httpResponse.bodyAsJsonObject();
                     String accessToken = response.getString(ACCESS_TOKEN_PARAMETER);
+                    // We store the token is option is enabled
```
leleueri, line 180: We should also do this at line 140, if the response type is "id_token token"

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/user/password/RedirectToResetAlreadyLoggedInHandler.java
```diff
@@ -0,0 +1,44 @@
...
+import io.vertx.reactivex.ext.web.RoutingContext;
+
+import static io.gravitee.am.gateway.handler.common.vertx.utils.UriBuilderRequest.CONTEXT_PATH;
+import static io.gravitee.am.gateway.handler.root.RootProvider.PATH_RESET_PASSWORD;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class RedirectToResetAlreadyLoggedInHandler implements Handler<RoutingContext> {
```
leleueri, line 32: IMHO this class comes from another feature and shouldn't be committed for this PR. Am I right ?

leleueri (changes requested):
The FranceConnect doesn't contains this new configuration settings.
FranceConnect is en OIDC provided even if it doesn't inherit from the "AbctractOIDCProvider" class

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/user/password/RedirectToResetAlreadyLoggedInHandler.java
remisultan: Indeed

gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/authentication/AbstractOpenIDConnectAuthenticationProvider.java
remisultan: For the `op_access_token ` filtered, you want me to do it here ? https://github.com/gravitee-io/gravitee-access-management/pull/1480

remisultan: Done

leleueri (approved):

PR #1480, opened 2021-12-07
feat: scope openid should not be used to get full profile information

remisultan:
closes: https://github.com/gravitee-io/issues/issues/6516

⚠️ This is a BREAKING CHANGE. openid may need to be associated with more scopes in the future


leleueri (changes requested):
LGTM but as it is a breaking change, I would like a PR in the documentation before closing this one
https://github.com/gravitee-io/gravitee-docs/blob/develop/pages/am/3.x/breaking-changes/breaking-changes.adoc

PR #1507, opened 2021-12-14
fix(plugins): do not expose sensitive configuration

tcompiegne:
fixes gravitee-io/issues#6734

related PRs:
- https://github.com/gravitee-io/gravitee-am-identityprovider-kerberos/pull/12
- https://github.com/gravitee-io/gravitee-notifier-slack/pull/17
- https://github.com/gravitee-io/gravitee-notifier-email/pull/27

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/AlertNotifierServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/AlertNotifierServiceProxy.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.AlertNotifierService;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface AlertNotifierServiceProxy extends AlertNotifierService {
+}
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/BotDetectionServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/BotDetectionServiceProxy.java
@@ -1,0 +1,24 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.BotDetectionService;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface BotDetectionServiceProxy extends BotDetectionService {}
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/CertificateServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/CertificateServiceProxy.java
@@ -1,0 +1,24 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.CertificateService;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface CertificateServiceProxy extends CertificateService {}
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/IdentityProviderServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/IdentityProviderServiceProxy.java
@@ -1,0 +1,25 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.IdentityProviderService;
+
+/**
+ * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface IdentityProviderServiceProxy extends IdentityProviderService {
+}
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ReporterServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ReporterServiceProxy.java
@@ -1,0 +1,24 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.ReporterService;
+
+/**
+ * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface ReporterServiceProxy extends ReporterService {}
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ServiceResourceServiceProxy.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ServiceResourceServiceProxy.java
@@ -1,0 +1,24 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service;
+
+import io.gravitee.am.service.ServiceResourceService;
+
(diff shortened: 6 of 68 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ReporterServiceProxyImpl.java
```diff
@@ -0,0 +1,146 @@
...
+                .switchIfEmpty(Single.error(new ReporterPluginSchemaNotFoundException(oldReporter.getType())))
+                .map(schema -> {
+                    JsonNode updatedConfigurationNode = objectMapper.readTree(updateReporter.getConfiguration());
+                    JsonNode oldConfigurationNode = objectMapper.readTree(oldReporter.getConfiguration());
+                    JsonNode schemaNode = objectMapper.readTree(schema);
+                    if (schemaNode.has("properties")) {
+                        schemaNode.get("properties").fields().forEachRemaining(entry -> {
+                            if (entry.getValue().has("sensitive") &&
+                                    entry.getValue().get("sensitive").asBoolean()) {
+                                if (SENSITIVE_VALUE.equals(updatedConfigurationNode.get(entry.getKey()).asText())) {
```
YoranSys, line 136: Hi, in order to limit the risk of overwriting in case of false manipulation (deletion of a * for example in the ui) is it possible at the front end to make the displayed value of the field in the form not be the real value (null value?). And check here that the value is null or only space maybe.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderServiceProxyImpl.java
```diff
@@ -0,0 +1,170 @@
...
+                .switchIfEmpty(Single.error(new IdentityProviderPluginSchemaNotFoundException(oldIdentityProvider.getType())))
+                .map(schema -> {
+                    JsonNode updatedConfigurationNode = objectMapper.readTree(updateIdentityProvider.getConfiguration());
+                    JsonNode oldConfigurationNode = objectMapper.readTree(oldIdentityProvider.getConfiguration());
+                    JsonNode schemaNode = objectMapper.readTree(schema);
+                    if (schemaNode.has("properties")) {
+                        schemaNode.get("properties").fields().forEachRemaining(entry -> {
+                            if (entry.getValue().has("sensitive") &&
+                                    entry.getValue().get("sensitive").asBoolean()) {
+                                if (SENSITIVE_VALUE.equals(updatedConfigurationNode.get(entry.getKey()).asText())) {
```
YoranSys, line 160: Hi, in order to limit the risk of overwriting in case of false manipulation (deletion of a * for example in the ui) is it possible at the front end to make the displayed value of the field in the form not be the real value (null value?). And check here that the value is null or only space maybe.

gravitee-am-identityprovider/gravitee-am-identityprovider-inline/src/main/resources/schemas/schema-form.json
```diff
@@ -34,7 +34,8 @@
             "title": "Password",
             "description": "Password",
             "type" : "string",
-            "widget": "password"
+            "widget": "password",
+            "sensitive": true
```
leleueri, line 38: The password field of the inline provider is marked as sensitive but it is not filtered.
The REST response contains the clear value. Probably because the users defiend in the inline provider configuration are inside an array

leleueri (changes requested):
* The password filtering isn't working for inline provider
* The Certificate plugins contains two password fields and should be handled by the SensitiveProxy class
* The `gravitee-am-identityprovider-kerberos` EE module should be updated to mark password as sensitive (like LDAP idp). As for the inline provider, the kerberos sensistive element will be in an inner object.

Finally, we should take into account the comment of YoranSys about the update. Indeed, If I remove a single character an update is performed. Maybe simply test if the new value only contains the same character used to mask the sensitive data.

gravitee-am-identityprovider/gravitee-am-identityprovider-inline/src/main/resources/schemas/schema-form.json
remisultan: @leleueri 
> The password filtering isn't working for inline provider
Since the inline providers purpose is for `ease-to-use` and testing the solution, this is hasn't been done.

> The Certificate plugins contains two password fields and should be handled by the SensitiveProxy class

I will add it

> The gravitee-am-identityprovider-kerberos EE module should be updated to mark password as sensitive (like LDAP idp). As for the inline provider, the kerberos sensistive element will be in an inner object.

I will also look into it

> Finally, we should take into account the comment of YoranSys about the update. Indeed, If I remove a single 
character an update is performed. Maybe simply test if the new value only contains the same character used to  ​mask the sensitive data.

My thought was to check  
```java
fieldValue.trim().matches("^(\*+)$") 
```
but wanted to discuss first

leleueri: :+1:

remisultan:
> Since the inline providers purpose is for ease-to-use and testing the solution, this is hasn't been done.

It's also possible that sometimes the default inline isn't changed at all. So we've secured the `inline` idp.
However this required some custom development since we're dealing with a different structure from the schema and the actual configuration. 

Username is the unique key per Inline provider. If the username changes or that we delete a user. A pop up message will appear to warn the end-user

Screenshot 2021-12-30 at 12 01 15

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/AlertNotifierServiceProxy.java
```diff
@@ -0,0 +1,44 @@
...
+import io.gravitee.am.service.model.PatchAlertNotifier;
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Single;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface AlertNotifierServiceProxy {
```
leleueri, line 32: I wondering if we have to redefine all methods for Proxy interfaces, why not just inherit from the service interface ?
`AlertNotifierServiceProxy extends AlertNotifierService`

Wdyt?

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/BotDetectionServiceProxy.java
```diff
@@ -0,0 +1,54 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface BotDetectionServiceProxy {
```
leleueri, line 31: same question as the AlertNotifierServiceProxy one

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/CertificateServiceProxy.java
```diff
@@ -0,0 +1,63 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface CertificateServiceProxy {
```
leleueri, line 31: same question as the AlertNotifierServiceProxy one

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/IdentityProviderServiceProxy.java
```diff
@@ -0,0 +1,73 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface IdentityProviderServiceProxy {
```
leleueri, line 32: same question as the AlertNotifierServiceProxy one

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ReporterServiceProxy.java
```diff
@@ -0,0 +1,64 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface ReporterServiceProxy {
```
leleueri, line 31: same question as the AlertNotifierServiceProxy one

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/ServiceResourceServiceProxy.java
```diff
@@ -0,0 +1,42 @@
...
+import io.reactivex.Completable;
+import io.reactivex.Flowable;
+import io.reactivex.Maybe;
+import io.reactivex.Single;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public interface ServiceResourceServiceProxy {
```
leleueri, line 31: same question as the AlertNotifierServiceProxy one

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/BotDetectionServiceProxyImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Autowired
+    private BotDetectionPluginService botDetectionPluginService;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Override
+    public Maybe<BotDetection> findById(String id) {
+        return botDetectionService.findById(id).flatMapSingle(this::filterSensitiveData).toMaybe();
```
leleueri, line 55: As `botDetectionService.findById(id)` returns a `Maybe`, it can be empty. In that case `flatMapSingle` will throw an error.
We should do the same tihng as https://github.com/gravitee-io/gravitee-access-management/pull/1507/files#diff-2eed3cdefb0682c04dfda0b09f2eca8ff3fb8b64a1b56b7d8e8ed14cf6a20e1aR84

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/CertificateServiceProxyImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+    @Autowired
+    private CertificatePluginService certificatePluginService;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Override
+    public Maybe<Certificate> findById(String id) {
+        return certificateService.findById(id).flatMapSingle(this::filterSensitiveData).toMaybe();
```
leleueri, line 55: As `certificateService.findById(id)` returns a Maybe, it can be empty. In that case `flatMapSingle` will throw an error.
We should do the same tihng as https://github.com/gravitee-io/gravitee-access-management/pull/1507/files#diff-2eed3cdefb0682c04dfda0b09f2eca8ff3fb8b64a1b56b7d8e8ed14cf6a20e1aR84

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ReporterServiceProxyImpl.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public Flowable<Reporter> findByDomain(String domain) {
+        return reporterService.findByDomain(domain).flatMapSingle(this::filterSensitiveData);
+    }
+
+    @Override
+    public Maybe<Reporter> findById(String id) {
+        return reporterService.findById(id).flatMap(reporter -> filterSensitiveData(reporter).toMaybe());
```
leleueri, line 63: As `reporterService.findById(id)` returns a Maybe, it can be empty. In that case flatMapSingle will throw an error.
We should do the same tihng as https://github.com/gravitee-io/gravitee-access-management/pull/1507/files#diff-2eed3cdefb0682c04dfda0b09f2eca8ff3fb8b64a1b56b7d8e8ed14cf6a20e1aR84

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ServiceResourceServiceProxyImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+    @Autowired
+    private ServiceResourceService serviceResourceService;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Override
+    public Maybe<ServiceResource> findById(String id) {
+        return serviceResourceService.findById(id).flatMapSingle(this::filterSensitiveData).toMaybe();
```
leleueri, line 55: As findById(id) returns a Maybe, it can be empty. In that case flatMapSingle will throw an error.
We should do the same tihng as https://github.com/gravitee-io/gravitee-access-management/pull/1507/files#diff-2eed3cdefb0682c04dfda0b09f2eca8ff3fb8b64a1b56b7d8e8ed14cf6a20e1aR84

leleueri (changes requested):
The REST ReportersResource doesn't use the ReporterServiceProxy, even if currently we can't add Mongo or JDBC reporter, we should use Proxy to not forget using it when a new kind of reporter with sensitive data will be created

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ServiceResourceServiceProxyImpl.java
remisultan: Done

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ReporterServiceProxyImpl.java
remisultan: Done

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/CertificateServiceProxyImpl.java
remisultan: Done

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/BotDetectionServiceProxyImpl.java
remisultan: Done

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/AlertNotifierServiceProxy.java
remisultan: I guess this makes sense.
Having a separated interface makes the separation of concerns more distinct. 
However I believe that whenever we expand a new method, it will be both use in the gateway and the management. 

If we start in the future to have a lot of methods only used in the service and not in the proxy for management, we'll have to rethink that choice.

In the meantime, that sounds like a good idea.

PR #1508, opened 2021-12-15
fix:Do not throw IllegalArgumentException if the a certificat is inva…

leleueri:
…id for WebAuthn to allow domain startup

fixes gravitee-io/issues#6737


remisultan (approved):
LGTM

PR #1514, opened 2021-12-15
fix: CWE-326: Inadequate Encryption Strength

YoranSys:
Implement secure HTTPS communication TLSv1.3.
Consider using latest TLSv1.2 instead of TLS.

https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html#sslcontext-algorithms


leleueri:
Thanks you for you PR (sorry for this late answer)
We have rework the MongoDB connection settings for the 3.18.0 and made this option configurable through the gravitee.yaml

PR #1535, opened 2021-12-27
refactor: remove find after create / update in mongo implementation

remisultan:
This was just a spike where we are removing the findById when not necessary on mongo collection for gatling tests.


remisultan:
duplicate: https://github.com/gravitee-io/gravitee-access-management/pull/1668

PR #1543, opened 2021-12-31
fix: change JS to make SMS & EMAIL MFA work

leleueri:
fixes gravitee-io/issues#6830


remisultan (approved):
LGTM

Let's make sure that the document ready is also applied correctly with Voice Call MFA Factor in 3.14

PR #1570, opened 2022-01-05
feat(management): add lock user endpoint API

remisultan:
closes: https://github.com/gravitee-io/issues/issues/6785

To test:

Ask a token: 
```bash
curl -vvv --location -g --request POST 'http://localhost:8093/management/auth/token' \                                                
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'grant_type=password' \
--data-urlencode 'username=username' \
--data-urlencode 'password=password' --header 'Authorization: Basic Base64Encode(username:password)'
```

Lock the user
```bash
curl -vvv -XPOST \ 
http://localhost:8093/management/organizations/<orgId>/environments/<envId>/domains/<domainId>/users/<userId>/lock \
-H 'Authorization: Bearer <token>'
```

Try to login.

If the user was locked already with Brute Force Login. It locks the user and removes any login attempt counter.


leleueri (changes requested):
Is a user has been locked due to "brut force" detection, he may receive an email to unlock the account. In this case, If the admin call the lock endpoint on this user the user should be indefinitly locked but using the link provided by the email, the user can redefine a password and unlock the account.

Also, a postman test may be useful

tcompiegne (approved):

PR #1620, opened 2022-01-21
fix: delete all user by domain when a domain is rmeoved

leleueri:
fixes gravitee-io/issues#6987


tcompiegne (changes requested):
I think we can implement the deleteByReference method for all the repositories involved with the domain deletion.

tcompiegne (approved):

PR #1635, opened 2022-01-28
chore: upgrade cockpit-connector to 2.0.2

leleueri:
fixes gravitee-io/issues#7020


leleueri:
This commit introduce a regression for user that user java parameters ( -Dcockpit.ws.endpoints;..)

PR #1644, opened 2022-02-01
feat: content security policy

remisultan:
closes: https://github.com/gravitee-io/issues/issues/6716

Example of CSP configuration in `gravitee.yml`

Please suggest in terms of gateway url endpoint. I've put them all but we might not need them everywhere

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/CSPHandlerFactory.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/CSPHandlerFactory.java
@@ -1,0 +1,79 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/CSPHandlerFactory.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public CSPHandler getObject() {
+        if (isNull(INSTANCE)) {
+            var reportOnly = environment.getProperty(HTTP_CSP_REPORT_ONLY, Boolean.class);
+            var directives = getDirectives();
+            if (isNull(reportOnly) && isNull(directives)) {
+                INSTANCE = new NoOpCspHandler();
+            } else {
+                INSTANCE = new CspHandlerImpl(reportOnly, directives);
```
leleueri, line 52: Does it make sens to create a handler with reportOnly without directives?

remisultan: I dont think so, but you would expect that if one of these are set, the handler is correctly instanciated ? 🤔

PR #1651, opened 2022-02-03
fix: try to load UserProvider if the id is missing from the map

leleueri:
fixes gravitee-io/issues#7035

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
@@ -354,7 +354,7 @@ public class IdentityProviderManagerImpl extends AbstractService<IdentityProvide
 
     @Override
     public boolean userProviderExists(String identityProviderId) {
-        return userProviders.containsKey(identityProviderId);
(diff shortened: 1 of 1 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
```diff
@@ -354,7 +354,7 @@ protected boolean useJdbcRepositories() {
 
     @Override
     public boolean userProviderExists(String identityProviderId) {
-        return userProviders.containsKey(identityProviderId);
+        return userProviders.containsKey(identityProviderId) ? true : getUserProvider(identityProviderId).blockingGet() != null;
```
tcompiegne, line 357: Not a huge fan of this blockingGet(), maybe relying on the plugin manager will be more elegant.

tcompiegne (changes requested):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
tcompiegne: Should not possible right now due to the setInternalStatus method. Approve this change.

PR #1652, opened 2022-02-03
Issues/#6836 password expiration policy

TrikyGeeky:
fixes https://github.com/gravitee-io/issues/issues/6836

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
@@ -205,6 +205,9 @@ public class UserAuthenticationServiceImpl implements UserAuthenticationService 
         }
         // set roles
         existingUser.setDynamicRoles(principal.getRoles());
+        if (existingUser.getLastPasswordReset() == null) {
+            existingUser.setLastPasswordReset(existingUser.getUpdatedAt() == null ? new Date() : existingUser.getUpdatedAt());
+        }
 
         Map<String, Object> additionalInformation = ofNullable(principal.getAdditionalInformation()).orElse(Map.of());
         removeOriginalProviderOidcTokensIfNecessary(existingUser, afterAuthentication, additionalInformation);
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordServiceImpl.java
@@ -15,8 +15,10 @@
  */
 package io.gravitee.am.service.impl;
 
+import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.PasswordSettings;
 import io.gravitee.am.model.User;
+import io.gravitee.am.model.oidc.Client;
 import io.gravitee.am.password.dictionary.PasswordDictionary;
 import io.gravitee.am.service.PasswordService;
 import io.gravitee.am.service.validators.password.PasswordValidator;
@@ -25,6 +27,8 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 
+import java.util.Calendar;
+import java.util.Optional;
 import java.util.stream.Stream;
 
 import static java.util.function.Predicate.not;
@@ -71,5 +75,29 @@ public class PasswordServiceImpl implements PasswordService {
                 throw validator.getCause();
             });
         }
+    }
+
+    /**
+     * Check the user password status
+     * @param user Authenticated user
+     * @param client Application
+     * @param domain current domain injected by Spring
+     * @return True if the password has expired or False if not
+     */
+    public boolean checkAccountPasswordExpiry(User user, Client client, Domain domain) {
+        Optional<PasswordSettings> passwordSettings = PasswordSettings.getInstance(client, domain);
+
+        /** If the expiryDate is null or set to 0 so it's disabled */
+        if (passwordSettings.isEmpty() ||
+                (passwordSettings.get().getExpiryDuration() == null
+                                || passwordSettings.get().getExpiryDuration() <= 0)) {
+            return false;
+        }
+
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(user.getLastPasswordReset());
+        calendar.add(Calendar.DAY_OF_MONTH, passwordSettings.get().getExpiryDuration());
+
+        return calendar.compareTo(Calendar.getInstance()) < 0;
     }
 }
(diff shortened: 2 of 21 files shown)


gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/test/java/io/gravitee/am/identityprovider/mongo/authentication/MongoAuthenticationProviderTest.java
```diff
@@ -41,6 +41,13 @@
     @Autowired
     private AuthenticationProvider authenticationProvider;
 
+    /**
+     * This is a workaround for Macs with M1 ships
+     * Link to the github issue:
+     * https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/issues/337#issuecomment-794303570
+     * */
+    static { System.setProperty("os.arch", "i686_64"); }
+
```
remisultan, line 50: Should this be really here ?
Isn't it something we should just mention in the concerned maven plugin ?
```
-Dos.arch= i686_64
```

remisultan (commented):
This is opened to discussion I am wondering if we should not use the PasswordService and create a PasswordValidator for this.

We don't need to go through the whole chain of password validation but that would make sense it could be there and have a dedicated validation for it

https://github.com/gravitee-io/gravitee-access-management/blob/e917799acc54f1e6930de931d1ee36c883aaf71e/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordServiceImpl.java

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
```diff
@@ -184,10 +186,29 @@ private boolean isIndefinitelyLocked(User user) {
...
+     * @param user Authenticated user
+     * @return True if the password has expired or False if not
+     */
+    private boolean checkAccountPasswordExpiry(User user) {
+        /** If the expiryDate is set to 0 so it's disabled */
+        if(domain.getPasswordSettings() == null || (domain.getPasswordSettings() != null && domain.getPasswordSettings().getExpiryDuration() == 0))
+            return false;
+
+        Calendar calendar = Calendar.getInstance();
+        calendar.setTime(user.getLastPasswordReset());
```
leleueri, line 206: the lastPasswordReset is null for users that exist before the 3.16 release.
When the user is read from the database, maybe we should define the lastPasswordReset to user creation date to avoid this. You did it into the UserSevice but we have to do it also here (or at the Repository level)

gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/test/java/io/gravitee/am/identityprovider/mongo/authentication/MongoAuthenticationProviderTest.java
leleueri: Indeed, maybe we can do it at mvn level.

leleueri (changes requested):
The password policy may also be configured at the application level. (see picture here after)
Currently, the UI doesn't expose this option for the application password policy.

image

gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/test/java/io/gravitee/am/identityprovider/mongo/authentication/MongoAuthenticationProviderTest.java
TrikyGeeky: yes we can, I tried it and it works fine in mvn level when calling "mvn clean install" just add "-Dos.arch=i686_64"

remisultan (changes requested):
> 
> This is open to discussion I am wondering if we should not use the PasswordService and create a PasswordValidator for this.
> 
> We don't need to go through the whole chain of password validation but that would make sense it could be there and have a dedicated validation for it

Can we consider this ? 
Having all the password policies put at the same place could be easier

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
```diff
@@ -181,13 +181,38 @@ private boolean isIndefinitelyLocked(User user) {
...
         return Completable.complete();
     }
 
+    /**
+     * Check the user password status
+     * @param user Authenticated user
+     * @return True if the password has expired or False if not
+     */
+    private boolean checkAccountPasswordExpiry(User user, Client client) {
+        if ((client != null && client.getPasswordSettings() == null) && (domain != null && domain.getPasswordSettings() == null)) {
```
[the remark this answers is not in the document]
remisultan: Also I believe this cold be a Validator and implemented in PasswordService

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordServiceImpl.java
```diff
@@ -72,4 +76,26 @@ public void validate(String password, PasswordSettings passwordSettings, User us
...
+     * @param user Authenticated user
+     * @return True if the password has expired or False if not
+     */
+    public boolean checkAccountPasswordExpiry(User user, Client client, Domain domain) {
+        Optional<PasswordSettings> passwordSettings = PasswordSettings.getInstance(client, domain);
+
+        /** If the expiryDate is null or set to 0 so it's disabled */
+        if (passwordSettings.isEmpty() ||
+                (passwordSettings.get().getExpiryDuration() == null
+                                || (passwordSettings.get().getExpiryDuration() != null && passwordSettings.get().getExpiryDuration() <= 0))) {
```
leleueri, line 91: No need to test `passwordSettings.get().getExpiryDuration() != null` here as it is already tested just before.
If it is null, the test on the `value <= 0` will not be executed

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
```diff
@@ -181,13 +190,27 @@ private boolean isIndefinitelyLocked(User user) {
...
+    private Completable checkAccountStatus(User user, Client client) {
         if (!user.isEnabled()) {
             return Completable.error(new AccountDisabledException("Account is disabled for user " + user.getUsername()));
+        } else if (passwordService.checkAccountPasswordExpiry(user, client, domain)) {
+            return Completable.error( new AccountPasswordExpiredException("Account's password is expired "));
         }
         return Completable.complete();
     }
 
+    private boolean checkAccountPasswordExpiry(User user) {
```
leleueri, line 202: Now that we have the method in the PasswordService, this one is not useful anymore

leleueri (changes requested):

leleueri:
I addition of the two minor code changes, can you resolve the conflicts please?

leleueri (approved):

PR #1681, opened 2022-02-17
chore: remove slack mention from successful notification

leleueri:


tcompiegne (changes requested):
Why not removing the whole - slack/notify - section ?

PR #1694, opened 2022-02-22
fix: set the right value for internal attribute on a user when he is …

leleueri:
…provisionned through SCIM

fixes gravitee-io/issues#7177


farmborough (approved):
LGTM @leleueri 🤓

PR #1701, opened 2022-03-01
feat: validate user code

farmborough:
Fixes gravitee-io/issues#7153.

I've opted to provide the `user_code` via the `additionalProperties` map on the User class. This has the benefit of not requiring a change to the UI at this point.


TrikyGeeky:
Can you please remove this part of code from DynamicClientRegistrationServiceImpl in order for tests to not fail : 
```
if (request.getBackchannelUserCodeParameter() != null && request.getBackchannelUserCodeParameter().isPresent() && request.getBackchannelUserCodeParameter().get()) {
                return Single.error(new InvalidClientMetadataException("Unsupported backchannel_user_code_parameter"));
            }
```

PR #1708, opened 2022-03-07
feat(recovery-code): recovery code endpoint implemented

ashraf706:
A new endpoint 'api/recovery_code' added to fetch user recoery code.
Also removed recovery codes from user factor list.
Relates to: https://github.com/gravitee-io/issues/issues/7147


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/resources/AccountFactorsEndpointHandler.java
```diff
@@ -556,4 +582,22 @@ private int generateInitialMovingFactor(User endUser) {
...
+                .stream()
+                .map(enrolledFactor -> {
+                    EnrolledFactorSecurity security = enrolledFactor.getSecurity();
+                    if (security != null && security.getAdditionalData().containsKey(RECOVERY_CODE)) {
+                        security.setAdditionalData(null);
+                        security.setValue(null);
+                    }
+                    return enrolledFactor;
+                }).collect(Collectors.toList());
+    }
```
tcompiegne, line 602: Why not just apply a filter on the security type ? 

```
return user.getFactors()
                .stream()
                .filter(enrolledFactor -> !RECOVERY_CODE.equals(enrolledFactor.getSecurity().getType())
````

tcompiegne (changes requested):
General comment : 

Might be better to generate the recovery code in the POST /recovery_code endpoint.

PR #1716, opened 2022-03-11
feat: cypress for ciba user code

farmborough:
Fixes gravitee-io/issues#7239.

I've implemented a fairly basic test of CIBA using the accept-all notifier - essentially a recreation of the Postman tests. You can run Cypress by entering `npm test` in a terminal from the gravitee-am-cypress directory (see the readme I've created, also https://docs.cypress.io/guides/guides/command-line).  You will need AM running as well as the `CibaHttpNotifier`. Ideally these would be started by whatever testing framework we're running against, but I've left that for another issue.

I should mention that the ciba test is slightly flaky - occasionally the request for an access token will fail because AM hasn't yet updated the status of the original auth request, resulting in a 400 - `slow_down` response. This even happens with a CIBA auth request interval of 1. It's possible to tell a Cypress request to not fail a test on the status code and retry (up to four attempts) but in these circumstances, for some reason AM still hasn't updated the authorisation status after four attempts (response is still `authorization_pending`. I'm averse to adding a `wait` command in the test but wanted to see what others think first.


gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        email: "pat@test.com",
+        username: "pat@test.com",
+        password: "Password123",
+        additionalInformation: {
+            user_code: "1234"
+        }
+    };
+
+    before(() => {
+        requestAdminAccessToken().should(response => expect(response.status).to.eq(200))
```
leleueri, line 83: :thinking: is it possible to have something more fluent with cypress ?
I'm afraid that this "multi level" calls become hard to maintain.

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        password: clientSecret
+    },
+    form: true,
+    body: {
+        scope: "openid",
+        login_hint: loginHint,
+        user_code: userCode
+    },
+    failOnStatusCode: false
+})
```
ctschacher, line 61: I would suggest keeping these types of functions out of the spec file, especially as the functions mentioned here are general in nature and can be reused in other spec files.
In APIM we have a _commands_ folder directly under the _cypress_ folder to store those functions. Could be a good idea to keep it the same across products.

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
ctschacher: To avoid this kind of nesting you could also do something like this:
```
requestAdminAccessToken()
      .then( response => (accessToken = response.body.access_token))
      .then(() => createDomain(accessToken, "cibatest", "A test domain for CIBA"))
      .then((response) => { 
[...]
```

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+                    password: clientSecret
+                },
+                form: true,
+                body: {
+                    auth_req_id: auth_req_id,
+                    grant_type: "urn:openid:params:grant-type:ciba",
+                }
+            }).should(response => {
+                expect(response.status).to.eq(200);
+            });
```
ctschacher, line 158: Since the test is called 'should receive an access token' you probably want to verify that it actually receives a token (`response.body.access_token`), in addition to status code 200.

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
farmborough: These are specific to CIBA so I don't think it would be a good strategy to keep everything under one directory. I think these can be moved to a separate file but within the same `ciba` directory.

farmborough: That does apply to the management requests, however. Cheers.

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
farmborough: That's much better.

gravitee-am-cypress/cypress/integration/oidc/ciba/ciba.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    let domain;
+    let client;
+    let auth_req_id;
+    let cibaConfig;
+    const user: User = {
+        firstName: "Pat",
+        lastName: "Test",
+        email: "pat@test.com",
+        username: "pat@test.com",
+        password: "Password123",
```
patelami, line 76: Would it be easier to reuse this username and password in other tests if we declare it in fixtures?

PR #1727, opened 2022-03-15
fix: define list of expiry threshold in days to have in a single plac…

leleueri:
…e the total number of notifications for a given cert with the different thresholds

fixes gravitee-io/issues#7271


tcompiegne (changes requested):
gravitee.yml must be updated to reflect that change

tcompiegne (approved):

PR #1735, opened 2022-03-22
Issues/#5925 certificate avoid code duplication

farmborough:
Fixes https://github.com/gravitee-io/issues/issues/5925.

### DIFF
--- a/gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
+++ b/gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
@@ -32,8 +32,6 @@ import java.util.Set;
 public class JavaKeyStoreProvider extends AbstractCertificateProvider implements InitializingBean {
     @Autowired
     private JavaKeyStoreConfiguration configuration;
-    @Autowired
-    private CertificateMetadata certificateMetadata;
 
     @Override
     public void afterPropertiesSet() throws Exception {
--- a/gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
+++ b/gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
@@ -15,143 +15,39 @@
  */
 package io.gravitee.am.certificate.javakeystore.provider;
 
-import com.nimbusds.jose.jwk.JWKSet;
-import com.nimbusds.jose.jwk.KeyUse;
-import io.gravitee.am.certificate.api.CertificateFormat;
-import io.gravitee.am.certificate.api.CertificateKey;
+import io.gravitee.am.certificate.api.AbstractCertificateProvider;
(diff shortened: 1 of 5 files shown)


gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
```diff
@@ -15,258 +15,68 @@
...
-    private SignatureAlgorithm signature = SignatureAlgorithm.RS256;
-    private io.gravitee.am.certificate.api.Key certificateKey;
-    private List<CertificateKey> certificateKeys;
-
+public class JavaKeyStoreProvider extends AbstractCertificateProvider implements InitializingBean {
     @Autowired
     private JavaKeyStoreConfiguration configuration;
-
     @Autowired
     private CertificateMetadata certificateMetadata;
```
leleueri, line 36: Is it possible to considere this attribute to be part of the Abstract class as it is also used into the PKCS12Provider ?

leleueri (changes requested):

gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
```diff
@@ -15,258 +15,68 @@
...
-                        .get());
+    public CertificateMetadata certificateMetadata() {
+        return certificateMetadata;
     }
 
     @Override
-    public Single<List<CertificateKey>> publicKeys() {
-        return Single.just(certificateKeys);
+    protected String getStorepass() {
+        return configuration.getStorepass();
```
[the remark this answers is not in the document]
farmborough: I tried that first but ran this caused an error when the certificate json was deserialised. The problem is that the java certificate schema and PKCS12 schemas are identical, apart from the key property itself (it's `content` for PKCS12, 'jks` for Java). The way around that was to change the schemas as well, although that then hints that the two certificate implementations are redundant.

gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
farmborough: I can do that, so long as adding the Spring dependency to that module isn't a problem?

gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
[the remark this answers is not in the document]
leleueri: ok thanks you for the details

gravitee-am-certificate/gravitee-am-certificate-javakeystore/src/main/java/io/gravitee/am/certificate/javakeystore/provider/JavaKeyStoreProvider.java
leleueri: As Spring is used on REST API & GW, I don't think it is a problem IMHO.

leleueri (approved):

PR #1738, opened 2022-03-23
refactor: plugin handler and manager

remisultan:
relates to: https://github.com/gravitee-io/issues/issues/6642


remisultan:
We don't need to refactor on previous versions.
Let's target it for the latest

PR #1755, opened 2022-03-28
fix: invalidate user token on reset password

leleueri:
fixes gravitee-io/issues#7365


gravitee-am-ui/src/app/domain/components/account/account-settings.component.html
```diff
@@ -170,6 +170,15 @@ <h5>Reset password</h5>
           </mat-form-field>
         </div>
       </div>
+      <div fxLayout="column">
+        <mat-slide-toggle
+          (change)="enableInvalidateTokens($event)"
+          [checked]="isInvalidateTokensEnabled()" [disabled]="readonly">
+          Invalidate Tokens
+        </mat-slide-toggle>
+        <mat-hint style="font-size: 75%;">Invalidate user's tokens after a resert password.</mat-hint>
```
ashraf706, line 179: type in line 179. should be "reset" instead of "resert"

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -66,6 +68,8 @@
 
     private static final String DEFAULT_IDP_PREFIX = "default-idp-";
 
+    protected Logger logger = LoggerFactory.getLogger(this.getClass());
```
ashraf706, line 71: line 71: could be private instead of protected

ashraf706 (approved):

PR #1804, opened 2022-04-11
feat(users): add new REGISTRATION_CONFIRMATION flow

tcompiegne:
closes gravitee-io/issues#7470


gravitee-am-service/src/main/resources/flow/am-schema.json
```diff
@@ -27,7 +27,8 @@
           "login": "LOGIN",
           "consent": "CONSENT",
           "register": "REGISTER",
-          "reset_password": "RESET PASSWORD"
+          "reset_password": "RESET PASSWORD",
+          "registration_confirmation": "REGISTRATION CONFIRMATION"
```
remisultan, line 31: `registration_confirmation`

Needs also to be added to enums above.

remisultan (changes requested):

gravitee-am-service/src/main/resources/flow/am-schema.json
remisultan: Also the unit test on the number of flows don't pass

tcompiegne: FP

remisultan (approved):

PR #1805, opened 2022-04-11
fix: search user management api may provide duplicates

remisultan:
fixes: https://github.com/gravitee-io/issues/issues/7439

The duplicates may have been due to the search filter not being kept


remisultan:
> Couple of things noticed:
> 
> 1. Cannot select first page of the results once 2nd or any other page selected for multi page results. Showed Remi the findings.

Fixed

> 2. More importantly I couldn't recreate customer problem. Customer was using Postman with different "size" settings (as described in the issue). I failed to recreate the problem and suspecting that this fix may not solve customer problem.

Tried also with a SCIM and haven't been able to reproduce

farmborough:
@tcompiegne has pointed out that the original support ticket for this did not use the UI but was getting the issue from the management api. Also the affected version was 3.12.6.

tcompiegne (approved):

PR #1824, opened 2022-04-25
chore: updated snyk scan for application path

sursak123:
chore: updated snyk scan for application path using snyk cli

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -69,19 +69,9 @@ orbs:
   gravitee: gravitee-io/gravitee@1.0
   keeper: gravitee-io/keeper@0.6.2
   slack: circleci/slack@4.8.1
-  snyk: snyk/snyk@1.1.2
   am: gravitee-io/gravitee-am@1.0
 
 commands:
-  docker-scan:
-      steps:
-          - snyk/scan:
-              organization: gravitee.io
-              monitor-on-build: false
-              severity-threshold: critical
-              docker-image-name: ${IMAGE_TAG}
-              fail-on-issues: false  
-
   notify-snyk-result:
       steps:
           - slack/notify:
@@ -99,7 +89,7 @@ commands:
                          "type": "button",
                          "text": {
                            "type": "plain_text",
-                           "text": "Click here for scan details",
+                           "text": "Vulnerable paths found - Click here for scan details",
                            "emoji": true
                          },
                          "value": "click_me_123",
@@ -109,7 +99,7 @@ commands:
                       }
                     ]
                   }
-                event: always   
+                event: fail   
 jobs:
   keeper:
     docker:
@@ -367,6 +357,13 @@ jobs:
           at: /tmp
       - setup_remote_docker
       - run:
+          name: install-nodesource-nodejs-snyk
+          command: |     
+            curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
+            sudo apt install nodejs
+            sudo npm install snyk -g                          
+            snyk auth $SNYK_TOKEN           
+      - run:
           name: Compute Tag for Docker images
           command: |
             # Compute tag with branch name, commmit SHA and timestamp. Then slugify it to be sure it can be handled by Docker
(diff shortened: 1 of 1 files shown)


.circleci/workflows.yml
```diff
@@ -377,26 +374,16 @@ jobs:
...
-            ALLOCATION_ID=$(echo "${FAILED_STEP}" | jq -r '.allocation_id')
-            STEP=$(echo "${FAILED_STEP}" | jq -r '.step')
-            curl -o slack_response.txt "https://circleci.com/api/v1.1/project/${API_URL}/output/${STEP}/0?file=true&allocation-id=${ALLOCATION_ID}&circle-token=$CIRCLE_API_TOKEN"
-            sed -i 's/\r$//g' slack_response.txt
-            SLACK_MESSAGE=$(cat slack_response.txt | tail -n 15 | sed '$!s/$/\\n/' | tr -d '\n')
-            echo "export SLACK_MESSAGE='${SLACK_MESSAGE}'" >> $BASH_ENV     
-      - notify-snyk-result 
+            SLACK_MESSAGE1="--AM Docker Images-- ${IMAGE_TAG},"
+            echo "export SLACK_MESSAGE='${SLACK_MESSAGE1}'" >> $BASH_ENV   
+            snyk container test ${IMAGE_TAG} --app-vulns
```
leleueri, line 386: with this change, does it mean that the pipeline will fail here if there are some vuln issues ?
int the previous implementation using the snyk orb, we go to the next step in case of failure https://github.com/gravitee-io/gravitee-access-management/pull/1824/files#diff-7bf091450b804ea25dc3d11ac28c50202d19178f7b53f54a02ad59c946afa6a7L83

.circleci/workflows.yml
```diff
@@ -409,22 +396,9 @@ jobs:
...
-            STEP=$(echo "${FAILED_STEP}" | jq -r '.step')
-            STEP=$(echo $STEP | cut -d' ' -f2)
-            curl -o slack_response.txt "https://circleci.com/api/v1.1/project/${API_URL}/output/${STEP}/0?file=true&allocation-id=${ALLOCATION_ID}&circle-token=$CIRCLE_API_TOKEN"
-            sed -i 's/\r$//g' slack_response.txt
-            SLACK_MESSAGE=$(cat slack_response.txt | tail -n 15 | sed '$!s/$/\\n/' | tr -d '\n')
-            echo "export SLACK_MESSAGE='${SLACK_MESSAGE}'" >> $BASH_ENV     
-      - notify-snyk-result
+            SLACK_MESSAGE2="${SLACK_MESSAGE} ${IMAGE_TAG}, "
+            echo "export SLACK_MESSAGE='${SLACK_MESSAGE2}'" >> $BASH_ENV
+            snyk container test ${IMAGE_TAG} --app-vulns
```
leleueri, line 401: with this change, does it mean that the pipeline will fail here if there are some vuln issues ?
int the previous implementation using the snyk orb, we go to the next step in case of failure https://github.com/gravitee-io/gravitee-access-management/pull/1824/files#diff-7bf091450b804ea25dc3d11ac28c50202d19178f7b53f54a02ad59c946afa6a7L83

.circleci/workflows.yml
```diff
@@ -439,22 +413,10 @@ jobs:
...
-            STEP=$(echo "${FAILED_STEP}" | jq -r '.step')
-            STEP=$(echo $STEP | cut -d' ' -f3)
-            curl -o slack_response.txt "https://circleci.com/api/v1.1/project/${API_URL}/output/${STEP}/0?file=true&allocation-id=${ALLOCATION_ID}&circle-token=$CIRCLE_API_TOKEN"
-            sed -i 's/\r$//g' slack_response.txt
-            SLACK_MESSAGE=$(cat slack_response.txt | tail -n 15 | sed '$!s/$/\\n/' | tr -d '\n')
-            echo "export SLACK_MESSAGE='${SLACK_MESSAGE}'" >> $BASH_ENV      
-      - notify-snyk-result
+            SLACK_MESSAGE3="${SLACK_MESSAGE}${IMAGE_TAG} images are tested for application vulnerabilities using snyk"
+            echo "export SLACK_MESSAGE='${SLACK_MESSAGE3}'" >> $BASH_ENV
+            snyk container test ${IMAGE_TAG} --app-vulns
```
leleueri, line 418: with this change, does it mean that the pipeline will fail here if there are some vuln issues ?
int the previous implementation using the snyk orb, we go to the next step in case of failure https://github.com/gravitee-io/gravitee-access-management/pull/1824/files#diff-7bf091450b804ea25dc3d11ac28c50202d19178f7b53f54a02ad59c946afa6a7L83

leleueri (commented):
Short question about the behaviour when vuln is detected.
I guess that if a vuln is detected in one of the images, the next step of the pipeline will not be executed... can you confirm that point ?

if any, in that case this is a issue IMHO as the PR will not be deployed on our nightly env and QA will not be able to test the feature.

.circleci/workflows.yml
sursak123: Pipeline will not faile. it will always runs and just notifies when there is vulnerability.

.circleci/workflows.yml
sursak123: This will go to next step even if image test fails.  This is using snyk cli as only cli has app vulnerability feature.  So in the step i added when : always which will run even previous step fails.

leleueri (approved):

PR #1832, opened 2022-04-27
fix: add failure handler for ResetPassword endpoint

leleueri:
fixes gravitee-io/issues#7574


tcompiegne (changes requested):
I think we can re-use the ErrorHandler here ... to avoid duplication : 

```
rootRouter.route(PATH_RESET_PASSWORD)
  .failureHandler(new ErrorHandler(PATH_RESET_PASSWORD));
```

Maybe one day we should rename to RedirectErrorHandler ...

tcompiegne (approved):

PR #1839, opened 2022-04-28
chore: rename http resource to http resource factor

ashraf706:
Updated UI files to reflect the name change

### DIFF
--- a/gravitee-am-ui/src/app/domain/settings/resources/resource/resource.component.ts
+++ b/gravitee-am-ui/src/app/domain/settings/resources/resource/resource.component.ts
@@ -89,10 +89,10 @@ export class ResourceComponent implements OnInit {
       });
   }
 
-  isHttpResource() {
-    if (this.resourceSchema) {
-      const resourceId = this.resourceSchema.id;
-      return resourceId ? resourceId.includes('HttpResourceConfiguration') : false;
+  isHttpFactorResource() {
+    if (this.resource) {
+      const type = this.resource.type;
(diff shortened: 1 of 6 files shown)


tcompiegne (changes requested):
I'm tempted to set the name of the plugin to : `http-factor-am-resource` to keep consistency with the others plugins. Which means we have to change the id value of the plugin.

gravitee-am-ui/src/app/domain/settings/resources/resource/resource.component.ts
```diff
@@ -89,10 +89,10 @@ export class ResourceComponent implements OnInit {
       });
   }
 
-  isHttpResource() {
+  isHttpFactorResource() {
     if (this.resourceSchema) {
       const resourceId = this.resourceSchema.id;
-      return resourceId ? resourceId.includes('HttpResourceConfiguration') : false;
+      return resourceId ? resourceId.includes('HttpFactorResourceConfiguration') : false;
```
tcompiegne, line 95: Why not test the resource `type` ? like the previous example : 

```
 return type ? type.includes('http-factor-am-resource') : false;
```

tcompiegne (changes requested):

gravitee-am-ui/src/app/domain/settings/resources/resource/resource.component.ts
ashraf706: changed

tcompiegne (approved):

PR #1841, opened 2022-05-02
chore: sonarcloud on frontend

remisultan:
relates to: https://github.com/gravitee-io/issues/issues/7590

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -3,11 +3,11 @@ version: 2.1
 parameters:
   gio_action:
     type: enum
-    enum: [release, publish_rpms, pull_requests]
+    enum: [ release, publish_rpms, pull_requests ]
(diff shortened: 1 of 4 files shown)


.circleci/workflows.yml
```diff
@@ -562,6 +562,27 @@ jobs:
           root: .
           paths:
             - ./*
+  fe_version_build:
```
leleueri, line 565: can we replace `fe` by `front` or `ui` ?
It takes me few minutes to translate `fe` in `front end` :)

leleueri (changes requested):

.circleci/workflows.yml
remisultan: Done 🚀

leleueri (approved):

PR #1844, opened 2022-05-03
chore: create pull_request_template

omatthewsgravitee:


remisultan (changes requested):
Maybe we could have something a bit more templaty like the issues ? 🤔 
https://github.com/gravitee-io/issues/issues/7563
Could be also useful for the community ?

omatthewsgravitee:
I didn't realise you could just add .md to the file. Is this any better?

remisultan (approved):

PR #1845, opened 2022-05-03
fix(audits): use the user technical ID field when fetching user audit logs

tcompiegne:
fixes gravitee-io/issues#7619


tcompiegne:
@leleueri I will need your help regarding SQL index names.

leleueri (changes requested):
Code is OK but some audits are missing into the user history (LOGOUT for example)
Probably due to the fact that for logout, only the alternativeId is field using the user internalId.

also, not related to this issue but the actor link is broken for logout event:
image

leleueri (approved):

PR #1851, opened 2022-05-12
feat: convert Cypress tests to Jest

farmborough:
## :id: 
fixes gravitee-io/issues#7592

## :pencil2: 

This work transforms the commands from gravitee-io/issues#7592 that used Cypress to instead use Jest and a library called **Supertest**. I meant to completely remove Cypress but I feel that we still need it for pure UI testing, which Jest is not designed for. The main differences between Cypress and Jest that I've found are that Cypress has a more comprehensive API and is more user friendly, but Jest offers greater simplicity and is easier and quick to run that Cypress. For example, it's now straightforward to run test from a run configuration in Intellij.

Promises can still be used with Jest and Supertest, but I found using ES6's **async/await** syntax to be more reliable.

I've also had to rejig the test environment configuration that Cypress provided pretty much out of the box, but I think the result is probably cleaner and simpler. We now use Node (which Jest runs on) to store and retrieve test environment variables.

## :books: 
I've updated the README file.


gravitee-am-test/api/commands/management/application-management-commands.ts
```diff
@@ -0,0 +1,52 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import {getDomainManagerUrl} from "./utils";
+
+const request = require('supertest');
+
+export const createApplication = (domainId, accessToken, body, status: number = 201) =>
+    request(getApplicationUrl(domainId, null))
```
farmborough, line 21: Annoyingly, it seems you have to call the http method function (get, post etc) directly after `request` which makes it unfeasible to have a common `request` used by all these other functions (as was the case with Cypress).

gravitee-am-test/api/config/ci.config.js
```diff
@@ -13,7 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-export interface BasicAuthentication {
-  username: string;
-  password: string;
-}
+module.exports = {
```
farmborough, line 16: I've reduced the config files to two in order to keep things simple. There needs to be two files per config for Jest: this one is passed as the `--config=` CLI argument and it points to another file `ci.setup.js` which contains the actual test environment variables.

remisultan (commented):
I followed the README and wasn't able to run the tests correctly.

image

npm version: 6.14.11
node version: v14.15.5

remisultan (approved):

gravitee-am-test/README.adoc
```diff
@@ -0,0 +1,46 @@
...
+The AM integration tests run on Jest using Supertest to make calls to the AM APIs.
+To run the tests from a terminal you need to be in the `gravitee-am-test` directory.
+
+==== Build
+
+Execute `npm install`. You usually only need to do this once, i.e. after cloning the AM repo or when dependencies have changed.
+
+==== Run tests with default config
+
+Executing `npm test` or `jest` will run all the tests. To run individual tests use `npm run` | `jest -t`, followed by the name of the spec file.
```
ashraf706, line 12: instead of `jest` perhaps the use has to run 'jest --config=api/config/dev.config.js' after executing 'npm install jest'.

leleueri (approved):

PR #1862, opened 2022-05-17
fix: update user firstname, lastname & displayname on user Update and…

leleueri:
… Connection

fixes gravitee-io/issues#7531


farmborough (approved):

remisultan (approved):
Tested and works like a charm

LGTM

PR #1863, opened 2022-05-17
fix: update user data from IDP response

farmborough:
## :id:  
fixes gravitee-io/issues#7647

This fixes a bug where `UserAuthenticationService` was not parsing IDP responses for key fields such as email, username etc., and was only updating a user's additional properties. With this fix, a user profile will be updated whenever corresponding values are updated at the IDP.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
@@ -256,6 +256,7 @@ public class UserAuthenticationManagerImpl implements UserAuthenticationManager 
                     Map<String, Object> additionalInformation = user.getAdditionalInformation() == null ? new HashMap<>() : new HashMap<>(user.getAdditionalInformation());
                     additionalInformation.put("source", authProvider);
                     additionalInformation.put(Parameters.CLIENT_ID, client.getId());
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
```diff
@@ -256,7 +256,6 @@ private Maybe<UserAuthentication> loadUserByUsername0(Client client, Authenticat
                     Map<String, Object> additionalInformation = user.getAdditionalInformation() == null ? new HashMap<>() : new HashMap<>(user.getAdditionalInformation());
                     additionalInformation.put("source", authProvider);
                     additionalInformation.put(Parameters.CLIENT_ID, client.getId());
-                    ((DefaultUser) user).setAdditionalInformation(additionalInformation);
```
leleueri, line 259: removing this line may lead to regression as we test in line 256 if the additionalINformation is null. If so, we create a new HashMap but we do not set this map into the User object

leleueri (changes requested):
In additon of the comment about the additionalInformation, I wondering if this PR is a duplicate of https://github.com/gravitee-io/issues/issues/7531

farmborough:
> gravitee-io/issues#7531

I think it's a slightly different fix as this issue deals with updates originating from the IDP.

leleueri:
After discussing with @farmborough same issue as https://github.com/gravitee-io/issues/issues/7531.
As 7531 will be merge in 3.15.x we close this one.

farmborough:
After chatting with @leleueri this does indeed appear to be a duplicate and is already fixed by https://github.com/gravitee-io/issues/issues/7531.

PR #1866, opened 2022-05-18
Issues/#7658 add jest to the ci pipeline

farmborough:
## :id: 

Fixes https://github.com/gravitee-io/issues/issues/7658

## :pencil2: 

Circle CI workflows added for Jest running against AM with Mongo and Postgresql.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -232,6 +232,127 @@ jobs:
             - ./gravitee-am-management-api-standalone-*.zip
             - ./gravitee-am-webui-*.zip
 
+  run_jest_tests_mongo:
+    docker:
+      - image: cimg/openjdk:17.0.1-node
+      - image: mongo:4.2
+    resource_class: small
+    steps:
+      - attach_workspace:
+          at: /tmp/workspace
+      - checkout
+      - run:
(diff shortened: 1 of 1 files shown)


.circleci/workflows.yml
```diff
@@ -232,6 +232,138 @@ jobs:
...
+            npm --prefix gravitee-am-test run ci
+      - slack/notify:
+          branch_pattern: ".+"
+          event: fail
+          mentions: "@am_team"
+          template: basic_fail_1
+      - slack/notify:
+          branch_pattern: ".+"
+          event: pass
+          mentions: "@am_team"
```
leleueri, line 282: Maybe we should remove the @am_team notification in case of success, wdyt?
Having the success message into the chanel should be enough. We want to be notified only for failure. (if you are agree, can you also do it for postman executions? )

leleueri (commented):
I leave a comment about slack notifications.
If you are not agree, you can merge this PR.

leleueri (approved):

.circleci/workflows.yml
farmborough: Completely agree 👍

PR #1874, opened 2022-05-23
fix(gateway): append client_id in URL generated for email template

leleueri:
fixes gravitee-io/issues#7734

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
@@ -203,13 +203,16 @@ public class EmailServiceImpl implements EmailService {
 
         Map<String, Object> params = new HashMap<>();
         params.put("user", new UserProperties(user));
-        params.put("url", redirectUrl);
         params.put("token", token);
         params.put("expireAfterSeconds", expiresAfter);
         params.put("domain", new DomainProperties(domain));
+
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
```diff
@@ -203,14 +203,17 @@ private Map<String, Object> prepareEmailParams(User user, Client client, Integer
...
         Map<String, Object> params = new HashMap<>();
         params.put("user", new UserProperties(user));
-        params.put("url", redirectUrl);
         params.put("token", token);
         params.put("expireAfterSeconds", expiresAfter);
         params.put("domain", new DomainProperties(domain));
+
         if (client != null) {
             params.put("client", new ClientProperties(client));
+            redirectUrl += "&client_id=" + client.getClientId();
```
remisultan, line 212: Should we check if the Url ends with `?` or not ?

leleueri: I would say no because we append the `?token` at line 202.
https://github.com/gravitee-io/gravitee-access-management/pull/1874/files#diff-d54c661c53c63e61bf2db7cc46349b856551ed8dc56fb009b2fd583cd20a2d27R202

wdyt ?

remisultan: fine be me then 👍

PR #1877, opened 2022-05-24
fix: addition of gateway setting to manage nonce value for inline js …

leleueri:
…script

fixes gravitee-io/issues#7724

This PR introduce a new setting for the CSP Headetrs managed by the gateway in order to generate a nonce value for each inline js script embedded into HTML templates.

Into the gravitee.yaml, the configuration parameter is available in the CSP section :
```
http:
  csp:
    script-inline-nonce: true|false
```

Into the values.yaml of the AM helm chart, the configuration parameter is available in the CSP section for the gateway settings :
```
gateway:
  http:
    csp:
      script-inline-nonce: true|false
```

## :memo: Test scenarios 

Enable this value and define directive in order to allow other resources. Then play with an Application to test all template pages.

Here is the directive I had to defined. 

```
      - "script-src http://localhost:8092/ https://cdn.jsdelivr.net/npm/@fingerprintjs/fingerprintjs@3/dist/fp.min.js;"
      - "style-src http://localhost:8092/ 'unsafe-inline'"
      - "img-src * data: blob: 'unsafe-inline';"
```

:warning:  Pay attention to the MFA enrollment in particular as there are more changes on this page

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/csp/CspHandlerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/csp/CspHandlerImpl.java
@@ -16,13 +16,16 @@
 
 package io.gravitee.am.gateway.handler.common.vertx.web.handler.impl.csp;
 
+import io.gravitee.am.common.utils.SecureRandomString;
 import io.gravitee.am.gateway.handler.common.vertx.web.handler.CSPHandler;
 import io.vertx.reactivex.ext.web.RoutingContext;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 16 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/csp/CspHandlerImpl.java
```diff
@@ -16,13 +16,22 @@
...
 
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Random;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
```
ashraf706, line 32: few unused import such as StandardCharsets, Base64, Random, pattern etc.

ashraf706 (approved):

PR #1880, opened 2022-05-25
fix: provide option to apply UserMapper on HTTPUserProvider

leleueri:
fixes gravitee-io/issues#7530

The purpose of this issue is to add an option to apply the UserMapper configuration for each UserProvider methods (for HTTP provider only)

image

This can help if the HTTP backend can't change the output to match the one provided by the AuthenticationProvider endpoint.

:warning: If EL are used for the Mapper, only profile is available into the EL context.


remisultan (approved):
LGTM

PR #1892, opened 2022-06-01
Issues/#7648 common bundle

leleueri:
fixes gravitee-io/issues#7648

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,6 +24,7 @@
         <groupId>io.gravitee</groupId>
         <artifactId>gravitee-parent</artifactId>
         <version>20.0</version>
+        <relativePath></relativePath>
     </parent>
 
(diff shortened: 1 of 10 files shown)


sebdevaux (approved):

pom.xml
```diff
@@ -133,6 +136,11 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
```
tcompiegne, line 139: Why we need this dependency ? beware typo : ${junit-jupiter-engine.version}} (double brackets)

tcompiegne (changes requested):

pom.xml
leleueri: We don't, it is something that I try in order to be able to generate idea:module with maven, I will remove it.

tcompiegne (approved):

PR #1924, opened 2022-06-21
fix: update accountNonLocked attribute after successful login

leleueri:
fixes gravitee-io/issues#7831

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
@@ -202,6 +202,7 @@ public class UserAuthenticationServiceImpl implements UserAuthenticationService 
         if (afterAuthentication) {
             existingUser.setLoggedAt(new Date());
             existingUser.setLoginsCount(existingUser.getLoginsCount() + 1);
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
```diff
@@ -202,6 +202,7 @@ private Single<User> update(User existingUser, io.gravitee.am.identityprovider.a
         if (afterAuthentication) {
             existingUser.setLoggedAt(new Date());
             existingUser.setLoginsCount(existingUser.getLoginsCount() + 1);
+            existingUser.setAccountNonLocked(true);
```
remisultan, line 205: Should we also clear out the `accountLockedUntil` also ?

remisultan: In the management-api when we allow to unlock via API we wipe the data.
https://github.com/gravitee-io/gravitee-access-management/blob/1f5ccbc384cd3a6d96527c6dab7ca9ef4f1e97cd/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java#L377

But this is for intended `lock` and `unlock`. Keeping the values while unlocking  at authentication would mean that the latest unlocking was not manual but the user recovered its own account.

Ok for me 👌

PR #1944, opened 2022-06-27
fix(reset-password): do not insert if force registration is disabled

tcompiegne:
closes gravitee-io/issues#7911


remisultan (approved):
LGTM

PR #1947, opened 2022-06-27
fix: do not reject the FactorPlugin schema if plugin isn't installed

leleueri:
fixes gravitee-io/issues#7881


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/plugins/FactorPluginResource.java
```diff
@@ -70,10 +69,9 @@ public void getSchema(@PathParam("factor") String factorId,
 
         // Check that the authenticator exists
         factorPluginService.findById(factorId)
```
ashraf706, line 71: perhaps `factorPluginService.findById(factorId)` not required anymore as in the `flatmap` we do not consider the returned value. Tests are running fine without calling `findById`

ashraf706 (approved):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/plugins/FactorPluginResource.java
leleueri: fp

PR #1948, opened 2022-06-27
fix: more ip and user agent consent check in handlers

remisultan:
https://github.com/gravitee-io/issues/issues/7919

This PR makes more check of IP and user agent in handlers


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/provider/impl/UserAuthProviderImpl.java
```diff
@@ -37,6 +37,8 @@
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import static io.gravitee.am.common.utils.ConstantKeys.DEVICE_ID;
```
leleueri, line 40: Still useful ?

leleueri (changes requested):

leleueri (approved):

leleueri:
As discuss with @remisultan we will open an issue to refactor these commons peaces of code

PR #1952, opened 2022-06-29
fix: always reload and refresh auths and resolvers on application page

remisultan:

https://github.com/gravitee-io/issues/issues/7929

This PR fixes the application not being refreshed when saved. 

### Idp settings

1. Change the application idp selection
2. Save the application 
3. Change to another tab
4. Come back to the app idp selection list
5. You should notice that the change app idp settings stays the same

### Redirect Uri

1. Go to the Application
2. In the Overview tab, notice the initiate the login flow redirect_uri 
3. Change the redirect_uri in Application > Settings
4. Come back to the Overview tab
5. Notice the redirect_uri has changed with the new settings

### DIFF
--- a/gravitee-am-ui/src/app/app-routing.module.ts
+++ b/gravitee-am-ui/src/app/app-routing.module.ts
@@ -850,7 +850,7 @@ export const routes: Routes = [
                           application: ApplicationResolver,
                           permissions: ApplicationPermissionsResolver
                         },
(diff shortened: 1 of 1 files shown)


tcompiegne (changes requested):
I wonder if it's not too overkill ... we could potentially face some latency issues as we refresh data every time now ...

remisultan:
I agree too, my second thought was basically to load it only at `application.component` and inject it in subcomponents.

PR #1954, opened 2022-06-29
 fix: inaccurate factor choice

ashraf706:
This fix considers client MFAs and user registered MFAs for showing alternative link in MFA challenge step.

## How to test
The issue 7928 contains steps to reproduce the problem. This fix makes sure that

1. Alternative MFAa links won't show if client has maximum 1 enabled MFA or or user has maximum 1 enrolled factor.
2. This fix also make sure that options in alternative MFAs are enrolled by the user and also enabled in the client.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeEndpoint.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeEndpoint.java
@@ -135,7 +135,7 @@ public class MFAChallengeEndpoint extends AbstractEndpoint implements Handler<Ro
             if (deviceId != null) {
                 routingContext.put(ConstantKeys.DEVICE_ID, deviceId);
             }
-            if (endUser.getFactors() != null && endUser.getFactors().size() > 1) {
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeEndpoint.java
```diff
@@ -442,4 +442,23 @@ private void saveDeviceAndRedirect(RoutingContext routingContext, Client client,
...
+        else if(client.getFactors() == null || client.getFactors().size() <= 1){
+            return false;
+        }
+        else {
+            final Set<String> clientFactorIds = client.getFactors();
+            final List<EnrolledFactor> activeEnrolledFactors = endUser.getFactors()
+                    .stream()
+                    .filter(enrolledFactor -> factorManager.get(enrolledFactor.getFactorId()) != null)
+                    .filter(enrolledFactor -> clientFactorIds.contains(enrolledFactor.getFactorId()))
+                    .collect(Collectors.toList());
```
remisultan, line 459: Maybe the `getEnabledFactors` could serve its purpose in a utility class ?

ashraf706: discussed with @remisultan and decided to leave it as it is for now.

PR #1968, opened 2022-07-05
fix: provide new option to force S256 as challenge method in PKCE

leleueri:
fixes gravitee-io/issues#7965

# How to test

* Create 4 apps (one for each type)
* Check for each app if the "Force S256 Challenge" toggle is active (should be if the Force PKCE is active)
image

## Test 1
* try to follow authentication steps described in the "overview" page of the tested app

## Test 2
* For SPA & Native App, try to request the` /authorization `endpoint using "`plain`" as value for the `code_challenge_method` (authent should fail with "`invalid code_challenge_method`" error)

## Test 3
* For SPA & Native App, switch off the "Force S256 Challenge" toggle and try to request the` /authorization `endpoint using "`plain`" as value for the `code_challenge_method` (should be OK)


tcompiegne (changes requested):
Can we change the `forceS256Challenge` name by `forceS256CodeChallengeMethod` ? to be consistent with the RFC naming convention ? 🙏

Thanks

tcompiegne (approved):

PR #1969, opened 2022-07-06
fix: do not log error when it is not necessary when user authenticates

remisultan:
fixes: https://github.com/gravitee-io/issues/issues/7984


leleueri (commented):
Do we have to remove the log or should we wrappe the call to `getInternalAuthenticationServiceException` in a supplier lambda method like it has been done in 3.18.x ?

tcompiegne:
It has been already defer in this commit : https://github.com/gravitee-io/gravitee-access-management/commit/52b5e6f896470fa58726372817b0eed98eced415

I think we can remove the log as it should be logged afte.

PR #1985, opened 2022-07-11
chore: implement mechanism to load i18n messages from filesystem and …

leleueri:
…use locale based on the Accept-Language header

fixes gravitee-io/issues#8038

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/LocaleHandler.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/LocaleHandler.java
@@ -1,0 +1,58 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.root.resources.handler;
+
+import io.gravitee.am.gateway.handler.vertx.view.thymeleaf.GraviteeMessageResolver;
+import io.vertx.core.Handler;
+import io.vertx.reactivex.ext.web.RoutingContext;
+
+import java.util.Locale;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class LocaleHandler implements Handler<RoutingContext> {
+
+    private final GraviteeMessageResolver messageResolver;
+
+    public LocaleHandler(GraviteeMessageResolver messageResolver) {
+        this.messageResolver = messageResolver;
+    }
+
+    @Override
+    public void handle(RoutingContext routingContext) {
+        // first evaluate AcceptedLanguage raw values
+        for(var language : routingContext.acceptableLanguages()) {
+            if (messageResolver.isSupported(new Locale(language.value()))) {
+                routingContext.put("lang", language.value());
+                routingContext.next();
+                return;
+            }
+        }
+
+        // if no raw values match, evaluate only the language code
+        for(var language : routingContext.acceptableLanguages()) {
+            if (messageResolver.isSupported(new Locale(language.tag()))) {
+                routingContext.put("lang", language.tag());
+                routingContext.next();
+                return;
+            }
+        }
+
+        routingContext.next();
+    }
+}
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/GraviteeMessageResolver.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/GraviteeMessageResolver.java
@@ -1,0 +1,103 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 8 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/LocaleHandler.java
```diff
@@ -0,0 +1,56 @@
...
+            }
+        }
+
+        for(var language : routingContext.acceptableLanguages()) {
+            if (messageResolver.isSupported(new Locale(language.tag()), true)) {
+                routingContext.put("lang", language.tag());
+                routingContext.next();
+                return;
+            }
+        }
```
tcompiegne, line 52: Why we need a double loop ? 

Why not using the routingContext.preferredLanguage() method ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/GraviteeMessageResolver.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        // create empty default properties to avoid NPE during message lookup
+        this.propertiesMap.putIfAbsent(DEFAULT_LOCALE, new Properties());
+
+        this.setOrder(0);
+    }
+
+    public boolean isSupported(Locale locale, boolean lax) {
+        return (!lax && this.propertiesMap.containsKey(locale.toString())) || (lax && this.propertiesMap.containsKey(locale.getLanguage()));
```
tcompiegne, line 83: I don't understand the purpose of the lax boolean.

tcompiegne (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/GraviteeMessageResolver.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.Optional.ofNullable;
+
+/**
+ * @author Eric LELEU (eric.leleu at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+public class GraviteeMessageResolver extends AbstractMessageResolver implements IMessageResolver {
```
ashraf706, line 43: `AbstractMessageResolver` already implements  `IMessageResolver`. Shouldn't it be ok if we remove `implements IMessageResolver`?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/LocaleHandler.java
leleueri: The Accept-Languages header value may contain multiple values with different priorities.
The preferredLanguage returns only the one with highest priority but we do not have guaranty that this preferredLanguage is managed by our dictionnary, so I go through the acceptableLanguages (that are shorted by priority) to get the most accurante language.

About the double loop, its purpose is to manage the language with country code (fr-FR or en-GB). 
for ex, if the Accept-Language contains `fr-FR` but not `fr` and if on AM the admin user only define `fr` as language, we will not be able to find a matching translation. As we do not have guaranty that the value 'fr-FR' will have the same priority as 'fr', I prefere doing a strict matching first and then fallback to a 'lax' matching

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/GraviteeMessageResolver.java
leleueri: Indeed, it is useless now as I call this method with the `language.tag()` to get only the language code and not the raw value (fr-FR) provided by `language.value()`

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/ThymeleafConfiguration.java
```diff
@@ -35,8 +40,11 @@ public class ThymeleafConfiguration {
     @Autowired
     private Vertx vertx;
 
+    @Value("${template.i18n:#{systemProperties['gravitee.home']}/templates/i18n}")
```
tcompiegne, line 43: We have already a `templates` section in the gravitee.yml file : 

```
# Mail templates
#templates:
#  path: ${gravitee.home}/templates
```

Not sure if we have to re-think that part 🤔

tcompiegne (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/view/thymeleaf/ThymeleafConfiguration.java
leleueri: OK, so even the form templates will be moved here. That make sens, specially for user that want to update default template . It will be possible for them to define a new custom default template without having to override templates for each domain. :+1:

leleueri: maybe forms instead of pages to be consistent with the UI :thinking:

tcompiegne: yes

leleueri: this refactor will be handled by https://github.com/gravitee-io/issues/issues/8114
I will FP the change to use the templates.path settings with i18n as subdirectory

tcompiegne (approved):

PR #1988, opened 2022-07-12
feat: define REST interface to manage translations at domain level

leleueri:
closes gravitee-io/issues#8067

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/TranslationsResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/TranslationsResource.java
@@ -1,0 +1,15 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/TranslationsResource.java
```diff
@@ -0,0 +1,218 @@
...
+    @ApiResponses({
+            @ApiResponse(code = 200, message = "List of languages for a security domain", response = Language.class, responseContainer = "List"),
+            @ApiResponse(code = 500, message = "Internal server error")})
+    public void list(
+            @PathParam("organizationId") String organizationId,
+            @PathParam("environmentId") String environmentId,
+            @PathParam("domain") String domain,
+            @Suspended final AsyncResponse response) {
+
+        //checkAnyPermission(organizationId, environmentId, domain, Permission.DOMAIN_TRANSLATIONS, Acl.LIST);
```
farmborough, line 64: Presumably these are commented out as placeholders?

leleueri: yes because the story is only about defining the REST API and the Data model.

gravitee-am-model/src/main/java/io/gravitee/am/model/Domain.java
```diff
@@ -396,6 +399,14 @@ public void setSaml(SAMLSettings saml) {
         this.saml = saml;
     }
 
+    public List<Language> getLanguages() {
+        return languages;
```
farmborough, line 403: Can this default to an empty list? Either `Collections.emptyList()` or List.of()`?

farmborough (approved):

leleueri:
> Maybe we can rely on a kind of "Dictionary" to store the languages :

Like this idea, it is a more REST approch.

PR #1989, opened 2022-07-13
fix(policies): add default value for the enroll mfa policy to enable saving

tcompiegne:
fixes gravitee-io/issues#8112


leleueri (commented):
Code is good and allow me to save when only one Factor is configured but I'm also able to save if I do not select a factor. Don't know if it is something that we can improve :thinking: 

also if the RecoveryCode factor is configured, it is proposed into the list. Is it expected ?

leleueri (approved):

leleueri:
> Code is good and allow me to save when only one Factor is configured but I'm also able to save if I do not select a factor. Don't know if it is something that we can improve thinking
> 
> also if the RecoveryCode factor is configured, it is proposed into the list. Is it expected ?

as discussed, we will open new issue for the behaviour described in this comment

PR #1992, opened 2022-07-14
fix: create a MetricsProvider for the Gateway to manage all metrics i…

leleueri:
…n a singleton. This avoid inaccurate data due to the multiple ClassLoaders used in AM

fixes gravitee-io/issues#8116


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
```diff
@@ -39,6 +39,7 @@
 import io.gravitee.am.model.idp.ApplicationIdentityProvider;
 import io.gravitee.am.model.oidc.Client;
 import io.gravitee.am.monitoring.metrics.CounterHelper;
+import io.gravitee.am.monitoring.provider.GatewayMetricProvider;
 import io.gravitee.am.repository.management.api.search.LoginAttemptCriteria;
 import io.gravitee.am.service.LoginAttemptService;
 import io.gravitee.am.service.PasswordService;
```
ashraf706, line 45: Some unused import from previous commit you worked in June. Cleaning would be a plus as you are already editing this file.

ashraf706 (approved):

PR #2018, opened 2022-08-01
fix: do not override the preferredLangague field by the locale claim

leleueri:
fixes gravitee-io/issues#8268

### How to test

You can use the acceptance tests explain here https://github.com/gravitee-io/issues/issues/8039
The case where the locale claims is updated should works

### DIFF
--- a/gravitee-am-model/src/main/java/io/gravitee/am/model/User.java
+++ b/gravitee-am-model/src/main/java/io/gravitee/am/model/User.java
@@ -337,12 +337,6 @@ public class User implements IUser {
     }
 
     public String getPreferredLanguage() {
-        if (preferredLanguage == null) {
-            // fall back to OIDC standard claims
-            if (getAdditionalInformation() != null && getAdditionalInformation().get(StandardClaims.LOCALE) != null) {
-                preferredLanguage = (String) getAdditionalInformation().get(StandardClaims.LOCALE);
-            }
-        }
         return preferredLanguage;
     }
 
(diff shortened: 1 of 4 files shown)


tcompiegne (changes requested):
Not a huge fan to have a parameter in the getter method.

Seems to only affected the SCIM part of AM, maybe do a change like this in the mapper : 

 ```
scimUser.setPreferredLanguage(get(additionalInformation, StandardClaims.LOCALE) == null ? user.getPreferredLanguage() : (!get(additionalInformation, StandardClaims.LOCALE).equals(user.getPreferredLanguage()) ? user.getPreferredLanguage() : get(additionalInformation, StandardClaims.LOCALE)));
```

leleueri:
> Not a huge fan to have a parameter in the getter method.
> 
> Seems to only affected the SCIM part of AM, maybe do a change like this in the mapper :
> 
> ```
> scimUser.setPreferredLanguage(get(additionalInformation, StandardClaims.LOCALE) == null ? user.getPreferredLanguage() : (!get(additionalInformation, StandardClaims.LOCALE).equals(user.getPreferredLanguage()) ? user.getPreferredLanguage() : get(additionalInformation, StandardClaims.LOCALE)));
> ```

It is not only about SCIM but also about UserProperties provided to the template engines.
IMHO we shouldn't persist/override the `preferredLanguage` using the standard locale claims (even in SCIM)

farmborough (approved):

PR #2040, opened 2022-08-12
fix: failed Jest and postman tests since theme feature integration

ashraf706:


farmborough (approved):
LGTM although please don't merge until those integration test pipelines all pass.

PR #2045, opened 2022-08-18
 fix: resolved pre-registration user login issue

ashraf706:
This fix checks existing user in db has the credentials or not. The check is reverted code which was deleted in previous commit.


remisultan (approved):
LGTM

PR #2059, opened 2022-08-31
feat(theme): display default gateway forms

ashraf706:
Shows default gateway forms if custom forms are not present.
This pull request also includes main.css and roboto font files.
The location of these files may be changed as required.

#### Issue: https://github.com/gravitee-io/issues/issues/8315

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/resources/FormsResourceTest.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/resources/FormsResourceTest.java
@@ -19,20 +19,19 @@ import io.gravitee.am.identityprovider.api.User;
 import io.gravitee.am.management.handlers.management.api.JerseySpringTest;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.Form;
-import io.gravitee.am.model.Template;
 import io.gravitee.am.model.ReferenceType;
+import io.gravitee.am.model.Template;
 import io.gravitee.am.service.exception.TechnicalManagementException;
 import io.gravitee.am.service.model.NewForm;
 import io.gravitee.common.http.HttpStatusCode;
 import io.reactivex.Maybe;
 import io.reactivex.Single;
+import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.Response;
-
-import java.io.IOException;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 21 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/forgot_password.html
```diff
@@ -46,7 +46,7 @@
                    th:name="${bot_detection_configuration.get('tokenParameterName')}" />
 
             <div>
-                <button type="submit" id="submitBtn" class="button primary">Sent</button>
+                <button type="submit" id="submitBtn" class="button submit-button">Send</button>
```
leleueri, line 49:
```suggestion
                <button type="submit" id="submitBtn" class="button primary">Send</button>
```

gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/resources/FormsResourceTest.java
```diff
@@ -59,6 +58,7 @@ public void shouldGetForm() throws IOException {
         mockForm.setReferenceId(domainId);
 
         doReturn(Maybe.just(mockForm)).when(formService).findByDomainAndTemplate(domainId, Template.LOGIN.template());
+        doReturn(Single.just(mockForm)).when(formService).getDefaultByDomainAndTemplate(domainId, Template.LOGIN.template());
```
leleueri, line 61: There is a missing TestCase, here `findByDomainAndTemplate` return a value so the result of `getDefaultByDomainAndTemplate` isn't used to generate the response.

We should have the same test but with `findByDomainAndTemplate` that return `Maybe.emtpy`

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/FormServiceImpl.java
```diff
@@ -297,6 +301,24 @@ public Completable delete(String domain, String formId, User principal) {
...
+        return Single.create(emitter -> {
+            try {
+                final String path = "views/default/" + template + ".html";
+                final InputStream resourceAsStream = this.getClass().getClassLoader().getResourceAsStream(path);
+                final String content = IOUtils.toString(resourceAsStream, defaultCharset());
+                final Form form = new Form(false, template);
+                form.setContent(content);
+                emitter.onSuccess(form);
+            } catch (Exception e) {
+                emitter.onError(new TechnicalManagementException("An error has occurred while trying load default form", e));
```
leleueri, line 317:
```suggestion
                emitter.onError(new TechnicalManagementException("An error has occurred while trying to load default form", e));
```

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/FormServiceImpl.java
```diff
@@ -297,6 +301,24 @@ public Completable delete(String domain, String formId, User principal) {
         return delete(ReferenceType.DOMAIN, domain, formId, principal);
     }
 
+    @Override
+    public Single<Form> getDefaultByDomainAndTemplate(String domain, String template) {
+        LOGGER.debug("Default form by domain {} and template {}", domain, template);
+
+        return Single.create(emitter -> {
+            try {
+                final String path = "views/default/" + template + ".html";
```
leleueri, line 310: Open question:

Should we introduce theme or domain in the directory tree to highlight the fact that these templates are the defaults coming from the gateway ?

leleueri (changes requested):
In addition of the

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/FormServiceImpl.java
ashraf706: done

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/FormServiceImpl.java
ashraf706: the `default` folder is created as the `view` folder contains a `login.html`. The sub-folder structure could be any appropriate directories. I have no reservation.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/forgot_password.html
ashraf706: done

gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/resources/FormsResourceTest.java
ashraf706: new test added as suggested

ashraf706:
> In addition of the comments here after, can you explain to me why you introduced the assets into the UI ?

Possibly we have already discussed it after standup among three of us. Happy to explain it again and then we can decide whether we can get the job done without these assets in the UI.

leleueri (approved):

leleueri (changes requested):
The new method introduced into the FormServiceImpl has no Unit Test.
Should at least test :
* Nominal case
* Case where the template is unknown

PR #2077, opened 2022-09-06
Issues/#8316 management override theme view mode load gateway assets

farmborough:
## :id: 
Fixes gravitee-io/issues#8316 

Enables template assets to be rendered in the theme preview.
Also fixed issue where the iframe was too short in Chromium browsers and some console errors that were caused by interaction with the DOM.

Has been integrated and manually tested against the other issues in the epic. clone and run 8123-management-override-theme-view-mode-integration to see this in action.


gravitee-am-ui/src/app/domain/settings/theme/theme.component.ts
```diff
@@ -305,20 +305,34 @@ export class DomainSettingsThemeComponent implements OnInit {
...
+        this.fixAssetUrl(doc, "script", "src");
       }
     }
   }
 
   private resizeIframe() {
     if (this.selectedTemplateContent && this.preview) {
-      this.preview.nativeElement.style.height = this.preview.nativeElement.contentWindow.document.body.scrollHeight + 20 + 'px';
+      const height = window.innerHeight * 0.80;
+      this.preview.nativeElement.style.height = height + 'px';;
```
ashraf706, line 335: double semicolon at the end of this line

gravitee-am-ui/src/app/domain/settings/theme/theme.component.ts
```diff
@@ -305,20 +305,34 @@ export class DomainSettingsThemeComponent implements OnInit {
     });
   }
 
+  private fixAssetUrl(doc: any, tag: string, urlAttribute: string) {
+    const tags = doc.getElementsByTagName(tag);
+    for (let i = 0; i < tags.length; i++) {
+      for (const attribute of tags[i].attributes) {
+        if(attribute.name === urlAttribute){
+          const url = attribute.value;
+          attribute.value = "/" + url;
```
leleueri, line 314: This doesn't work when the URL is an external URL to download a logo for example.
You should test if the `attribute.value` start with http before adding the slash.

leleueri (changes requested):
Some assets are useless:
* for CSS, only main.css is useful
* for font, only Robot & Material are useful

I also wondering if it is possible to move the theme assets into a subfolder (ex: `previews`) in order to make a distinction between assets used for the UI and assets used for the preview (so in the fixAssetUrl, we will have to replace `assets` by `assets/preview`

gravitee-am-ui/src/app/domain/settings/theme/theme.component.ts
farmborough, line 335:
```suggestion
      this.preview.nativeElement.style.height = height + 'px';
```

farmborough:
> Some assets are useless:
> 
> * for CSS, only main.css is useful
> * for font, only Robot & Material are useful
> 
> I also wondering if it is possible to move the theme assets into a subfolder (ex: `previews`) in order to make a distinction between assets used for the UI and assets used for the preview (so in the fixAssetUrl, we will have to replace `assets` by `assets/preview`)? 🤔

I like the idea of the preview folder but it should probably be captured as a separate issue. I got it working _mostly_ but it took extra work to account for the fact that some templates references assets with `../assets` instead of `assets/`. Even after that I was then seeing 404 errors when the browser was trying to load the unedited asset URLs, which I wasn't seeing when everything was under `assets`.

PR #2088, opened 2022-09-14
fix: avoid blockingGet into the SyncManager

leleueri:
fixes gravitee-io/issues#8377


leleueri:
This implementation is too complex.
Need to diagnose the root cause of the blocking more deeply to know if a more in deep refactoring could be usedul.
As a quick fix, I introduced a configurable timeout (see https://github.com/gravitee-io/gravitee-access-management/pull/2106)

PR #2090, opened 2022-09-14
feat(mfa): handle new type of MFA factors

tcompiegne:
closes gravitee-io/issues#8444


leleueri (approved):
I made some changes to make the plugin compatible with 3.15 and made some change into AM to manage resource filtering.

If these changes are OK  for you, you can merge :)

leleueri (changes requested):
the enrollment & challenge screen is the one of the TOTP that ask for an authenticator app. Maybe we should adapt it ? wydt ?

tcompiegne:
For the time being we can keep like this.

ashraf706 (approved):

leleueri (approved):

PR #2098, opened 2022-09-16
fix: remove login_hint on backToIdentifierFirst action

leleueri:
fixes gravitee-io/issues#8236


tcompiegne (changes requested):
why not change the link in the gateway ?

leleueri:
> why not change the link in the gateway ?

Your right, it should

ashraf706 (approved):

tcompiegne (approved):

PR #2118, opened 2022-09-26
fix: try to decode clientId before searching it during auth phase

leleueri:
fixes gravitee-io/issues#8501

## How to test

Create a service app with a clientId containing % character and try to get an access_token


leleueri:
the fix isn't good, we will postpone this fix for a future major version (to take time for testing)

PR #2121, opened 2022-09-26
fix: update live preview with default theme after reset

leleueri:
fixes gravitee-io/issues#8513


farmborough (approved):

gravitee-am-ui/src/app/domain/settings/theme/theme.component.ts
```diff
@@ -274,7 +274,14 @@ export class DomainSettingsThemeComponent implements OnInit {
               this.themes = [];
               this.theme = {};
               this.selectedColorPalette = null;
+              this.renderPreview();
             });
+          } else {
+            // if theme is is missing, reseting the theme should also reset the unpublished fields
```
ashraf706, line 280: double 'is'. Should be "if theme is missing......"

ashraf706 (approved):

PR #2137, opened 2022-09-28
8528: fixed sms phone number input

ashraf706:
The phone number drop down list is a div which becomes 100% in width including all children div. As a result input get masked by the div.

Had to manually add the class to fix the issue.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/assets/css/main.css
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/assets/css/main.css
@@ -226,6 +226,10 @@ form div {
     width: 100%;
 }
 
+.iti__flag-container {
+    width: auto;
+}
+
 .section, .form {
     display: flex;
     flex-direction: column;


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/assets/css/main.css
```diff
@@ -222,7 +222,7 @@ body, body * {
     flex-grow: 0;
 }
 
-form div {
+form div:not(.iti__flag-container, .iti__selected-flag, .iti__flag, .iti__arrow) {
```
tcompiegne, line 225: I wonder if it's not clearer to have a specific section for the intl-tel-input components :

```
.iti__flag-container, .iti__selected-flag, .iti__flag, .iti__arrow {
  width: auto;
}
```

tcompiegne (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/assets/css/main.css
ashraf706: simplified and works in chrome, firefox and safari with 

`.iti__flag-container {
  width: auto;
}`

leleueri (commented):
The fix works but when I press enter, the value is reset. This wasn't the cas in 3.18.x

PR #2152, opened 2022-10-03
fix: Clean up templates

farmborough:
fixes gravitee-io/issues#8548

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/error.html
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/error.html
@@ -30,7 +30,5 @@
             </div>
         </div>
     </div>
-
-    <script th:src="@{assets/material/material.min.js}"></script>
 </body>
 </html>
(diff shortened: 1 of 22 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/forgot_password.html
```diff
@@ -85,35 +85,6 @@
...
-<script th:inline="javascript" th:attr="nonce=${script_inline_nonce}">
- $(document).ready(function (){
-    $(".mdl-textfield__input").focus(function (){
-        if( !this.value ){
-            $(this).prop('required', true);
-            $(this).parent().addClass('is-invalid');
-        }
-    });
-
-    $(".mdl-button").filter(":button").click(function (event){
```
leleueri, line 102: This section is used to enable the Google ReCaptcha on the page if the application settings requires this feature, so the code of this click event should be adapt to make ti works IMHO. (FYI, reCaptcha is define in the following forms login, registration, forgot_password and id_first_login)

leleueri (changes requested):
In addition of the comment above, you forgot to duplicate the changes into the ManagementAPi part.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/forgot_password.html
farmborough: I did that at first, no problem.

leleueri (approved):

PR #2167, opened 2022-10-10
 AM-60: fixed Kafka audit reporter null pointer exception

ashraf706:
Null check added for audit reference type.

**How to test**
1. Follow the instruction in the jira issue [AM-60](https://graviteecommunity.atlassian.net/browse/AM-60) to run kafka locally and configure a kafka reporter
2. Create an application and a user. 
3. Now try to login with an invalid credential. 
4. Kafka reporter should publish the login failure message instead of the null pointer exception.

### DIFF
--- a/gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/DtoMapper.java
+++ b/gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/DtoMapper.java
@@ -74,7 +74,7 @@ public class DtoMapper {
   }
 
   public String map(ReferenceType referenceType) {
-    return referenceType.toString().toUpperCase();
+    return referenceType != null ? referenceType.toString().toUpperCase() : null;
   }
 
   public AuditEntityDto map(AuditEntity auditEntity) {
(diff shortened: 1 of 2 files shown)


gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/DtoMapper.java
```diff
@@ -82,7 +82,7 @@ public AuditEntityDto map(AuditEntity auditEntity) {
     dto.setId(auditEntity.getId());
     dto.setType(auditEntity.getType());
     dto.setReferenceId(auditEntity.getReferenceId());
-    dto.setReferenceType(this.map(auditEntity.getReferenceType()));
+    dto.setReferenceType(auditEntity.getReferenceType() != null ? this.map(auditEntity.getReferenceType()): null);
```
leleueri, line 85: The fix works but maybe we should do this test into the map method to be sure that other usage of this method using a null ReferenceType will not break the injestion of kafka message. wdyt ?

leleueri (changes requested):

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/DtoMapper.java
ashraf706: done, moved the null check as suggested.

leleueri (approved):

leleueri:
fixes gravitee-io/issues#8609

PR #2168, opened 2022-10-10
 AM-67: removed set primary from email factor

ashraf706:
Removed set primary option for email factor.


leleueri (approved):

leleueri:
fixes gravitee-io/issues#8544

PR #2169, opened 2022-10-10
feat: implement CSP directives on ManagementAPI

leleueri:
fixes gravitee-io/issues#8559

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/spring/security/SecurityConfiguration.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/spring/security/SecurityConfiguration.java
@@ -61,10 +61,14 @@ import org.springframework.web.cors.UrlBasedCorsConfigurationSource;
 
 import javax.servlet.Filter;
 import javax.servlet.http.HttpServletRequest;
+import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/spring/security/SecurityConfiguration.java
```diff
@@ -228,6 +235,33 @@ private HttpSecurity csrf(HttpSecurity security) throws Exception {
...
+                directives.add(DEFAULT_CSP_DIRECTIVE);
+            }
+            security.headers()
+                    .contentSecurityPolicy(directives.stream()
+                            .map(directive -> directive.trim().endsWith(";") ? directive : directive + ";")
+                            .collect(Collectors.joining(" ")));
+        }
+    }
+
+    private List<String> getDirectives() {
```
ashraf706, line 251: I am approving this pull request with this suggestion: better if we could avoid `break`. This below code should be equivalent 


private List<String> getDirectives() {

    final List<String> directives = new ArrayList<>();
    var value = "";
    var i = 0;
    do {
        var propertyKey = String.format(HTTP_CSP_DIRECTIVES, i);
        value = environment.getProperty(propertyKey, String.class);
        if (nonNull(value)) {
            directives.add(value);
            i++;
        }
    } while (nonNull(value));

    return directives;
}

ashraf706 (approved):

PR #2171, opened 2022-10-10
feat: allow to configure password history in management ui for domain…

leleueri:
… & application

fixes AM-29

# How to test

1. Update the password policy at domain level

image

2. Update the password policy at application level

image


ashraf706 (approved):

ashraf706:
Functionalities checked both mongo and jdbc and looks good. UI english language could be better. I have approved this pull request but suggesting @farmborough to check the English language in UI at least.

farmborough (changes requested):

PR #2175, opened 2022-10-11
fix: avoid overflow using integer divider onn timeMillis, prefer the …

leleueri:
…usage of Instant.plusSecond.

fixes AM-70


ashraf706 (approved):

leleueri:
fixes gravitee-io/issues#8560

PR #2188, opened 2022-10-19
AM-56-business-logic-validation-during-reset-password

farmborough:
## :id: 
Fix [AM 56](https://graviteecommunity.atlassian.net/browse/AM-56)

## :pencil2: 

A new `PasswordHistoryService` which provides functionality to `UserService` on create and password reset.


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
```diff
@@ -0,0 +1,135 @@
...
+     * @param userId           id of user for this password
+     * @param password         the password to add
+     * @param principal        user performing this action
+     * @param passwordSettings domain/application password settings
+     * @return Single containing {@link PasswordHistory} or an error if the password was already in the history.
+     */
+    public Single<PasswordHistory> addPasswordToHistory(ReferenceType referenceType, String referenceId, String userId, String password, User principal, PasswordSettings passwordSettings) {
+        LOGGER.debug("Adding password history entry for user {}", userId);
+        if (passwordSettings == null || !passwordSettings.isPasswordHistoryEnabled()) {
+            return Single.error(new IllegalArgumentException("Password history disabled."));
```
leleueri, line 74: Should not return an error in this case in my opinion, but return an empty value.
As `Single<>` can't accept null value, the return type for this method should be `Maybe`.

Otherwise you have to manage the Error in the upstream flow during the ResetPassword to avoid blocking the resetPassword action if the history is disabled. (see postman test error)

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
```diff
@@ -0,0 +1,135 @@
...
+     * Checks if a password is already in the history
+     *
+     * @param referenceType    the reference type (DOMAIN, APPLICATION)
+     * @param referenceId      id of the reference
+     * @param userId           id of user for this password
+     * @param password         the password to add
+     * @param passwordSettings domain/application password settings
+     * @return Single containing a {@link Boolean} {@code true} if the password is already present the allotted number of previous passwords.
+     */
+    public Single<Boolean> passwordAlreadyUsed(ReferenceType referenceType, String referenceId, String userId, String password, PasswordSettings passwordSettings) {
```
leleueri, line 111: Is this method used somewhere ?

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/PasswordHistoryServiceTest.java
```diff
@@ -0,0 +1,228 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@ExtendWith(MockitoExtension.class)
+class PasswordHistoryServiceTest {
```
leleueri, line 71: CodeCoverage isn't taking into account for this class, Quality Gate fails

https://sonarcloud.io/component_measures?id=gravitee-io_graviteeio-access-management&pullRequest=2188&metric=new_coverage&view=list

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
farmborough: It will be for the gateway reset password form.

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/PasswordHistoryServiceTest.java
farmborough: Might be because it's package private.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
farmborough: Yes I thought the same when I saw the failing jest tests.

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/PasswordHistoryServiceTest.java
farmborough: Hmm that is weird. I have Sonarlint in intellij as well and it actually recommends default access modifier for Junit 5 tests. I will investigate this.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -286,7 +301,11 @@ public Single<ResetPasswordResponse> resetPassword(Client client, User user, io.
                     // see https://github.com/gravitee-io/issues/issues/8407
                     return userProvider.findByUsername(user.getUsername())
                             .switchIfEmpty(Maybe.error(new UserNotFoundException(user.getUsername())))
-                            .flatMapSingle(idpUser -> userProvider.updatePassword(idpUser, user.getPassword()))
+                            .flatMapSingle(idpUser -> passwordHistoryService
+                                    .addPasswordToHistory(DOMAIN, domain.getReferenceId(), idpUser.getId(), idpUser.getCredentials(), principal, getPasswordSettings(client))
```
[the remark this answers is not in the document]
leleueri: Do we have to use the `idpUser.getId()` or the `user.getId()` ?
I was thinking of the `user.getId()` but maybe you have good arguments in favor of the `idpUser.getId()`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -188,6 +194,11 @@ public Single<RegistrationResponse> register(Client client, User user, io.gravit
                         .switchIfEmpty(Maybe.error(new UserProviderNotFoundException(source)))
                         .flatMapSingle(userProvider -> userProvider.create(convert(user)))
                         .flatMap(idpUser -> {
+                            //noinspection ResultOfMethodCallIgnored
+                            passwordHistoryService
+                                    .addPasswordToHistory(DOMAIN, domain.getReferenceId(), idpUser.getId(), idpUser.getCredentials(), principal, getPasswordSettings(client))
```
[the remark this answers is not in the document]
leleueri: Do we have to use the `idpUser.getId()` or the `user.getId()` ?
I was thinking of the `user.getId()` but maybe you have good arguments in favor of the `idpUser.getId()`.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
```diff
@@ -199,6 +204,11 @@ public Single<User> create(Domain domain, NewUser newUser, io.gravitee.am.identi
                                                         .doOnError(throwable -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class).principal(principal).type(EventType.USER_CREATED).throwable(throwable)))
                                                         .andThen(userProvider.create(convert(newUser)))
                                                         .map(idpUser -> {
+                                                            //noinspection ResultOfMethodCallIgnored
+                                                            passwordHistoryService
+                                                                        .addPasswordToHistory(DOMAIN, domain.getReferenceId(), idpUser.getId(), idpUser.getCredentials(), principal, domain.getPasswordSettings())
```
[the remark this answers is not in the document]
leleueri: Do we have to use the `idpUser.getId()` or the `user.getId()` ?
I was thinking of the `user.getId()` but maybe you have good arguments in favor of the `idpUser.getId()`.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
```diff
@@ -287,44 +297,47 @@ public Completable resetPassword(Domain domain, String userId, String password,
...
-                                        // as it is an admin action here, we want to force the user to login
-                                        .flatMap(updatedUser -> Single.defer(() -> tokenService.deleteByUserId(updatedUser.getId())
-                                                .toSingleDefault(updatedUser)
-                                                .onErrorResumeNext(err -> {
-                                                    logger.warn("Tokens not invalidated for user {} due to : {}", userId, err.getMessage());
-                                                    return Single.just(updatedUser);
-                                                })))
-                                        .doOnSuccess(user1 -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class).client(client).principal(principal).type(EventType.USER_PASSWORD_RESET).user(user)))
+                                final Client client = optClient.map(Application::toClient).orElse(new Client());
+                                return passwordHistoryService.addPasswordToHistory(DOMAIN, domain.getReferenceId(), user.getExternalId(), password, principal, domain.getPasswordSettings())
```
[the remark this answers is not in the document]
leleueri: Do we have to use the `idpUser.getId()` or the `user.getId()` ?
I was thinking of the `user.getId()` but maybe you have good arguments in favor of the `idpUser.getId()`.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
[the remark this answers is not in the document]
leleueri: When a user is created ugin the UI, you are able to select the app. So if the user definition contains a clientId, we should  check if the app has defined a PasswordPolicy

leleueri (changes requested):
In addition of the inner code comments, here is some others that I can't attach to the code reviewed:
* In the `io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java` the `confirmRegistation` method doesn't use the PasswordHistory service to register the user password. 
* I did some forgotPassword tests and I was abel to register a same password multiple times.

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/PasswordHistoryServiceTest.java
```diff
@@ -0,0 +1,231 @@
...
+        given(repository.findUserHistory(DOMAIN, REFERENCE_ID, userId)).willReturn(fromIterable(List.of()));
+        given(repository.create(any(PasswordHistory.class))).willReturn(Single.just(new PasswordHistory()));
+        given(passwordEncoder.encode(password)).willReturn("encrypted password");
+
+        var testObserver = service
+                .addPasswordToHistory(ReferenceType.DOMAIN, REFERENCE_ID, userId, password, new DefaultUser(), passwordSettings)
+                .test();
+        testObserver.awaitTerminalEvent();
+        testObserver.assertComplete();
+        testObserver.assertNoErrors();
```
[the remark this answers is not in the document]
farmborough: Yes I forgot to verify the mock.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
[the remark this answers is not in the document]
farmborough: I could be wrong but I was working on the assumption that the idp assigns the ID to the new user, so we have to wait until the user is created to get that assigned ID.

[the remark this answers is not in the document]
leleueri: IMHO, the only safe ID we should use as foreign key is the one of the `Users` collection (`user.getId()`).
The idpUser.getId() is used to initialize the externalId attribute of the user but as we do not control all the info provided by the IDPs and specially the UserMapper, we can't guaranty that the value will remains the same along the user lifecycle.

That's why I think the user ID should be used instead of the idpUser ID.

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/PasswordHistoryServiceTest.java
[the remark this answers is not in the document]
farmborough: Actually the mocker `passwordEncoder` verifies that is called on line 110 (via the argument captor).

farmborough:
> In addition of the inner code comments, here is some others that I can't attach to the code reviewed:
> 
> * In the `io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java` the `confirmRegistation` method doesn't use the PasswordHistory service to register the user password.
> * I did some forgotPassword tests and I was abel to register a same password multiple times.

I've found issues that came from the BCrypt encoder e.g the first password is said to not match a subsequent identical password. These should be fixed with the Argon2 encoder.

farmborough:
> > In addition of the inner code comments, here is some others that I can't attach to the code reviewed:
> > 
> > * In the `io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java` the `confirmRegistation` method doesn't use the PasswordHistory service to register the user password.
> > * I did some forgotPassword tests and I was abel to register a same password multiple times.
> 
> I've found issues that came from the BCrypt encoder e.g the first password is said to not match a subsequent identical password. These should be fixed with the Argon2 encoder.

Tell a lie. I was encoding the encoded password, rather than the raw one.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
```diff
@@ -0,0 +1,137 @@
...
+                             int passwordCount = passwordHistories.size();
+                             if (passwordCount >= passwordSettings.getOldPasswords()) {
+                                 passwordHistories.sort(comparing(PasswordHistory::getCreatedAt));
+                                 return repository.delete(passwordHistories.get(0).getId())
+                                                  .andThen(repository.create(getPasswordHistory(referenceType, referenceId, userId, password)));
+                             } else {
+                                 return repository.create(getPasswordHistory(referenceType, referenceId, userId, password));
+                             }
+                         })
+                         .doOnSuccess(a -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class)
```
leleueri, line 93: the Domain isn't set a reference of this audit logs, so it is not registered, also maybe we should set the User in order to be able to link the user for who the passwordHistory has been created.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
```diff
@@ -0,0 +1,137 @@
...
+                                                                           .principal(principal)
+                                                                           .type(EventType.PASSWORD_HISTORY_CREATED)))
+                         .doOnError(throwable -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class)
```
leleueri, line 96: the Domain isn't set a reference of this audit logs, so it is not registered, also maybe we should set the User in order to be able to link the user for who the passwordHistory has been created.

leleueri (changes requested):
The `UserService` into the package `io.gravitee.am.gateway.handler.scim.service.impl` need to use the `PasswordHistoryService` in order to update the password on user creation and update (in both case, the password may be missing).

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -286,7 +297,11 @@ public Single<ResetPasswordResponse> resetPassword(Client client, User user, io.
                     // see https://github.com/gravitee-io/issues/issues/8407
                     return userProvider.findByUsername(user.getUsername())
                             .switchIfEmpty(Maybe.error(new UserNotFoundException(user.getUsername())))
-                            .flatMapSingle(idpUser -> userProvider.updatePassword(idpUser, user.getPassword()))
+                            .flatMapSingle(idpUser -> passwordHistoryService
+                                    .addPasswordToHistory(DOMAIN, domain.getId(), user, user.getPassword() , principal, getPasswordSettings(client))
+                                    .isEmpty()
+                                    .flatMapMaybe(empty -> Maybe.just(new PasswordHistory()))
```
leleueri, line 303: I think we can replace it by `switchIfEmpty(Maybe.just(new PasswordHistory())`
wdyt?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -348,9 +359,12 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
                                         .doOnError(error -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class).principal(principal).user(existingUser).type(EventType.USER_UPDATED).throwable(error))));
                             }));
                 })
-                .map(user1 -> UserMapper.convert(user1, baseUrl, false))
+                .map(user1 -> {
+                    createPasswordHistory(domain, user1, rawPassword, principal);
```
leleueri, line 363: As it is an update, in order to reject the update if the password already exist we should do the same kind of call as the one made during resetPassword https://github.com/gravitee-io/gravitee-access-management/pull/2188/files#diff-c09299b3f56fdb15db97bbff8867279a03f6f636d5dcfcaabca730aade600366R300

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param referenceId      id of the reference
+     * @param user           id of user for this password
+     * @param rawPassword      unencrypted password provided by the user. Passed separate from the user object as its password is nulled after creation to avoid leakage.
+     * @param principal        user performing this action
+     * @param passwordSettings domain/application password settings
+     * @return Single containing {@link PasswordHistory} or an error if the password was already in the history.
+     */
+    public Maybe<PasswordHistory> addPasswordToHistory(ReferenceType referenceType, String referenceId, io.gravitee.am.model.User user, String rawPassword, User principal, PasswordSettings passwordSettings) {
+        LOGGER.debug("Adding password history entry for user {}", user);
+        if (passwordSettings == null || !passwordSettings.isPasswordHistoryEnabled()) {
```
leleueri, line 74: In SCIM password may be missing, we should also return a Maybe.empty if rawPassword is null. 
Or do not call this method into the upstream service if the password is null.

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/PasswordHistoryService.java
farmborough: Doesn't the code return an error when the password is empty?

`// check password
                                if (isInvalidUserPassword(userToUpdate.getPassword(), userToUpdate)) {
                                    return Single.error(new InvalidValueException("Field [password] is invalid"));
                                }`

leleueri: Not for an update and I have open an issue about that. 
Password is an optional field on SCIM spec and as we can declare a user linked to a social IDP, the password isn't required in this case.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -369,6 +399,10 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
                 });
     }
 
+    private PasswordSettings getPasswordSettings(Client client) {
```
farmborough, line 402:
```suggestion
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -369,6 +399,10 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
...
+        return PasswordSettings.getInstance(client, domain).orElse(null);
```
farmborough, line 403:
```suggestion
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -369,6 +399,10 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
...
+    }
```
farmborough, line 404:
```suggestion
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -313,10 +331,22 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
                                                                     return userProvider.create(UserMapper.convert(userToUpdate));
                                                                 } else {
                                                                     if (isNullOrEmpty(userToUpdate.getPassword())) {
-                                                                        return userProvider.update(userToUpdate.getExternalId(), UserMapper.convert(userToUpdate));
+                                                                        return createPasswordHistory(
```
leleueri, line 334: Maybe we should check the behaviour here.
If `isNullOrEmpty(userToUpdate.getPassword())` return true, why should we call createPasswordHistory instead of just updating the user ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -467,4 +501,23 @@ private Completable checkRoles(List<String> roles) {
...
+                .subscribe(passwordHistory -> LOGGER.debug("Created password history for user with ID {}", user),
+                           throwable -> LOGGER.debug("Failed to create password history", throwable));
+    }
+
+
+    @SuppressWarnings({"ResultOfMethodCallIgnored", "ReactiveStreamsUnusedPublisher"})
+    private Single<io.gravitee.am.identityprovider.api.User> createPasswordHistory(
+            Domain domain, io.gravitee.am.model.User user,
+            String rawPassword,
+            io.gravitee.am.identityprovider.api.User principal, Function<PasswordHistory, SingleSource<? extends io.gravitee.am.identityprovider.api.User>> function) {
```
leleueri, line 517: Open discussion:

I'm not a big fan of passing a reactive flow as parameter here. I do prefer a createPasswordHistory method that returns a Single/Maybe and continue the processing outside the method.
Something like :
```
createPasswordHistory()
  .map/flatMap(ph -> userProvider.update(user))
  .flapMap(user -> userProvider.updatePassword)
```
wdyt ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -313,10 +331,22 @@ public Single<User> update(String userId, User user, String idp, String baseUrl,
...
+                                                                        Function<PasswordHistory, SingleSource<? extends io.gravitee.am.identityprovider.api.User>> update = ph -> userProvider
+                                                                                .update(userToUpdate.getExternalId(), UserMapper.convert(userToUpdate))
                                                                                 .flatMap(updatedUser -> userProvider.updatePassword(updatedUser, user.getPassword()));
+                                                                        return createPasswordHistory(
+                                                                                domain,
+                                                                                userToUpdate,
+                                                                                rawPassword,
+                                                                                principal,
+                                                                                update);
```
leleueri, line 349: We may simplify all of this by something like: 

```
createPasswordHistory(domain, userToUpdate, rawPassword, principal)
  .flatMapSingle(ph -> userProvider.update(userToUpdate.getExternalId(), UserMapper.convert(userToUpdate)))
  .flatMap(updatedUser -> {
      if (isNullOrEmpty(user.getPassword())) {
         return userProvider.updatePassword(updatedUser, user.getPassword())
      }
     return Single.just(updatedUser);
   }
);
```

wdyt ?

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -467,4 +484,9 @@ private Completable checkRoles(List<String> roles) {
                 }).ignoreElement();
     }
 
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    private Maybe<PasswordHistory> createPasswordHistory(Domain domain, io.gravitee.am.model.User user, String rawPassword, io.gravitee.am.identityprovider.api.User principal) {
+        return passwordHistoryService
+                .addPasswordToHistory(DOMAIN, domain.getId(), user, rawPassword , principal, domain.getPasswordSettings());
```
leleueri, line 490: We are only managing the domain settings, we should use `PasswordSettings.getInstance()`

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
```diff
@@ -287,44 +292,47 @@ public Completable resetPassword(Domain domain, String userId, String password,
...
-                                        // as it is an admin action here, we want to force the user to login
-                                        .flatMap(updatedUser -> Single.defer(() -> tokenService.deleteByUserId(updatedUser.getId())
-                                                .toSingleDefault(updatedUser)
-                                                .onErrorResumeNext(err -> {
-                                                    logger.warn("Tokens not invalidated for user {} due to : {}", userId, err.getMessage());
-                                                    return Single.just(updatedUser);
-                                                })))
-                                        .doOnSuccess(user1 -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class).client(client).principal(principal).type(EventType.USER_PASSWORD_RESET).user(user)))
+                                final Client client = optClient.map(Application::toClient).orElse(new Client());
+                                return passwordHistoryService.addPasswordToHistory(DOMAIN, domain.getId(), user, password , principal, domain.getPasswordSettings())
```
leleueri, line 296: We are only using the domain settings, we should use` PasswordSetting.getInstance`

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
farmborough: Where is the reference for client/application in the scim user service?

leleueri: There is no difference in term of concept application is the same as client.
Client is used in the GW for legacy reason (and to keep the same naming as the one defined into the OAuth spec)

To convert an application object to a client object, you can do `application.toClient()`;

farmborough: I'm not seeing any application reference either.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
```diff
@@ -385,10 +406,10 @@ public Single<User> patch(String userId, PatchOp patchOp, String idp, String bas
 
                     // check password
                     if (isInvalidUserPassword(userToPatch.getPassword(), UserMapper.convert(userToPatch))) {
-                        return Single.error(new InvalidValueException("Field [password] is invalid"));
+                        return Single.error(new InvalidValueException(FIELD_PASSWORD_IS_INVALID));
                     }
 
-                    return update(userId, userToPatch, idp, baseUrl, principal);
+                    return update(userId, userToPatch, idp, baseUrl, principal, null);
```
leleueri, line 412: Instead of null, you should pass the client. 
In the same way as the update method, you have to provide the client as parameter at line 398 and then update the UserEnpoint to provide the value using the routingContext. like you did here https://github.com/gravitee-io/gravitee-access-management/pull/2188/files#diff-64e06ae1c40a661a2f1d9502a04eef338de3414e06558f42d0a6378191f7e49aR134

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/UserServiceImpl.java
farmborough: This is why I suggested to you that these SCIM changes should have been pulled out into a separate subtask - that should have become clear when I started having to change interface method signatures. I'm now making changes to code that itself was created from changes you suggested in this PR.

leleueri (approved):

PR #2194, opened 2022-10-21
fix: check the validity of request_uri parameter when it is a URL

leleueri:
also allow to define the restricted values through the UI and not only using DCR

fixes AM-61


leleueri:
fixes gravitee-io/issues#8532

PR #2197, opened 2022-10-24
feat(saml2): add option to sign SAML assertions and SAML response at SP level

tcompiegne:
closes AM-107

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/internal/model/ApplicationSAMLSettingsMongo.java
+++ b/gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/internal/model/ApplicationSAMLSettingsMongo.java
@@ -25,6 +25,8 @@ public class ApplicationSAMLSettingsMongo {
     private String attributeConsumeServiceUrl;
     private String singleLogoutServiceUrl;
     private String certificate;
+    private boolean wantResponseSigned = true;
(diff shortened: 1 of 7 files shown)


leleueri (commented):
Should we enable `wantResponseSigned` by default as a minimal security expectation and to keep consistency with the previous behaviour where the signature was apply by default ? :thinking:

leleueri (changes requested):
IHMO we should perform a check on application update to be sure that if one of these booleans is enable the SAML idp has been configured with a certificate ?

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/internal/model/ApplicationSAMLSettingsMongo.java
```diff
@@ -25,6 +25,8 @@ public class ApplicationSAMLSettingsMongo {
     private String attributeConsumeServiceUrl;
     private String singleLogoutServiceUrl;
     private String certificate;
+    private boolean wantResponseSigned;
```
leleueri, line 28: Should also defne the default value here  link https://github.com/gravitee-io/gravitee-access-management/pull/2197/files#diff-70a5fcd7cf0bbcedd548b7f8bd379a7768991e8b15b9d0032065bdac139de7f1R47 otherwise existing app will have the flag disabled.

leleueri (changes requested):

leleueri (approved):

PR #2211, opened 2022-10-31
AM-66 remove passwordlessOrigin from the factor data as it seems to b…

leleueri:
…e useless


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeEndpoint.java
```diff
@@ -233,7 +233,6 @@ private void verifyCode(RoutingContext routingContext) {
         if(factor.is(FIDO2)){
             factorData.put(ConstantKeys.PASSWORDLESS_CHALLENGE_KEY, routingContext.session().get(PASSWORDLESS_CHALLENGE_KEY));
             factorData.put(ConstantKeys.PASSWORDLESS_CHALLENGE_USERNAME_KEY, routingContext.session().get(PASSWORDLESS_CHALLENGE_USERNAME_KEY));
-            factorData.put(ConstantKeys.PASSWORDLESS_ORIGIN, domain.getWebAuthnSettings().getOrigin());
```
ashraf706, line 236: What will be the impact on FIDO2 plugin if we remote the origin? Here is the line in FIDO facto plugin:

https://github.com/gravitee-io/gravitee-am-factor-fido2/blob/master/src/main/java/com/graviteesource/am/factor/fido2/provider/Fido2FactorProvider.java#L32

leleueri: It will probably not work anymore as the origin is a element on which WebAuthn perform a control. I didn't find the reference of the ConstantKey.PASSWORDLESS_ORIGIN when I looked for it using intellij, that's why I removed it but as it was strange to me, I prefered to ask you and so I will reintroduce this line and only test the WebAuthSettings before accessing to the origin :) thanks

ashraf706 (approved):

PR #2214, opened 2022-11-03
feat: add maxLength validation

loriepisicchio:
N.A.

I noticed that the max length of the password was not checked by the front-end validation


leleueri (commented):
Thank Lorie but as mentionned into the JIRA issue https://graviteecommunity.atlassian.net/browse/AM-74 

We do not want to display this rule.

PR #2227, opened 2022-11-14
fix(account): verify MFA code for every OTP based factors and not only the PENDING_ACTIVATION ones

tcompiegne:
fixes AM-153

### DIFF
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/exception/authentication/AuthenticationException.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/exception/authentication/AuthenticationException.java
@@ -15,6 +15,8 @@
  */
 package io.gravitee.am.common.exception.authentication;
 
+import io.gravitee.common.http.HttpStatusCode;
+
 import java.util.Map;
 
 /**
@@ -51,6 +53,10 @@ public abstract class AuthenticationException extends RuntimeException {
     }
 
     public abstract String getErrorCode();
+
+    public int getHttpStatusCode() {
+        return HttpStatusCode.UNAUTHORIZED_401;
+    }
 
     public Map<String, String> getDetails() {
         return details;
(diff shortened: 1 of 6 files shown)


leleueri (approved):
Two minor changes but otherwise LGTM

gravitee-am-common/src/main/java/io/gravitee/am/common/exception/authentication/AuthenticationException.java
```diff
@@ -52,6 +52,10 @@ public AuthenticationException(String msg, Throwable throwable, Map<String, Stri
 
     public abstract String getErrorCode();
 
+    public int getHttpStatusCode() {
+        return 401;
```
leleueri, line 56: I would prefer a `HttpStatusCode.UNAUTHORIZED_401`

gravitee-am-common/src/main/java/io/gravitee/am/common/exception/mfa/InvalidCodeException.java
```diff
@@ -31,4 +31,9 @@ public InvalidCodeException(String msg) {
     public String getErrorCode() {
         return "invalid_mfa";
     }
+
+    @Override
+    public int getHttpStatusCode() {
+        return 403;
```
leleueri, line 37: I would prefer a `HttpStatusCode.FORBIDDEN_403`

leleueri (commented):
Two minor comments but LGTM

gravitee-am-common/src/main/java/io/gravitee/am/common/exception/authentication/AuthenticationException.java
tcompiegne: FP

gravitee-am-common/src/main/java/io/gravitee/am/common/exception/mfa/InvalidCodeException.java
tcompiegne: FP

PR #2229, opened 2022-11-14
AM-136 enable CSP / X-Frame-Options & X-XSS-Protection headers by def…

leleueri:
…ault

# How to test

Simply play with AM Console and the Gateway and be sure that :
* Everything still work :) 
* There is no error into the browser console

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -92,22 +92,21 @@
 #    reportOnly: false
 #    # set true if you want to include the directive 'script-src' with the nonce value generated for inline script.
 #    # to by marked by the nonce value, your inline script has to contain th:attr="nonce=${script_inline_nonce}"
-#    script-inline-nonce: false
+#    script-inline-nonce: true
 #    directives:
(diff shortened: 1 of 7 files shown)


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -92,22 +92,21 @@
...
+#    script-inline-nonce: true
 #    directives:
-#      - "default-src * data: blob: filesystem: about: ws: wss: 'unsafe-inline' 'unsafe-eval' 'unsafe-dynamic';"
-#      - "script-src * 'unsafe-inline' 'unsafe-eval';"
-#      - "connect-src * 'unsafe-inline';"
-#      - "img-src * data: blob: 'unsafe-inline';"
-#      - "frame-src *;"
-#      - "style-src * data: blob: 'unsafe-inline';"
-#      - "font-src * data: blob: 'unsafe-inline';"
+#       - "default-src 'self';"
```
ashraf706, line 97: is it intentional to remove `connect-src` from the updated directives?

ashraf706 (approved):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
leleueri: I put into the gravitee.yaml our default values, previously it was some examples to illustrate configuration capabilities.
So yes, it is normal, cusomter/user still have the option to override all of these values to define CSP directives according to their needs.

PR #2230, opened 2022-11-15
feat(account): add MFA sendChallenge method

tcompiegne:
closes AM-156

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/resources/AccountFactorsEndpointHandler.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/resources/AccountFactorsEndpointHandler.java
@@ -502,6 +502,56 @@ public class AccountFactorsEndpointHandler {
                 );
     }
 
+    /**
+     * Issue a challenge for the selected factor (SMS, Email mainly)
+     *
+     * @param routingContext the routingContext holding the current user
+     */
+    public void sendChallenge(RoutingContext routingContext) {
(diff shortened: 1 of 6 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/resources/AccountFactorsEndpointHandler.java
```diff
@@ -507,6 +507,51 @@ public void deleteRecoveryCode(RoutingContext routingContext) {
...
+                    .filter(enrolledFactor -> factorId.equals(enrolledFactor.getFactorId()))
+                    .findFirst();
+
+            if (!optionalEnrolledFactor.isPresent()) {
+                routingContext.fail(new FactorNotFoundException(factorId));
+                return;
+            }
+
+            final EnrolledFactor enrolledFactor = optionalEnrolledFactor.get();
+            sendChallenge(factorProvider, enrolledFactor, user, routingContext, sh -> {
```
leleueri, line 544: the sendChallenge method return a success if the factor doesn't manage the code emission.
As it is an explicit API call to send a code should we test `factorProvider.needChallengeSending()` before calling `sendChallenge` in order to throw a 400 status code (or 403 forbidden) in this specific case ?

ashraf706 (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/resources/AccountFactorsEndpointHandler.java
tcompiegne: FP - 400 response is now the default behavior

PR #2241, opened 2022-11-22
Am 129 password history jest testing

farmborough:
## :id: 
Fixes [https://gravitee.atlassian.net/browse/AM-129](AM-129)

## :pencil2: 
Tests for both the Management and Gateway reset password endpoints. Feel free to ignore the generate api `ts` files; the tests to review are `reset-password.jest.spec` and `password-history.jest.spec`.

### DIFF
--- a/gravitee-am-test/specs/gateway/reset-password.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/reset-password.jest.spec.ts
@@ -1,0 +1,272 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import fetch from "cross-fetch";
+import {afterAll, beforeAll, expect, jest} from "@jest/globals";
+import {requestAdminAccessToken} from "@management-commands/token-management-commands";
+import {createDomain, deleteDomain, patchDomain, startDomain} from "@management-commands/domain-management-commands";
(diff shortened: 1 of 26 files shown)


gravitee-am-test/specs/management/password-history.jest.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+});
+
+describe("Testing password history...", () => {
+    describe("when password history is enabled", () => {
+
+        beforeAll(async () => {
+            domain = await patchDomain(domain.id, accessToken, {
+                "passwordSettings": {
+                    "passwordHistoryEnabled": true,
+                    "oldPasswords": 3
```
leleueri, line 47: In order to fully validate the feature, maybe we should also test that after 3 reset password the first one can be reused? wdyt?

gravitee-am-test/specs/gateway/reset-password.jest.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+beforeAll(async () => {
+    const adminTokenResponse = await requestAdminAccessToken();
+    accessToken = adminTokenResponse.body.access_token;
+    domain = await createDomain(accessToken, "test-reset-password", "test Gateway reset password with password history");
+});
+
+describe("Gateway reset password", () => {
+    beforeAll(async () => {
+        domain = await patchDomain(domain.id, accessToken, {
```
leleueri, line 50: PasswordPolicy can be overided at application level.
In this simulation we only check that the domain settings is working but IMHO we should also have a dedicated test to validate the usage of the application settings.

In both case we should check that the first password can be reused after 3 resetPassword.

leleueri (changes requested):

gravitee-am-test/specs/management/password-history.jest.spec.ts
farmborough: Agree. but there isn't really time at this stage to add anything else to the test (given that I'm on support and the only person fully in the team today).

leleueri (approved):
Can be merge if test cases identified during the review are managed by the task "JEST: forgot password"

gravitee-am-test/specs/management/password-history.jest.spec.ts
farmborough: Done.

gravitee-am-test/specs/gateway/reset-password.jest.spec.ts
farmborough: Done. Application stuff can be done during the quality sprint.

PR #2242, opened 2022-11-23
fix(customClaims): add more user information attributes in the execution context

tcompiegne:
fixes AM-169


gravitee-am-model/src/main/java/io/gravitee/am/model/scim/Address.java
```diff
@@ -90,6 +90,10 @@ public Boolean isPrimary() {
         return primary;
     }
 
+    public Boolean getPrimary() {
```
ashraf706, line 93: curious, why do need `getPrimary` when we already have `isPrimary` method? Same thing you did for Attributes.java class as well?

ashraf706 (approved):

PR #2249, opened 2022-11-25
AM-140 propagate parameters to the registration callback

leleueri:
fixes gravitee-io/issues#8683


ashraf706 (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/user/password/ResetPasswordSubmissionEndpoint.java
```diff
@@ -81,8 +86,9 @@ public void handle(RoutingContext context) {
                 return;
             }
             // else, redirect to the custom redirect_uri
+            var redirectTo = appendQueryParameter(resetPasswordResponse.getRedirectUri(), queryParams);
```
ashraf706, line 89: don't think we need the method anymore, cleaner way will be:

`var redirectTo = keepParams ? ParamUtils.appendQueryParameter(registrationResponse.getRedirectUri(), queryParams) : registrationResponse.getRedirectUri();`

We are avoiding adding a method with same name; This is true for other classes as well

ashraf706 (approved):

PR #2278, opened 2022-12-06
feat(oauth2): add an option to not renew refresh token during refresh token flow

tcompiegne:
Closes AM-207


leleueri:
This PR is useless now that the PR https://github.com/gravitee-io/gravitee-access-management/pull/2277 answer to the user need

PR #2281, opened 2022-12-07
MFA integration tests added

ashraf706:
MFA enrol, rate limit, brute force, recovery code tests added

### DIFF
--- a/gravitee-am-test/package-lock.json
+++ b/gravitee-am-test/package-lock.json
@@ -29,7 +29,6 @@
         "har-validator": "5.1.5",
         "jest": "^28.0.3",
         "license-check-and-add": "^4.0.3",
-        "node-fetch": "2.6.7",
         "prettier": "2.5.1",
         "supertest": "^6.2.3",
         "test-jest": "^1.0.1",
@@ -4219,9 +4218,9 @@
       }
     },
     "node_modules/entities": {
-      "version": "4.3.1",
-      "resolved": "https://registry.npmjs.org/entities/-/entities-4.3.1.tgz",
-      "integrity": "sha512-o4q/dYJlmyjP2zfnaWDUC6A3BQFmVTX+tZPezK7k0GLSU9QYCauscf5Y+qcEPzKL+EixVouYDgLQK5H9GrLpkg==",
+      "version": "4.4.0",
+      "resolved": "https://registry.npmjs.org/entities/-/entities-4.4.0.tgz",
+      "integrity": "sha512-oYp7156SP8LkeGD0GF85ad1X9Ai79WtRsZ2gxJqtBuzH+98YUV6jkHEKlZkMbcrjJjIVJNIDP/3WL9wQkoPbWA==",
       "dev": true,
       "engines": {
         "node": ">=0.12"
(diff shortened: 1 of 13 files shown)


gravitee-am-test/package-lock.json
```diff
@@ -4219,9 +4218,9 @@
       }
     },
     "node_modules/entities": {
-      "version": "4.3.1",
-      "resolved": "https://registry.npmjs.org/entities/-/entities-4.3.1.tgz",
-      "integrity": "sha512-o4q/dYJlmyjP2zfnaWDUC6A3BQFmVTX+tZPezK7k0GLSU9QYCauscf5Y+qcEPzKL+EixVouYDgLQK5H9GrLpkg==",
+      "version": "4.4.0",
+      "resolved": "https://registry.npmjs.org/entities/-/entities-4.4.0.tgz",
+      "integrity": "sha512-oYp7156SP8LkeGD0GF85ad1X9Ai79WtRsZ2gxJqtBuzH+98YUV6jkHEKlZkMbcrjJjIVJNIDP/3WL9wQkoPbWA==",
```
farmborough, line 4223: Don't commit this file.

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
```diff
@@ -0,0 +1,689 @@
...
+}
+
+const createSMTPResource = async (domain, accessToken) => {
+    const smtp = await createResource(domain.id, accessToken, {
+        "type": "smtp-am-resource",
+        "configuration": "{\"host\":\"localhost\",\"port\":5025,\"from\":\"admin@test.com\",\"protocol\":\"smtp\",\"authentication\":false,\"startTls\":false}",
+        "name": "FakeSmtp"
+    });
+
+    expect(smtp).toBeDefined();
```
farmborough, line 415: This is tautological - for the purpose of this test, it doesn't add anything to assert that `smtp` is both defined and not null, as they are essentially the same thing. Also I would argue that you don't need to assert whether smtp is null or defined because you are then asserting on `smtp.id` which would fail were `smtp` null/undefined.

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
```diff
@@ -0,0 +1,689 @@
...
+const emailVerificationCode = "346147";
+
+jest.setTimeout(200000)
+
+beforeAll(async () => {
+    const adminTokenResponse = await requestAdminAccessToken();
+    accessToken = adminTokenResponse.body.access_token;
+    expect(accessToken).toBeDefined();
+    domain = await createDomain(accessToken, "mfa-test-domain", "test mfa");
+    expect(domain).toBeDefined();
```
farmborough, line 58: You should really be making assertions in set up code, furthermore I don't think it adds anything to the test to assert whether the domain has been created, whether it has an ID etc as the test would fail were that not true.

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
```diff
@@ -0,0 +1,689 @@
...
+
+            expect(authorize2.headers['location']).toBeDefined();
+            expect(authorize2.headers['location']).toContain(`${process.env.AM_GATEWAY_URL}/${domain.hrid}/mfa/challenge`);
+
+            /**
+             * The number of the get requests is based on the gateway gravtitee.yml 'mfa_rate' configuration
+             * These assertions will fail or need to be updated if 'mfa_rate' configuration is changed
+             */
+            const expectedCode = [200, 200, 302];
+            for (let i = 0; i < expectedCode.length; i++) {
```
farmborough, line 124: You can use a `forEach` here which would arguably be cleaner/less verbose.

farmborough (changes requested):

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
ashraf706: replaced with `for of` instead of `forEach` for a cleaner code. The anonymous function has to be a `async` in case of `forEach` and it makes this slightly difficult in this case.

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
ashraf706: yep, removed the unnecessary assertions

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
ashraf706: removed the unnecessary assertions.

gravitee-am-test/package-lock.json
ashraf706: this point is not clear to me, will discuss with you.

farmborough: Did you intend to commit the package-lock.json? It gets updated when you run `npm install` but we generally don't commit those changes unless there was a reason to. If it needs updating to run your test, that's fine.

farmborough (approved):

PR #2287, opened 2022-12-08
AM-215 reject forgot passwors on inactive user only if the ConfirmReg…

leleueri:
…istration is disabled

fixes graivtee-io/issues#8725


tcompiegne (approved):

tcompiegne (changes requested):
Missing the following condition :

- A user has been disabled and has completed its registration => Should be KO

tcompiegne (approved):

PR #2323, opened 2022-12-16
fix(ui): apply latest UX/UI recommendations for the sidenav and topnav components

tcompiegne:
fixes AM-227

### DIFF
--- a/gravitee-am-ui/src/app/components/sidenav/sidenav.component.html
+++ b/gravitee-am-ui/src/app/components/sidenav/sidenav.component.html
@@ -34,7 +34,9 @@
       </a>
     </gio-menu-list>
     <gio-menu-footer *ngIf="canDisplayOrganizationSettings()">
-      <a *ngFor="let item of footerMenuItems" [routerLink]="item.path" title="{{item.label}}">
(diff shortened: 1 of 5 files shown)


gravitee-am-ui/src/app/components/sidenav/sidenav.component.html
```diff
@@ -34,7 +34,10 @@
       </a>
     </gio-menu-list>
     <gio-menu-footer *ngIf="canDisplayOrganizationSettings()">
-      <a *ngFor="let item of footerMenuItems" [routerLink]="item.path" title="{{item.label}}">
+      <a *ngFor="let item of footerMenuItems"
+         [routerLink]="item.path"
+         title="{{item.label}}"
+         [matTooltip]="item.tooltip" [matTooltipPosition]="'right'">
```
leleueri, line 40: Is it for accessibilty reason that the title is defined to the label value ?
Due to title definition and the matTooltip definition we have two labels displayed with two different value.

leleueri: after a sync with UI/UX team, title has to be removed.

leleueri (approved):

PR #2345, opened 2022-12-21
 AM-46: removed client check for user update

ashraf706:
Client check is not necessary while updating a user from management.
 This check throws exception in case client is deleted.
 Hence removed the check and the tests


leleueri (commented):
The change works but I have a couple of questions.

* Do we have to adapt the UI information to inform that the app has been deleted ?

image

* Is it as simple as removing the check ?  I'm wondering if we should also reset the client attribute of all the users when the app is deleted. :thinking:  (and in this case, the previous question as no sens :man_shrugging:

ashraf706:
> The change works but I have a couple of questions.
> 
> * Do we have to adapt the UI information to inform that the app has been deleted ?
> 
> image
> 
> * Is it as simple as removing the check ?  I'm wondering if we should also reset the client attribute of all the users when the app is deleted. 🤔  (and in this case, the previous question as no sens 🤷‍♂️


For now I have removed the client id if the application name is not available. What do you think?

leleueri (approved):
LGTM, please open a new issue to manage the update when the application is removed.

PR #2347, opened 2022-12-23
 AM-133: added webauthn validation

ashraf706:
Validation added to make sure origin in not null or blank

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DomainServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DomainServiceImpl.java
@@ -69,6 +69,7 @@ import io.gravitee.am.service.exception.InvalidParameterException;
 import io.gravitee.am.service.exception.InvalidRedirectUriException;
 import io.gravitee.am.service.exception.InvalidRoleException;
 import io.gravitee.am.service.exception.InvalidTargetUrlException;
+import io.gravitee.am.service.exception.InvalidWebAuthnConfigurationException;
 import io.gravitee.am.service.exception.TechnicalManagementException;
 import io.gravitee.am.service.model.NewDomain;
 import io.gravitee.am.service.model.NewSystemScope;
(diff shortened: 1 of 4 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DomainServiceImpl.java
```diff
@@ -615,6 +616,13 @@ private Completable validateDomain(Domain domain) {
             }
         }
 
+        if (domain.getWebAuthnSettings() != null) {
+            final String origin = domain.getWebAuthnSettings().getOrigin();
+            if (origin == null || origin.isBlank()) {
```
leleueri, line 621: We should also test that it is a valid URL (starting by `http` or `https`)

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DomainServiceImpl.java
ashraf706: scheme checked in back end. From front end it looks line

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DomainServiceImpl.java
```diff
@@ -615,6 +618,20 @@ private Completable validateDomain(Domain domain) {
...
+                return Completable.error(new InvalidWebAuthnConfigurationException("Error: Invalid origin. Please provide a valid origin."));
+            }
+
+            final URI uri = UriBuilder.fromURIString(origin).build();
+            final List<String> schemes = Arrays.asList("http", "https");
```
leleueri, line 628: Moving this array into a constant attribute maybe better than creating an instance each time. But as it is only on Management (so low traffic), I will approve it :)

leleueri (approved):

PR #2364, opened 2023-01-03
chore(ci): trigger Docker Image and RPM build after the git tag [AM-12]

leleueri:


sebdevaux (approved):

leleueri:
Moved to draft as the config is falsy and we can do some common peace of code.

PR #2379, opened 2023-01-11
AM-314: error log added for code exchange failure

ashraf706:


leleueri (changes requested):
We should also implement this log for other Social providers:
* github
* franceconnect
* twitter
* facebook
* salesforce
* linkedin
* github

ashraf706:
> We should also implement this log for other Social providers:
> 
> * github
> * franceconnect
> * twitter
> * facebook
> * salesforce
> * linkedin


Done, added log for all of them

leleueri (approved):

PR #2387, opened 2023-01-12
chore: Added ExitOnOutOfMemoryError JVM option as default into gravit…

bozhidarstanishovski:
In case of OOM sometimes the JVM (vertx) does not respond, while the liveness and readiness probes are working in Kubernetes. This is a problem because Kubernetes does not restart the pods in this case. 

A Java option "-XX:+ExitOnOutOfMemoryError" is added by default in the gravitee startup scripts. JVM exits on the first occurrence of an out-of-memory error.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
@@ -88,6 +88,9 @@ JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
 # space for a full heap dump.
 #JAVA_OPTS="$JAVA_OPTS -XX:HeapDumpPath=$GRAVITEE_HOME/logs/heapdump.hprof"
 
+# JVM exits on the first occurrence of an out-of-memory error.
+JAVA_OPTS="$JAVA_OPTS -XX:+ExitOnOutOfMemoryError"
+
 # Disables explicit GC
 JAVA_OPTS="$JAVA_OPTS -XX:+DisableExplicitGC"
 
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
@@ -89,9 +89,13 @@ JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
 #JAVA_OPTS="$JAVA_OPTS -XX:HeapDumpPath=$GRAVITEE_HOME/logs/heapdump.hprof"
 
 # JVM exits on the first occurrence of an out-of-memory error.
-if [ "$GIO_EXIT_ON_OOM_ERROR" = "true" ]; then
-  JAVA_OPTS="$JAVA_OPTS -XX:+ExitOnOutOfMemoryError"
+if [ "$GIO_EXIT_ON_OOM_ERROR" = "false" ]; then
+  GIO_EXIT_ON_OOM_ERROR=""
+else
+  GIO_EXIT_ON_OOM_ERROR="-XX:+ExitOnOutOfMemoryError"
 fi
+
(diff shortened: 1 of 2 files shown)


leleueri (commented):
Is it something we want to set in the startup script or an option into the Chart ?

leleueri (changes requested):
With this change the user can't make a choise. The Exit will be executed on OOM whatever is the expectation of our user. IMHO, we should let the user take the decision by providing an option to simply enable or disable it.

I was expecting something like a new variable `GIO_JAVA_OPTS` that can be append to `JAVA_OPTS` and use the Chart to easily define the value of this `GIO_JAVA_OPTS`.

bozhidarstanishovski:
@leleueri I have changed the default behavior. New variable `GIO_EXIT_ON_OOM_ERROR` can be set to "true" if the user wants to append "-XX:+ExitOnOutOfMemoryError" to  `JAVA_OPTS`.

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
```diff
@@ -88,6 +88,11 @@ JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
 # space for a full heap dump.
 #JAVA_OPTS="$JAVA_OPTS -XX:HeapDumpPath=$GRAVITEE_HOME/logs/heapdump.hprof"
 
+# JVM exits on the first occurrence of an out-of-memory error.
+if [ "$GIO_EXIT_ON_OOM_ERROR" = "true" ]; then
```
maikzumstrull, line 92: I suggest flipping this logic, exiting on OOM unless this is explicitly set to false. Having the process hang around in an undefined state isn't useful. Defaults should be what's useful in most cases.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/bin/gravitee
```diff
@@ -88,6 +88,15 @@ JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
...
 
+# JVM exits on the first occurrence of an out-of-memory error.
+if [ "$GIO_EXIT_ON_OOM_ERROR" = "false" ]; then
+  GIO_EXIT_ON_OOM_ERROR=""
+else
+  GIO_EXIT_ON_OOM_ERROR="-XX:+ExitOnOutOfMemoryError"
+fi
+
+JAVA_OPTS="$JAVA_OPTS $GIO_EXIT_ON_OOM_ERROR"
+
```
leleueri, line 99:
```suggestion
GIO_EXIT_ON_OOM_ERROR=${GIO_EXIT_ON_OOM_ERROR:-true}
if [ "$GIO_EXIT_ON_OOM_ERROR" = "true" ]; then
  JAVA_OPTS="$JAVA_OPTS -XX:+ExitOnOutOfMemoryError"
fi
```

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/bin/gravitee
```diff
@@ -88,6 +88,15 @@ JAVA_OPTS="$JAVA_OPTS -XX:+HeapDumpOnOutOfMemoryError"
...
 
+# JVM exits on the first occurrence of an out-of-memory error.
+if [ "$GIO_EXIT_ON_OOM_ERROR" = "false" ]; then
+  GIO_EXIT_ON_OOM_ERROR=""
+else
+  GIO_EXIT_ON_OOM_ERROR="-XX:+ExitOnOutOfMemoryError"
+fi
+
+JAVA_OPTS="$JAVA_OPTS $GIO_EXIT_ON_OOM_ERROR"
+
```
leleueri, line 99:
```suggestion
GIO_EXIT_ON_OOM_ERROR=${GIO_EXIT_ON_OOM_ERROR:-true}
if [ "$GIO_EXIT_ON_OOM_ERROR" = "true" ]; then
  JAVA_OPTS="$JAVA_OPTS -XX:+ExitOnOutOfMemoryError"
fi
```

PR #2392, opened 2023-01-16
fix: reset login attempts using client internal ID and not the value …

leleueri:
fixes gravitee-io/issues#8812

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
@@ -323,7 +323,7 @@ public class UserServiceImpl implements UserService {
                 .flatMap(user1 -> {
                     LoginAttemptCriteria criteria = new LoginAttemptCriteria.Builder()
                             .domain(user1.getReferenceId())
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -323,7 +323,7 @@ public Single<ResetPasswordResponse> resetPassword(Client client, User user, io.
                 .flatMap(user1 -> {
                     LoginAttemptCriteria criteria = new LoginAttemptCriteria.Builder()
                             .domain(user1.getReferenceId())
```
remisultan, line 325: Should we use `client.getDomain()` as well ?

leleueri: I prefere to keep user.ref as the loginAttempt is related to the user (and both user & client are linked to the domain anyway). We can't rely on the user.client because this field is updated with the last application on which the user signed-in but a user can't move to another domain.

remisultan: Fair enough 👍

remisultan (approved):

PR #2397, opened 2023-01-19
Am 180 mfa with totp and email

farmborough:
## :id: 

Fixes [AM-180](https://gravitee.atlassian.net/browse/AM-180)

## :pencil2: 

Added MFA TOTP using a "virtual' authenticator to generate codes.

[AM-180]: https://gravitee.atlassian.net/browse/AM-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/mfa.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/mfa.jest.spec.ts
@@ -24,6 +24,8 @@ import {createFactor} from "@management-commands/factor-management-commands";
 import {createResource} from "@management-commands/resource-management-commands";
 import {extractXsrfTokenAndActionResponse, getWellKnownOpenIdConfiguration, logoutUser, performFormPost, performGet, performPost} from "@gateway-commands/oauth-oidc-commands";
 import {clearEmails, getLastEmail} from "@utils-commands/email-commands";
+import {TOTP} from "otpauth";
+import * as faker from "faker";
 
 const cheerio = require('cheerio');
 
@@ -41,6 +43,8 @@ let recoveryCodeFactor;
 let recoveryCodeTestApp;
 let bruteForceTestApp;
 let emailTestApp;
+let totpFactor;
+let totpApp;
 
 const mfaChallengeAttemptsResetTime = 1
 const validMFACode = '333333'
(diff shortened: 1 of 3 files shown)


gravitee-am-test/specs/gateway/mfa.jest.spec.ts
```diff
@@ -271,8 +302,6 @@ describe("MFA", () => {
             expect(authorize2.headers['location']).toContain(`${process.env.AM_GATEWAY_URL}/${domain.hrid}/mfa/challenge`);
 
             const successfulVerification = await verifyFactor(authorize2, validMFACode, smsFactor);
-            expect(successfulVerification.headers['location']).toBeDefined();
-            expect(successfulVerification.headers['location']).toContain(`${process.env.AM_GATEWAY_URL}/${domain.hrid}/oauth/authorize`);
```
leleueri, line 275: Instead of removing these two checks, why not doing them into the verifyFactor method (at least the last one to test the right Location value) ?

leleueri (changes requested):
IMHO, some test cases are missing (agree with the fact that the issue description is light)
* We currently test the enrollment step but what about a MFA challenge on a second connection (with factor already enrolled)
* we do not test the case where multiple factor are available (can be provisionned either by selfaccount api or by using the "Enrollment factor policy"), we should create a user with multiple factor and try to switch the method

also we should include the recovery code


Maybe a new issue to create to cover the cases with recovery code and alternative factor selection, but the authentication test with a factor already enrolled should be done in this one IMHO.

gravitee-am-test/specs/gateway/mfa.jest.spec.ts
farmborough: See line 671.

leleueri: ok, my bad. :+1:

farmborough:
> IMHO, some test cases are missing (agree with the fact that the issue description is light)
> 
> * We currently test the enrollment step but what about a MFA challenge on a second connection (with factor already enrolled)
> * we do not test the case where multiple factor are available (can be provisionned either by selfaccount api or by using the "Enrollment factor policy"), we should create a user with multiple factor and try to switch the method
> 
> also we should include the recovery code
> 
> Maybe a new issue need to be created to cover the cases with recovery code and alternative factor selection, but the authentication test with a factor already enrolled should be done in this one IMHO.

I agree but that wasn't the goal of this particular piece of work. It was merely to add an OTP MFA test. There simply isn't time to add those things to the MFA tests and it's out of scope for this work also.

leleueri (approved):

PR #2427, opened 2023-02-07
feat: make sure service level updates username

remisultan:

https://gravitee.atlassian.net/browse/AM-445

This PR ensures that the username is update where necessary if possible

## :memo: Test Scenarios

`curl -vvv -X POST http://localhost:8093/management/auth/token -H 'authorization: Basic base64(username:password)'`

Will get you an access token

```curl -vvv -X PATCH -H "content-type: application/json" -H "Authorization: Bearer <access-token>" http://localhost:8093/management/organizations/DEFAULT/environments/DEFAULT/domains/<domainId>/users/<userId>/username -d '{ "username"  : "<newUsername>" }' ```

- Test if the username respects the username pattern -- should fail
- Test if the username already exists -- should fail
- Finally update a user with its actual username -- should succeed
- Try updating the username of a non UserProvider IDP (Everything but Mongo/Jdbc/Http) -- should fail

### DIFF
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/UserProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/UserProvider.java
@@ -37,7 +37,13 @@ public interface UserProvider extends Service<UserProvider> {
 
     Single<User> update(String id, User updateUser);
 
-    Completable updateUsername(String id, String username);
+    default Single<User> updateUsername(User user, String username) {
+        if (username == null || username.isEmpty()) {
+            return Single.error(new IllegalArgumentException("Username required for UserProvider.updatePassword"));
+        }
+        ((DefaultUser) user).setUsername(username);
+        return this.update(user.getId(), user);
+    }
 
     default Single<User> updatePassword(User user, String password) {
         if (password == null || password.isEmpty()) {
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/UserProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/UserProvider.java
@@ -39,7 +39,7 @@ public interface UserProvider extends Service<UserProvider> {
 
     default Single<User> updateUsername(User user, String username) {
         if (username == null || username.isEmpty()) {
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: gravitee-am-identityprovider/gravitee-am-identityprovider-jdbc/src/main/java/io/gravitee/am/identityprovider/jdbc/user/JdbcUserProvider.java [hunk 1]
```diff
@@ -343,6 +343,32 @@ public Single<User> update(String id, User updateUser) {
                 });
     }
 
+    @Override
+    public Completable updateUsername(String id, String username) {
+        if (Strings.isNullOrEmpty(username)) {
+            return Completable.error(new IllegalArgumentException("Username required for UserProvider.updatePassword"));
```
mariano-fernandez: `UserProvider.updatePassword`? I'm guessing a small copy/paste oversight 😜

gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/UserProvider.java
```diff
@@ -37,6 +37,8 @@ default Maybe<User> findByEmail(String email) {
 
     Single<User> update(String id, User updateUser);
 
+    Completable updateUsername(String oldUsername, String username);
```
mariano-fernandez: The first parameter is called `oldUsername` in the interface and `id` in all implementations
I think they should be in line

remisultan: This is an old revision. 
The last commit 

```
 default Single<User> updateUsername(User user, String username) {
        if (username == null || username.isEmpty()) {
            return Single.error(new IllegalArgumentException("Username required for UserProvider.updateUsername"));
        }
        ((DefaultUser) user).setUsername(username);
        return this.update(user.getId(), user);
    }
```

gravitee-am-identityprovider/gravitee-am-identityprovider-jdbc/src/main/java/io/gravitee/am/identityprovider/jdbc/user/JdbcUserProvider.java
remisultan: Definitely ! 😆

aurelien-pacaud (approved):

PR #2456, opened 2023-02-15
AM-232: redirected to callback url for pre-login exit

ashraf706:
Align with the behaviour for post-login exit when there is only one IDP and the login form is hidden

Detailed of this issue can be found here: https://gravitee.atlassian.net/browse/AM-232

What to test:
I have added an Excel shit with before and after modification behaviours. Only one cell (D9) behaviour has been changed after this fix. Please ping me for further details. Thanks


remisultan (approved):
LGTM

Usage is fine

PR #2459, opened 2023-02-16
AM-447-AM-IDP-When-using-an-internal-API-for-AM-no-validation-of-the-requests-payload-it-provided

farmborough:
## :id: 

Fixes [AM-447](https://gravitee.atlassian.net/browse/AM-447)

## :pencil2: 

Created a util class that checks JSON documents for negative values. 

## :memo: Test scenarios 

See the JIRA for an example IDP configuration that can be used to test this. Essentially, the steps to test are

1. Create an IDP (Suggest HTTP or In-memory as it's one of the easiest to configure)
2. Locate the IDP at the identities mongo table
3. Modify the 'configuration' field JSON to include invalid numeric values, e.g. ```"timeout" : -1000```

[AM-447]: https://gravitee.atlassian.net/browse/AM-447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/utils/JsonNodeValidator.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/utils/JsonNodeValidator.java
@@ -1,0 +1,56 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service.impl.utils;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.NumericNode;
+
+/**
+ * Validates the values of {@link JsonNode}s.
+ */
+public class JsonNodeValidator {
+
+    private JsonNodeValidator(){}
+
+    /**
+     * Validates a JSON document (in the form of a {@link JsonNode}) for 'correctness' of values.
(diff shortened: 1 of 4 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/utils/JsonNodeValidator.java
```diff
@@ -0,0 +1,56 @@
...
+                    case BIG_INTEGER:
+                        isNegative = jsonNode.asInt() < 0;
+                        break;
+                    case FLOAT:
+                    case DOUBLE:
+                    case BIG_DECIMAL:
+                        isNegative = jsonNode.asDouble() < 0;
+                }
+                if (isNegative) {
+                    throw new IllegalArgumentException();
```
mariano-fernandez, line 51: I would add a message to the exception to give some context

mariano-fernandez (changes requested):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/utils/JsonNodeValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Validates a JSON document (in the form of a {@link JsonNode}) for 'correctness' of values.
+     * This is currently restricted to ensuring numeric values are not negative; future enhancement might involve using regex
+     * or similar to validate specific string values.
+     *
+     * @param idpConfig {@link JsonNode} representing a JSON document.
+     */
+    public static void validateConfiguration(JsonNode idpConfig) {
+        idpConfig.elements().forEachRemaining(jsonNode -> {
```
remisultan, line 36: Should we handle also when a field is required ?

remisultan (commented):
The validity on the IDP configuration depends on the class defined on every IDP plugin.
Shouldn't we try to serialize it with the implementation of this configuration ?

farmborough:
> The validity on the IDP configuration depends on the class defined on every IDP plugin. Shouldn't we try to serialize it with the implementation of this configuration ?

I think what you're suggesting would be a much more thorough/better form of validation but I would argue it requires its own jira as an improvement, rather than this bug fix. I'm just validating on negative numbers but we could do with using regex or similar to provide more comprehensive validation, but that would be significantly more work.

mariano-fernandez (approved):

farmborough:
Sonarcloud wasn't detecting the junit 5 test so have bypassed the quality gate.

PR #2461, opened 2023-02-22
AM-449: Prevent null usernames with cockpit handler

farmborough:
Fixes [AM-449](https://gravitee.atlassian.net/browse/AM-449)

## :pencil2: 
Added a null check to `UserCommandHandler` which is Cockpit's route into AM when creating organisation users.

[AM-449]: https://gravitee.atlassian.net/browse/AM-449?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/UserCommandHandler.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/UserCommandHandler.java
@@ -17,7 +17,6 @@ package io.gravitee.am.management.service.impl.commands;
 
 import io.gravitee.am.common.oidc.StandardClaims;
 import io.gravitee.am.management.service.OrganizationUserService;
-import io.gravitee.am.management.service.UserService;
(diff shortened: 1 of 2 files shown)


passionne (approved):
LGTM

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/commands/UserCommandHandler.java
```diff
@@ -57,8 +57,10 @@ public Command.Type handleType() {
 
     @Override
     public Single<UserReply> handle(UserCommand command) {
-
         UserPayload userPayload = command.getPayload();
+        if (userPayload.getUsername() == null) {
+            return Single.error(new IllegalArgumentException("Username must not be null"));
```
leleueri, line 62: **Question**: Instead of an IllegalArgumentException, I wondering if returning a UserReply bean with the ERROR status  could work. :thinking: (with log WARN to trace the reason of the error)

PR #2470, opened 2023-03-01
fix: use onsuccess/onerror mechanism of subscribe

aurelien-pacaud:
https://gravitee.atlassian.net/browse/AM-194


remisultan (approved):
LGTM

PR #2481, opened 2023-03-08
chore: do not use Dozer to convert User to JdbcUser

leleueri:
Related to https://gravitee.atlassian.net/browse/AM-505

Dozer consume a lot of CPU, this change has demonstrate that removing dozer increase drastically the performance.

This change has been tested on 3.19.x, make sure that all the user fields are managed if new user attributes have been introduced between 3.19.x and 3.21.x


remisultan (approved):
LGTM

passionne (approved):
Test passed 😁

PR #2489, opened 2023-03-13
fix: user is unlocked if the username is updated

aurelien-pacaud:
https://gravitee.atlassian.net/browse/AM-385

Unlock the user when his username is changed

## :memo: Test scenarios 

1. Create a user
2. Enable the brut force detection at security domain level or application level (Settings -> User Accounts -> Login)
3. Try to connect with the previous user with a wrong password
4. Go to the user (Settings -> Users -> Your user)
5. The user is locked
6. Change the username
7. The user is unlocked

### DIFF
--- a/gravitee-am-model/src/main/java/io/gravitee/am/model/User.java
+++ b/gravitee-am-model/src/main/java/io/gravitee/am/model/User.java
@@ -771,4 +771,15 @@ public class User implements IUser {
         putAdditionalInformation(StandardClaims.ADDRESS, address);
     }
 
+    public void updateUsername(String username) {
+        setUsername(username);
+        setLastUsernameReset(new Date());
+        unlockUser();
+    }
+
+    public void unlockUser() {
+        setAccountNonLocked(true);
+        setAccountLockedAt(null);
+        setAccountLockedUntil(null);
+    }
 }
--- a/gravitee-am-ui/src/app/domain/settings/users/user/profile/profile.component.ts
+++ b/gravitee-am-ui/src/app/domain/settings/users/user/profile/profile.component.ts
@@ -127,9 +127,7 @@ export class UserProfileComponent implements OnInit {
       .subscribe(res => {
         if (res) {
           this.userService.unlock(this.domainId, this.user.id).subscribe(() => {
-            this.user.accountNonLocked = true;
-            this.user.accountLockedAt = null;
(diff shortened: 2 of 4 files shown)


gravitee-am-model/src/main/java/io/gravitee/am/model/User.java
```diff
@@ -771,4 +771,15 @@ public void setAddress(Map<String, Object> address) {
         putAdditionalInformation(StandardClaims.ADDRESS, address);
     }
 
+    public void updateUsername(String username) {
+        setUsername(username);
+        setLastUsernameReset(new Date());
+        unlockUser();
```
remisultan, line 777: Shouldn't we have a `setUpdatedAt` ?
Or this is done by the underlying `getUserService()` method ? (both domain and organization)

aurelien-pacaud: Done by the underlying `getUserService()` method.

https://github.com/gravitee-io/gravitee-access-management/blob/fd7f78ee47d337420320456912ff31d16b217fd4/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/OrganizationUserServiceImpl.java#L114

https://github.com/gravitee-io/gravitee-access-management/blob/fd7f78ee47d337420320456912ff31d16b217fd4/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/UserServiceImpl.java#L134

remisultan (approved):
LGTM

gravitee-am-ui/src/app/domain/settings/users/user/profile/profile.component.ts
```diff
@@ -127,9 +127,7 @@ export class UserProfileComponent implements OnInit {
       .subscribe(res => {
         if (res) {
           this.userService.unlock(this.domainId, this.user.id).subscribe(() => {
-            this.user.accountNonLocked = true;
-            this.user.accountLockedAt = null;
-            this.user.accountLockedUntil = null;
+            this.unlockUser(this.user);
```
ppbeauchemin, line 130: can't we do this.user.unlockUser() and remove private unlockUser(user) function?

aurelien-pacaud: It's not possible (there is no type User) but you are right I should have something like that

PR #2491, opened 2023-03-13
fix: full distribution path naming

remisultan:
relates to:
https://github.com/gravitee-io/gravitee-access-management/pull/2421


passionne (approved):
🙏

PR #2492, opened 2023-03-13
fix: login count and last login date not updated when user is updated

aurelien-pacaud:
https://gravitee.atlassian.net/browse/AM-453

Not reset login count and last login date when user data are updated

## :memo: Test scenarios 

1. Login with  a user
2. Update firstname of the user (Settings -> Users -> Your User)
3. Check that `Logins count` and `Last login` don't change (right column of the screen)


remisultan (approved):
As we discussed @aurelien-pacaud I'd like to make about a test case scenario.
If it's good then LGTM

PR #2493, opened 2023-03-14
Improve validation for dictionary create and update

farmborough:
Fixes [AM-511](https://gravitee.atlassian.net/browse/AM-511)

## :pencil2: 

Updated the DictionaryAlreadyExists exception, added an InvalidDictionaryException and added additional checks for blank name and locale (was already checking locale but "" is a valid locale in the current JRE).

[AM-511]: https://gravitee.atlassian.net/browse/AM-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/I18nDictionaryService.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/I18nDictionaryService.java
@@ -28,7 +28,6 @@ import io.gravitee.am.service.AuditService;
 import io.gravitee.am.service.EventService;
 import io.gravitee.am.service.exception.AbstractManagementException;
 import io.gravitee.am.service.exception.DictionaryAlreadyExistsException;
-import io.gravitee.am.service.exception.InvalidLocaleException;
 import io.gravitee.am.service.exception.TechnicalManagementException;
 import io.gravitee.am.service.model.NewDictionary;
 import io.gravitee.am.service.model.UpdateI18nDictionary;
@@ -45,10 +44,8 @@ import org.springframework.context.annotation.Lazy;
 import org.springframework.stereotype.Component;
 
 import java.util.Date;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
-import java.util.stream.Stream;
 
 import static io.gravitee.am.common.audit.EventType.I18N_DICTIONARY_CREATED;
 import static io.gravitee.am.common.audit.EventType.I18N_DICTIONARY_UPDATED;
@@ -76,7 +73,7 @@ public class I18nDictionaryService {
                 .isEmpty()
                 .map(isEmpty -> {
                     if (!isEmpty) {
-                        throw new DictionaryAlreadyExistsException(newDictionary.getName(), newDictionary.getLocale());
+                        throw new DictionaryAlreadyExistsException(newDictionary.getLocale());
                     } else {
                         var dictionary = new I18nDictionary();
                         dictionary.setId(RandomString.generate());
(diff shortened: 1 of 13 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/I18nDictionaryService.java
```diff
@@ -142,17 +145,22 @@ public Single<I18nDictionary> update(ReferenceType referenceType, String referen
...
-                                throw new DictionaryAlreadyExistsException(updateDictionary.getName(), updateDictionary.getLocale());
+                                throw new DictionaryAlreadyExistsException(updateDictionary.getLocale());
                             }
                             return existingDictionary;
                         })
                 )
                 .flatMapSingle(oldDictionary -> {
                     var toUpdate = new I18nDictionary(oldDictionary);
                     if (updateDictionary.getName() != null) {
+                        validateName(updateDictionary.getName());
```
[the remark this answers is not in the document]
farmborough: > Same as above

With a PUT the name and/or locale may be null. I think the only way to do this with annotations is for a custom `ConstraintValidator`. WDYT?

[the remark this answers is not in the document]
aurelien-pacaud: In that case, if it's normal to have a null value for name and/or locale we can't use annotation easily. Bu, instead of testing null in the if and after test the name with validateName we can use validateName directly in the if, no ?

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/annotations/IfNotNullThenNotBlankValidatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
```
mariano-fernandez, line 27: Ups, copyright is declared twice in this class 😁

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/dictionary/LocaleValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.Locale;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
```
mariano-fernandez, line 30: ♻️ 😉

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/annotations/IfNotNullThenNotBlankValidatorTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @MethodSource("validValues")
+    void isValid(String value) {
+        assertTrue(validator.isValid(value, null));
+    }
+
+    public Stream<String> validValues() {
+        return Stream.of("blah", "_", "  test  ", null);
+    }
+    @ParameterizedTest
+    @MethodSource("invalidValues")
```
mariano-fernandez, line 61: I think that this should do the trick, sparing us the `invalidValues` method
```suggestion
    @ValueSource(strings = {"", "   "})
```

remisultan (approved):

PR #2496, opened 2023-03-15
fix: ensure gravitee-am-spring-web have javadoc

passionne:

When we added the gravitee-am-spring-web package with content of
org.springframework:spring-web:5.3.18 via a maven-dependency-plugin:unpack.
The deploy to maven central failed because this module doesn't have javadoc.
The hack on the hack here is to add a Dummy.java class to generate
sources and add a little doc which explain the reason.


remisultan (approved):
image
Managed to generate the source file

PR #2522, opened 2023-03-29
chore: add mergify configuration

remisultan:

https://gravitee.atlassian.net/browse/AM-533

This PR adds mergify configuration to the repository

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,96 @@
+pull_request_rules:
+  - name: Apply patches on `master`
+    conditions:
+      - label=apply-on-master
+    actions:
+      backport:
+        branches:
+          - master
+        assignees:
+          - "{{ author }}"
+        body: |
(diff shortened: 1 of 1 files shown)


passionne:
Correct me if I am wrong, this configuration will create a merge from one feature to each maintained branch ? (for now: master, 3.21.x, etc.)
In that case, with the same change on each maintained branch. When we will merge for example 3.20.x into 3.21.x, we will have a conflict with the same change on both release branch :confused:

passionne (approved):
Regarding the ticket and expectation, this PR do the job.

Nevertheless, I think we should add a documentation about how to use-it. May-be one linked to our git work flow.

remisultan:
@passionne 
Thanks.

I took inspiration from @gaetanmaisse and what APIM implemented.
I agree that it should be documented and I think this is the plan for the next coming weeks

PR #2741, opened 2023-07-05
feat(accountLinking): adapt code to support the Account Linking policy

tcompiegne:
closes AM-653


ashraf706 (approved):

ashraf706:
@tcompiegne , I have approved this PR as it works as expected. One observation is the account link itself. For an example, the following user's idp is the default one. A linked account entry is created once the user logged in successfully.

Screenshot 2023-08-22 at 12 27 48

PR #2798, opened 2023-07-21
feat(ui): use license resource to bind gravitee unlock popup

gcusnieux:
https://gravitee.atlassian.net/browse/AM-729


gravitee-am-ui/src/app/domain/applications/application/advanced/factors/mfa-activate/mfa-activate.component.html
```diff
@@ -33,9 +33,9 @@ <h5>Activate MFA</h5>
     <form>
       <mat-form-field appearance="outline" floatLabel="always" fxFlex="40" style="margin-left: 5px;">
         <mat-select placeholder="Mode" name="mode" [(ngModel)]="this.currentMode" (selectionChange)="applyModeChange($event)">
-          <mat-option *ngFor="let mode of modes" [value]="mode" [gioLicense]="modeLicenseMetadata(mode)">
+          <mat-option *ngFor="let mode of modes" [value]="mode" [gioLicense]="mode.feature">
             {{ mode.label }}
-            <mat-icon *ngIf="mode.ee" svgIcon="gio:lock"></mat-icon>
+            <mat-icon *ngIf="mode?.isMissingFeature$ | async" svgIcon="gio:lock"></mat-icon>
```
mariano-fernandez, line 38: is the check `?` in the variable `mode` really necessary? if yes, then maybe it also needs to be added to the other uses of the variable?

gravitee-am-ui/src/app/components/gio-license/gio-license.service.ts
```diff
@@ -149,11 +153,40 @@ const featureMoreInformationData = {
...
+  }
+
+  isMissingFeature$(feature: Feature): Observable<boolean> {
+    if (feature == null || feature.deployed === true) {
+      return of(false);
+    }
+    if (feature.deployed === false) {
+      return of(true);
+    }
+    return this.getLicense$().pipe(map((license) => license == null || license.features.find((feat) => feat === feature) == null));
```
passionne, line 187:
```suggestion
    return this.getLicense$().pipe(any((license) => license == null || license.features.includes(feature));
```
Does it make the same things ?

gcusnieux: > Does it make the same things ?

No. but this yes :
```
return this.getLicense$().pipe(any((license) => license == null || !license.features.includes(feature));
```

passionne (approved):

gravitee-am-ui/src/app/components/gio-license/gio-license.service.ts
passionne: I think that `any` and `includes` is easier to read. Nevertheless, it works for both, so no issue 😊

mariano-fernandez (approved):

gravitee-am-ui/src/app/domain/settings/saml2/saml2.component.ts
```diff
@@ -55,11 +59,11 @@ export class Saml2Component implements OnInit {
       this.publicKeys(this.domainSamlSettings.certificate);
     }
 
-    this.pluginMetadata = {'deployed': false, 'feature': 'am-idp-saml2'};
     const samlPlugin = this.route.snapshot.data['identities']['am-idp-saml'];
```
aelamrani, line 62: I think this part is wrong as this plugin is not the gateway idp but the idp saml

gravitee-am-ui/src/app/domain/settings/saml2/saml2.component.ts
```diff
@@ -55,11 +59,11 @@ export class Saml2Component implements OnInit {
...
     if (samlPlugin != null) {
-      this.pluginMetadata.deployed = samlPlugin.deployed;
+      this.saml2Feature.deployed = samlPlugin.deployed;
     }
+    this.isMissingSaml2Feature$ = this.licenseService.isMissingFeature$(this.saml2Feature);
```
aelamrani, line 66: I would rather than use a resolver to get the license details in order to check if the plugin is deployed or not

gravitee-am-ui/src/app/domain/settings/saml2/saml2.component.ts
```diff
@@ -35,13 +37,15 @@ export class Saml2Component implements OnInit {
   editMode: boolean;
   certificates: any[] = [];
   certificatePublicKeys: any[] = [];
-  pluginMetadata: any;
+  saml2Feature: Feature = { deployed: false, feature: 'am-idp-saml2'};
```
aelamrani, line 40: The name of the plugin is am-gateway-handler-saml-idp and not am-idp-saml2

aelamrani (commented):
It does not work locally, I am keep seeing the popup even on an EE config with a valid license and the plugin deployed

PR #2806, opened 2023-07-24
fix: retrieve the correct email templates

mariano-fernandez:

https://gravitee.atlassian.net/browse/AM-721

The method `findByDomainAndClientId` does not search an application by its id, but by auth `clientId`.
Hence, this has never worked properly.

This PR brings back the use of `emailTemplateService.findByClientAndTemplate` that correctly searches by domain and application.

## :memo: Test scenarios 

Everything is described in the issue.


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailManagerImpl.java
```diff
@@ -73,9 +73,6 @@ public class EmailManagerImpl extends AbstractService<EmailManager> implements E
     @Autowired
     private EventManager eventManager;
 
-    @Autowired
-    private ApplicationService applicationService;
```
leleueri, line 77: Import becomes useless L28

leleueri (approved):

PR #2840, opened 2023-08-03
chore: add linters for syntax, license and format

gcusnieux:
https://gravitee.atlassian.net/browse/CJ-566


mariano-fernandez (commented):
Awesome to have this in place. Now, can we add a job in the CI to enforce the linter to be successful before merging a PR?

mariano-fernandez (approved):

remisultan:
Several things:

- Can we have the configuration and the actual linting in 2 separated commits ?
That way we can back port ourselves the configuration on maintenance branches without having to diff on code

- I see that the js/ts linter is applied on gravitee-am-ui only. Could we do it also on `gravitee-am-test` which contains jest tests ?


Also I believe that the Java formatter will appear on the upgrade bom with Spring 6 (which is not expected in this PR I 
believe) ?

PR #2843, opened 2023-08-07
Update Changelog Generation

gaetanmaisse:

Update the script generating the changelog to match the new format and target the documentation website instead of the internal changelog.

-> To move further, we need to align filenames for APIM and AM first: https://github.com/gravitee-io/gravitee-platform-docs/pull/21

### DIFF
--- a/release/helpers/changelog-helper.mjs
+++ b/release/helpers/changelog-helper.mjs
@@ -1,23 +1,23 @@
 /**
- * Get the Ascii doc formatted changelog for input issues
+ * Get the MarkDown doc formatted changelog for input issues
  *
  * @param title {string}
- * @param issues {Array<{id: string, fields: Array<{customfield_10115: string, summary: string}>}>}
+ * @param issues {Array<{id: string, key: string, githubIssue: string, summary: string, components: Array<string>, issueTypeName: string, statusName: string}>}
  */
 export function getChangelogFor(title, issues) {
-  const authorizedIssueTypes = ['Public Bug', 'Story'];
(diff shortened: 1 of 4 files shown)


release/helpers/changelog-helper.mjs
```diff
@@ -1,23 +1,23 @@
...
- * @param issues {Array<{id: string, fields: Array<{customfield_10115: string, summary: string}>}>}
+ * @param issues {Array<{id: string, key: string, githubIssue: string, summary: string, components: Array<string>, issueTypeName: string, statusName: string}>}
  */
 export function getChangelogFor(title, issues) {
   const authorizedIssueTypes = ['Public Bug', 'Story'];
 
   const filteredIssues = issues
     .filter((issue) => {
-      return !!issue.fields.issuetype.name && authorizedIssueTypes.includes(issue.fields.issuetype.name);
+      return !!issue.issueTypeName && authorizedIssueTypes.includes(issue.issueTypeName);
```
remisultan, line 12: Should we add `Public Security` or any other tasks ?

I am wondering if we should have field that is automatically set on JIRA side regarding so we can include anything without maintaining this array (or any config file)?

```
const authorizedIssueTypes = ['Public Bug', 'Story'];
```

gaetanmaisse: I included the `Public Security` and I do agree with you about having something shared between APIM and AM. I think we will need a Jira master there :)

gaetanmaisse:
BTW, I tested it locally and it works :)

PR #2849, opened 2023-08-08
feat: email allow list to protect from impersonation

remisultan:

https://gravitee.atlassian.net/browse/AM-649

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -220,6 +220,11 @@ email:
   from: noreply@my.domain
   username: user@my.domain
   password: password
+  # List of allowed from
+  allowedfrom:
+    - ${email.from}
+    # - *@mydomain.org
+    # - *@*.*
 #  properties:
 #    auth: true
 #    starttls.enable: true
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailServiceImpl.java
@@ -37,6 +37,7 @@ import io.gravitee.am.service.i18n.ThreadLocalDomainDictionaryProvider;
 import io.gravitee.am.service.impl.I18nDictionaryService;
 import io.gravitee.am.service.reporter.builder.AuditBuilder;
 import io.gravitee.am.service.reporter.builder.EmailAuditBuilder;
+import io.gravitee.am.service.validators.email.EmailDomainValidator;
 import io.netty.handler.codec.http.QueryStringDecoder;
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Maybe;
@@ -75,19 +76,19 @@ public class EmailServiceImpl implements EmailService, InitializingBean {
     private final Integer registrationVerifyExpireAfter;
     private final String certificateExpirySubject;
 
-    private EmailManager emailManager;
+    private final EmailManager emailManager;
 
-    private io.gravitee.am.service.EmailService emailService;
+    private final io.gravitee.am.service.EmailService emailService;
 
-    private Configuration freemarkerConfiguration;
+    private final Configuration freemarkerConfiguration;
 
-    private AuditService auditService;
+    private final AuditService auditService;
 
-    private JWTBuilder jwtBuilder;
+    private final JWTBuilder jwtBuilder;
 
-    private DomainService domainService;
+    private final DomainService domainService;
 
-    private I18nDictionaryService i18nDictionaryService;
+    private final I18nDictionaryService i18nDictionaryService;
 
     private final ThreadLocalDomainDictionaryProvider dictionaryProvider;
 
@@ -142,9 +143,8 @@ public class EmailServiceImpl implements EmailService, InitializingBean {
             Email email = prepareEmail(domain, client, template, emailTemplate, user);
             // send email
             sendEmail(email, user);
-
-                return email;
-            });
+            return email;
+        });
     }
 
     private Completable refreshDomainDictionaries(Domain domain) {
--- a/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -281,6 +281,11 @@ email:
   from: noreply@my.domain
   username: user@my.domain
   password: password
+  # List of allowed from
+  allowedfrom:
+    - ${email.from}
+    # - *@mydomain.org
+    # - *@*.*
 #  properties:
 #    auth: true
 #    starttls.enable: true
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/email/EmailDomainValidatorImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/email/EmailDomainValidatorImpl.java
@@ -1,0 +1,50 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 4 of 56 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/validators/flow/policy/SendEmailPolicyValidator.java
```diff
@@ -0,0 +1,60 @@
...
+import static java.util.Optional.ofNullable;
+
+/**
+ * @author Rémi SULTAN (remi.sultan at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+@Component
+public class SendEmailPolicyValidator implements PolicyValidator {
+
+    private final Logger LOGGER = LoggerFactory.getLogger(SendEmailPolicyValidator.class);
```
mariano-fernandez, line 35:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(SendEmailPolicyValidator.class);
```

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/flow/policy/SendEmailPolicyValidator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public Boolean validate(Step element) {
+        if (POLICY_AM_SEND_EMAIL.equalsIgnoreCase(element.getPolicy())) {
+            try {
+                var tree = objectMapper.readTree(element.getConfiguration());
+                final String from = ofNullable(tree.get(FROM_KEY)).map(JsonNode::textValue).orElse(null);
+                return emailDomainValidator.validate(from);
+            } catch (Exception e) {
+                LOGGER.debug("An unexpected error has occurred", e);
```
mariano-fernandez, line 54: If something unexpected happens, we want to know about it. I would probably log this with `error` level

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/FlowValidatorImplTest.java
```diff
@@ -0,0 +1,149 @@
...
+        var observer = flowValidator.validate(flow).test();
+        observer.awaitDone(10, TimeUnit.SECONDS);
+        observer.assertComplete().assertNoErrors();
+
+        verify(emailPolicyValidator, times(2)).validate(any());
+    }
+
+    @ParameterizedTest
+    @MethodSource("params_that_must_invalidate_step")
+    void must_invalidate_step(String invalidStepName, int expectedInvocations) {
```
mariano-fernandez, line 69: Can variables declaration be adjusted to have consistency? In the same method there are variables declared with `final`, some of them without it and other with `var`.
In general I prefer to use `var` and leave the compiler to do the work for me. But that's just personal preference 😄

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailServiceImpl.java
```diff
@@ -137,14 +138,13 @@ public Maybe<Email> send(Domain domain, Application client, io.gravitee.am.model
 
     private Maybe<Email> prepareAndSend(Domain domain, Application client, io.gravitee.am.model.Template template, User user) {
         // get raw email template
-        return getEmailTemplate(template, user).map(emailTemplate -> {
+        return getEmailTemplate(template, user).flatMap(emailTemplate -> {
```
leleueri, line 141: Question: Why are you switching to flatMap and so Initialize a Maybe whereas the previous implementation was OK IMO.

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -281,6 +281,8 @@ email:
   from: noreply@my.domain
   username: user@my.domain
   password: password
+  allowlist:
```
leleueri, line 284: Not sure about the name, it is not obvious that this entry refer to the allowed `from` values.

If better name can't be found, maybe a comment maybe useful.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/flow/policy/SendEmailPolicyValidator.java
leleueri: Agree, should be at least a WARN

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
leleueri: Also the helm chart will have to be updated to take into account this new setting

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/EmailDomainValidatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+        assertEquals(expectedResult, validator.validate(email));
+    }
+
+    private static Stream<Arguments> params_that_must_validate_email(){
+        return Stream.of(
+                Arguments.of("test@example.com", true),
+                Arguments.of("another@mail.com", true),
+                Arguments.of("invalid.email", false),
+                Arguments.of("", false),
+                Arguments.of(null, false)
```
leleueri, line 68: Can we add another test with  `evil@mail.com` as all invalid cases are reallying on a non valid email address ?

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -220,6 +220,8 @@ email:
   from: noreply@my.domain
   username: user@my.domain
   password: password
+  allowlist:
+    - "*@*.*"
```
leleueri, line 224: see comments on the [gravitee.yaml](https://github.com/gravitee-io/gravitee-access-management/pull/2849/files#diff-12f73c7e177a2eda64da20512fd027158e798e496fabb691acf7aa8adb302b40) in the management-api service

leleueri (changes requested):
In addition of all the comments, here is two general ones.

First I have a doubt about the way the error is propagated to the REST Api.
As far as I undestand, we just inform that the flow configuration is invalid but we do not specify that it is due to the email settings. Maybe something to improve here.

Second, IMO this PR is not complete as we do not perform control on the **from** value defined in custom email templates (domain & application level)

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/email/EmailDomainValidatorImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+    private static List<Pattern> initializeAllowList(ConfigurableEnvironment environment) {
+        List<Pattern> allowList = new ArrayList<>();
+        for (int i = 0; true; i++) {
+            var propertyKey = String.format(EMAIL_ALLOW_LIST, i);
+            var value = environment.getProperty(propertyKey, String.class);
+            if (isNull(value)) {
+                break;
+            }
+            allowList.add(Pattern.compile(WildcardUtils.toRegex(value)));
```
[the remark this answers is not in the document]
remisultan: Usually if it fails it we just let it crash

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/flow/policy/SendEmailPolicyValidator.java
remisultan: Let's go with this then

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailServiceImpl.java
remisultan: I implemented something there but removed it after. I'll rollback to thoses changes

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
remisultan: It's true we have the helm charts in the repo now 👍 
Good catch thanks

gravitee-am-service/src/test/java/io/gravitee/am/service/validators/FlowValidatorImplTest.java
remisultan: Will do

mariano-fernandez (approved):

PR #2870, opened 2023-08-28
fix: unique constraints added for username

ashraf706:
Brief summary:

Unique constraints are added in:
Users table and newly created idp_xxx tables as necessary.
As a result there won't be any duplicate username in the users table. 

What to check:
Postgresql already had unique constraints and had no issue. For rest of the JDBC or mongodb the easiest way to test would be running the "CreateUsers" gatling test under "gravitee-am-performance" package. 

- Run "CrateDomain" gatling test
- Change "SimulationSettigns.scala" such as

1. val NUMBER_OF_USERS = Integer.getInteger("number_of_users", 12)
2. val AGENTS = Integer.getInteger("agents", 3)
3. Remove "${index}" from username around line number 87.
The goal here to create a duplicate user. These changes allow 3 threads to try to create user with same username in total 4 times. Feel free to change the variable value as necessary.

- Run "CreateUsers" gatling test
- The expectation is there will be no duplicate user with same username

Testing mongoDB and at least one JDBC should be sufficient.
I have written few queries to delete duplicate username entries from the exiting tables (if requres), I will add them in a slab page. The reason is; exception will be thrown during startup due to unique constraints if table already has duplicate data.

### DIFF
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/user/MongoUserProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/user/MongoUserProvider.java
@@ -16,6 +16,7 @@
 package io.gravitee.am.identityprovider.mongo.user;
 
 import com.google.common.base.Strings;
+import com.mongodb.client.model.IndexOptions;
 import com.mongodb.client.model.Updates;
 import com.mongodb.reactivestreams.client.MongoCollection;
 import io.gravitee.am.common.oidc.StandardClaims;
@@ -38,6 +39,7 @@ import org.bson.conversions.Bson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Import;
 
 import java.security.SecureRandom;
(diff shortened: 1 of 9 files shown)


gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/AbstractUserRepository.java
```diff
@@ -67,9 +69,13 @@ public abstract class AbstractUserRepository<T extends UserMongo> extends Abstra
     protected static final String FIELD_EMAIL = "email";
     protected static final String FIELD_EMAIL_CLAIM = "additionalInformation.email";
     protected static final String FIELD_EXTERNAL_ID = "externalId";
+    private static final String OLD_INDEX_REFERENCE_TYPE_REFERENCE_ID_USERNAME_SOURCE = "referenceType_1_referenceId_1_username_1_source_1";
```
mariano-fernandez, line 72: I found it confusing to name the constant `OLD_`. If I understand correctly, the index will be deleted and re-created using the same name but with an extra constraint

mariano-fernandez (commented):
I was also wondering, is there already a way to communicate this fix to customers? Because if the upgrade fails, they'll need to take manual actions on their database and remove/rename duplicated users

gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/user/MongoUserProvider.java
```diff
@@ -53,22 +55,27 @@
  */
 @Import({MongoAuthenticationProviderConfiguration.class})
 public class MongoUserProvider extends MongoAbstractProvider implements UserProvider {
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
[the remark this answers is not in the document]
ashraf706: done

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/AbstractUserRepository.java
ashraf706: done, "OLD_" removed from the name

mariano-fernandez: Class `MongoUserProvider` also has a constant named `OLD_` 😉

ashraf706: updated :D

mariano-fernandez (approved):

PR #2872, opened 2023-08-29
fix: add toggle to disable parameter encoding

mariano-fernandez:

https://gravitee.atlassian.net/browse/AM-737

This PR introduces a hack 😞. Please be as harsh as possible with reviews. I don't like this code and I'm open to any suggestion 🙏 .

So far, when computing redirect URLs in AM, the `static` class `UriBuilderRequest` decodes and encodes all parameters.

When parameters aren't url-encoded, this is not a big deal. But when parameters are already url-encoded, we introduce a few issues.

Since the method is `UriBuilderRequest.resolveProxyRequest(...)` is used all around AM, it's hard to tell if removing the extra decoding will not break stuff.

This PR introduces the parameter `legacy.openid.sanitizeParametersEncoding` that, when set to `false`, disables the extra decoding.

For the future, we should remove the hack and any extra handling of parameters, only decoding parameters when they are received and encoding them when they leave AM Handlers.

## :memo: Test scenarios 

The issue contains a reproduction scenario.

I manually tested the fix in the following flows:
- Normal login flow
- Wrong password on normal login flow
- Login flow with `Identifier-first` enabled.
- Wrong password on ID first flow
- Navigation to Password reset page and back.
- Navigation to Register page and back.
- MFA with an SMS mock factor
- Wrong code with MFA mock factor.

I'm not even sure where this should be documented

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/AccountProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-account/src/main/java/io/gravitee/am/gateway/handler/account/AccountProvider.java
@@ -69,6 +69,8 @@ public class AccountProvider extends AbstractService<ProtocolProvider> implement
     @Autowired
     private RateLimiterService rateLimiterService;
 
+    private boolean sanitizeParametersEncoding;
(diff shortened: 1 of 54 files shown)


mariano-fernandez:
Closing this PR in favor of a cleaner solution. PR https://github.com/gravitee-io/gravitee-access-management/pull/2885

PR #2885, opened 2023-09-07
fix: add sanitizeParametersEncoding toggle

mariano-fernandez:

https://gravitee.atlassian.net/browse/AM-737

This PR introduces a hack 😞.

So far, when computing redirect URLs in AM, the `static` class `UriBuilderRequest` decodes and encodes all parameters.

When parameters aren't url-encoded, this is not a big deal. But when parameters are already url-encoded, we introduce a few issues.

Since the method is `UriBuilderRequest.resolveProxyRequest(...)` is used all around AM, it's hard to tell if removing the extra decoding will not break stuff.

This PR introduces the parameter `legacy.openid.sanitizeParametersEncoding` that, when set to `false`, disables the extra decoding.

For the future, we should remove the hack and any extra handling of parameters, only decoding parameters when they are received and encoding them when they leave AM Handlers.

## :memo: Test scenarios 

The issue contains a reproduction scenario.

I manually tested the fix in the following flows:
- Normal login flow
- Wrong password on normal login flow
- Login flow with `Identifier-first` enabled.
- Wrong password on ID first flow
- Navigation to Password reset page and back.
- Navigation to Register page and back.
- MFA with an SMS mock factor
- Wrong code with MFA mock factor.

I'm not even sure where this should be documented

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/StaticEnvironmentProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/StaticEnvironmentProvider.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.common.utils;
+
+import org.springframework.core.env.Environment;
+
+public class StaticEnvironmentProvider {
+
+    private static Environment env = null;
+    private static Boolean sanitizeParametersEncoding = null;
+
+    public static void setEnvironment(Environment environment) {
+        env = environment;
+        // reset cached values
+        sanitizeParametersEncoding = null;
+    }
+
+    public static boolean sanitizeParametersEncoding() {
+        if (sanitizeParametersEncoding != null) {
+            return sanitizeParametersEncoding;
+        }
+        sanitizeParametersEncoding = getEnvironmentProperty("legacy.openid.sanitizeParametersEncoding", boolean.class, true);
+        return sanitizeParametersEncoding;
+    }
+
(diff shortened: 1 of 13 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/StaticEnvironmentProvider.java
```diff
@@ -0,0 +1,42 @@
...
+        // reset cached values
+        sanitizeParametersEncoding = null;
+    }
+
+    public static boolean sanitizeParametersEncoding() {
+        if (sanitizeParametersEncoding != null) {
+            return sanitizeParametersEncoding;
+        }
+        sanitizeParametersEncoding = getEnvironmentProperty("legacy.openid.sanitizeParametersEncoding", boolean.class, true);
+        return sanitizeParametersEncoding;
```
passionne, line 36:
```suggestion
        if (sanitizeParametersEncoding == null) {
            sanitizeParametersEncoding = getEnvironmentProperty("legacy.openid.sanitizeParametersEncoding", boolean.class, true);
        }        
        return sanitizeParametersEncoding;
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/test/java/io/gravitee/am/gateway/handler/common/utils/StaticEnvironmentProviderTest.java
```diff
@@ -0,0 +1,46 @@
...
+    @BeforeClass
+    public static void setup() {
+        when(env.getProperty("legacy.openid.sanitizeParametersEncoding", boolean.class, true))
+                .thenReturn(false)
+                .thenReturn(true);
+        StaticEnvironmentProvider.setEnvironment(env);
+    }
+
+    @Test
+    public void sanitizeParametersEncoding_environment_returns_value() {
```
passionne, line 39:
```suggestion
    public void sanitizeParametersEncoding_environment_returns_cached_value() {
```

passionne (approved):
I added a few comments, but no blocker for me.
Thank you, it is a better way to handle this issue.

leleueri (changes requested):
Code is OK but there is a regression when the legacy option is set to false for the use case describe here https://github.com/gravitee-io/issues/issues/7889.

image

mariano-fernandez:
@leleueri good catch! I fixed that one and found another missing toggle. What a lovely plate of 🍝  😅 
Changes in a new commit. [`5aad56f` (#2885)](https://github.com/gravitee-io/gravitee-access-management/pull/2885/commits/5aad56f040fb427a8ee3da3438058ff7d003f823)

mariano-fernandez:
@leleueri I think you wanted to do some more testing. Any news regarding that? the sprint is arriving to an end and maybe is something we want to include in the scheduled release

leleueri:
> @leleueri I think you wanted to do some more testing. Any news regarding that? the sprint is arriving to an end and maybe is something we want to include in the scheduled release

Still have an issue

image

leleueri (approved):

PR #2886, opened 2023-09-07
fix: Potential bug in IDTokenServiceImpl

passionne:
backport of #2883


mariano-fernandez (approved):
Nothing to add regarding the change. But I saw that you added the mergify label for `4.0.x`. Isn't the change coming from that branch? Or maybe I'm missing something 😊 

Also, now we have a brand new `4.1.x`, so I guess this will also need to be added there.

I already updated the labels. But don't hesitate to adjust them again if I got something wrong 😉

PR #3011, opened 2023-09-26
feat(helm): Allow users to define extra manifests

sebdevaux:
Allow users to define extra manifests


passionne (commented):
Could we had a unit test like on APIM?
https://github.com/gravitee-io/gravitee-api-management/blob/master/helm/tests/extra-manifests_test.yaml

leleueri (approved):

PR #3046, opened 2023-09-29
chore: update changelog generation process

leleueri:
The update enhances the changelog generation process in multiple ways. Firstly, it uses the new documentation website and ensures that directory and branch names follow a more uniform naming pattern. Also, it optimizes how the changelog information is extracted from Jira by filtering out irrelevant issues and fetching additional details, enhancing the document's content. Temporary files used in the process are now being stored in a local '.tmp' folder, which is ignored by Git, keeping the repository clean.

Cherry-pick of 074895e0b

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ target/
 .idea/
 *.iml
 .DS_Store
+.tmp/
 
 # eclipse
 .settings/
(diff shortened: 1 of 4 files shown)


ashraf706 (approved):

leleueri:
We do not run IntegTests as it is only changes in the release process scripts

PR #3051, opened 2023-10-02
am-762: fixed liquibase driver issue for MSSQLServer

ashraf706:
Added parameter when SSL is disabled.


gravitee-am-repository/gravitee-am-repository-jdbc-api/src/main/java/io/gravitee/am/repository/jdbc/provider/utils/TlsOptionsHelper.java
```diff
@@ -30,21 +30,25 @@
...
+                case MYSQL:
                     builder = mysqlOptions(builder, environment, prefix);
                     break;
-                case "mariadb":
+                case MARIADB:
                     builder = mariadbOptions(builder, environment, prefix);
                     break;
-                case "sqlserver":
+                case SQLSERVER:
                     builder = sqlServerOptions(builder, environment, prefix);
```
leleueri, line 52: **Suggestion**: in the same way we force the SSL to false for JDBC if the settings is not set in the gravitee.yaml, I think we should also force the SSL option to false for R2DBC. wdyt ?

ashraf706: Done for R2DBC section.

leleueri (approved):

PR #3086, opened 2023-10-09
fix: provide MongoDB migration script to rename duplicate users based…

leleueri:
… on the loginsCount, loggedAt and createdAt attributes

 this script provide a dryRun mode (enabled by default) to check the action before performing updates into DB

 to disable the dryRun mode, change the dryRun variable to false at the beginning of the script

fixe AM-942

gravitee-io/issues#9288


passionne (approved):
LGTM

PR #3091, opened 2023-10-10
fix: use the usernameField configuration to rename the username into …

leleueri:
…the IDP collection

related-to AM-942

# description

The field which contains the username maybe different from "username" for collection that are not the system ones. This fix, read the value from the configuration.


aelamrani (approved):

leleueri:
we script the integration tests as we updated a script that is not related to Java code

PR #3092, opened 2023-10-10
fix: improve mongo script to recreate indexes

leleueri:
fixe AM-944


passionne (approved):

leleueri:
didn't pass all the checks because the change impact a script that is not used for testing

PR #3096, opened 2023-10-10
fix: add commitQuorum value and fix usename unique constraint definition

leleueri:
fixe AM-944


passionne (approved):
Good to know: https://www.mongodb.com/docs/manual/reference/command/createIndexes/#command-fields

leleueri:
no processed the IT because the script is not used into them

PR #3097, opened 2023-10-10
fix: manage migration of OrganizationUsers duplicates

leleueri:
fixes AM-942

## Description

Manage Organization duplicate during execution of migration script.
This update only manage cockpit users and organization users created with the default gravitee ognization user idp


leleueri:
Integ test not executed as the changes are only targeting a mongo script

PR #3121, opened 2023-10-18
fix: evaluate password on server side for ResetPassword and ConfirmRe…

leleueri:
…gistration to allow the validation of the UserProfileInformation rule

fixes AM-873

fixes gravitee-io/issues#9276

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/resources/views/default/registration.html
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/resources/views/default/registration.html
@@ -170,6 +170,6 @@
 
     const passwordSettings = /*[[${passwordSettings}]]*/ null;
 </script>
-<script th:if="${error == null && success == null}" th:src="@{assets/js/password-validation-v2.js}"></script>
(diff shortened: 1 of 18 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/resources/views/default/registration.html
```diff
@@ -170,6 +170,6 @@
 
     const passwordSettings = /*[[${passwordSettings}]]*/ null;
```
tcompiegne, line 171: Missing ```const passwordValidation = /*[[${passwordValidation}]]*/ null;``` no ?

tcompiegne (changes requested):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/resources/views/default/registration.html
leleueri: No, this is the variable used to store the GW URL that the v3 script need to call in order to valdiate the password.
As for registration page we rely on the form fields to evaluate the "User Profile Information" rule, we do not need for this value and the GW will not provide this variable during the template rendering

PR #3168, opened 2023-10-25
AM-1013 email verification disable confirm

ashraf706:
**How To Test**

Steps are written in the JIRA issue

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
@@ -41,6 +41,8 @@ import io.gravitee.am.service.i18n.GraviteeMessageResolver;
 import io.gravitee.am.service.reporter.builder.AuditBuilder;
 import io.gravitee.am.service.reporter.builder.EmailAuditBuilder;
 import io.vertx.rxjava3.core.MultiMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -50,8 +52,13 @@ import java.io.StringReader;
 import java.io.StringWriter;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 
+import static io.gravitee.am.common.oauth2.Parameters.CLIENT_ID;
 import static io.gravitee.am.common.web.UriBuilder.encodeURIComponent;
 import static io.gravitee.am.service.utils.UserProfileUtils.preferredLanguage;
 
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
```diff
@@ -60,7 +69,10 @@
  * @author GraviteeSource Team
  */
 public class EmailServiceImpl implements EmailService, InitializingBean {
+    private static final Logger logger = LoggerFactory.getLogger(EmailServiceImpl.class);
 
+    public static final String CLIENT_ID = "client_id";
+    public static final String REDIRECT_URI = "redirect_uri";
```
leleueri, line 75: **todo:** We should use the constants defined into the oauth2.Parameters class https://github.com/gravitee-io/gravitee-access-management/blob/master/gravitee-am-common/src/main/java/io/gravitee/am/common/oauth2/Parameters.java

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
```diff
@@ -232,8 +244,12 @@ private Map<String, Object> prepareEmailParams(User user, Client client, Integer
...
         }
 
-        if (client != null) {
-            queryParams.add("client_id", encodeURIComponent(client.getClientId()));
+        if (client != null && !queryParams.contains(CLIENT_ID)) {
+            queryParams.add(CLIENT_ID, encodeURIComponent(client.getClientId()));
+        }
+
+        if(queryParams.contains(REDIRECT_URI)) {
+            encodeAndReplaceRedirectURIQueryParam(queryParams);
```
leleueri, line 252: **thought**: I'm not sure that this is the fix we should apply :thinking: I wondering instead with we provide a redirect_uri parameter to the validationRegistration endpoint as it not used... wdyt ?

leleueri: I just check, and for the forgotPassword for example, we are not providing the parameters coming from the request. That make sens IMO, I wondering why it has been done in another way for registrationValidation... I think we should do the same as forgotPassword. wdyt ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
ashraf706: done

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
ashraf706: code updated

leleueri (approved):

PR #3182, opened 2023-10-31
feat(helm): backport HTTP 2 configuration from gravitee.yml to helm values

passionne:

https://gravitee.atlassian.net/browse/DEVOPS-283

Previously the `http.alpn` option was hardcoded as `true` in `gravitee.yml`.
Now we add the option `gateway.http.alpn` to configure it in helm `values.yml`
which is set to `true` by default for backward compatibility.

## :memo: Test scenarios 

See unit test


sebdevaux (commented):
Could you please update the changelog and chart.yaml?

sebdevaux (approved):

leleueri (approved):

PR #3184, opened 2023-10-31
refactor(helm): rework email configuration to use ident

passionne:

https://gravitee.atlassian.net/browse/DEVOPS-180

* rework email configuration in helm `values.yml` in `smtp.email`
* rework also the `notifiers.email`

Note: previously, some properties was different between helm values and
`gravitee.yml` file:

* `starttlsEnable` to `starttls.enable`
* `sslTrust` to `ssl.trust`
* `sslProtocols` to `ssl.protocols`

And on another side, the `notifiers` email configuration in
`gravitee.yml` use these camelcase properties...

This refactoring take care to be backward compatible with old and new
configuration. (see unit test)

* fix also the conversion from `starttlsEnable` to `startTLSEnabled`
* remove unwanted properties from `smtp` when include to `notifiers`
* flatten the `.smtp.properties` into `api.notifiers.email` directly

## :memo: Test scenarios 

see unittest

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -25,3 +25,6 @@ annotations:
   artifacthub.io/changes: |
     - Make optional HTTP2 request processing via `gateway.http.alpn` set at `true` by default.
     - Manage consent setting to log IP & User-Agent into the audits.
+    - Allow users to define extra manifests
+    - refactoring of email configuration now `smtp.email` and `api.notifiers.email` transpose helm values to `gravitee.yml` with some process to be backward compatible.
+
--- a/helm/templates/api/api-configmap.yaml
+++ b/helm/templates/api/api-configmap.yaml
@@ -246,65 +246,18 @@ data:
   {{- end }}
 
     notifiers:
-      email: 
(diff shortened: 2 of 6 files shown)


leleueri (changes requested):
**issue:** Except if I missed something, your commit changes the behaviour. Previously, it was possible to initialize the notifier part using the smtp section to avoid settings duplications. Looks like it is not the case anymore or at least there is not test case to validate this point.

passionne:
> **issue:** Except if I missed something, your commit changes the behaviour. Previously, it was possible to initialize the notifier part using the smtp section to avoid settings duplications. Looks like it is not the case anymore or at least there is not test case to validate this point.

Nope, this is the test which ensure notifiers configuration is well merged with smtp part: https://github.com/gravitee-io/gravitee-access-management/pull/3184/files#diff-43ca7103cf4aef604c0a1982d4f40c84d5465229ecf0570ec0e33821b22bcb53R189
And the notifiers part can overwrite smtp part 😁

leleueri (approved):

PR #3217, opened 2023-11-09
AM-867: conditionally print log based on exception type

ashraf706:


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/CookieSessionHandler.java
```diff
@@ -121,10 +122,26 @@ public void handle(RoutingContext context) {
...
+                    })
+                    .doFinally(context::next)
+                    .subscribe(
+                            success -> {
+                            }, error -> {
+                            }
+                    );
+        } catch (Exception ex) {
+            logger.error("Some exception reported");
+        }
```
leleueri, line 144: **suggestion**: all unexpected exception should be logged at WARN, we should restrict the INFO level only for expired or prematured token IMO.

```suggestion
        sessionObs
                .doFinally(context::next)
                .subscribe(
                   success -> logger.trace("Session restored successfully"),
                   error -> {
                        final Throwable cause = error.getCause();
                        if (cause instanceof PrematureJWTException | error instanceof ExpiredJWTException) {
                            logger.info("Unable to restore the session: {}", cause.getMessage());
                        } else {                            
                          logger.warn("Unable to restore the session: {}", cause.getMessage());
                        }
                      }   
                );
```

ashraf706: updated as suggested

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/CookieSessionHandler.java
```diff
@@ -121,10 +123,23 @@ public void handle(RoutingContext context) {
...
+                                final Throwable cause = error.getCause();
+                                if (cause instanceof PrematureJWTException | error instanceof ExpiredJWTException) {
+                                    logger.info("Unable to restore the session: {}", cause.getMessage());
+                                } else {
+                                    logger.warn("Unable to restore the session: {}", cause.getMessage());
+                                }
+                            }
+                    );
+        } catch (Exception ex) {
+            logger.error("Some exception reported");
```
leleueri, line 141: **todo:** I don't see where an exception can be thrown here, can you explain ? If an exception can be raised, I think the message should be changed and the exception should be provided as parameter of the error method to log the stacktrace as it will be an unexpected error. wdyt ?

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/handler/impl/CookieSessionHandler.java
ashraf706: This try catch is pointless. Removed.

leleueri (approved):

PR #3237, opened 2023-11-23
fix: improve default password policy to follow latest OWASP recommend…

leleueri:
…ations

fixes AM-526


aurelien-pacaud:
I trying to apply the same pattern to our instance but I found "an issue". 

The default PasswordValidation implementation validate if the password size is less or equal to 64: https://github.com/gravitee-io/gravitee-access-management/blob/491e8cc652650be1c0c21145a52a700d283f64b9/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/password/impl/DefaultPasswordValidatorImpl.java#L49

ashraf706 (approved):

PR #3239, opened 2023-11-27
Add option to choose deployment type

Oicho:
## :pencil2: Allow to deploy api and gateway as statefulset
This is following our discussion with alban leduc

A few unclear things I want your input since I do not know how you are managing  releases.
- I kept the default for gateway as deployment on apim it's a statefulset but I do not want to impact gravitee user. I'll amend my pr if needed
- On various files  there's a bunch of trailing whitespaces by default vscode remove (rightfully so IMO). I can readd them if you want for a cleaner diff
- I took the liberty of bumping the chart version to 4.1.6 since I saw a 4.1.5 in the change low

## :man: @gravitee-aleduc  FYI

FYI I did a quick before and after helm template to make sure there was no change by default


helm/templates/gateway/gateway-deployment.yaml
```diff
@@ -3,7 +3,7 @@
 {{- $serviceAccount := include "am.serviceAccount" . -}}
 {{- $logbackVolumeName := include "gateway.logbackVolumeName" . -}}
 apiVersion: apps/v1
-kind: Deployment
+kind: {{ .Values.gateway.type }}
```
benoitgravitee, line 6: Same as above

leleueri:
Hi,
First of all, thanks for you proposal and we are sorry for the late answer.
After internal discussions, we will decline your change proposal. Management API and Gateway are stateless, so in our opinion it make no sens to propose a StatefulSet deployment.

PR #3263, opened 2023-12-06
AM-52: reset password policy settings

mstasgravitee:
AM-52

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
@@ -17,6 +17,7 @@ package io.gravitee.am.service.model;
 
 import io.gravitee.am.model.CorsSettings;
 import io.gravitee.am.model.Domain;
+import io.gravitee.am.model.PasswordSettings;
 import io.gravitee.am.model.SelfServiceAccountManagementSettings;
 import io.gravitee.am.model.VirtualHost;
 import io.gravitee.am.model.account.AccountSettings;
@@ -220,8 +221,14 @@ public class PatchDomain {
             }
         }
 
-        if (this.passwordSettings != null) {
-            this.passwordSettings.ifPresent(ps -> toPatch.setPasswordSettings(ps.patch(toPatch.getPasswordSettings())));
+        if (passwordSettings != null && passwordSettings.isPresent()) {
+            var ps = passwordSettings.get();
+            if (ps.getInherited() != null && ps.getInherited().isPresent() && Boolean.TRUE.equals(ps.getInherited().get())) {
+                toPatch.setPasswordSettings(null);
+                setPasswordSettings(Optional.empty());
+            } else {
+                toPatch.setPasswordSettings(ps.patch(toPatch.getPasswordSettings()));
+            }
         }
 
         if (this.getSaml() != null && this.getSaml().isPresent()) {
@@ -260,7 +267,7 @@ public class PatchDomain {
             requiredPermissions.add(Permission.DOMAIN_SETTINGS);
         }
 
-        if(alertEnabled != null && alertEnabled.isPresent()) {
+        if (alertEnabled != null && alertEnabled.isPresent()) {
             requiredPermissions.add(Permission.DOMAIN_ALERT);
         }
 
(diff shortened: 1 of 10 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
```diff
@@ -220,8 +221,14 @@ public Domain patch(Domain _toPatch) {
...
-        if (this.passwordSettings != null) {
-            this.passwordSettings.ifPresent(ps -> toPatch.setPasswordSettings(ps.patch(toPatch.getPasswordSettings())));
+        if (passwordSettings != null && passwordSettings.isPresent()) {
+            var ps = passwordSettings.get();
+            if (ps.getInherited() != null && ps.getInherited().isPresent() && Boolean.TRUE.equals(ps.getInherited().get())) {
+                toPatch.setPasswordSettings(null);
+                setPasswordSettings(Optional.empty());
+            } else {
+                toPatch.setPasswordSettings(ps.patch(toPatch.getPasswordSettings()));
+            }
```
leleueri, line 231: **todo**: Not sure about this way to manage the reset. You are considering the presence of the inherited boolean but everywhere else into the code for patch/update action we consider the Optional.isEmpty() as the condition to reset the value to null. 

Basically, in the request payload:
* if the field is missing, the Java bean will have a null for this attribute,  in that case we do not change the value
* if the field is set to null, the Java bean will have a Optional.empty() for this attribute,  in that case we set the value to null
* if the field is not null, the Java bean will have a Optional.from() for this attribute,  in that case we change the value with the content of the Optional.

The change should be something like this IMO to be consistent

```suggestion
        if (passwordSettings != null) {
            if (passwordSettings.isEmpty()) {
                toPatch.setPasswordSettings(null);
            } else {
                toPatch.setPasswordSettings(this.passwordSettings.get().patch(toPatch.getPasswordSettings()));
            }
```
wdyt ?

gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
```diff
@@ -260,7 +267,7 @@ public Set<Permission> getRequiredPermissions() {
             requiredPermissions.add(Permission.DOMAIN_SETTINGS);
         }
 
-        if(alertEnabled != null && alertEnabled.isPresent()) {
+        if (alertEnabled != null && alertEnabled.isPresent()) {
```
leleueri, line 270: **question**: are we sure that this change will not break the current behaviour ? :thinking: This change will probably allow to disable the Alert feature een if the user doesn't have the DOMAIN_ALERT permission. I think this change should be revert. (I can see that we test the `isPresent` in the previous statement but don't know if we have to apply the same for alerts)

gravitee-am-ui/src/app/domain/settings/password-policy/domain-password-policy.component.ts
```diff
@@ -132,12 +133,38 @@ export class DomainPasswordPolicyComponent implements OnInit {
...
+    this.dialogService
+      .confirm('Reset password policy', 'Are you sure you want to reset your password policy settings and use the default one?')
+      .subscribe((res) => {
+        if (res) {
+          this.reset();
+        }
+      });
+  }
+  reset(): void {
+    this.domainService.patchPasswordSettings(this.domainId, { passwordSettings: { inherited: true } }).subscribe(() => {
```
leleueri, line 155: **todo**: based on the comment on gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java we should set passwordSettings to null.

```suggestion
    this.domainService.patchPasswordSettings(this.domainId, { passwordSettings: null }).subscribe(() => {
```

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
mstasgravitee: It has been added only a space to fix format here.

leleueri: :tired_face: sorry about that one

gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchDomain.java
mstasgravitee: image
image

leleueri: ok, as discussed we can rely on null.

PR #3280, opened 2023-12-12
fix: refactor JDBC autheitcationProvider to avoid COnnection leak

leleueri:
fixes AM-1186

gravitee-io/issues#9426


gravitee-am-identityprovider/gravitee-am-identityprovider-jdbc/src/main/java/io/gravitee/am/identityprovider/jdbc/JdbcAbstractProvider.java
```diff
@@ -106,4 +113,34 @@ protected void computeMetadata(Map<String, Object> claims) {
...
+
+    protected Flowable<Result> query(String sql, Object... args) {
+        return Single.fromPublisher(connectionPool.create())
+                .toFlowable()
+                .flatMap(connection ->
+                        query(connection, sql, args)
+                                .doFinally(() -> Completable.fromPublisher(connection.close()).subscribe()));
+    }
+
+    protected Flowable<Result> query(Connection connection, String sql, Object... args) {
```
mstasgravitee, line 125: Could we make this method private to not allow using it and remove "selectUserByUsername" JdbcUserProvider the one with connection parameter in JdbcUserProvider?

leleueri: We can't make it private as it is used by the JdbcUserProvider implementation during the provider startup. 
I removed the selectUserByUsername.

mstasgravitee: yes, that's true, but it can be removed and replaced by the query without a connection param.

leleueri: yes, but for efficiency, I do prefer keeping this code as it is.

PR #3302, opened 2023-12-15
fix: change the overview page for apps to highlight the fact clientSe…

leleueri:
…cret need to be provided

fixes AM-1198

### DIFF
--- a/gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
+++ b/gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
@@ -90,41 +90,29 @@
             <p>An extra step is required if the response_type=code, exchange the authorization_code for an access_token.</p>
             <p><b>Note:</b> remember to adapt the request to use the client secret you copied during the application creation</p>
           </blockquote>
-          <div class="code">
-            <button
-              mat-icon-button
-              matSuffix
-              matTooltip="Copy to clipboard"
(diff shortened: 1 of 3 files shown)


gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
```diff
@@ -91,37 +91,14 @@ <h2>Initiate the Login Flow</h2>
...
-              (click)="safeAuthorizationHeader = authorizationHeader"
-            >
-              <mat-icon>visibility</mat-icon>
-            </button>
 
             <pre #pre_authorization_code_exchange class="multiline">
 curl -X POST \
   {{ baseUrl }}/oauth/token \
-  -H 'Authorization: Basic {{ safeAuthorizationHeader }}' \
+  -u '{{ clientId }}:&lt;add you secret here&gt;' \
```
mstasgravitee, line 101: thought: there is no info that it has to be encoded with base64. Would it be better to change it to "******" or "access_token" and then describe what it means above in the text? 

Regarding this line, 249 and 161

leleueri: the '-u' option will encode in Base64. I will update the "add your secret here" with a "<_client_secret_>"

ashraf706 (approved):

PR #3303, opened 2023-12-15
fix: preserve the clientSecret on application PATCH action when app w…

leleueri:
…as create befor the 4.2.0

fixes AM-1199

### DIFF
--- a/gravitee-am-service/src/test/java/io/gravitee/am/service/ApplicationServiceTest.java
+++ b/gravitee-am-service/src/test/java/io/gravitee/am/service/ApplicationServiceTest.java
@@ -1015,6 +1015,42 @@ public class ApplicationServiceTest {
     }
 
     @Test
+    public void update_clientCredentials_keep_clientSecret() {
+        // if application has been created before 4.2, the clientSecret field must be preserved until
+        final String APP_ID = "appId";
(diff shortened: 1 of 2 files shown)


gravitee-am-service/src/test/java/io/gravitee/am/service/ApplicationServiceTest.java
```diff
@@ -1014,6 +1014,42 @@ public void update_clientCredentials_ok() {
         verify(applicationRepository, times(1)).update(any(Application.class));
     }
 
+    @Test
+    public void update_clientCredentials_keep_clientSecret() {
+        // if application has been created before 4.2, the clientSecret field must be preserved until
```
mstasgravitee, line 1019: why must it be preserved?

leleueri: for backward compatibility.
If an application create before AM 4.2 is updated, the client secret have to be perserved in the OAuth2 settings to make it works with GW 4.1. If the clientSecret is renewed then the clientSecert into oauth2 settings can be reset to null.

mstasgravitee (approved):

PR #3335, opened 2023-12-20
fix: provide the right parameters for an authorization_code flow

leleueri:
fixes AM-1198

### DIFF
--- a/gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
+++ b/gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
@@ -98,7 +98,7 @@
             <pre #pre_authorization_code_exchange class="multiline">
 curl -X POST \
               {{ baseUrl }}/oauth/token \
-  -u '{{ clientId }}:&lt;<i>client_secret</i>&gt;' \
+  -u '{{ clientId }}:|client_secret|' \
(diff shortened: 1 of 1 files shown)


gravitee-am-ui/src/app/domain/applications/application/overview/overview.component.html
```diff
@@ -112,7 +112,9 @@ <h2>Initiate the Login Flow</h2>
 curl -X POST \
               {{ baseUrl }}/oauth/token \
   -H 'Content-Type: application/x-www-form-urlencoded' \
-  -d 'grant_type=client_credentials&client_id={{ clientId }}&client_secret=&lt;<i>client_secret</i>&gt;'</pre
+  -d 'grant_type=authorization_code&code=|code|&redirect_uri={{ redirectUri }}&client_id={{
+                clientId
+              }}&client_secret=&lt;<i>client_secret</i>&gt;{{ getCodeVerifier() }}'</pre
```
mstasgravitee, line 117: though: 

should it be one way to say that there is a variable to replace?
if we have for client secret: &lt;<i>client_secret</i>&gt;
then the same for code: &lt;<i>code</i>&gt;

What do you think?

mstasgravitee (approved):

leleueri:
ignore IT execution as only on UI page has been updated

PR #3352, opened 2023-12-21
fix: revert change that introduce regression

leleueri:
related to AM-1194


leleueri:
ignore IT as only a page in the UI has been updated

leleueri (commented):
this change need to be rolled back. It has been apply in all mergify PR as Integration tests were failing due to this addition.

ashraf706 (approved):

PR #3366, opened 2023-12-27
AM-1200: do not allow null username

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-1200


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractUserService.java
```diff
@@ -180,10 +182,13 @@ public Maybe<User> findByExternalIdAndSource(ReferenceType referenceType, String
     @Override
     public Single<User> create(ReferenceType referenceType, String referenceId, NewUser newUser) {
         LOGGER.debug("Create a new user {} for {} {}", newUser, referenceType, referenceId);
+        if (StringUtils.isBlank(newUser.getUsername())) {
+            return Single.error(new UserInvalidException("Field [username] is required"));
+        }
```
leleueri, line 187: **nitpick (non-blocking)**: I understand you are doing this check to avoid a useless DB request but this check is done into the `create` method line 233. Maybe we can remove this one :thinking:  wdyt ?

leleueri (approved):
**General thought.**

I wondering if we should move the `StringUtils.isBlank(user.getUsername())` check you added into the UserValidator method and create two methods inside this validator (validate for create action and a validate for update action, both relying on a single validate method which takes in parameter a boolan to determine if the null value need to be checked for the username) and call the right method based on the context to centralize the user validation checks :thinking: 

I know UserValidator is sometimes called late into the user creation process but it looks more maintenable to have everything into the validator instread of spread this check accross the services...

PR #3371, opened 2023-12-29
AM-1136: don't allow to upload expired certificate

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-1136

The backend logic is implemented and errors are adjusted to show proper popup messages on the frontend side. There are no additional fronted changes due to the lack of new mockups. They can be adjusted as part of a new task.

The changes include:
- check the certificate expiration date
- valid certificate password
- clear form when an error occurs 

p12 / jks expired
image

p12 / jks invalid password
image

p12 ok
p12-ok

jks ok
jks-ok

update error
 image
 image
image


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -490,48 +465,56 @@ public Single<Certificate> rotate(String domain, User principal) {
...
+            default -> throw new CertificateException("Unsupported certification type: " + type);
+        };
+        store.load(new ByteArrayInputStream(cert), storePass.toCharArray());
+        var expiryDate = ((X509Certificate) store.getCertificate(alias)).getNotAfter();
+        if (Instant.now().isAfter(expiryDate.toInstant())) {
+            throw new CertificateException("Uploading certificate is already expired");
+        }
+        store.deleteEntry(alias);
+        return expiryDate;
+    }
```
leleueri, line 481: **TODO:** I don't think we have to reimplement the KeyStore loading here. We should instead provide a "isValid" method into the CertificateManager that will create a CertificateProvider and then reject the cert if the expiration date is not valid. https://github.com/gravitee-io/gravitee-access-management/blob/833fb5e34162991f5d30fd43d9a198237d24ecae/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/CertificateManagerImpl.java#L152

We can't use the loadCertificate as it is but we can extract the certificate provider creation to use the plugin instead of reimplementing the logic. In that way, if we introduce a new certificate plugin type, we will not forget to update this switch
```
            var providerConfig = new CertificateProviderConfiguration(certificate);
            var certificateProvider = certificatePluginManager.create(providerConfig);
```

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -180,71 +193,57 @@ public Flowable<Certificate> findAll() {
...
+                    }
+                })
+                .flatMap(certificate -> {
+                    try {
+                        isValid(certificate);
+                        return certificateRepository.create(certificate);
+                    } catch (Exception e) {
+                        return Single.error(e);
+                    }
+                })
```
leleueri, line 240: **suggestion**: instead of 'isValid' which return a void, maybe we should name it 'check' or 'validate' and make it returns a Certificate. In that way, we will be able to write it like this;
```suggestion
                .map(this::validate)
                .flatMap(certificateRepository::create)
```

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -268,83 +267,65 @@ public CertificateSchema getSchema() {
...
+                    return Single.just(certificateToUpdate);
+                })
+                .flatMap(certificate -> {
+                    try {
+                        isValid(certificate);
+                        return certificateRepository.update(certificate);
+                    } catch (Exception e) {
+                        return Single.error(e);
+                    }
+                })
```
leleueri, line 320: same comment as line 240

leleueri (changes requested):
Minor comments about the code but I can't approve until integration tests pass

postman/collections/graviteeio-am-openid-ciba-collection.json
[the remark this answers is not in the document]
mstasgravitee: fixing integration tests is not done yet

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
mstasgravitee: it will be wrong error handling: 
java.lang.NullPointerException: The single returned by the mapper is null

leleueri (commented):
LGTM but are you sure that the certificate form is sucessfully reset in case of error ?

leleueri (approved):

PR #3378, opened 2024-01-05
chore(gatling): addition of a simulations to create and use multiple …

leleueri:
…domains

fix AM-1223 - simulation
fix AM-1224 - adition of grafana dashboard


leleueri:
by pass validation as it is purely techincal and used for testing purpose, we do not provide this script in a release

PR #3380, opened 2024-01-08
fix: add javax-annotation dependency to initialize Handler in AE Comm…

leleueri:
…andHandlerMng

fixes AM-1230

gravitee-io/issues#9454

### DIFF
--- a/gravitee-am-service/pom.xml
+++ b/gravitee-am-service/pom.xml
@@ -180,6 +180,11 @@
             <groupId>jakarta.inject</groupId>
             <artifactId>jakarta.inject-api</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


gravitee-am-service/pom.xml
```diff
@@ -180,6 +180,11 @@
             <groupId>jakarta.inject</groupId>
             <artifactId>jakarta.inject-api</artifactId>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/javax.annotation/javax.annotation-api -->
```
ashraf706, line 183: suggestion: perhaps the mavenrepository url is not intentional

ashraf706 (approved):

PR #3383, opened 2024-01-09
fix: bypass BodyHandler on GET method

leleueri:
fixes AM-1022

gravitee-io/issues#9352


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
```diff
@@ -787,7 +787,8 @@ private void staticHandler(Router router) {
     }
 
     private void bodyHandler(Router router) {
-        router.route().handler(BodyHandler.create());
+        router.route()
+                .handler(new io.gravitee.am.gateway.handler.root.resources.handler.ConditionalBodyHandler(environment));
```
ashraf706, line 791: Suggestion: perhaps you can import the class and then it will be just `new ConditionalBodyHandler`

ashraf706 (approved):

PR #3396, opened 2024-01-15
AM-1219: fix sync duplicated events

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-1219

before
before
after
after

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/CertificateManagerImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/CertificateManagerImpl.java
@@ -47,8 +47,8 @@ import java.util.concurrent.ConcurrentMap;
 public class CertificateManagerImpl extends AbstractService<CertificateManager> implements CertificateManager, EventListener<CertificateEvent, Payload> {
 
     private static final Logger logger = LoggerFactory.getLogger(CertificateManagerImpl.class);
-    private static final long retryTimeout = 10000;
-    private ConcurrentMap<String, CertificateProvider> certificateProviders = new ConcurrentHashMap<>();
+    private static final long RETRY_TIMEOUT = 10000;
+    private final ConcurrentMap<String, CertificateProvider> certificateProviders = new ConcurrentHashMap<>();
 
     @Autowired
     private CertificatePluginManager certificatePluginManager;
@@ -112,7 +112,7 @@ public class CertificateManagerImpl extends AbstractService<CertificateManager> 
                 return Maybe.empty();
             }
             // retry
-            while (certificateProvider == null && System.currentTimeMillis() - startTime < retryTimeout) {
+            while (certificateProvider == null && System.currentTimeMillis() - startTime < RETRY_TIMEOUT) {
                 certificateProvider = certificateProviders.get(certificateId);
             }
             return certificateProvider == null ? Maybe.empty() : Maybe.just(certificateProvider);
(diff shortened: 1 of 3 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -306,10 +305,10 @@ public Single<Certificate> update(String domain, String id, UpdateCertificate up
...
                 .map(this::getValid)
                 .flatMap(certificate -> certificateRepository.update(certificate))
-                // create event for sync process
-                .flatMap(certificate1 -> {
-                    Event event = new Event(Type.CERTIFICATE, new Payload(certificate1.getId(), ReferenceType.DOMAIN, certificate1.getDomain(), Action.UPDATE));
-                    return eventService.create(event).flatMap(__ -> Single.just(certificate1));
+                .doOnSuccess(certificate -> {
+                    auditService.report(AuditBuilder.builder(CertificateAuditBuilder.class).principal(principal).referenceId(domain).referenceType(ReferenceType.DOMAIN).type(EventType.CERTIFICATE_UPDATED).certificate(certificate));
+                    Event event = new Event(Type.CERTIFICATE, new Payload(certificate.getId(), ReferenceType.DOMAIN, certificate.getDomain(), Action.UPDATE));
+                    eventService.create(event).subscribe();
```
leleueri, line 311: **question**: do we really want to fire&forget the event creation ? :thinking: If an error occures the event may not be persist in DB and so the GW will never see the change. Before this update, an error will be display on the UI, after this change, the UI will display a successful response but the GWs will not see the updates...

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -232,10 +232,10 @@ public Single<Certificate> create(String domain, NewCertificate newCertificate,
                 })
                 .map(this::getValid)
                 .flatMap(certificate -> certificateRepository.create(certificate))
-                // create event for sync process
-                .flatMap(certificate -> {
+                .doOnSuccess(certificate -> {
```
leleueri, line 235: question: do we really want to fire&forget the event creation ? 🤔 If an error occures the event may not be persist in DB and so the GW will never see the change. Before this update, an error will be display on the UI, after this change, the UI will display a successful response but the GWs will not see the updates...

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -320,33 +319,33 @@ public Single<Certificate> update(String domain, String id, UpdateCertificate up
...
-                    // create event for sync process
+                .flatMap(certificate -> certificateRepository.delete(certificateId).andThen(Single.just(certificate)))
+                .doOnSuccess(certificate -> {
+                    auditService.report(AuditBuilder.builder(CertificateAuditBuilder.class).principal(principal).type(EventType.CERTIFICATE_DELETED).certificate(certificate));
                     Event event = new Event(Type.CERTIFICATE, new Payload(certificate.getId(), ReferenceType.DOMAIN, certificate.getDomain(), Action.DELETE));
-                    return Completable.fromSingle(certificateRepository.delete(certificateId)
-                            .andThen(eventService.create(event)))
-                            .doOnComplete(() -> auditService.report(AuditBuilder.builder(CertificateAuditBuilder.class).principal(principal).type(EventType.CERTIFICATE_DELETED).certificate(certificate)))
-                            .doOnError(throwable -> auditService.report(AuditBuilder.builder(CertificateAuditBuilder.class).principal(principal).type(EventType.CERTIFICATE_DELETED).throwable(throwable)));
+                    eventService.create(event).subscribe();
```
leleueri, line 336: **question**: do we really want to fire&forget the event creation ? 🤔 If an error occures the event may not be persist in DB and so the GW will never see the change. Before this update, an error will be display on the UI, after this change, the UI will display a successful response but the GWs will not see the fact that this certificate has been deleted...

leleueri (approved):
No need to submit another review if you accept the proposed changes

PR #3418, opened 2024-01-22
AM-1091: use common banner for license notification

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-1091

## :id: Reference related Pull Request. 
https://github.com/gravitee-io/gravitee-api-management/pull/6441

Screenshot from 2024-01-22 09-05-08

Screenshot from 2024-01-22 08-59-47

Screenshot from 2024-01-22 08-59-34


gravitee-am-ui/src/app/components/sidenav/sidenav.component.ts
```diff
@@ -76,48 +75,11 @@ export class SidenavComponent implements OnInit, OnDestroy {
     this.licenseService
       .getLicense$()
```
jourdiw, line 76: If you want, you can use the method `getExpirationDate$()` to return just the expiration date.

jourdiw (approved):

gravitee-am-ui/src/app/components/sidenav/sidenav.component.ts
mstasgravitee: as we discussed: getExpirationDate$() uses 'expiresAt' instead 'expirationDate' from the backend License object. It should be unified somehow to not make a risk of a field change and break frontend.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/LicenseResource.java
```diff
@@ -51,7 +51,7 @@ public void get(@Suspended final AsyncResponse response) {
                 .tier(platformLicense.getTier())
                 .packs(platformLicense.getPacks())
                 .features(platformLicense.getFeatures())
-                .expirationDate(isLicenseExpirationNotifierEnabled ? platformLicense.getExpirationDate() : null)
+                .expiresAt(isLicenseExpirationNotifierEnabled ? platformLicense.getExpirationDate() : null)
```
jourdiw, line 54: Non-blocking: It would be best to refactor `isLicenseExpirationNotifierEnabled` into another endpoint. That way we can return the true expiration date of the license with this endpoint, and can recuperate the info if the license notification is enabled from another endpoint.

jourdiw (approved):

PR #3426, opened 2024-01-23
AM-1249: added missing rate limit audit log

ashraf706:
How to test:
Please check the jira issue AM-1249


gravitee-am-service/src/main/java/io/gravitee/am/service/exception/RateLimitException.java
```diff
@@ -27,6 +29,6 @@ public RateLimitException(String message) {
 
     @Override
     public int getHttpStatusCode() {
-        return 0;
+        return OK_200;
```
mstasgravitee, line 32: Why is the 200 code here? Would not the 429 code number be better than that?

ashraf706: done, didn't know about this 429 :)

PR #3427, opened 2024-01-23
fix: apply timeout on blockingGet made in the REST API

leleueri:
fixes am-2133

gravitee-io/issues#9476


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/JWTAuthenticationFilter.java
```diff
@@ -81,6 +83,9 @@ public class JWTAuthenticationFilter extends AbstractAuthenticationProcessingFil
     @Value("${jwt.cookie-domain:}")
     private String jwtCookieDomain;
 
+    @Value("${http.blockingGet.timeoutMillis:120000}")
+    private long blockingGetTimeoutMillis = -1;
```
mstasgravitee, line 87: Why are there 2 different defaults?

leleueri: it was to be sure that the default value in the annotation was used.
updated

mstasgravitee (approved):

PR #3449, opened 2024-01-30
chore: update release note generation script

leleueri:
Improve Release Note generation script:

* avoid filtering on JIRA issue status (the value depend of the language preferences of the account used to generate the token, so it may not be Done but for example Terminé(e) in french)
* addition of "Public Improvement" as issue to include into the release note

### DIFF
--- a/release/helpers/changelog-helper.mjs
+++ b/release/helpers/changelog-helper.mjs
@@ -2,17 +2,17 @@
  * Get the MarkDown doc formatted changelog for input issues
  *
  * @param title {string}
- * @param issues {Array<{id: string, key: string, githubIssue: string, summary: string, components: Array<string>, issueTypeName: string, statusName: string}>}
+ * @param issues {Array<{id: string, key: string, githubIssue: string, summary: string, components: Array<string>, issueTypeName: string, statusKey: string}>}
  */
 export function getChangelogFor(title, issues) {
-  const authorizedIssueTypes = ['Public Bug', 'Public Security', 'Story'];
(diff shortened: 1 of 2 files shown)


release/helpers/changelog-helper.mjs
```diff
@@ -5,15 +5,12 @@
...
 export function getChangelogFor(title, issues) {
-  const authorizedIssueTypes = ['Public Bug', 'Public Security', 'Story'];
+  const authorizedIssueTypes = ['Public Bug', 'Public Security', 'Public Improvement', 'Story'];
 
   const filteredIssues = issues
     .filter((issue) => {
       return !!issue.issueTypeName && authorizedIssueTypes.includes(issue.issueTypeName);
     })
-    .filter((issue) => {
-      return issue.statusName === 'Done';
```
mstasgravitee, line 15: Should we add more conditions instead of removing the line completely? The done status is interesting for us only.

release/helpers/changelog-helper.mjs
```diff
@@ -5,14 +5,14 @@
...
-  const authorizedIssueTypes = ['Public Bug', 'Public Security', 'Story'];
+  const authorizedIssueTypes = ['Public Bug', 'Public Security', 'Public Improvement', 'Story'];
 
   const filteredIssues = issues
     .filter((issue) => {
       return !!issue.issueTypeName && authorizedIssueTypes.includes(issue.issueTypeName);
     })
     .filter((issue) => {
-      return issue.statusName === 'Done';
+      return issue.statusName === 'done';
```
mstasgravitee, line 15: use key instead of statusName

mstasgravitee (approved):

PR #3511, opened 2024-02-12
fix Bug in IDTokenServiceImpl and with changes with MultiMap in Autho…

igorRosinskyi:
…rizationEnd


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/AuthorizationEndpoint.java
```diff
@@ -117,6 +118,14 @@ private void doRedirect(RoutingContext context, AuthorizationRequest request, Au
             // and thus are transmitted via the HTTP POST method to the Client.
             // Prepare context to render post form.
             final MultiMap queryParams = RequestUtils.cleanParams(response.params());
+            logger.info("RequestUtils.getCleanedQueryParams");
+            logger.info("queryParams.size {}", queryParams.size());
```
mstasgravitee, line 122: Would not be better to use one logger.debug() there? Do we need it at all?

igorRosinskyi: Changed to debug. Thanks.

mstasgravitee (approved):

leleueri:
Thanks you for the PR but I don't understand the purpose of this change.
Looks like it is to fix the same issue as https://github.com/gravitee-io/issues/issues/9179

PR #3588, opened 2024-02-26
fix: remove usage of legacy adaptiveAuthenticationRule

leleueri:
related-to AM-1141

fixes AM-2758


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/loginattempt/LoginAttemptHandler.java
```diff
@@ -80,7 +80,8 @@ public void handle(RoutingContext routingContext) {
 
     private String getAdaptiveRule(Client client) {
```
mstasgravitee, line 81: the method could be renamed

mstasgravitee (approved):

PR #3605, opened 2024-03-01
factors settings in postman tests

mstasgravitee:


postman/collections/graviteeio-am-login-collection-app-version.json
```diff
@@ -3674,7 +3674,7 @@
 											"    var location = postman.getResponseHeader('Location');",
 											"    let domain = pm.environment.get(\"domainHrid\");",
 											"",
-											"    tests['Redirect to mfa enroll'] = location.includes(pm.environment.get('gateway_url') + '/' + domain + '/mfa/enroll') && location.includes('client_id=' + pm.environment.get('clientId4'));",
+											"    tests['Redirect to mfa challenge'] = location.includes(pm.environment.get('gateway_url') + '/' + domain + '/mfa/challenge');",
```
leleueri, line 3677: I don't think it the right fix. The test is testing that we can't sent a request with the "accept_enrollmen" flag to false.
If such request is provided when user can't skip the enrollment, we should come back on the enrollment page.

leleueri (changes requested):

PR #3607, opened 2024-03-01
chore: update mergify config to manage 4.3.x branch

leleueri:


lgw-gravitee (approved):

leleueri:
do not execute all the tests as it is only merfigy config

PR #3623, opened 2024-03-12
fix: http redirection to https

endekovski10:
## :id: 
https://gravitee.atlassian.net/browse/AM-3084

Previously, our default configuration say that even if you enable TLS, we don't redirect HTTP to HTTPS.
By default (without our overwrite config) nginx says:

```
By default the controller redirects (308) to HTTPS if TLS is enabled for that ingress.
```
https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-side-https-enforcement-through-redirect

So both case:

- customer keep default and use HTTP
- customer only enable TLS

will be handled correctly with default nginx configuration.

As we add nginx.ingress.kubernetes.io/ssl-redirect: "false" which is ok by default without TLS.
This mean that even if we enable TLS, the redirection is still disabled.

My proposal is to remove our default setting and keep the nginx default one to make both case (TLS or not) work properly:
enable redirection only if TLS is used.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -704,10 +704,10 @@ gateway:
       - am.example.com
     annotations:
       kubernetes.io/ingress.class: nginx
-      nginx.ingress.kubernetes.io/ssl-redirect: "false"
       nginx.ingress.kubernetes.io/enable-rewrite-log: "true"
       kubernetes.io/app-root: /auth
       kubernetes.io/rewrite-target: /auth
+      # nginx.ingress.kubernetes.io/ssl-redirect: "false"
       # ingress.kubernetes.io/configuration-snippet: "etag on;\nproxy_pass_header ETag;\nproxy_set_header if-match \"\";\n"
       # kubernetes.io/tls-acme: "true"
     #tls:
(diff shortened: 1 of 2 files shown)


helm/values.yaml
```diff
@@ -704,10 +704,10 @@ gateway:
       - am.example.com
     annotations:
       kubernetes.io/ingress.class: nginx
-      nginx.ingress.kubernetes.io/ssl-redirect: "false"
```
mstasgravitee, line 707: why is it removed instead of setting it as true?

pom.xml
```diff
@@ -31,7 +31,7 @@
     <artifactId>gravitee-am-parent</artifactId>
     <packaging>pom</packaging>
 
-    <version>4.0.17-SNAPSHOT</version>
+    <version>4.0.18-SNAPSHOT</version>
```
mstasgravitee, line 34: why is this version updated?

mstasgravitee (changes requested):

helm/values.yaml
passionne: As mentioned in the PR comment, if we force it to true, then if you want to use HTTP without TLS, it will fail.
The proposal here is to keep default config to let both case: with or without TLS works.

pom.xml
passionne: Sorry, this was a mistake in the PR commits. Fixed.

leleueri (changes requested):
Can you please add this change into the changelogs?

mstasgravitee (approved):

leleueri (approved):

PR #3635, opened 2024-03-14
AM-2161: on login=prompt authorized used should be removed from the s…

lgw-gravitee:
…ession


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/AuthorizationRequestParseParametersHandler.java
```diff
@@ -119,12 +118,18 @@ private void parsePromptParameter(RoutingContext context) {
...
-                    context.clearUser();
+                    clearUser(context);
                 }
             }
         }
     }
 
+    private void clearUser(RoutingContext ctx){
+        ctx.clearUser();
+        ctx.session().remove(STRONG_AUTH_COMPLETED_KEY);
```
leleueri, line 129: Beware in 4.3, this one may not be enough.

leleueri (commented):
LGTM but I wondering if we can instead of just removing the key from the session, clear completly the session. :thinking: 

Anyway, I think this fix (if a complete session cleaning) is not possible, can only be apply on 4.1.x & 4.2.x branch, another Should be open for 4.3.x and master as I'm not sure if removing the STRONG_AUTH key will be enough in 4.3.

PR #3643, opened 2024-03-15
fix: save factor metadata on challenge step if the provider is using …

leleueri:
…a moving factor

     adapt the MFA audit event type based on the challenge context (enrollment or challenge)

fixes AM-2853


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeEndpoint.java
```diff
@@ -336,19 +337,19 @@ private Handler<AsyncResult<Void>> verifyHandler(RoutingContext routingContext,
             }
             // save enrolled factor if needed and redirect to the original url
             routingContext.session().put(ConstantKeys.MFA_FACTOR_ID_CONTEXT_KEY, factorId);
-            if (isEnrolling(enrolledFactor)) {
+            if (enrolling || factorProvider.useVariableFactorSecurity(factorContext)) {
```
mstasgravitee, line 340: can you add a test for this case?

mstasgravitee (approved):

PR #3661, opened 2024-03-18
AM-2847 default factor glitch

lgw-gravitee:
https://gravitee.atlassian.net/browse/AM-2847

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchFactorSettings.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchFactorSettings.java
@@ -36,14 +36,14 @@ public class PatchFactorSettings {
 
     public FactorSettings patch(FactorSettings _toPatch) {
         FactorSettings toPatch = _toPatch == null ? new FactorSettings() : new FactorSettings(_toPatch);
-        if ((hasText(toPatch.getDefaultFactorId()) && defaultFactorId.isEmpty())
-                || (toPatch.getApplicationFactors() == null || toPatch.getApplicationFactors().isEmpty()) && (applicationFactors.isEmpty() || applicationFactors.get().isEmpty())) {
-            throw new InvalidParameterException("Default factor is not set");
-        }
         SetterUtils.safeSet(toPatch::setDefaultFactorId, getDefaultFactorId());
         getApplicationFactors().ifPresent(patchApplicationFactorSettings -> toPatch.setApplicationFactors(
                 applicationFactorsPath(toPatch.getApplicationFactors(), patchApplicationFactorSettings)
         ));
+        boolean hasFactors = toPatch.getApplicationFactors() != null && !toPatch.getApplicationFactors().isEmpty();
+        if(hasFactors && !hasText(toPatch.getDefaultFactorId())){
+            throw new InvalidParameterException("Default factor is not set");
+        }
         return toPatch;
     }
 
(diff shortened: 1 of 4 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchFactorSettings.java
```diff
@@ -36,14 +36,14 @@ public class PatchFactorSettings {
...
-        if ((hasText(toPatch.getDefaultFactorId()) && defaultFactorId.isEmpty())
-                || (toPatch.getApplicationFactors() == null || toPatch.getApplicationFactors().isEmpty()) && (applicationFactors.isEmpty() || applicationFactors.get().isEmpty())) {
-            throw new InvalidParameterException("Default factor is not set");
-        }
         SetterUtils.safeSet(toPatch::setDefaultFactorId, getDefaultFactorId());
         getApplicationFactors().ifPresent(patchApplicationFactorSettings -> toPatch.setApplicationFactors(
                 applicationFactorsPath(toPatch.getApplicationFactors(), patchApplicationFactorSettings)
         ));
+        boolean noFactors = toPatch.getApplicationFactors() == null || toPatch.getApplicationFactors().isEmpty();
+        if(!noFactors && !hasText(toPatch.getDefaultFactorId())){
```
leleueri, line 44: suggestion: can we have a 'hasFactors' boolean instead of 'noFactors', in order to avoid a ("not" noFactors) ?  this is not a blocker but I think is will be more readable. wdyt ?

lgw-gravitee: Of course, that's a valid point.

leleueri (commented):
code looks ok.
2 questions:
* Postman tests are failing, I restart them but as multiple postman run fail on RecoveryCode scenario, we mlay have something to fix there. (either the Postman is invalid or this fix break something)
* do we need unit test for thisn change?

postman/collections/graviteeio-am-self-account-management-collection-app-version.json
```diff
@@ -1806,7 +1806,7 @@
 										],
 										"body": {
 											"mode": "raw",
-											"raw": "{\n    \"factors\": [      \n       \"{{recoveryCodeFactorId}}\"\n    ],\n    \"settings\": {\n        \"mfa\": {\n            \"factor\": {\n                \"applicationFactors\": [ {\n                    \"id\": \"{{recoveryCodeFactorId}}\"\n                }]\n            }\n        }\n    }\n}"
+											"raw": "{\n    \"factors\": [      \n       \"{{recoveryCodeFactorId}}\"\n    ],\n    \"settings\": {\n        \"mfa\": {\n            \"factor\": {\n                \"applicationFactors\": [ {\n                    \"id\": \"{{recoveryCodeFactorId}}\"\n                }],\n                \"defaultFactorId\":\"{{recoveryCodeFactorId}}\"\n            }\n        }\n    }\n}"
```
leleueri, line 1809: **NOTE**: we shouldn't be able to define a recoverycode as default factor, this is an issue to fix. Can you please confirm that point with Product and open a Private bug if I'm right.

leleueri (approved):
LGTM but put a comment as IMO your update of postman test highligth another issue.

postman/collections/graviteeio-am-self-account-management-collection-app-version.json
lgw-gravitee: Yes, it is. I will create one

PR #3668, opened 2024-03-19
AM-2873: update particles

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-2873

### DIFF
--- a/gravitee-am-ui/package.json
+++ b/gravitee-am-ui/package.json
@@ -38,7 +38,7 @@
     "@angular/router": "16.2.12",
     "@ctrl/ngx-codemirror": "^7.0.0",
     "@gravitee/ui-components": "3.45.2",
-    "@gravitee/ui-particles-angular": "11.0.1",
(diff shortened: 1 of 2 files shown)


gravitee-am-ui/package.json
```diff
@@ -38,7 +38,7 @@
     "@angular/router": "16.2.12",
     "@ctrl/ngx-codemirror": "^7.0.0",
     "@gravitee/ui-components": "3.45.2",
-    "@gravitee/ui-particles-angular": "11.0.1",
+    "@gravitee/ui-particles-angular": "12.0.1",
```
leleueri, line 41: Looks like the v12.0 is targeting Angular 17, in our 4.3.x branch, we are still in angular 16. is it an issue ?

image

mstasgravitee: yes, it may be, we should update the angular first

mstasgravitee: updated to 11.3.1 instead

leleueri (approved):

PR #3709, opened 2024-03-22
fix: allow to set a String array as aud custom value

leleueri:
this claims can be a list according to the spec

  WARN: to make this array usable by introspection endpoint and don't break anything, the client_id need to be the first entry

fixes AM-2865

gravitee-io/issues#9614


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/service/token/TokenServiceTest.java
```diff
@@ -121,6 +124,38 @@ public void shouldCreate() {
         verify(refreshTokenRepository, never()).delete(anyString());
     }
 
+    @Test
```
mstasgravitee, line 127: should it be a test with an empty list too?

leleueri: Made a change into the service code to force the first entry to be the client_id what ever is the result of the ExpressionLanguage.

mstasgravitee (approved):

PR #3711, opened 2024-03-22
AM-2885 verbose error message

lgw-gravitee:
https://gravitee.atlassian.net/browse/AM-2885

When request JSON is malformed, the reposne body contained technical details of application. This would mask it.


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/ManagementApplication.java
```diff
@@ -55,6 +48,7 @@ public ManagementApplication() {
         register(ValidationExceptionMapper.class);
         register(JsonMappingExceptionMapper.class);
         register(WebApplicationExceptionMapper.class);
+        register(JsonParseExceptionMapper.class);
```
mstasgravitee, line 51: is it possible to register a default error in case there is no match to these specified?

lgw-gravitee: well, there is a ThrowableMapper, but it seems not to catch ex in case of malformed json.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/JsonParseExceptionMapper.java
```diff
@@ -0,0 +1,35 @@
...
+import io.gravitee.common.http.HttpStatusCode;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.Provider;
+
+@Provider
+public class JsonParseExceptionMapper extends AbstractExceptionMapper<JsonParseException> {
+    @Override
+    public Response toResponse(JsonParseException e) {
+        return Response
```
leleueri, line 29: suggestion: it maybe useful to get the exception stacktrace/message in a log (at least DEBUG) wdyt ?

mstasgravitee (changes requested):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/ManagementApplication.java
lgw-gravitee: its not only the ThrowableMapper case. All mappers put into the response message body the exception message. Should we adjust all of them? wdyt? @leleueri  @mstasgravitee

mstasgravitee (approved):

PR #3774, opened 2024-04-02
fix: remove disabled applications

podlesrafal:
Prevent loading application if is disabled. 
AM-2797

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientManagerImpl.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientManagerImpl.java
```diff
@@ -1,12 +1,12 @@
...
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
lgw-gravitee, line 9: Run  mvn license:format

lgw-gravitee: And if you find a way how to prevent IDE from formatting it I would appreciate knowing :P

leleueri (changes requested):
TODO: Can you please add Unit test to be sure that this filtering will remain in place if we update this class in the future ?

leleueri (approved):

PR #3786, opened 2024-04-03
AM-2926 verbose error messages vol 2

lgw-gravitee:
https://gravitee.atlassian.net/browse/AM-2926

Tested request:

```
PATCH http://{AM_HOST}/management/organizations/DEFAULT/environments/DEFAULT/domains/{DOMAIN_ID}
{"webAuthnSettings":{"certificates":{}"origin":"http://localhost:8092","relyingPartyId":"localhost","forceRegistration":true}}
```
Request is malformed - missing “,” after “certificates”:{}

The result:

```
Unexpected character ('"' (code 34)): was expecting comma to separate Object entries
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 40] (through reference chain: io.gravitee.am.service.model.PatchDomain["webAuthnSettings"])
```

Expected result:
{"message":"Malformed request body","http_status":400}


leleueri (commented):
suggestion: Looks like the 3 exceptions are inheriting from the same JacksonException.
Maybe we can do a single mapper using this super class ?

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/JsonProcessingExceptionMapper.java
```diff
@@ -0,0 +1,46 @@
...
+import io.gravitee.common.http.HttpStatusCode;
+
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import jakarta.ws.rs.ext.Provider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
```
podlesrafal, line 30: change author

lgw-gravitee: Its not mine:P Git handled it in wrong way :P

lgw-gravitee: I guess GIT handled it in wrong way. Im not creator of that class

lgw-gravitee:
> suggestion: Looks like the 3 exceptions are inheriting from the same JacksonException. Maybe we can do a single mapper using this super class ?

The issue is that there are other JsonParseExceptionMapper and JsonMappingExceptionMapper that take precedence over the general JacksonExceptionMapper. They need to be overridden.

lgw-gravitee:
I manged to configure app without adding default exception mappers. Now its working with general exception mapper

leleueri (approved):

PR #3796, opened 2024-04-04
AM-2915 - Password policies selection dialog

lgw-gravitee:
https://gravitee.atlassian.net/browse/AM-2915

image

image

### DIFF
--- a/gravitee-am-ui/src/app/domain/settings/password-policies/password-policies-idp-select-dialog/password-policies-idp-select-dialog.component.html
+++ b/gravitee-am-ui/src/app/domain/settings/password-policies/password-policies-idp-select-dialog/password-policies-idp-select-dialog.component.html
@@ -1,0 +1,72 @@
+<!--
+
+    Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


gravitee-am-ui/src/app/domain/settings/password-policies/password-policies-idp-select-dialog/password-policies-idp-select-dialog.component.html
```diff
@@ -0,0 +1,72 @@
...
+        <div fxFlex="initial" fxLayoutGap="5" fxLayoutAlign="start center"><mat-icon>info_outline</mat-icon></div>
+        <div fxFlex fxLayoutGap="5">
+          <p>An Identity Provider (IdP) can only by linked to one password policy.</p>
+          <p>Selecting a different IdP will unlink it form its current policy.</p>
+        </div>
+      </div>
+      <app-password-policies-idp-select-table
+        [showAssociation]="true"
+        [rows]="data.linkedIdps"
+        [withSelectionWarning]="true"
```
leleueri, line 62: **question**: do we need this input ? Do you think we can rely only on showAssociation as it look like the warning is only when we select an already associated idp?

leleueri (commented):
LGTM, just made a comment.

gravitee-am-ui/src/app/domain/settings/password-policies/password-policies-idp-select-dialog/password-policies-idp-select-dialog.component.html
lgw-gravitee: I have the impression that those properties are coincidentally similar. Substituting 'showAssociation' for 'withSelectionWarning' in 
`<mat-icon class="warning-icon" *ngIf="withSelectionWarning && selectionModel.get(row.id)">info_outline</mat-icon>`
might not make sense to a 'code vistior'. WDYT?

podlesrafal (approved):

PR #3808, opened 2024-04-05
AM-2912 List password policies 

podlesrafal:


gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.html
```diff
@@ -15,13 +15,15 @@
...
-  <button mat-raised-button color="primary" type="button" style="margin-top: 10px" [routerLink]="['new']">Create Password Policy</button>
+<div *ngIf="isEmpty()" class="gv-page-container">
+  <div fxFlex fxLayout="column" fxLayoutAlign="center center">
+    <app-emptystate
+      message="Password policies will appear here"
+      subMessage="Start configuring your password policies here"
+      icon="donut_large"
+    ></app-emptystate>
+    <button mat-raised-button color="primary" type="button" style="margin-top: 10px" [routerLink]="['new']">Create Password Policy</button>
+  </div>
```
leleueri, line 26: **question**: is it on purpose you define double `div` here ? I was thinking a single div `<div *ngIf="isEmpty()" class="gv-page-container" fxFlex fxLayout="column" fxLayoutAlign="center center">` was enough

leleueri (commented):
Small comment, you can do the change if  with

gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.html
podlesrafal: When list of elements was not empty, then on before get elements from backend those classes was applied.
With before fix:
Screenshot 2024-04-05 at 14 39 50
With after fix:
Screenshot 2024-04-05 at 14 42 42

PR #3834, opened 2024-04-09
feat: backed for password policy list

podlesrafal:
AM-2892

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/PasswordPolicyEntity.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/PasswordPolicyEntity.java
@@ -19,6 +19,8 @@ import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
+import java.util.List;
+
 @Getter
 @Setter
 @NoArgsConstructor
@@ -26,5 +28,5 @@ public class PasswordPolicyEntity {
     private String id;
     private String name;
     private Boolean isDefault;
-    private Integer idpCount;
+    private List<String> idpsNames;
 }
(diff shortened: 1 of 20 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/PasswordPolicyEntity.java
```diff
@@ -19,12 +19,14 @@
...
+
 @Getter
 @Setter
 @NoArgsConstructor
 public class PasswordPolicyEntity {
     private String id;
     private String name;
     private Boolean isDefault;
-    private Integer idpCount;
+    private List<String> idpsNames;
```
leleueri, line 31: **thought**; should we hav to provide a list of FilteredIdentityProvider to have at least the name, the id and the type of the IDP. In any cases, I this it should be better to have internalId instead of the name in a way to retrieve the right IDP as the name may not be unique

gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.html
```diff
@@ -39,10 +39,10 @@ <h1>Password Policies</h1>
       </ngx-datatable-column>
       <ngx-datatable-column name="Identity Providers" [flexGrow]="1" [resizeable]="false" [sortable]="false">
         <ng-template let-row="row" ngx-datatable-cell-template>
-          <span matBadge="{{ row.idpCount }}"></span>
+          <div matBadge="{{ row.idpsNames.length }}" matTooltip="{{ getTooltipText(row.id) }}"></div>
```
leleueri, line 42: **question**: what if the idpsNames is null or undefined

gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.ts
```diff
@@ -32,6 +32,7 @@ import { IdpDataModel } from './password-policies-idp-select-dialog/password-pol
 })
 export class PasswordPoliciesComponent implements OnInit {
   domain: any = {};
+  test: 'tootip';
```
leleueri, line 35: **question**: is it still used ?

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/PasswordPolicyEntity.java
leleueri: ok

leleueri (approved):

PR #3886, opened 2024-04-15
feat: implement REST Endpoint to manage PasswordPolicy deletion

leleueri:
fixes AM-3006

related to AM-2896

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/PasswordPolicyResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/PasswordPolicyResource.java
@@ -34,12 +34,14 @@ import io.swagger.v3.oas.annotations.tags.Tag;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotNull;
 import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.DELETE;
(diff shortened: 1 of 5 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/PasswordPolicyResource.java
```diff
@@ -107,4 +109,30 @@ public void update(
                 .subscribe(response::resume, response::resume);
     }
 
+    @DELETE
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(summary = "Delete a password policy",
+            operationId = "deletePasswordPolicy",
+            description = "User must have the DOMAIN_SETTINGS[UPDATE] permission on the specified domain " +
+                    "or DOMAIN_SETTINGS[UPDATE] permission on the specified environment " +
```
podlesrafal, line 117: **question**: shouldn't be [DELETE]?

leleueri: I asked the question to myself. I put UPDATE because even if it is a REST resource deletion, we are updating the domain settings (how the password are validated), I also check where this DOMAIN_SETTINGS permission was used and we only use the READ and UPDATE ACLs never the DELETE one

podlesrafal (approved):

PR #3890, opened 2024-04-15
feat: set default policy in ui

podlesrafal:
PR to feature branch AM-2893.

### DIFF
--- a/gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.ts
+++ b/gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.ts
@@ -17,6 +17,7 @@ import { Component, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 
 import { PasswordPolicyService } from '../../../services/password-policy.service';
+import { SnackbarService } from '../../../services/snackbar.service';
 
 import { PasswordPolicy } from './domain-password-policies.model';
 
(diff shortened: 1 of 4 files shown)


gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.ts
```diff
@@ -48,11 +51,19 @@ export class PasswordPoliciesComponent implements OnInit {
...
+
   protected getTooltipText(id: string): string {
     const idpsNames = this.rows.find((pp) => pp.id === id).idpsNames;
     if (idpsNames === undefined || idpsNames.length === 0) {
       return null;
     }
     return 'Used in following Identity Providers: ' + idpsNames.join(', ');
   }
+
+  protected selectDefault(id: string) {
```
mstasgravitee, line 63:
```suggestion
  protected selectDefault(id: string): void {
```

mstasgravitee (approved):

gravitee-am-ui/src/app/domain/settings/password-policies/domain-password-policies.component.ts
```diff
@@ -48,11 +51,19 @@ export class PasswordPoliciesComponent implements OnInit {
...
+    this.passwordPolicyService.update(this.domain.id, id, { defaultPolicy: true }).subscribe({
```
leleueri, line 64: **Suggestion**: I would expect to have a **setDefault** method which will call a dedicated endpoint on "password-policies" resoruce in a way that a single call will be responsible to reset the default flag and add the falg on the nex resource.

PR #3892, opened 2024-04-15
feat: manage migration process for Domain password settings

leleueri:
fixes AM-2894


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DomainPasswordPoliciesUpgrader.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Autowired
+    private PasswordPolicyService passwordPolicyService;
+
+    @Override
+    protected Single<Boolean> processUpgrade(String instanceOperationId, SystemTask task, String previousOperationId) {
+        return updateSystemTask(task, (SystemTaskStatus.ONGOING), previousOperationId)
+                .flatMap(updatedTask -> {
+                    if (updatedTask.getOperationId().equals(instanceOperationId)) {
+                        return migrateDomainPasswordSettings(updatedTask);
```
mstasgravitee, line 61:
```suggestion
                        return migrateDomainPasswordSettings(updatedTask).andThen(Single.just(true)).onErrorResumeNext(err -> Single.just(false));
```

I would suggest returning Completable in migrateDomainPasswordSettings and then mapping it to single only once.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DomainPasswordPoliciesUpgrader.java
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                });
+    }
+
+    private SingleSource<? extends Boolean> migrateDomainPasswordSettings(SystemTask task) {
+        return domainService.findAll()
+                .flatMapCompletable(domains ->
+                        Flowable.fromIterable(domains)
+                                .flatMapCompletable(domain -> {
+                                    if (domain.getPasswordSettings() != null) {
```
mstasgravitee, line 73: what do you think about using Single.merge?

The type is Single and it can be switched to Completable only at the end like this:

  ```
 private Completable migrateDomainPasswordSettings(SystemTask task) {
        return domainService.findAll()
                .flatMap(domains -> {
                    var list = new ArrayList<Single<PasswordPolicy>>();
                    for(var domain : domains) {
                        if (domain.getPasswordSettings() != null) {
                            final var policy = domain.getPasswordSettings().toPasswordPolicy();
                            policy.setName(PASSWORD_POLICY_NAME_DEFAULT);
                            policy.setDefaultPolicy(true);
                            policy.setReferenceId(domain.getId());
                            policy.setReferenceType(ReferenceType.DOMAIN);
                            Date now = new Date();
                            policy.setCreatedAt(now);
                            policy.setUpdatedAt(now);
                            list.add(passwordPolicyService.create(policy, null));
                        }
                    }
                    return Single.merge(list).toList();
                })
                .doOnError(err -> updateSystemTask(task, (SystemTaskStatus.FAILURE), task.getOperationId()).subscribe())
                .flatMapCompletable(results -> updateSystemTask(task, SystemTaskStatus.SUCCESS, task.getOperationId()).ignoreElement());
    }
```

What do you think?

Comment (leleueri) [reply gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DomainPasswordPoliciesUpgrader.java]: I added a new method at domainService to return a Flowable instead.

File: gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/upgrades/DomainPasswordPoliciesUpgraderTest.java [hunk 3]
```diff
@@ -0,0 +1,138 @@
...
+        verify(systemTaskRepository, times(1)).findById(any());
+        verify(domainService, never()).listAll();
+    }
+
+    @Test
+    public void should_ignore_domain_without_passwordSettings() {
+        initializeSystemTask();
+        var domainNoPasswordSettings = new Domain();
+
+        when(domainService.listAll()).thenReturn(Flowable.just(domainNoPasswordSettings));
```
mstasgravitee, line 85: I recommend adding more domains here and verifying how many updates have been performed.

leleueri: I have created a dedicated test with 3 domains (2 with settings, 1 without) to test that all domains are manage properly.
I prefere keeping a test with single case in addition of this one to ease the identification of future regression

mstasgravitee (approved):

PR #3896, opened 2024-04-16
fix: avoid exception when an event has a unknown type

leleueri:
this change allowes to keep synchronization process working

 when a upper version introduce new type of event

related-to AM-2894


podlesrafal (approved):
ok

PR #3943, opened 2024-04-22
fix: token attribute

mstasgravitee:
fixes AM-3068

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/OrganizationUserServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/OrganizationUserServiceImpl.java
@@ -177,20 +177,25 @@ public class OrganizationUserServiceImpl extends AbstractUserService<Organizatio
                     }
                     return accessTokenRepository.create(token)
                             .map(created -> created.toCreateResponse(rawToken))
-                            .flatMap(this::getAccountAccessTokenWithIssuerUsername);
+                            .flatMap(i -> prepareTokenToGet(i, true));
                 });
     }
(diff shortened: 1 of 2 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/OrganizationUserServiceImpl.java
```diff
@@ -177,20 +177,24 @@ public Single<AccountAccessToken> generateAccountAccessToken(User user, NewAccou
...
-        return token.issuerId() == null ? Single.just(token) :
+    private Single<AccountAccessToken> prepareTokenToGet(AccountAccessToken token) {
+        return prepareTokenToGet(token, false);
+    }
+
+    private Single<AccountAccessToken> prepareTokenToGet(AccountAccessToken token, boolean showToken) {
+        return token.issuerId() == null ? Single.just(token.toBuilder().token(showToken ? token.token() : null).build()) :
                 userRepository.findById(token.issuerId()).defaultIfEmpty(new User())
-                        .map(user -> token.toBuilder().issuerUsername(user.getUsername()).build());
+                        .map(user -> token.toBuilder().token(showToken ? token.token() : null).issuerUsername(user.getUsername()).build());
```
podlesrafal, line 197: Suggestion: you can extract `showToken ? token.token() : null` to variable, since you are using this two times.

mstasgravitee: updated

podlesrafal (approved):

PR #3966, opened 2024-04-26
fix: NPE when applying MFA to app with legacy/broken MFA config

mregulski-gravitee:
fixes: AM-3065, #9685

Avoid NPE when an application has MFA factors set to null

### DIFF
--- a/gravitee-am-model/src/main/java/io/gravitee/am/model/FactorSettings.java
+++ b/gravitee-am-model/src/main/java/io/gravitee/am/model/FactorSettings.java
@@ -15,17 +15,27 @@
  */
 package io.gravitee.am.model;
 
-import java.util.List;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
+import java.util.List;
+import java.util.Objects;
+
 @Getter
 @Setter
 @NoArgsConstructor
+@AllArgsConstructor
+@Builder
 public class FactorSettings {
     private String defaultFactorId;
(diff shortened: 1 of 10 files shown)


gravitee-am-model/src/main/java/io/gravitee/am/model/FactorSettings.java
```diff
@@ -15,18 +15,28 @@
...
 @Setter
 @NoArgsConstructor
+@AllArgsConstructor
+@Builder
 public class FactorSettings {
     private String defaultFactorId;
     private List<ApplicationFactorSettings> applicationFactors = List.of();
 
+    public List<ApplicationFactorSettings> getApplicationFactors() {
+        return Objects.requireNonNullElseGet(applicationFactors, List::of);
```
mstasgravitee, line 37: Is it required? It can throw NullPointerException. If yes, why not: 
return applicationFactors != null ? applicationFactors : List.of();

mregulski-gravitee: `requireNonNullElseGet()` does exactly what your comment says :)
The reason this safeguard is required is that when `FactorSettings` gets instantiated by deserialization, e.g. from the database, then the field defaults are ignored. 

The linked issue describes exactly such case: the application is represented in the DB as 
```
{
  "factor": {},
  ...
}
```
Deserialization effectively does the equivalent of: 
```java
var factor = new FactorSettings();
factor.setDefaultFactorId(null);
factor.setApplicationFactors(null);
```
thus leading to NPE and a 500 response.

mstasgravitee (approved):

PR #3967, opened 2024-04-29
fix: make job not to use maven cache with corrupted lib

lgw-gravitee:
corrupted lib: com/github/eirslett/node/16.10.0/node-16.10.0-linux-x64.tar.gz
fixes release :) hopefully.

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -1146,7 +1146,7 @@ jobs:
     steps:
       - checkout
       - restore-maven-job-cache:
-          jobName: release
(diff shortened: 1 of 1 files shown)


passionne (changes requested):
Please update the key on all usage on `restore-maven-job-cache` and `save-maven-job-cache`. (a replace all should do the job 😊.

mstasgravitee:
please add version labels

passionne (approved):

mstasgravitee (approved):

PR #3982, opened 2024-04-30
AM-2994 allow making email optional for new users created via management api

mregulski-gravitee:
AM-2994

Introduces a new config property in `gravitee.yml`: `user.email.required: (boolean)`, set to `true` by default.
If `false`, management API will allow creating users without email addresses. 

The new property is exposed on a new config endpoint: `configuration/users/emailRequired`, allowing UI to adapt appropriately.


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/configuration/ConfigurationResource.java
```diff
@@ -85,4 +86,12 @@ public void getSpelGrammar(@Suspended final AsyncResponse response) {
             .subscribe(response::resume, response::resume);
     }
 
+    @GET
+    @Path("users/emailRequired")
```
mstasgravitee, line 90: Path has to be lowercase: "users/email-required"

Shouldn't we think about more generic configuration endpoint for easy adaptability in the future? What do you think?

mregulski-gravitee: I agree on both points.
The more general endpoint, might be a good idea (we could then just load it once in `main.ts`, together with `constants.json` and `build.json`), but I don't think we currently have enough configuration that needs such an endpoint to know how to best structure it.
I think for now it's good enough that these few config elements are grouped under a single resource.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/configuration/ConfigurationResource.java
```diff
@@ -85,4 +86,12 @@ public void getSpelGrammar(@Suspended final AsyncResponse response) {
             .subscribe(response::resume, response::resume);
     }
 
+    @GET
+    @Path("users/email-required")
+    @Produces(jakarta.ws.rs.core.MediaType.APPLICATION_JSON)
```
mstasgravitee, line 91: produces JSON in annotation but it is returned boolean

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/platform/configuration/ConfigurationResource.java
```diff
@@ -79,10 +83,19 @@ public void getAlertServiceStatus(@Suspended final AsyncResponse response) {
...
+                .subscribe(response::resume, response::resume);
+    }
+
+    @GET
+    @Path("users/email-required")
+    @Produces(jakarta.ws.rs.core.MediaType.APPLICATION_JSON)
+    public void getUserEmailRequired(@Suspended final AsyncResponse response) {
+        var emailRequired = environment.getProperty(UserEmail.PROPERTY_USER_EMAIL_REQUIRED, boolean.class, true);
+        Single.just(Map.of("emailRequired", emailRequired))
+                .subscribe(response::resume, response::resume);
```
mstasgravitee, line 98:
```suggestion
        response.resume(Map.of("emailRequired", emailRequired));
```

mstasgravitee (approved):

PR #3990, opened 2024-05-02
chore: update node version

podlesrafal:
fixes: AM-3077


mstasgravitee (approved):

tcompiegne (changes requested):
As it should be also compatible with the 4.3.x version, we should add the apply-on-4-* labels for this PR

PR #4004, opened 2024-05-06
AM-3002 - Introduced iteration rounds for SHA password encoding

lgw-gravitee:


podlesrafal (approved):

gravitee-am-service/src/test/java/io/gravitee/am/service/authentication/crypto/password/SHAMD5PasswordEncoderTest.java
```diff
@@ -33,17 +36,21 @@ protected PasswordEncoder getEncoder() {
         return this.messageDigestPasswordEncoder;
     }
 
-    @Test
```
mstasgravitee, line 36: I would suggest adding a test with a negative round

mstasgravitee (approved):

PR #4009, opened 2024-05-07
fix: add typeadapter for gson serialization

podlesrafal:
fixes: AM-3137


gravitee-am-jwt/src/main/java/io/gravitee/am/jwt/DefaultJWTBuilder.java
```diff
@@ -98,7 +109,7 @@ public String sign(JWT payload) {
             if (issuer != null && !payload.containsKey(Claims.iss)) {
                 payload.setIss(issuer);
             }
-            SignedJWT signedJWT = new SignedJWT(header, JWTClaimsSet.parse(payload));
+            SignedJWT signedJWT = new SignedJWT(header, JWTClaimsSet.parse(GSON.toJson(payload)));
```
mstasgravitee, line 112: is it not possible to use ObjectMapper?

podlesrafal: There is no Jackson library imported to use ObjectMapper

mstasgravitee (approved):

PR #4035, opened 2024-05-15
AM-2999 Uploading Certificates for mTLS

lgw-gravitee:
AM-2999

image

image

image

### DIFF
--- a/gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/ConfigurationCertUtils.java
+++ b/gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/ConfigurationCertUtils.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.certificate.api;
+
+import com.nimbusds.jose.util.JSONObjectUtils;
+import lombok.extern.slf4j.Slf4j;
+
+import java.text.ParseException;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Lukasz GAWEL (lukasz.gawel at graviteesource.com)
+ * @author GraviteeSource Team
+ */
+
+@Slf4j
+public final class ConfigurationCertUtils {
+
+    public static List<String> extractUsageFromCertConfiguration(String configuration) {
+        if (configuration == null) {
+            return List.of();
+        }
+        try {
+            Map<String, Object> cfg = JSONObjectUtils.parse(configuration);
+            List<String> uses = JSONObjectUtils.getStringList(cfg, "use");
+            return uses == null ? List.of() : uses;
+        } catch (ParseException e) {
+            log.warn("Problem at parsing certificate configuration, msg={}", e.getMessage());
+            return List.of();
+        }
+    }
+}
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/CertificatesResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/CertificatesResource.java
@@ -64,6 +64,8 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static io.gravitee.am.certificate.api.ConfigurationCertUtils.extractUsageFromCertConfiguration;
+
 /**
  * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
  * @author GraviteeSource Team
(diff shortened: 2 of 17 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/CertificatesResource.java
```diff
@@ -257,6 +259,15 @@ private CertificateEntity filterCertificateInfos(Certificate certificate, int ce
...
         }
+        if (certificate.getConfiguration() != null) {
+            final JsonObject config = JsonObject.mapFrom(Json.decodeValue(certificate.getConfiguration(), HashMap.class));
+            if (config != null && config.getJsonArray("use") != null) {
+                filteredCertificate.setUsage(config.getJsonArray("use").stream().map(Object::toString).toList());
+            } else {
+                filteredCertificate.setUsage(List.of());
+            }
+        }
+
```
leleueri: **Suggestion/Question**: Wondering if the logic to extract the _Use_ values should be moved into the **certificate-api** module into a Configuration Utility class :thinking:

gravitee-am-ui/src/app/domain/settings/certificates/certificates.component.ts
```diff
@@ -139,7 +157,10 @@ export class DomainSettingsCertificatesComponent implements OnInit {
 
   rotateCertificate() {
     this.dialogService
-      .confirm('Rotate key', 'Are you sure you want to generate a new system certificate ?')
+      .confirm(
+        'Rotate key',
+        'Are you sure you want to generate a new system certificate?\nThis will only effect the latest System Certificate created.',
```
leleueri, line 162:
```suggestion
        'Are you sure you want to generate a new system certificate?\nThis will only affect the latest System Certificate created.',
```
Your change is consistent with figma design, but 'effect' looks strange to me :thinking:

leleueri (commented):
Question: Wondering if we should prevent the mTLS usage for "system" certificate. As they are self signed certificates generated by AM, it make no sens to mark them as mTLS. It may be better UX to reject creation/update i this case. wdyt ?

lgw-gravitee:
> **Question**: Wondering if we should prevent the mTLS usage for "system" certificate. As they are self signed certificates generated by AM, it make no sens to mark them as mTLS. It may be better UX to reject creation/update i this case. wdyt ?

Yes. It should be a part of next story https://gravitee.atlassian.net/browse/AM-2998 where we will use uploaded certs. Here, we can

gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/ConfigurationCertUtils.java
```diff
@@ -0,0 +1,46 @@
...
+    public static List<String> extractUsageFromCertConfiguration(String configuration) {
+        if (configuration == null) {
+            return List.of();
+        }
+        try {
+            Map<String, Object> cfg = JSONObjectUtils.parse(configuration);
+            List<String> uses = JSONObjectUtils.getStringList(cfg, "use");
+            return uses == null ? List.of() : uses;
+        } catch (ParseException e) {
+            log.debug("Problem at parsing certificate configuration", e);
```
leleueri, line 42: **question**: should it be a WARN ? I'm OK to return an empty list but having a parsing exception here is maybe something we whould be worried as it should no happen. wdyt ?

leleueri (approved):
LGTM, I let you decide if the log level need to be changed.

gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/resources/CertificatesResourceTest.java
```diff
@@ -68,7 +70,64 @@ public void shouldGetCertificates() {
...
+
+
+        doReturn(Maybe.just(mockDomain)).when(domainService).findById(domainId);
+        doReturn(Flowable.just(mockCertificate, mockCertificate2, mockCertificate3, mockCertificate4)).when(certificateService).findByDomain(domainId);
+        doReturn(Flowable.empty()).when(applicationService).findByCertificate(anyString());
+
+        final Response response = target("domains").path(domainId).path("certificates").request().get();
+        assertEquals(HttpStatusCode.OK_200, response.getStatus());
+
+        final List<LinkedHashMap> responseEntity = readEntity(response, List.class);
```
podlesrafal, line 114: There is readListEntity method. You can use it like this:  readListEntity(reponse, CertificateEntity.class)

gravitee-am-model/src/main/java/io/gravitee/am/model/Certificate.java
```diff
@@ -18,6 +18,7 @@
 import io.swagger.v3.oas.annotations.media.Schema;
 
 import java.util.Date;
+import java.util.List;
```
podlesrafal, line 21: remove

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -18,6 +18,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.Lists;
```
podlesrafal, line 21: remove unused imports

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/CertificateServiceImpl.java
```diff
@@ -120,6 +126,7 @@ public class CertificateServiceImpl implements CertificateService {
     private static final String STORE_PASS = "storepass";
     private static final String ALIAS = "alias";
     private static final String NAME = "name";
+    private static final String USAGE = "use";
```
podlesrafal, line 129: remove

gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/ConfigurationCertUtils.java
lgw-gravitee: You're right, I've changed to warn.

PR #4045, opened 2024-05-16
fix: don't require a redirect_uri param in the reset password flow

mregulski-gravitee:
AM-3187, #9735

Keep track of what flow the MFA challenge is invoked in, and allow redirect_uri to not be provided where it's not used (i.e. in the password reset flow)

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
@@ -265,12 +265,12 @@ public class EmailServiceImpl implements EmailService, InitializingBean {
     }
 
     private Optional<TokenPurpose> getTokenPurpose(io.gravitee.am.model.Template template) {
-        var purpose = switch (template) {
-            case RESET_PASSWORD -> TokenPurpose.RESET_PASSWORD;
-            case REGISTRATION_VERIFY -> TokenPurpose.REGISTRATION_VERIFY;
-            default -> null;
+        return switch (template) {
+            case RESET_PASSWORD -> Optional.of(TokenPurpose.RESET_PASSWORD);
+            case REGISTRATION_VERIFY -> Optional.of(TokenPurpose.REGISTRATION_VERIFY);
+            // not UNSPECIFIED, because if the token has no particular purpose, we don't want it to contain this claim
+            default -> Optional.empty();
         };
-        return Optional.ofNullable(purpose);
 
     }
 
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
@@ -23,6 +23,7 @@ import io.gravitee.am.common.email.Email;
 import io.gravitee.am.common.email.EmailBuilder;
 import io.gravitee.am.common.jwt.Claims;
 import io.gravitee.am.common.jwt.JWT;
+import io.gravitee.am.common.jwt.TokenPurpose;
 import io.gravitee.am.common.utils.ConstantKeys;
 import io.gravitee.am.gateway.handler.common.email.EmailManager;
 import io.gravitee.am.gateway.handler.common.email.EmailService;
@@ -41,8 +42,6 @@ import io.gravitee.am.service.i18n.GraviteeMessageResolver;
 import io.gravitee.am.service.reporter.builder.AuditBuilder;
 import io.gravitee.am.service.reporter.builder.EmailAuditBuilder;
 import io.vertx.rxjava3.core.MultiMap;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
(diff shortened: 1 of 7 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -44,25 +52,58 @@ public class RedirectUriValidationHandler implements Handler<RoutingContext> {
...
+                .map(UserToken::getToken);
     }
 
     @Override
     public void handle(RoutingContext context) {
         final Client client = context.get(ConstantKeys.CLIENT_CONTEXT_KEY);
+        getOperation(context)
+                .subscribe(operation -> {
+                            // proceed redirect_uri parameter
+                            parseRedirectUriParameter(context, client, operation);
```
leleueri, line 76: **thought**: Are you sure it will work in case of failure ? I would move this method call into a `map` operator to simply manage the context::next or context::fail into the subscription. In the current implementation, my fear is if the redirect url is not valid, vertx will never be inform about the failure :thinking: maybe I'm wrong

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -32,9 +40,9 @@
  * The authorization server validates the request to ensure that all parameters are valid.
  * If the request is valid, the authorization server authenticates the resource owner and obtains
  * an authorization decision (by asking the resource owner or by establishing approval via other means).
- *
+ * <p>
```
mstasgravitee, line 43:
```suggestion
 *
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -32,9 +40,9 @@
...
  * See <a href="https://tools.ietf.org/html/rfc6749#section-4.1.1">4.1.1. Authorization Request</a>
- *
+ * <p>
```
mstasgravitee, line 45:
```suggestion
 *
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
mregulski-gravitee: you're right, that's exactly what caused some tests to fail

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
mregulski-gravitee: We *do* publish the javadocs, so I don't think it makes sense to remove automatic formatting fixes.

Before:
image

After:
image

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/impl/EmailServiceImpl.java
```diff
@@ -256,11 +259,21 @@ private Map<String, Object> prepareEmailParams(User user, Client client, Integer
...
 
+    private Optional<TokenPurpose> getTokenPurpose(io.gravitee.am.model.Template template) {
+        var purpose = switch (template) {
+            case RESET_PASSWORD -> TokenPurpose.RESET_PASSWORD;
+            case REGISTRATION_VERIFY -> TokenPurpose.REGISTRATION_VERIFY;
+            default -> null;
+        };
+        return Optional.ofNullable(purpose);
+
+    }
```
remisultan, line 275:
```suggestion
        return switch (template) {
            case RESET_PASSWORD -> Optional.of(TokenPurpose.RESET_PASSWORD);
            case REGISTRATION_VERIFY -> Optional.of(TokenPurpose.REGISTRATION_VERIFY);
            default -> Optional.empty();
        };
    }
```
Also is there a difference in the logic between `TokenPurpose.UNSPECIFIED` and `null` ?

mregulski-gravitee: `EmailService` uses `null` because if we don't know what the purpose is it doesn't make sense to put the claim in the token.
On the other hand, `RedirectUriValidationHandler` logic gets simpler if it can assume that every token has a non-null purpose - that's why `TokenPurpose.of(String)` returns `UNSPECIFIED` instead of null/`Optional.empty()`

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/RedirectUriValidator.java
```diff
@@ -49,4 +58,8 @@ public void validate(Client client, String requestedRedirectUri, BiConsumer<Stri
             checkMethod.accept(requestedRedirectUri, registeredClientRedirectUris);
         }
     }
+
+    private boolean requiresRedirectUri(TokenPurpose operation) {
+        return operation != TokenPurpose.RESET_PASSWORD;
```
leleueri, line 63: **suggestion**: could be nice to specify in a comment the reason why there is an exception for reset_password.

leleueri (approved):

PR #4077, opened 2024-05-21
fix: make sure client secret are force to null into the execution con…

leleueri:
…text

fixes AM-3240


mregulski-gravitee:
The same sanitization is done in 2 more places:
`io.gravitee.am.gateway.handler.root.resources.endpoint.logout.LogoutEndpoint#logout`
`io.gravitee.am.gateway.handler.root.resources.endpoint.logout.LogoutCallbackEndpoint#logout`

Might be worth updating these too, for consistency.

leleueri:
> The same sanitization is done in 2 more places: `io.gravitee.am.gateway.handler.root.resources.endpoint.logout.LogoutEndpoint#logout` `io.gravitee.am.gateway.handler.root.resources.endpoint.logout.LogoutCallbackEndpoint#logout`
> 
> Might be worth updating these too, even if just for consistency.

Missed your comment and I merge the PR sorry. 
Good catch I will open another PR to fix it

PR #4110, opened 2024-05-24
add cloud rebrand properties

gcusnieux:

https://gravitee.atlassian.net/browse/CJ-1221

Screenshot 2024-05-28 at 09 29 41
Screenshot 2024-05-28 at 09 29 37

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/CockpitAuthenticationFilter.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/CockpitAuthenticationFilter.java
@@ -154,41 +154,45 @@ public class CockpitAuthenticationFilter extends GenericFilterBean {
     }
 
     private boolean enabled() {
-        return configuration.getProperty("cockpit.enabled", Boolean.class, false);
+        return getProperty("cockpit.enabled", "cloud.enabled", Boolean.class, false);
     }
 
     /**
      * Cockpit keystore type for client certificate (mtls) and jwt signature verification.
      */
     private String keyStoreType() {
-        return getProperty("cockpit.connector.ws.ssl.keystore.type", "cockpit.keystore.type", null);
+        return getProperty("cockpit.keystore.type","cloud.connector.ws.ssl.keystore.type", null);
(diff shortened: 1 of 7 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/CockpitAuthenticationFilter.java
```diff
@@ -154,41 +154,45 @@ public void doFilter(jakarta.servlet.ServletRequest request, jakarta.servlet.Ser
     }
 
     private boolean enabled() {
-        return configuration.getProperty("cockpit.enabled", Boolean.class, false);
+        return getProperty("cockpit.enabled", "cloud.enabled", Boolean.class, false);
```
leleueri, line 157: **question**: do we have to test the property starting with cockpit first ? :thinking:  I was thinking the cloud one should be tested first and then fallback to the cockpit one at it is legacy. (this comment is for all the changes related to this getProperty method)

gcusnieux: Hi @leleueri 
We don't wish to introduce any breaking change which is why cockpit configuration takes precedence over cloud conf.

leleueri: ok, as there is a fallback IMO this is not a BC but lets keep it like this.

leleueri (approved):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/filter/CockpitAuthenticationFilter.java
gcusnieux: Not really, depending of the installation type. 
If a user have `cockpit.enabled = true` and process to upgrade, the user will have both `cockpit.enabled = true` and `cloud.enbaled = false`

PR #4119, opened 2024-05-27
chore: remove unused ldap dependency

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-196

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,11 +102,9 @@
         <commons-lang.version>2.6</commons-lang.version>
         <swagger.version>2.2.15</swagger.version>
         <hibernate-validator.version>8.0.1.Final</hibernate-validator.version>
-        <ldaptive.version>1.2.4</ldaptive.version>
         <javax.servlet-api.version>3.1.0</javax.servlet-api.version>
         <thymeleaf.version>3.1.2.RELEASE</thymeleaf.version>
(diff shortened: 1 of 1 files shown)


leleueri (commented):
Though: which part of AM is using ldaptive outside of the LDAP and Kerberos IDP  ?

mstasgravitee:
> **question**: which part of AM is using ldaptive outside of the LDAP and Kerberos IDP ?

It is not used outside of these plugins. If you are sure we can safely remove them I can update the PR.

leleueri:
> > **question**: which part of AM is using ldaptive outside of the LDAP and Kerberos IDP ?
> 
> It is not used outside of these plugins. If you are sure we can safely remove them I can update the PR.

IMO, it is safe to remove it if ldaptive is part of the runtime libe embedded into the LDAP & Kerberos plugin

mstasgravitee:
> > > **question**: which part of AM is using ldaptive outside of the LDAP and Kerberos IDP ?
> > 
> > 
> > It is not used outside of these plugins. If you are sure we can safely remove them I can update the PR.
> 
> IMO, it is safe to remove it if ldaptive is part of the runtime libe embedded into the LDAP & Kerberos plugin

removed

leleueri (approved):

PR #4183, opened 2024-05-31
feat: migrate upgraders to node-services-upgrader framework

mregulski-gravitee:
implements AM-941


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/UpgraderOrder.java
```diff
@@ -19,22 +19,26 @@
...
+    public static final int DEFAULT_IDP_UPGRADER = 11;
+    public static final int DEFAULT_REPORTER_UPGRADER = 12;
+    public static final int APPLICATION_IDENTITY_PROVIDER_UPGRADER = 13;
+    public static final int SYSTEM_CERTIFICATE_UPGRADER = 14;
+    public static final int APPLICATION_FACTOR_UPGRADER = 15;
+    public static final int DOMAIN_PASSWORD_POLICIES_UPGRADER = 16;
+
+    private UpgraderOrder() {
+        throw new UnsupportedOperationException("utility class, don't instantiate");
+    }
```
leleueri, line 43: **question**: why moving it to a class instead of interface ?

helm/templates/api/api-deployment.yaml
```diff
@@ -112,6 +112,8 @@ spec:
               containerPort: {{ .Values.api.http.services.core.http.port | default "18093" }}
             {{- end }}
           env:
+            - name: upgrade.mode
```
leleueri, line 115: issue: as far as I know env var containing dot character is not allowed (at least not on all systems) and the convention to provide a value coming from gravitee.yaml as an environment variable is to suffix it by gravitee as described here https://documentation.gravitee.io/am/getting-started/configuration/configure-am-gateway#environment-variables. So IMO it should be: `gravitee_upgrade_mode`

helm/templates/api/api-upgrader-job.yaml
```diff
@@ -96,7 +96,7 @@ spec:
             - name: {{ .Values.api.service.internalPortName }}
               containerPort: {{ .Values.api.service.internalPort }}
           env:
-            - name: gravitee_upgrade_mode
+            - name: upgrade.mode
```
leleueri, line 99: **issue**: see previous comment for helm/templates/api/api-deployment.yaml

helm/tests/api/api_upgrader_test.yaml
```diff
@@ -24,7 +24,7 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].env
           value:
-            - name: gravitee_upgrade_mode
+            - name: upgrade.mode
```
leleueri, line 27: **issue**: see previous comment for helm/templates/api/api-deployment.yaml

leleueri (changes requested):
Need to have a deeper look but here is some comments

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/UpgraderOrder.java
mregulski-gravitee: Sonar marks interfaces with only constants as an issue

helm/templates/api/api-deployment.yaml
mregulski-gravitee: That would be more consistent with all the other properties, but - we can't use `gravitee_upgrade_mode` without changes in gravitee-node. 
That's because the [`UpgraderConfiguration`](https://github.com/gravitee-io/gravitee-node/blob/0df61fa414b74d983c1fb8437c6976064b46f5c9/gravitee-node-services/gravitee-node-services-upgrader/src/main/java/io/gravitee/node/services/upgrader/spring/UpgraderConfiguration.java#L45) that actually handles this property, uses raw system environment & properties, not the Spring `Environment` processed by `io.gravitee.node.container.spring.env.EnvironmentPropertySourceBeanProcessor`.

leleueri: indeed, I was not aware of this choise in gravitee node.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/UpgraderOrder.java
leleueri: hum... ok...

leleueri:
Before approving, we need to test it

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DomainPasswordPoliciesUpgrader.java
```diff
@@ -43,13 +45,18 @@ public class DomainPasswordPoliciesUpgrader extends SystemTaskUpgrader {
     private static final String TASK_ID = "domain_password_settings_migration";
     private static final String UPGRADE_NOT_SUCCESSFUL_ERROR_MESSAGE =
             "Domain Password Policy can't be upgraded, other instance may process them or an upgrader has failed previously";
+    @SuppressWarnings("squid:S2068") // not a hardcoded password
     public static final String PASSWORD_POLICY_NAME_DEFAULT = "default";
```
mstasgravitee, line 49: It would be better to rename is to NAME_DEFAULT. SuppressWarnings will not be needed

mregulski-gravitee:
> Before approving, we need to test it

Scenarios I've checked:
1. run with upgrade.mode=true, check app exits, run with upgrade.mode=false, app runs normally, upgraders are not run again
1b. run with uprade.mode=true on empty SQL db - to make sure upgrade mode works if the upgraders table doesn't exist yey
2. run 3.21.x on an empty database, create a domain, add an app, change the password policy, then run scenario 1 on the same DB

This definitely needs an independent check though. The helm chart could also be checked (not sure how reliable Helm's unit tests are)

leleueri:
> > Before approving, we need to test it
> 
> Scenarios I've checked:
> 
> 1. run with upgrade.mode=true, check app exits, run with upgrade.mode=false, app runs normally, upgraders are not run again
>    1b. run with uprade.mode=true on empty SQL db - to make sure upgrade mode works if the upgraders table doesn't exist yey
> 2. run 3.21.x on an empty database, create a domain, add an app, change the password policy, then run scenario 1 on the same DB
> 
> This definitely needs an independent check though. The helm chart could also be checked (not sure how reliable Helm's unit tests are)

Sorry for the delay.
Did simple test (not migrating version but only executing the mAPI on my master DB).
It looks working fine for Mongo but on PSQL, I hae this error

image

leleueri (changes requested):
**TODO**: can you plese add a UpgraderRecord repository test to be sure the Repository layer is working for all backend implementation ?
https://github.com/gravitee-io/gravitee-access-management/tree/master/gravitee-am-repository/gravitee-am-repository-tests/src/test/java/io/gravitee/am/repository/management/api

leleueri (approved):
Tested locally by starting Mongo & PSQL.
Looks OK.

A proper validation using HelmChart have to be done but it can be merge

PR #4184, opened 2024-05-31
fix: update scim Operation class to manage multi valued elements usin…

leleueri:
…g Azure AD syntaxe

 the syntax is not consistent with the RFC 7644 as it is using filter 'emails[type eq "work"].value'

 instead of a path with an array

fixes AM-3041

gravitee-io/issues#9674

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/model/Operation.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/model/Operation.java
@@ -155,16 +155,28 @@ public abstract class Operation {
             JsonNode parentNode = node.get(getPath().getAttributePath());
             if (getPath().getSubAttribute() != null) {
                 if (parentNode != null) {
-                    ((ObjectNode) parentNode).set(getPath().getSubAttribute(), value);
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/model/Operation.java
```diff
@@ -155,16 +155,28 @@ public void apply(ObjectNode node) {
...
-                    node.putObject(getPath().getAttributePath()).set(getPath().getSubAttribute(), value);
+                    if (MultiValuedAttributes.isListOfObjects(getPath().getAttributePath())) {
+                        fillObjectAttributes(node.putArray(getPath().getAttributePath()).addObject());
+                    } else {
+                        node.putObject(getPath().getAttributePath()).set(getPath().getSubAttribute(), value);
+                    }
                 }
             } else if (parentNode == null) {
-                node.set(getPath().getAttributePath(), value);
+                if (MultiValuedAttributes.isListOfSingular(getPath().getAttributePath())) {
```
mstasgravitee, line 171: why isn't the function fillObjectAttributes used here?

leleueri: This is because Singular value are not complex object, this is either a simple string or numberic.

mstasgravitee (approved):

PR #4209, opened 2024-06-04
AM-3292 mtls openid delete cert

lgw-gravitee:
If client auth method equals to client_secret_basic or  client_secret_post there is no use of certificate.
By removing it on create/update for configurtion, the certificate would be able to be deleted as its won't be associated with any IDP.

### DIFF
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/oidc/OpenIDConnectConfigurationUtils.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/oidc/OpenIDConnectConfigurationUtils.java
@@ -16,6 +16,7 @@
 package io.gravitee.am.identityprovider.api.oidc;
 
 import com.nimbusds.jose.util.JSONObjectUtils;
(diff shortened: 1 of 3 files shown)


podlesrafal (approved):

gravitee-am-identityprovider/gravitee-am-identityprovider-api/src/main/java/io/gravitee/am/identityprovider/api/oidc/OpenIDConnectConfigurationUtils.java
```diff
@@ -40,4 +41,21 @@ public static Optional<String> extractCertificateId(String configuration) {
             return Optional.empty();
         }
     }
+
+    public static String validateClientAuthMethod(String configuration) {
```
leleueri, line 45: **suggestion**: Looking at the method behaviour it more a _sanitizeClientAuthMethod_ than _validateClientAuthMethod_

```suggestion
    public static String sanitizeClientAuthMethod(String configuration) {
```

lgw-gravitee: I changed it.

PR #4211, opened 2024-06-04
fix: prevent create and update policy with default flag.

podlesrafal:
Only way to set policy as default is by calling endpoint /default

### DIFF
--- a/gravitee-am-service/src/test/java/io/gravitee/am/service/PasswordPolicyServiceTest.java
+++ b/gravitee-am-service/src/test/java/io/gravitee/am/service/PasswordPolicyServiceTest.java
@@ -95,7 +95,7 @@ public class PasswordPolicyServiceTest {
     private User principal;
 
     @Test
-    public void shouldCreate_noDefaultPolicy() {
+    public void shouldCreate_noExistingDefaultPolicy() {
         PasswordPolicy newPasswordPolicy = createPolicy();
 
         when(passwordPolicyRepository.create(any())).thenAnswer(answer -> Single.just(answer.getArguments()[0]));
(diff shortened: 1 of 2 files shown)


gravitee-am-service/src/test/java/io/gravitee/am/service/PasswordPolicyServiceTest.java
```diff
@@ -267,6 +292,64 @@ public void shouldUpdate() {
...
+                        !passwordPolicy.getUpdatedAt().equals(existingPolicy.getUpdatedAt()) &&
+                        passwordPolicy.getReferenceId().equals(DOMAIN_ID) &&
+                        passwordPolicy.getReferenceType().equals(ReferenceType.DOMAIN) &&
+                        passwordPolicy.getName().equals(updatePasswordPolicy.getName()) &&
+                        passwordPolicy.getDefaultPolicy().equals(existingPolicy.getDefaultPolicy()));
+
+        verify(auditService).report(ArgumentMatchers.argThat(builder -> builder.build(MAPPER).getOutcome().getStatus().equals(Status.SUCCESS)));
+        verify(eventService).create(ArgumentMatchers.argThat(evt -> evt.getType().equals(Type.PASSWORD_POLICY) && evt.getPayload().getAction().equals(Action.UPDATE)));
+    }
+
```
leleueri, line 352: suggestion: could be nice to mututalize this test as the only difference is the boolean values used at the beginning to define the defaultPolicy. wdyt?

leleueri (approved):

gravitee-am-service/src/test/java/io/gravitee/am/service/PasswordPolicyServiceTest.java
podlesrafal: I would stick with current solution for better readability

PR #4230, opened 2024-06-11
fix: Remove more general duplicated "user" block in Helm chart

lgw-gravitee:
fixes AM-3313, also updated with values not present in Helm chart, but in gravitee.yml: mfaVerifyAttempt and webAuthn

### DIFF
--- a/helm/templates/gateway/gateway-configmap.yaml
+++ b/helm/templates/gateway/gateway-configmap.yaml
@@ -306,51 +306,6 @@ data:
       url: https://{{index .Values.gateway.ingress.hosts 0 }}{{ .Values.gateway.ingress.path }}
     {{- end }}
 
-   {{- if .Values.userManagement }}
-    # User management configuration
-    user:
-    {{- if .Values.userManagement.email }}
-      email:
-{{ toYaml .Values.userManagement.email | indent 8 }}
-    {{- end }}
-
-    {{- if .Values.userManagement.password }}
-      password:
-{{ toYaml .Values.userManagement.password | indent 8 }}
-    {{- end }}
-
-    {{- if .Values.userManagement.name }}
-      name:
-{{ toYaml .Values.userManagement.name | indent 8 }}
-    {{- end }}
-
-    {{- if .Values.userManagement.username }}
-      username:
-{{ toYaml .Values.userManagement.username | indent 8 }}
-    {{- end }}
-
-    {{- if .Values.userManagement.registration }}
-      registration:
-{{ toYaml .Values.userManagement.registration | indent 8 }}
-    {{- end }}
-
-    {{- if .Values.userManagement.resetPassword }}
-      resetPassword:
-{{ toYaml .Values.userManagement.resetPassword | indent 8 }}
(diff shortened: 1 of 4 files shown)


leleueri (changes requested):
**TODO:** can you please add unit test to validate the final config file ?

lgw-gravitee:
I wasnt aware there are tests for helm charts ;)
anyway, I added them and chcanged it according to your comment.

helm/templates/gateway/gateway-configmap.yaml
leleueri, line 1: **TODO**: looks like we have the same block on management-api configmap https://github.com/gravitee-io/gravitee-access-management/blob/master/helm/templates/api/api-configmap.yaml#L319 can you please update this one too ?

helm/tests/gateway/configmap_user_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+        email: test@test.com
+        password: password
+        name: name
+        username: luk
+        registration: 1
+        resetPassword: 2
+        blockedAccount: 3
+        mfaChallenge: 4
+        mfaVerifyAttempt: 5
+        webAuthn: 6
```
leleueri, line 18: **TODO**: is it possible to test at lest one complexe entry like it should be in a valid config?
ex:
```
  email:
    policy:
      pattern: ^[a-zA-Z0-9_+-]+(?:\.[a-zA-Z0-9_+-]+)*@(?:[a-zA-Z0-9-]+\.)+[a-zA-Z]{2,15}$
```

helm/tests/gateway/configmap_user_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+    asserts:
+      - hasDocuments:
+          count: 1
+      - isKind:
+          of: ConfigMap
+      - matchRegex:
+          path: data.[gravitee.yml]
+          pattern: "user:\n( *[a-zA-Z]*: .*\n)* *email: test@test.com"
```
leleueri, line 26: **suggestion**: Looks like we can use a text block as regex. It should help for readability and indentation validation.
ex : https://github.com/gravitee-io/gravitee-api-management/blob/master/helm/tests/gateway/confiigmap_db_less_test.yaml#L31

```
      - matchRegex:
          path: data.[gravitee.yml]
          pattern: |
            email:
              policy:
                pattern: ...
```

leleueri (changes requested):

helm/tests/gateway/configmap_user_test.yaml
lgw-gravitee: fixed

helm/tests/gateway/configmap_user_test.yaml
lgw-gravitee: added

helm/templates/gateway/gateway-configmap.yaml
lgw-gravitee: updated

leleueri (approved):

PR #4255, opened 2024-06-14
feat: publish some domain event to the Organization Audit logs

leleueri:
Only DOMAIN_CREATED, DOMAIN_DELETED are sent in addition of DOMAIN_UPDATED when the name or the enabled attributes change

 This commit changes the referenceType and refenceId of the target entry for domain event (previously it was referencing the domain but it is more relevant to have the env

 DOMAIN_CREATED & DOMAIN_DELETED are only sent on the Org level reporters not on the domain reporters anymore

fixes AM-3269

### DIFF
--- a/gravitee-am-ui/src/app/domain/settings/audits/audits.component.html
+++ b/gravitee-am-ui/src/app/domain/settings/audits/audits.component.html
@@ -169,12 +169,20 @@
       </ngx-datatable-column>
       <ngx-datatable-column [sortable]="true" name="Target" [flexGrow]="2" prop="target.sortDisplayName">
         <ng-template let-row="row" ngx-datatable-cell-template>
-          <a *ngIf="row.target" [routerLink]="getTargetUrl(row)" [queryParams]="getTargetParams(row)"
-            >{{ row.target?.displayName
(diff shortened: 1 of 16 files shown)


gravitee-am-ui/src/app/domain/settings/audits/audits.component.html
```diff
@@ -169,12 +169,21 @@ <h1>Audit log</h1>
...
             }}<span *ngIf="row.target.alternativeId">
               | <small>{{ row.target?.alternativeId }}</small></span
             ></a
           >
+          <span *ngIf="row.target && isDomainAuditOnOrganizationLevel(row)">
+            {{ row.target?.displayName
+            }}<span *ngIf="row.target.alternativeId">
+              | <small>{{ row.target?.alternativeId }}</small></span
+            >
+          </span>
```
lgw-gravitee, line 186: suggestion: put intermediate div element with ngIf="row.target". Then in "a" and "span" the conditions simplify.

lgw-gravitee (approved):

mregulski-gravitee (approved):

PR #4268, opened 2024-06-19
feat: added schema registry to kafka reporter

podlesrafal:


gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
```diff
@@ -34,123 +35,133 @@
...
-  @Autowired
-  private Node node;
+    @Autowired
+    private Node node;
 
-  private final DtoMapper dtoMapper;
+    private KafkaProducer<String, AuditMessageValueDto> producer;
 
-  private KafkaProducer<String, AuditMessageValueDto> producer;
+    public KafkaAuditReporter(KafkaReporterConfiguration config) {
```
mregulski-gravitee, line 72: it would make sense to include other injected fields (and to make them final), to avoid mixing 2 approaches to injection in 1 class

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/resources/json/schema.json
```diff
@@ -0,0 +1 @@
...","title":"Not included"},{"type":"string"}]},"nodeHostname":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]}},"definitions":{"AuditAccessPointDto":{"type":"object","additionalProperties":false,"properties":{"id":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"alternativeId":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"displayName":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"ipAddress":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"userAgent":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]}}},"AuditEntityDto":{"type":"object","additionalProperties":false,"properties":{"id":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"alternativeId":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"type":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"displayName":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"referenceType":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]},"referenceId":{"oneOf":[{"type":"null","title":"Not included"},{"type":"string"}]}}}}}
```
leleueri, line 1: **question**: is there any reason why you didn't provide the schema in petty format?

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
```diff
@@ -34,123 +35,133 @@
...
+            log.debug("Producer is null, ignore reportable");
+        }
+    }
+
+    @Override
+    public void afterPropertiesSet() {
+        if (context == null) {
+            context = GraviteeContext.defaultContext(null);
+        }
+    }
```
mregulski-gravitee, line 98: If you replaced field-level `@Autowired` with constructor injection, we could remove this method as well.
Also isn't this dead code? I'm not sure it's actually possible for `@Autowired` field to be null **after** injection - won't Spring complain about unmatched dependency?

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/KafkaReporterConfiguration.java
```diff
@@ -25,43 +27,15 @@
...
  */
+@Setter
+@Getter
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
 public class KafkaReporterConfiguration implements ReporterConfiguration {
 
   private String bootstrapServers;
   private String topic;
   private String acks;
+  private String schemaRegistryUrl;
```
leleueri, line 38: **question**: how do we manage authentication on this registry ?

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
```diff
@@ -34,123 +35,133 @@
...
+            properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, JacksonSerializer.class);
+        }
+        properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class);
+
+        List<Map<String, String>> additionalProperties = this.config.getAdditionalProperties();
+        if (additionalProperties != null && !additionalProperties.isEmpty()) {
+            additionalProperties.forEach(claimMapper -> {
+                String option = claimMapper.get("option");
+                String value = claimMapper.get("value");
+                properties.put(option, value);
```
mregulski-gravitee, line 158:
```suggestion
            additionalProperties.forEach(property-> {
                String option = property.get("option");
                String value = property.get("value");
                properties.put(option, value);
```

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/dto/AuditAccessPointDto.java
```diff
@@ -15,50 +15,27 @@
...
-    this.alternativeId = alternativeId;
-  }
+@Setter
+@Getter
+public class AuditAccessPointDto {
 
-  public String getDisplayName() {
-    return this.displayName;
-  }
+    @JsonProperty
```
mregulski-gravitee, line 26: I'm pretty sure that by default Jackson serializes everything it finds a public gettter/setter for - `@JsonProperty` isn't required unless you're customizing something

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/dto/AuditEntityDto.java
```diff
@@ -15,60 +15,30 @@
...
-    this.type = type;
-  }
+@Setter
+@Getter
+public class AuditEntityDto {
 
-  public String getDisplayName() {
-    return displayName;
-  }
+    @JsonProperty
```
mregulski-gravitee, line 26: not sure it's needed (see [AuditAccessPointDto](https://github.com/gravitee-io/gravitee-access-management/pull/4268/files#diff-18abfb814783921c68a3df61da906d61e0ec78f51569d2eda90cd70b7b12d2dc))

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/dto/AuditMessageValueDto.java
```diff
@@ -15,198 +15,63 @@
...
-  }
-
-  public AuditEntityDto getActor() {
-    return actor;
-  }
-
-  public void setActor(AuditEntityDto actor) {
-    this.actor = actor;
-  }
+    @JsonProperty
```
mregulski-gravitee, line 33: not sure it's needed (see [AuditAccessPointDto](https://github.com/gravitee-io/gravitee-access-management/pull/4268/files#diff-18abfb814783921c68a3df61da906d61e0ec78f51569d2eda90cd70b7b12d2dc))

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/kafka/KafkaJsonSerializer.java
```diff
@@ -0,0 +1,37 @@
...
+        super();
+        configureObjectMapper();
+    }
+
+    private void configureObjectMapper() {
+        ObjectMapper mapper = this.objectMapper();
+        mapper.registerModule(new JavaTimeModule());
+        mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+        mapper.disable(SerializationFeature.WRITE_DATE_TIMESTAMPS_AS_NANOSECONDS);
+        mapper.disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
```
mregulski-gravitee, line 35: is there a reason we don't want these settings when we're not using schemas (i.e. in `JacksonSerializer`)

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/java/io/gravitee/am/reporter/kafka/JUnitConfiguration.java
```diff
@@ -35,31 +40,56 @@
...
 
-    @Override
-    public void onApplicationEvent(ContextClosedEvent contextClosedEvent) {
-        if(this.kafkaContainer != null){
-            this.kafkaContainer.stop();
-        }
+    @Bean
+    public SchemaRegistryContainer schemaRegistryContainer(KafkaContainer kafkaContainer) {
+        schemaRegistryContainer = new SchemaRegistryContainer(CONFLUENT_PLATFORM_VERSION);
+        schemaRegistryContainer.withNetwork(NETWORK).withKafka(kafkaContainer).withReuse(true).start();
```
mregulski-gravitee, line 59: `withReuse()` isn't needed (and won't work anyway wihout an additrional global properties file) - please remove from both containers

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/java/io/gravitee/am/reporter/kafka/JUnitConfiguration.java
```diff
@@ -35,31 +40,56 @@
...
+    private static final Network NETWORK = Network.newNetwork();
+    private static final String CONFLUENT_PLATFORM_VERSION = "7.6.1";
 
     @Bean
-    public KafkaContainer kafkaContainer(){
-        this.kafkaContainer = new KafkaContainer(
-            DockerImageName.parse("confluentinc/cp-kafka:6.1.1"));
+    public KafkaContainer kafkaContainer() {
+        kafkaContainer = new KafkaContainer(
+                DockerImageName.parse("confluentinc/cp-kafka:" + CONFLUENT_PLATFORM_VERSION))
```
mregulski-gravitee, line 50: question: would it make sense to do something like
`cofluentVersion = Sytem.getProperty("confluentVersion", CONFLUENT_PLATFORM_VERSION)` instead? It would allow us to run tests with a different version without recompiling, like what we're doing with repositories tests. Not sure if/how useful it would be.

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/resources/json/schema.json
mregulski-gravitee, line 1: any reason against having the schema formatted? 😅

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/liquibase/UsernameUniquenessMigration.java
```diff
@@ -107,8 +107,7 @@ public SqlStatement[] generateStatements(Database database) throws CustomChangeE
 
                 if (!referenceUser.getSource().equalsIgnoreCase("gravitee") && !referenceUser.getSource().equalsIgnoreCase("cockpit")) {
                     logger.error("Organization Username '{}' migration only manages gravitee & cockpit identity providers",
-                            referenceUser.getUsername(),
-                            referenceUser.getSource());
```
mregulski-gravitee, line 111: not sure how user's `source` here relates to kafka 😀 why is it removed?

leleueri: agreed, these two params need to be kept. Just adapt the log message to display the source at the end ("`Organization Username '{}' migration only manages gravitee & cockpit identity providers but source is {}`")

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporterWithSchemaRegistryITests.java
```diff
@@ -0,0 +1,139 @@
...
+        valueDeserializer.configure(config, false);
+
+        KafkaConsumer<String, AuditMessageValueDto> kafkaConsumer = new KafkaConsumer<>(config, keyDeserializer, valueDeserializer);
+        kafkaConsumer.subscribe(Collections.singletonList(this.configuration.getTopic()));
+
+        ConsumerRecords<String, AuditMessageValueDto> poll = kafkaConsumer.poll(Duration.of(10, ChronoUnit.SECONDS));
+        assertNotNull(poll);
+        assertEquals(1, poll.count());
+
+        ConsumerRecord<String, AuditMessageValueDto> record = poll.iterator().next();
```
leleueri, line 116: question: where do we check that the schema id is present into the record ?

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/kafka/KafkaJsonSerializer.java
podlesrafal: Same was in non-registry version. I will keep it to have compatibility.

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/resources/json/schema.json
podlesrafal: Kafka to reduce payload remove all unnecessary characters and sends payload in pure string. Thats why this is not formatted.

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporterWithSchemaRegistryITests.java
podlesrafal: There is specified `schema.registry.url` in consumer properties. If this wouldn't be provided, than consumer don't know how to validate te message and throw an error. When we consume this message as a string, we can see at the beginning few characters that correspond to schema Id.

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/resources/json/schema.json
podlesrafal: This is how schema registry store schemas.

leleueri: does it means we can't have the "formatted" version in the repo because the registry will not flatten the content before checking if the schema is already known?

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/KafkaReporterConfiguration.java
podlesrafal: To add authentication we can use additionalProperties. For example to use basic auth we have to set `basic.auth.credentials.source=USER_INFO`
`basic.auth.user.info=username:password`
Using additional Properties ssl certificate can be set as well.

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/resources/json/schema.json
podlesrafal: changed

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/kafka/KafkaJsonSerializer.java
leleueri: what is the behaviour of the KafkaJsonSerializer if the registry url is not present, doesn't it simply serialize the content ? If it is the same output as the current Jackon maybe we can just keep the kafka one :thinking:

leleueri (changes requested):
two points:

* I don't understand how the schema is registered, I was thinking the client will send the schema to the registry to declare it if the schema was missing but I can't find how it is done.
* locally, when I try to initialize a kafka reporter, I have this error:

```
11:08:30.049 [Thread-19] [] ERROR i.g.a.p.h.a.c.ProviderPluginManager - An unexpected error occurs while loading
java.lang.NoSuchMethodException: io.gravitee.am.reporter.kafka.audit.KafkaAuditReporter.<init>()
	at java.base/java.lang.Class.getConstructor0(Class.java:3585)
	at java.base/java.lang.Class.getDeclaredConstructor(Class.java:2754)
	at io.gravitee.am.plugins.handlers.api.core.ProviderPluginManager.createInstance(ProviderPluginManager.java:94)
	at io.gravitee.am.plugins.handlers.api.core.ProviderPluginManager.createProvider(ProviderPluginManager.java:67)
	at io.gravitee.am.plugins.handlers.api.core.ProviderPluginManager.createProvider(ProviderPluginManager.java:62)
	at io.gravitee.am.plugins.reporter.core.ReporterPluginManager.create(ReporterPluginManager.java:69)
```

leleueri:
> * I don't understand how the schema is registered, I was thinking the client will send the schema to the registry to declare it if the schema was missing but I can't find how it is done.

Answer to myself, the schema is created on the fly by the client based on the object serialization output

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/kafka/KafkaJsonSerializer.java
podlesrafal: KafkaJsonSchemaSerializer checks if schema.registry.url is present. If not, then it will throw an exception. This is the flow of KafkaJsonSchemaSerializer: connect to schema registry -> fetch id of schema or register new schema and fetch id of schema -> attach schemaId on the beggining of message -> convert object to json format and attach to message. https://github.com/confluentinc/schema-registry/blob/4e09a7c01e34b5a8509306ae7b70f9eb804ae9ca/json-schema-serializer/src/main/java/io/confluent/kafka/serializers/json/AbstractKafkaJsonSchemaSerializer.java#L161C5-L165C40

podlesrafal: We cannot use the same serializer, because Jackson only convert POJO into json string

gravitee-am-reporter/gravitee-am-reporter-kafka/src/test/java/io/gravitee/am/reporter/kafka/JUnitConfiguration.java
podlesrafal: There will be only one version tested. This is not the same case as repository.

leleueri (approved):

PR #4301, opened 2024-06-25
feat: manage response_mode parameter to follow https://openid.net/spe…

leleueri:
…cs/oauth-v2-multiple-response-types-1_0.html

fixes AM-3336

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/AuthorizationRequestFailureHandler.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/AuthorizationRequestFailureHandler.java
@@ -19,9 +19,10 @@ import io.gravitee.am.common.exception.oauth2.InvalidRequestObjectException;
 import io.gravitee.am.common.exception.oauth2.OAuth2Exception;
 import io.gravitee.am.common.exception.oauth2.RedirectMismatchException;
 import io.gravitee.am.common.oauth2.Parameters;
+import io.gravitee.am.common.oauth2.ResponseMode;
+import io.gravitee.am.common.utils.ConstantKeys;
 import io.gravitee.am.common.web.UriBuilder;
 import io.gravitee.am.gateway.handler.common.jwt.JWTService;
-import io.gravitee.am.common.utils.ConstantKeys;
 import io.gravitee.am.gateway.handler.common.vertx.utils.UriBuilderRequest;
 import io.gravitee.am.gateway.handler.oauth2.exception.JWTOAuth2Exception;
 import io.gravitee.am.gateway.handler.oauth2.resources.request.AuthorizationRequestFactory;
@@ -55,6 +56,7 @@ import static io.gravitee.am.common.oauth2.GrantType.JWT_BEARER;
 import static io.gravitee.am.gateway.handler.common.vertx.utils.UriBuilderRequest.CONTEXT_PATH;
 import static io.gravitee.am.service.utils.ResponseTypeUtils.isHybridFlow;
 import static io.gravitee.am.service.utils.ResponseTypeUtils.isImplicitFlow;
+import static org.springframework.util.StringUtils.hasLength;
 
 /**
  * If the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid,
@@ -226,9 +228,13 @@ public class AuthorizationRequestFailureHandler implements Handler<RoutingContex
         query.computeIfAbsent("error_description", val -> errorDescription);
         query.computeIfAbsent(Parameters.STATE, val -> authorizationRequest.getState());
 
-        boolean fragment = !isDefaultErrorPage(authorizationRequest.getRedirectUri(), errorPath) &&
-                (isImplicitFlow(authorizationRequest.getResponseType()) || isHybridFlow(authorizationRequest.getResponseType()));
+        boolean fragment = !isDefaultErrorPage(authorizationRequest.getRedirectUri(), errorPath) && requiresFragment(authorizationRequest);
         return append(authorizationRequest.getRedirectUri(), query, fragment);
+    }
+
+    private boolean requiresFragment(AuthorizationRequest authorizationRequest) {
+        return (!hasLength(authorizationRequest.getResponseMode()) && (isImplicitFlow(authorizationRequest.getResponseType())) || isHybridFlow(authorizationRequest.getResponseType()))
+                || ResponseMode.FRAGMENT.equals(authorizationRequest.getResponseMode());
     }
 
     private String append(String base, Map<String, String> query, boolean fragment) throws URISyntaxException {
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/HybridResponse.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/HybridResponse.java
@@ -21,8 +21,6 @@ import io.gravitee.am.common.web.UriBuilder;
 import io.gravitee.am.gateway.handler.oauth2.service.token.Token;
 import io.vertx.rxjava3.core.MultiMap;
 
-import java.net.URISyntaxException;
-
 /**
  * When using the Hybrid Flow, Authentication Responses are made in the same manner as for the Implicit Flow, as defined in Section 3.2.2.5, with the exception of the differences specified in this section.
  * These Authorization Endpoint results are used in the following manner:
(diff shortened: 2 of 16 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/AuthorizationRequestFailureHandler.java
```diff
@@ -227,7 +228,9 @@ private String buildRedirectUri(String error, String errorDescription, Authoriza
         query.computeIfAbsent(Parameters.STATE, val -> authorizationRequest.getState());
 
         boolean fragment = !isDefaultErrorPage(authorizationRequest.getRedirectUri(), errorPath) &&
-                (isImplicitFlow(authorizationRequest.getResponseType()) || isHybridFlow(authorizationRequest.getResponseType()));
+                (isImplicitFlow(authorizationRequest.getResponseType()) ||
+                        isHybridFlow(authorizationRequest.getResponseType()) ||
+                        ResponseMode.FRAGMENT.equals(authorizationRequest.getResponseMode()));
```
lgw-gravitee, line 233: suggestion: looks like the same condition is used in 3 places. maybe it can be encapsulated?

lgw-gravitee (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/IDTokenResponse.java
```diff
@@ -47,6 +45,7 @@ public void setIdToken(String idToken) {
     @Override
     public String buildRedirectUri() {
         UriBuilder uriBuilder = UriBuilder.fromURIString(getRedirectUri());
+        // idToken required fragment
```
tcompiegne, line 48: Are we sure about this ? 

The comment about mention this : ```unless a different Response Mode was specified.```

tcompiegne (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/HybridResponse.java
```diff
@@ -70,6 +68,7 @@ public void setCode(String code) {
     @Override
     public String buildRedirectUri() {
         UriBuilder uriBuilder = UriBuilder.fromURIString(getRedirectUri());
+        // Hybrid response requires Fragment
```
tcompiegne, line 71: Not sure if we should also consider the response mode here ... Difficult to read in the spec if the response_mode override the default response mode of the Flow. 

WDYT ?

tcompiegne (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/IDTokenResponse.java
leleueri: According to https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes
For id_token response type:

* `The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used` 

My interpretation is only fragment is allowed or another mode like jwt. wdyt ?


`3.  ID Token Response Type
This section registers a new Response Type, the id_token, in accordance with the stipulations in the OAuth 2.0 specification, Section 8.4. The intended purpose of the id_token is that it MUST provide an assertion of the identity of the Resource Owner as understood by the Authorization Server. The assertion MUST specify a targeted audience, e.g. the requesting Client. However, the specific semantics of the assertion and how it can be validated are not specified in this document.

id_token
When supplied as the response_type parameter in an OAuth 2.0 Authorization Request, a successful response MUST include the parameter id_token. The Authorization Server SHOULD NOT return an OAuth 2.0 Authorization Code, Access Token, or Access Token Type in a successful response to the grant request. If a redirect_uri is supplied, the User Agent SHOULD be redirected there after granting or denying access. The request MAY include a state parameter, and if so, the Authorization Server MUST echo its value as a response parameter when issuing either a successful response or an error response. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.
Returning the id_token in a fragment reduces the likelihood that the id_token leaks during transport and mitigates the associated risks to the privacy of the user (Resource Owner).`

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/HybridResponse.java
leleueri: same as for id_token,

For all of them we have this sentence : `The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used`


`
This section defines combinations of the values code, token, and id_token, which are each individually registered Response Types.

code token
When supplied as the value for the response_type parameter, a successful response MUST include an Access Token, an Access Token Type, and an Authorization Code. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.

code id_token
When supplied as the value for the response_type parameter, a successful response MUST include both an Authorization Code and an id_token. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.

id_token token
When supplied as the value for the response_type parameter, a successful response MUST include an Access Token, an Access Token Type, and an id_token. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.

code id_token token
When supplied as the value for the response_type parameter, a successful response MUST include an Authorization Code, an id_token, an Access Token, and an Access Token Type. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.
`

leleueri: I just checking also the JARM rfc, query.jwt is forbidden also if token or id_token are part of the response_type 
https://openid.net/specs/oauth-v2-jarm.html#name-response-mode-queryjwt

Will amend the validation method to check these combinaison if it is not already done.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/IDTokenResponse.java
tcompiegne: ```Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.```

Not sure how to read this ... but I still think we should rely on the supplied response mode.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/HybridResponse.java
tcompiegne: Same thing as above : 

```Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.```

Not sure how to read this ... but I still think we should rely on the supplied response mode.

leleueri: OK after digging into the differents specs, here is the conclusion:

If `token` or `id_token` are part of the `response_type`, then the `response_mode` has to be `fragment` (or `fragment.jwt`) for security reason as spotted by https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/response/IDTokenResponse.java
leleueri: If token or id_token are part of the response_type, then the response_mode has to be fragment (or fragment.jwt) for security reason as spotted by https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Security.

I will amend the PR.

tcompiegne (approved):

PR #4304, opened 2024-06-26
Feat/am 3268 inheritable reporters

mregulski-gravitee:
AM-3268
AM-3346

This PR allows setting **non-system**, **org-level** reporters as "inherited". Inherited reporters are automatically added to all existing & future domain within the organization and pick up all events from these domains.

In management API, the org-level reporter's instance is re-used for all domains. Adding/deleting a domain doesn't trigger a re-deployment of the reporter, this part of configuration is updated in-place on the running reporter.
In the gateway, each domain gets a copy of this domains' reporters and inherited org's reporters.

All these instances are linked to a single configuration object (entry in `reporters` table/collection), so updating it in the org's settings updates all relevant instances.
There's no limit on the number of inherited reporters in the domain.

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
@@ -48,7 +48,6 @@ import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.context.annotation.Primary;
 import org.springframework.core.env.Environment;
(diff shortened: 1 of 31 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
```diff
@@ -201,19 +204,18 @@ public Single<Reporter> update(Reference reference, String reporterId, UpdateRep
                     if (!oldReporter.isSystem() || isUpgrader) {
                         reporterToUpdate.setConfiguration(updateReporter.getConfiguration());
                     }
+                    if (oldReporter.getReference().type() == ReferenceType.ORGANIZATION && !oldReporter.isSystem()) {
+                        // only non-system org reporters can be inherited
+                        reporterToUpdate.setInherited(updateReporter.isInherited());
+                    }
```
leleueri, line 210: **Question**: do we have to reject the request with 400 Bad Request if inherite is set to true for "system reporter" or domain reporter ? IMO better to notify the user of the error instead of silently accept ther request and ignore the value.

leleueri (approved):

PR #4325, opened 2024-07-01
fix: missing transaction id in context

podlesrafal:


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
```diff
@@ -745,6 +750,8 @@ private void sessionAndCookieHandler(Router router) {
     private void authFlowContextHandler(Router router) {
         // Login endpoint
         AuthenticationFlowContextHandler authenticationFlowContextHandler = new AuthenticationFlowContextHandler(authenticationFlowContextService, environment);
+        TransactionIdHandler transactionIdHandler = new TransactionIdHandler(transactionHeader);
+        router.route(GET, PATH_LOGIN).handler(transactionIdHandler);
```
mstasgravitee, line 754: should we add this also for other routes MFA for example?

podlesrafal: This happens only on the flow when policy was added to webauthn or login

PR #4334, opened 2024-07-03
fix: send registration confirmation form when user has no password

podlesrafal:
fixes: AM-3400

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
@@ -377,11 +377,15 @@ public class UserServiceImpl extends AbstractUserService<io.gravitee.am.service.
                                     .map(Optional::of)
                                     .defaultIfEmpty(Optional.empty())
                                     .doOnSuccess(optClient -> {
-                                        var template = getTemplate(domain1, optClient);
-                                        emailService.send(domain1, optClient.orElse(null), template, user)
-                                                .doOnSuccess(__ -> auditService.report(AuditBuilder.builder(UserAuditBuilder.class).principal(principal).type(resoleEventType(template)).user(user)))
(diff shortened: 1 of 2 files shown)


mstasgravitee (approved):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/UserServiceImpl.java
```diff
@@ -570,12 +570,15 @@ private void createPasswordHistory(Domain domain, Application client, User user,
                            throwable -> logger.debug("Failed to create password history", throwable));
     }
 
-    private static Template getTemplate(Domain domain, Optional<Application> optClient) {
+    private static Template getTemplate(Domain domain, Optional<Application> optClient, User user) {
+        //Users without password should be redirected to registration confirmation
+        if (user.getPassword() == null) {
```
leleueri, line 575: question: is it really the fix to provide :thinking: ? User password will most of the time be null except during the preRegistration, if the admin click on the "send email" button the password will be null and we will not be able to determine if we have to send confirmation or verification email

podlesrafal: I dont really get your use case. Those are the test cases I checked before PR : https://gravitee.atlassian.net/browse/AM-3400?focusedCommentId=31036 can you find your use case here?

leleueri: settings:

- account verification via email enabled
- user registration enabled on the app

use case:

- on the GW, user click on register my acount on the login form
- on the registration form, he provides info (including email and password)
- he received an email (the link is about validate the email as the password as been provided
- I do not click on this link (simulation of link expiry)
- on the admin UI, I'm going to the user profile
- click on "send confirmation"
- email received but the link send me to the "confirmation form" on which I have to provide pwd, I should receive the validate email link

podlesrafal: fixed

leleueri (approved):

PR #4360, opened 2024-07-05
fix: validate OAuth2 settings on HTTP Resource

leleueri:
the schema form can't apply condition on oauth settings to require them

 this fix apply validation on the resource configuration json element

fixes AM-3383

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/resource/http/HttpResourceValidatorImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/resource/http/HttpResourceValidatorImpl.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/validators/resource/http/HttpResourceValidatorImpl.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final String FIELD_AUTH = "auth";
+    public static final String FIELD_TYPE = "type";
+    public static final String OAUTH2 = "oauth2";
+    public static final String FIELD_ENDPOINT = "endpoint";
+    public static final String FIELD_CLIENTID = "clientId";
+    public static final String FIELD_CLIENT_SECRET = "clientSecret";
+    private final ObjectMapper objectMapper;
+    public static final String HTTP_AM_RESOURCE = "http-am-resource";
+    public static final String INVALID_OAUTH_CONF = "OAuth config is missing or invalid";
+
```
lgw-gravitee, line 48: **wondering** I understand that static fields are public because they are used in unit tests, but I'm wondering if we shouldn't use literals there. If you changed value in HTTP_AM_RESOURCE, the test would pass?

leleueri: I changed the visibility

lgw-gravitee (approved):

PR #4369, opened 2024-07-08
fix: close kafka producer connection without delay.

podlesrafal:
Workaround for bug in kafka vertx library https://github.com/vert-x3/vertx-kafka-client/issues/271

### DIFF
--- a/gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
+++ b/gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
@@ -32,10 +32,12 @@ import io.gravitee.reporter.api.Reportable;
 import io.gravitee.reporter.api.Reporter;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
+import io.vertx.core.Context;
 import io.vertx.core.Vertx;
 import io.vertx.kafka.client.producer.KafkaProducer;
 import io.vertx.kafka.client.producer.KafkaProducerRecord;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.clients.producer.Producer;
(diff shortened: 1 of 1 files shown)


gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
```diff
@@ -110,7 +112,17 @@ protected void doStart() {
     protected void doStop() throws Exception {
         super.doStop();
         if (this.producer != null) {
-            this.producer.close();
+            // Need to do unwrap producer because of bug in kafka vertx library: https://github.com/vert-x3/vertx-kafka-client/issues/271
+            try (Producer<String, AuditMessageValueDto> producer = this.producer.unwrap()) {
```
leleueri, line 116: **question**: is it usefull to use a try resources here as you are calling the close method in the try block ?

podlesrafal: unwrap is autoclosable. Thats why it is in try-with-resource.

leleueri: yes I know, but my point is the close provided by the unwrap is closing the same resource (KafkaWriteStream<K, V> stream) as the one you are closing in the try statement. So you are closing two times the same thing.

If there is no concurrent issue for doing this, then it is OK.

leleueri (approved):

gravitee-am-reporter/gravitee-am-reporter-kafka/src/main/java/io/gravitee/am/reporter/kafka/audit/KafkaAuditReporter.java
podlesrafal: I got it. Removed try block to not call close twice.

PR #4397, opened 2024-07-12
fix: add audit log with cause message

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-3413

configure Http Provider message
configure

Audit log
audit_log

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
@@ -106,7 +106,9 @@ public class UserAuthenticationManagerImpl implements UserAuthenticationManager 
 
         var applicationIdentityProviders = getApplicationIdentityProviders(client);
         if (isNull(applicationIdentityProviders) || applicationIdentityProviders.isEmpty()) {
-            return Single.error(() -> getInternalAuthenticationServiceException(client));
+            var error = getInternalAuthenticationServiceException(client);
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
```diff
@@ -150,6 +157,7 @@ public Single<User> authenticate(Client client, Authentication authentication, b
                 .doOnError(throwable -> {
                     gatewayMetricProvider.incrementFailedAuth(false);
                     eventManager.publishEvent(AuthenticationEvent.FAILURE, new AuthenticationDetails(authentication, domain, client, throwable));
+                    auditService.report(new AuthenticationAuditBuilder().client(client).throwable(throwable));
```
leleueri, line 160: **issue**: IMO This change is just duplicating the audit as currently the `USER_LOGIN` audit entry is managed by the `AuthenticationEventListener` which is listening the `AuthenticationEvent.FAILURE` published just before the `auditService.report` call you added. The change in the AuditBuilder should be enough to provide the info. wdyt ?

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationManagerImpl.java
mstasgravitee: When I was testing I got only one audit, It was no duplicated.

leleueri (approved):

PR #4410, opened 2024-07-12
feat: AM-3270 MTLS for LDAP

lgw-gravitee:


podlesrafal:
According to your description, there should be usage of `loadDefaultTrustStore` when variable is missing. I cannot see usage of this. Will it be used in the future or you missing this fallback in `load` method?

lgw-gravitee:
the descrption is obsolete. It fallbacks to JVM truststore by default, so no need to code it on our side.

podlesrafal:
So do we need `loadDefaultTrustStore()`? I cant find usage of it

lgw-gravitee:
it is used in LDAP plugin

gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/DefaultTrustStoreProvider.java
```diff
@@ -16,24 +16,19 @@
...
-
-    private DefaultTrustStoreProvider() {
-    }
+import lombok.experimental.UtilityClass;
 
-    public static String defaultTrustStorePassword() {
-        return System.getProperty(JAVA_DEFAULT_TRUSTSTORE_PASSWORD_PROPERTY);
-    }
+@UtilityClass
+public final class DefaultTrustStoreProvider {
```
podlesrafal, line 22: using @UtilityClass, you dont need to set this class as final, as well as all fields as a static. https://projectlombok.org/features/experimental/UtilityClass

podlesrafal (approved):

gravitee-am-certificate/gravitee-am-certificate-api/src/main/java/io/gravitee/am/certificate/api/DefaultTrustStoreProvider.java
lgw-gravitee: agreed to leave as it is.

PR #4437, opened 2024-07-16
Allow disabling default reporter for new domains via configuration

mregulski-gravitee:

AM-3442

This PR contains 2 parts:

1. split DomainService into DomainReadService (minimal set of read-only
 operations used by the gateway), and DomainService that extends the ReadService with
 management (write) methods. This improves code coherence by making sure that the **entire** domain creation process is properly handled by the `DomainService#create` method.

2. Actually introducing the new config setting. There's also a change in audit log search, as it no longer can assume that every domain has at least 1 reporter with search capabilities.


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/AuditServiceImpl.java
```diff
@@ -49,7 +50,9 @@ public class AuditServiceImpl implements AuditService {
     @Override
     public Single<Page<Audit>> search(ReferenceType referenceType, String referenceId, AuditReportableCriteria criteria, int page, int size) {
         try {
-            return getReporter(referenceType, referenceId).search(referenceType, referenceId, criteria, page, size);
+            return Maybe.fromOptional(getReporter(referenceType, referenceId))
+                    .flatMapSingle(r -> r.search(referenceType, referenceId, criteria, page, size))
+                    .switchIfEmpty(Single.just(new Page<>()));
```
leleueri, line 55: **suggestion**: to avoid useless rx subscription and publisher convertion, why not doing something like this ?

`getReporter(x,y).map(r -> r.search(referenceType, referenceId, criteria, page, size)).orElse(Single.just(new Page))`

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/AuditServiceImpl.java
```diff
@@ -65,7 +68,9 @@ public Single<Page<Audit>> search(String domain, AuditReportableCriteria criteri
     @Override
     public Single<Map<Object, Object>> aggregate(String domain, AuditReportableCriteria criteria, Type analyticsType) {
         try {
-            return getReporter(domain).aggregate(ReferenceType.DOMAIN, domain, criteria, analyticsType);
+            return Maybe.fromOptional(getReporter(domain))
+                    .flatMapSingle(r ->r.aggregate(ReferenceType.DOMAIN, domain, criteria, analyticsType))
+                    .switchIfEmpty(Single.just(Map.of()));
```
leleueri, line 73: **suggestion**: same as previous comment

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/AuditServiceImpl.java
```diff
@@ -75,7 +80,8 @@ public Single<Map<Object, Object>> aggregate(String domain, AuditReportableCrite
     @Override
     public Single<Audit> findById(ReferenceType referenceType, String referenceId, String auditId) {
         try {
-            return getReporter(referenceType, referenceId).findById(referenceType, referenceId, auditId)
+            return Maybe.fromOptional(getReporter(referenceType, referenceId))
+                    .flatMap(r->r.findById(referenceType, referenceId, auditId))
```
leleueri, line 84: this one is OK as we want to throw an error if the search fails

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -359,6 +360,15 @@ public Single<Domain> create(String organizationId, String environmentId, NewDom
...
                 })
+
+                .onErrorResumeNext(ex -> {
+                    if (ex instanceof AbstractManagementException) {
+                        return Single.error(ex);
+                    }
+
+                    LOGGER.error("An error occurred while trying to create a domain", ex);
+                    return Single.error(new TechnicalManagementException("An error occurred while trying to create a domain", ex));
+                })
```
leleueri, line 371: **question**: what is the impact of moving this from the end of the reactive flow to here if the next steps are failing ?

mregulski-gravitee: looks like someting got lost in the rebase - resumeNext should stay at the end, there's no reason to move it

mregulski-gravitee:
Re: sonarqube: sonar hihglioghted all pre-existing uncovered conditions in domainServiceImpl as new uncovered code. I don't think it makes sense to cover all conditions of this service within the scope of this PR

leleueri (changes requested):

helm/templates/api/api-configmap.yaml
```diff
@@ -297,7 +297,10 @@ data:
     applications:
       {{- .Values.applications | toYaml | nindent 6 }}
     {{- end }}
-
+    {{- if .Values.api.domains.reporters }}
+    reporters:
+      {{- .Values.api.domains.reporters | toYaml | nindent 6 }}
+    {{- end }}
```
leleueri, line 303: **issue**: Should be under domains section in the same way as identities. so the indent should be 8. In addition can you add unit test for this config ?

leleueri (changes requested):

leleueri (approved):

PR #4468, opened 2024-07-18
feat: option to remove additional parameters from token response

podlesrafal:


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -108,6 +110,10 @@ public class TokenServiceImpl implements TokenService {
     @Autowired
     private AuditService auditService;
 
+    @Setter
+    @Value("${token.response.mapAdditionalParameters:true}")
```
leleueri, line 114: **discussion**: not sure about this key, maybe something like handlers.oidc.token.response.strict could be better (strict = true, only oauth2 standard values are provided, if false keep the current behaviour)

My main concern is about the first par of the key "token.", I would prefer to have a section entry which can be use later for other protocol (SAML, SCIM...) to keep a kind of logic if we have to adapt with a settings the behaviour of another endpoint

Each protocol are in a "gateway-handler" sub module (handler-oidc, handler-scim...) and we already have a `handlers` section into the [gravitee.yaml ](https://github.com/gravitee-io/gravitee-access-management/blob/master/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml#L332)

```
handlers:
  request:
    transaction:
      header: X-Gravitee-Transaction-Id
  oidc: (or oauth2)
    token:
      response:
        strict: true|false
```

wdyt ?

tcompiegne: We have also the `legacy` section ... : 

```
legacy:
  openid:
   sanitizeParametersEncoding:
   always_enhance_scopes: 
   openid_scope_full_profile:
   filterCustomPrompt:
```

leleueri: legacy is an option if we want to change the defautl AM behaviour to return only well known attributes in the response and give a way to keep previous behaviour if it breaks the deployment.

because this PR has been made because some framewok/lib reject additional fields, we were more in a logic to keep the behaviour as it is currently as the default one and give an option to limit the result to wellknown attributes. The reason is nothing prevent into the specification the response to have more attributes and a client should ignore the additional one. But maybe our reasoning is wrong on this point.

leleueri (approved):

PR #4482, opened 2024-07-22
fix: block override token standard response parameter.

podlesrafal:
Fixes: AM-3596

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/Token.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/Token.java
@@ -21,6 +21,7 @@ import lombok.Setter;
 import java.io.Serializable;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/Token.java
```diff
@@ -86,6 +87,10 @@ public abstract class Token implements Serializable {
      */
     public static final String UPGRADED = "upgraded";
 
+    public static List<String> getStandardParameters(){
+        return List.of(ACCESS_TOKEN, TOKEN_TYPE, EXPIRES_IN, REFRESH_TOKEN);
```
leleueri, line 91: issue: at least id_token is missing, please check in the OpenId specification if some other attributes can be present

leleueri (changes requested):
PLease also apply this fix on 4.4.x, 4.3.x, 4.2.x, 4.1.x as this bug is present in all version.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/Token.java
podlesrafal: adde id_token and checked the documentation. Only id_token was missing.

leleueri: Why not adding id_token here  ?

podlesrafal: Because this is not part of the Token class . In my feeling getStandardParameter should return only fields that are in this class.

leleueri (approved):

PR #4485, opened 2024-07-22
AM-3583 

lgw-gravitee:
added jsonform null string chema property to handle empty values submission

fixes AM-3583


lgw-gravitee:
ok, you're right. Its the simpliest solution. Ill close this PR.

PR #4532, opened 2024-07-29
fix: remove mock factor when go back on enroll mfa 2

podlesrafal:
Fixes AM-3627 - remove not needed code
PR on branches will be updated with this fix.


leleueri (commented):
**TODO**: can you please also manage the template into the management-api module ?

leleueri (approved):

PR #4557, opened 2024-07-31
fix: hide sensitive additional properties from api

mregulski-gravitee:
AM-3453

This PR hides "sensitive" user properties from being viewed.
These properties will still be readable in policies, EL, etc, but the REST API and consequently the UI will see a placeholder (`●●●●●●●●`) instead of the real value.
The sensitive properties can still be deleted/updated normally.

Currently the only properties marked sensitive are the properties used for storing original tokens from OIDC providers:
`op_access_token` and `op_id_token`.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/UserEntity.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/UserEntity.java
@@ -15,8 +15,6 @@
  */
 package io.gravitee.am.management.handlers.management.api.model;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import io.gravitee.am.model.User;
 import lombok.Getter;
 import lombok.Setter;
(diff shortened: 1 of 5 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/UserEntity.java
```diff
@@ -58,27 +66,39 @@ public UserEntity(User user) {
...
+            return properties.entrySet()
+                    .stream()
+                    .map(AdditionalProperty::wrap)
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+        }
+
+        private static Map.Entry<String, AdditionalProperty> wrap(Map.Entry<String, Object> e) {
+            return Map.entry(e.getKey(), new AdditionalProperty(e.getValue(), SENSITIVE_ADDITIONAL_PROPERTIES.contains(e.getKey()), false));
+        }
     }
```
leleueri, line 102: **question**: is it still useful ?

leleueri (changes requested):
Can you please add a simple test to ensure that the behaviour is correct.

leleueri (approved):

PR #4597, opened 2024-08-06
feat: AM-3525: support device lookup by users' externalId + source

mregulski-gravitee:
AM-3525

This PR introduces a UserId record to wrap the internal and/or external id together to use for entities that can be linked to a user by any of these ids.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DevicesResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DevicesResource.java
@@ -18,8 +18,8 @@ package io.gravitee.am.management.handlers.management.api.resources.organization
 import io.gravitee.am.management.handlers.management.api.resources.AbstractUsersResource;
 import io.gravitee.am.management.service.DomainService;
 import io.gravitee.am.model.Acl;
+import io.gravitee.am.model.ReferenceType;
 import io.gravitee.am.model.User;
-import io.gravitee.am.model.UserId;
 import io.gravitee.am.model.permissions.Permission;
 import io.gravitee.am.service.DeviceService;
 import io.gravitee.am.service.exception.DomainNotFoundException;
@@ -72,14 +72,15 @@ public class DevicesResource extends AbstractUsersResource {
     public void list(
             @PathParam("organizationId") String organizationId,
             @PathParam("environmentId") String environmentId,
-            @PathParam("domain") String domain,
-            @PathParam("user") String user,
+            @PathParam("domain") String domainId,
+            @PathParam("user") String userId,
             @Suspended final AsyncResponse response) {
 
-        checkAnyPermission(organizationId, environmentId, domain, Permission.DOMAIN_USER_DEVICE, Acl.LIST)
-                .andThen(domainService.findById(domain)
-                        .switchIfEmpty(Maybe.error(new DomainNotFoundException(domain)))
-                        .flatMapSingle(__ -> this.deviceService.findByDomainAndUser(domain, UserId.internal(user)).toList()))
+        checkAnyPermission(organizationId, environmentId, domainId, Permission.DOMAIN_USER_DEVICE, Acl.LIST)
+                .andThen(domainService.findById(domainId)
+                        .switchIfEmpty(Maybe.error(new DomainNotFoundException(domainId)))
+                        .flatMapSingle(domain -> this.userService.findById(ReferenceType.DOMAIN, domain.getId(), userId))
+                        .flatMapSingle(user -> this.deviceService.findByDomainAndUser(domainId, user.getFullId()).toList()))
                 .subscribe(response::resume, response::resume);
     }
 
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DevicesResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DevicesResource.java
@@ -16,11 +16,12 @@
 package io.gravitee.am.management.handlers.management.api.resources.organizations.environments.domains;
 
 import io.gravitee.am.management.handlers.management.api.resources.AbstractUsersResource;
+import io.gravitee.am.management.service.DomainService;
 import io.gravitee.am.model.Acl;
 import io.gravitee.am.model.User;
+import io.gravitee.am.model.UserId;
 import io.gravitee.am.model.permissions.Permission;
 import io.gravitee.am.service.DeviceService;
-import io.gravitee.am.management.service.DomainService;
 import io.gravitee.am.service.exception.DomainNotFoundException;
 import io.gravitee.common.http.MediaType;
 import io.reactivex.rxjava3.core.Maybe;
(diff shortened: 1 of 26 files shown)


gravitee-am-model/src/main/java/io/gravitee/am/model/UserId.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.model;
+
+public record UserId(String id, String externalId, String source) {
+    public UserId {
+        if (externalId == null ^ source == null) {
```
mstasgravitee, line 20:
```suggestion
        if (externalId == null || source == null) {
```

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DevicesResource.java
```diff
@@ -78,7 +79,7 @@ public void list(
         checkAnyPermission(organizationId, environmentId, domain, Permission.DOMAIN_USER_DEVICE, Acl.LIST)
                 .andThen(domainService.findById(domain)
                         .switchIfEmpty(Maybe.error(new DomainNotFoundException(domain)))
-                        .flatMapSingle(__ -> this.deviceService.findByDomainAndUser(domain, user).toList()))
+                        .flatMapSingle(__ -> this.deviceService.findByDomainAndUser(domain, UserId.internal(user)).toList()))
```
leleueri, line 82: **todo**: we need to search by (internalId or (externalId and source)) to be sure to retrieve all the devices for a single user. So you have to search the user first in the same way as we do for the domain to build the fullUserId.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcDeviceRepository.java
```diff
@@ -48,39 +50,49 @@ public class JdbcDeviceRepository extends AbstractJdbcRepository implements Devi
...
-                                .and(where(USER_FIELD).is(user))
+                                .and(userMatches(userId))
                                 .and(where(EXPIRES_AT_FIELD).greaterThanOrEquals(now))
                 ))
                 .all())
                 .map(this::toEntity);
     }
 
+    private Criteria userMatches(UserId user) {
+        if (user.isExternal()) {
```
leleueri, line 80:
```suggestion
        if (user.hasExternal()) {
```

Suggest to rename it `hasExternal` as we also check using the internalId in this case

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/MongoDeviceRepository.java
```diff
@@ -69,19 +72,23 @@ public void init() {
...
-        var query = and(eq(FIELD_REFERENCE_ID, referenceId), eq(FIELD_REFERENCE_TYPE, referenceType.name()), eq(FIELD_USER_ID, user), gte(FIELD_EXPIRES_AT, new Date()));
+    public Flowable<Device> findByReferenceAndUser(ReferenceType referenceType, String referenceId, UserId user) {
+        var query = and(eq(FIELD_REFERENCE_ID, referenceId), eq(FIELD_REFERENCE_TYPE, referenceType.name()), userIdMatches(user), gte(FIELD_EXPIRES_AT, new Date()));
         var devicePublisher = rememberDeviceMongoCollection.find(query);
         return Flowable.fromPublisher(devicePublisher).map(this::convert);
 
     }
 
+    private Bson userIdMatches(UserId user) {
+        return or(eq(FIELD_USER_ID, user.id()), and(eq(FIELD_USER_EXTERNAL_ID, user.externalId()), eq(FIELD_USER_SOURCE, user.source())));
```
leleueri, line 83: **todo**: why are you managing the userIdMatches method in a different manner than for JDBC?

Here is what you did for JDBC
```
if (user.isExternal()) {
     return where(USER_ID_FIELD).is(user.id()).or(where(USER_EXTERNAL_ID_FIELD).is(user.externalId()).and(USER_SOURCE_FIELD).is(user.source()));
        } else {
            return where(USER_ID_FIELD).is(user.id());
        }
    }
```

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/AbstractJdbcRepository.java
```diff
@@ -37,13 +39,17 @@
...
+import static org.springframework.data.relational.core.query.Criteria.where;
+
 /**
  * @author Eric LELEU (eric.leleu at graviteesource.com)
  * @author GraviteeSource Team
  */
 public abstract class AbstractJdbcRepository {
+    public static final String USER_ID_FIELD = "user_id";
+    public static final String USER_EXTERNAL_ID_FIELD = "user_external_id";
+    public static final String USER_SOURCE_FIELD = "user_source";
```
mstasgravitee, line 51:
```suggestion
    private static final String USER_ID_FIELD = "user_id";
    private static final String USER_EXTERNAL_ID_FIELD = "user_external_id";
    private static final String USER_SOURCE_FIELD = "user_source";
```

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_5_0/4.5.0-device-external-user-id.yml
```diff
@@ -0,0 +1,25 @@
...
+      validCheckSum: ANY
+      preConditions:
+        - onFail: MARK_RAN
+      changes:
+        - addColumn:
+            tableName: devices
+            columns:
+              - column:
+                  name: user_external_id
+                  type: nvarchar(255)
```
mstasgravitee, line 14:
```suggestion
                  type: nvarchar(64)
```

mstasgravitee (approved):

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/AbstractJdbcRepository.java
mregulski-gravitee: with another package structure these could be package-private, but I think they should be _at least_ protected - so that concrete repo implementation can reuse them. If they're left public that allows using these for `@Column` annotations for these fields, which might be good for keeping consistency beetween different entities sharing fields.

tcompiegne (changes requested):
Do we really need to have these two more fields ? 

I wonder if we can stick to only the userId which might have a new format for the resilience mode. 

Happy to discuss this internally.

mregulski-gravitee:
> Do we really need to have these two more fields ?
> 
> I wonder if we can stick to only the userId which might have a new format for the resilience mode.
> 
> Happy to discuss this internally.

It **would** be painful to have to manage 3 fields all over the place, that's why I've introduced the `UserId` value object. All the domain/business logic code uses the `UserId` abstraction. Whether it stored as 3 fields or magically formatted string is an implementation detail of the repository layer.
I'm strongly opposed to using "just a formatted `String`" as a field/param/return value in the business logic. The AM code uses Strings way too much for non-interchangeable values, making it hell to actually track usages or any more complex logic.
On the DB side, my opinion is that 3 fields are easier to work with than a custom-formatted string (e.g. for some future queries). It also makes it obvious **that** there are multiple parts and **what** these parts are.

PR #4604, opened 2024-08-08
feat: add ids for actor and target in audit

mstasgravitee:
https://gravitee.atlassian.net/browse/AM-3521


leleueri (commented):
**remark**: for kafka reporter, the map of attributes is not used. Don't think we can add this map for kafka specially because of the schema management and because in future if we want to go to avro as serialization protocol I'm not sure we can specify map of <string, object> (we an specify a map but all values need to be in the same type) 

So either, we set explicitly the externalId and sourceId as field in the AuditEntity class and all reporter need to be adapt to manage these two fields. Or we only adapt kafkareporter to extract from the map these two values to set them explicitly as field into AuditEntityDto

podlesrafal:
TODO: add liquibase file to add new fields in audit collection

gravitee-am-reporter/gravitee-am-reporter-jdbc/src/main/java/io/gravitee/am/reporter/jdbc/audit/JdbcAuditReporter.java
```diff
@@ -534,8 +538,11 @@ private Mono<Long> prepateInsertEntity(Audit audit, AuditEntity entity, String f
         insertEntitySpec = addQuotedField(insertEntitySpec, COL_TYPE, entity.getType(), String.class);
         insertEntitySpec = addQuotedField(insertEntitySpec, COL_DISPLAY_NAME, entity.getDisplayName(), String.class);
         insertEntitySpec = addQuotedField(insertEntitySpec, COL_REFERENCE_TYPE, entity.getReferenceType() == null ? null : entity.getReferenceType().name(), String.class);
-        insertEntitySpec = addQuotedField(insertEntitySpec, COL_REFERENCE_ID, entity.getReferenceId(), String.class);
+        insertEntitySpec = addQuotedField(insertEntitySpec, COL_EXTERNAL_ID, entity.getReferenceId(), String.class);
+        insertEntitySpec = addQuotedField(insertEntitySpec, COL_SOURCE_ID, entity.getSourceId(), String.class);
```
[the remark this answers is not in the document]
mstasgravitee: fixed

gravitee-am-reporter/gravitee-am-reporter-api/src/main/java/io/gravitee/am/reporter/api/audit/model/AuditEntity.java
```diff
@@ -57,6 +57,10 @@ public class AuditEntity {
      */
     private String referenceId;
 
+    private String externalId;
```
tcompiegne, line 60: Those two fields are very specific to a User and not the others type of Audits (App, Domain, IdP, ...) I'm not sure this is the best solution to add the values at the root level of an Audit ...

tcompiegne (changes requested):

gravitee-am-reporter/gravitee-am-reporter-api/src/main/java/io/gravitee/am/reporter/api/audit/model/AuditEntity.java
tcompiegne: I conscious that by using the attributes it may be difficult to search / filters on those values ...

gravitee-am-ui/src/app/domain/settings/audits/audit/audit.component.html
```diff
@@ -42,11 +42,16 @@ <h1>Audit log detail</h1>
...
           <mat-label>Actor</mat-label>
-          <input matInput placeholder="Actor" disabled [value]="audit.actor?.displayName" />
+          <input matInput placeholder="Actor" disabled [value]="getActorDetails()" />
           <mat-hint>Who triggered the event</mat-hint>
         </mat-form-field>
+        <mat-form-field *ngIf="audit.target" appearance="outline" floatLabel="always">
+          <mat-label>Target</mat-label>
+          <input matInput placeholder="Target" disabled [value]="getTargetDetails()" />
+          <mat-hint>The target object</mat-hint>
+        </mat-form-field>
```
leleueri, line 54: TODO: can you please keep only the displayName as it was before ? The change you propose is not really readable so I would prefer to provide the display name only.

leleueri (changes requested):

gravitee-am-ui/src/app/domain/settings/audits/audit/audit.component.html
mstasgravitee: fixed

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/builder/AuditBuilder.java
```diff
@@ -158,7 +167,7 @@ public T oldValue(Object oldValue) {
 
     public T principal(User principal) {
         if (principal != null) {
-            setActor(principal.getId(), EntityType.USER, principal.getUsername(), getDisplayName(principal), getReferenceType(principal), getReferenceId(principal));
+            setActor(principal.getId(), EntityType.USER, principal.getUsername(), getDisplayName(principal), getReferenceType(principal), getReferenceId(principal), null, null);
```
tcompiegne, line 170: Will it not be better to add a new method parameters for setActor and setTarget with these two extra information ? 

I'm not a big fan of having (null, null) as parameters everywhere. 

Thanks

tcompiegne (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/builder/AuditBuilder.java
mstasgravitee: fixed

leleueri (approved):

PR #4666, opened 2024-08-22
fix: redirect_url not required for registration confirmation

podlesrafal:
Fixes: https://gravitee.atlassian.net/browse/AM-3718


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/EmailServiceImpl.java
```diff
@@ -247,6 +248,10 @@ private Map<String, Object> prepareEmailParams(Domain domain, Application client
             claims.put(Claims.AUD, user.getClient());
         }
 
+        if (template == io.gravitee.am.model.Template.REGISTRATION_CONFIRMATION) {
```
mstasgravitee, line 251: do we need this? 

In management we have something like this
     getTokenPurpose(template)
                .ifPresent(purpose -> claims.put(ConstantKeys.CLAIM_TOKEN_PURPOSE, purpose));

mstasgravitee (approved):

PR #4673, opened 2024-08-23
[AM-3651] fix: On jwt-bearer flow the external-id is not known, thus creating g…

lgw-gravitee:
…is is not possible.

fixes AM-3651

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/subject/SubjectManagerV2.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/subject/SubjectManagerV2.java
@@ -30,7 +30,10 @@ import lombok.extern.slf4j.Slf4j;
 import org.springframework.util.StringUtils;
 
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 import java.util.UUID;
+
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/subject/SubjectManagerV2.java
```diff
@@ -55,7 +58,9 @@ public String generateSubFrom(User user) {
 
     @Override
     public String generateInternalSubFrom(User user) {
-        return user.getSource() + SEPARATOR + user.getExternalId();
+        return Optional.ofNullable(user.getAdditionalInformation())
+                .map(map -> (String) map.get(GIO_INTERNAL_SUB))
+                .orElseGet(() -> user.getSource() + SEPARATOR + user.getExternalId());
```
leleueri, line 63: **discussion**: I don't think this is the place to test and fallback on the gio_internal_sub :thinking: And I'm not even sure that we have to fallback to the additional property as it is very specific to the ExtensionGrant plugin... I'm afraid this change will bring more issues in the future...

lgw-gravitee: Impleneted accordingly

leleueri (approved):

PR #4695, opened 2024-08-28
fix: prevent npe when collecting null value

podlesrafal:
fixes: https://gravitee.atlassian.net/browse/AM-3733


leleueri (commented):
can you please add simple test to demonstrate the behaviour ? (evenif it looks ok)

podlesrafal:
added test

leleueri (approved):

PR #4744, opened 2024-09-05
fix: fix device delete to use userId

podlesrafal:
fixes: https://gravitee.atlassian.net/browse/AM-3737


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DeviceServiceImpl.java
```diff
@@ -95,13 +95,8 @@ public Single<Device> create(String domain, String client, UserId user, String r
     public Completable delete(String domain, UserId user, String deviceId, User principal) {
         return deviceRepository.findById(deviceId)
                 .switchIfEmpty(Maybe.error(new DeviceNotFoundException(deviceId)))
-                .flatMapCompletable(device -> {
-                    if (DOMAIN.equals(device.getReferenceType()) && device.getReferenceId().equals(domain) && device.getUserId().equals(user)) {
-                        return deviceRepository.delete(deviceId).andThen(Completable.complete());
-                    } else {
-                        return Completable.error(new DeviceNotFoundException(deviceId));
-                    }
-                }).onErrorResumeNext(ex -> {
```
leleueri, line 104: Question: I guess the bug is link to this condition where we are checking that the device is linked to the right domain and user. I do not understand why it is not wroking as in nominal mode (I mean no in degraded mode), the device will be linked to the right userId. So IMO this test should pass. In addition as we want to remove the externalId & the source from the device entity, the bug may not be present after that...

podlesrafal: The thing is that Device has UserId object. The previous version compare `device.getUserId().equals(user)` so here is comparison of UserId object and String with userId from request. I dont know where this userid comes from. There can be situation when the device was created in degraded mode and the UserId object will not contain `id` but `externalId` and `source` so I cannot do like this: `device.getUserId().getId().equals(user)`

leleueri: You can, as I said we will remove the externalId & sourceId from the device (AM-3753). As the device entity is managed by the CP database, in degraded mode it will not be possible to create such entity

podlesrafal: ok, I didnt know that. I will apply the change.

leleueri (approved):

PR #4813, opened 2024-09-17
chore: rewrite discovery oidc postman collection to jest test

podlesrafal:
AM-3775


leleueri (approved):
**Thought**: Could be nice to see if there is a way to generate PKCS12 certificate using Jest in order to avoid static B64 payload

PR #4816, opened 2024-09-17
fix: make jdbc user repository to perform upserts on user update

lgw-gravitee:
fixes AM-3875

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcUserRepository.java
+++ b/gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcUserRepository.java
@@ -74,6 +74,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcUserRepository.java
```diff
@@ -842,14 +842,21 @@ public Single<User> update(User item, UpdateActions updateActions) {
 
         Mono<Long> action = update.fetch().rowsUpdated();
 
-
         if (updateActions.updateRequire()) {
-            action = deleteChildEntities(item.getId(), updateActions).then(action);
+            if (item.getId() != null) {
```
leleueri, line 846: **remark**: shouldn't be possible to have null ID in this method. If there is a case, then maybe we should throw an error at the beginning of the update method to avoid side effect issue with the update request
` update = addQuotedField(update, USER_COL_ID, item.getId(), String.class);`  wdyt ?

lgw-gravitee: valid point. I also added one test

leleueri (approved):

PR #4892, opened 2024-09-24
fix: upgrade gravitee-node to fix cache settings issue

leleueri:
fixes AM-3974


leleueri:
Gravitee-node has been upgraded by https://github.com/gravitee-io/gravitee-access-management/pull/4842

PR #4896, opened 2024-09-25
fix: handle missing code verifier gracefully, error handling improvement

mregulski-gravitee:
AM-3982

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
@@ -17,9 +17,10 @@ package io.gravitee.am.gateway.handler.root.resources.auth.provider;
 
 import io.gravitee.am.common.exception.authentication.BadCredentialsException;
 import io.gravitee.am.common.exception.authentication.LoginCallbackFailedException;
+import io.gravitee.am.common.exception.oauth2.InvalidRequestException;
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
```diff
@@ -116,6 +117,14 @@ public void authenticate(RoutingContext context, JsonObject authInfo, Handler<As
             endUserAuthentication.getContext().set(Claims.USER_AGENT, RequestUtils.userAgent(context.request()));
         }
 
+        if (context.queryParams().contains(ConstantKeys.ERROR_PARAM_KEY)) {
+            var error = context.queryParams().get(ConstantKeys.ERROR_PARAM_KEY);
+            var description = context.queryParams().get(ConstantKeys.ERROR_DESCRIPTION_PARAM_KEY);
+            var message = error + (description != null ? ":" + description : "");
+            context.fail(new InvalidRequestException(message));
+            return;
+        }
```
leleueri, line 126: **thought**: is it really safe to had this code :thinking: ? if a login failed the user is redirect back to the login page, (url will contains error), so in the second attempt error param will be present no ?

mregulski-gravitee: This code is meant to display a correct error in case PKCE is misconfigured.

If the upstream IDP forces PKCE, but the provider configuration has no `codeChallengeMethod`, the user will get redirected back to the login page with "social_authentication_failed Missing authorization code" (without any information about the error from the upstream IDP).
If that's the expected behaviour, I'll remove this change.

PR #4941, opened 2024-10-03
feat(helm): allow configuring security section of management api conf…

mregulski-gravitee:
…ig via Helm values

AM-4057

### DIFF
--- a/helm/templates/api/api-configmap.yaml
+++ b/helm/templates/api/api-configmap.yaml
@@ -396,7 +396,16 @@ data:
       {{- end }}
       # Allows to define if cookie secure only (default false)
       cookie-secure: {{ .Values.api.jwt.cookie.secure | default false }}
-
+    security:
+      defaultAdmin: {{ .Values.api.security.defaultAdmin | default true }}
+      accountAccessTokens:
+        encoder:
+          settings:
+            rounds: {{ .Values.api.security.accountTokenBcryptRounds | default 10 }}
+      {{- if .Values.api.security.providers }}
+      providers:
+        {{- toYaml .Values.api.security.providers | nindent 8 }}
+      {{- end}}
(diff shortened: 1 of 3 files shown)


helm/templates/api/api-configmap.yaml
```diff
@@ -396,7 +396,16 @@ data:
...
+    security:
+      defaultAdmin: {{ .Values.api.security.defaultAdmin | default true }}
+      accountAccessTokens:
+        encoder:
+          settings:
+            rounds: {{ .Values.api.security.accountTokenBcryptRounds | default 10 }}
+      {{- if .Values.api.security.providers }}
+      providers:
+        {{- toYaml .Values.api.security.providers | nindent 8 }}
+      {{- end}}
```
leleueri, line 408: suggestion: I would do a test on the **security** section as you did for `.Values.api.security.providers` line 405. In that way we are expecting the same structure as the one in the gravitee.yaml. And in future if new entries in the security section are added, they will be taken into account automatically. wdyt ?

helm/values.yaml
```diff
@@ -436,6 +436,30 @@ api:
...
+        users:
+          - user:
+            username: admin
+            #email:
+            firstname: Administrator
+            lastname: Administrator
+            ## Passwords are encoded using BCrypt
+            ## Password value: adminadmin
+            password: $2a$10$NG5WLbspq8V1yJDzUKfUK.oum94qL/Ne3B5fQCgekw/Y4aOEaoFZq
+            role: ORGANIZATION_OWNER
```
leleueri, line 462: if you accept the comment made for the api-configmap.yalm this section will have to be updated for the accountToken.

leleueri (changes requested):
Can you please add unit test ?

leleueri (approved):

PR #4943, opened 2024-10-04
fix: AM-4047 introspection on refresh token

lgw-gravitee:
fixes AM-4047

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
@@ -25,7 +25,7 @@ import io.gravitee.am.common.utils.ConstantKeys;
 import io.gravitee.am.common.utils.RandomString;
 import io.gravitee.am.common.utils.SecureRandomString;
 import io.gravitee.am.gateway.handler.common.jwt.JWTService;
-import io.gravitee.am.gateway.handler.common.oauth2.IntrospectionTokenService;
+import io.gravitee.am.gateway.handler.common.oauth2.IntrospectionTokenFacade;
 import io.gravitee.am.gateway.handler.context.ExecutionContextFactory;
 import io.gravitee.am.gateway.handler.oauth2.exception.InvalidGrantException;
 import io.gravitee.am.gateway.handler.oauth2.service.request.OAuth2Request;
@@ -99,7 +99,7 @@ public class TokenServiceImpl implements TokenService {
     private TokenManager tokenManager;
 
     @Autowired
-    private IntrospectionTokenService introspectionTokenService;
+    private IntrospectionTokenFacade introspectionTokenFacade;
 
     @Override
     public Maybe<Token> getAccessToken(String token, Client client) {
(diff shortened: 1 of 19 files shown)


mregulski-gravitee (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -126,11 +126,17 @@ public Maybe<Token> getRefreshToken(String refreshToken, Client client) {
...
-        return introspectionTokenService.introspect(token, false)
-                .map(this::convertAccessToken);
+    public Single<Token> introspect(Token token) {
+        if(token instanceof RefreshToken){
+            return introspectionTokenFacade.introspectRefreshToken(token.getValue())
+                    .map(this::convertRefreshToken);
+        } else {
+            return introspectionTokenFacade.introspectAccessToken(token.getValue())
+                    .map(this::convertAccessToken);
+        }
```
leleueri, line 136: **discussion**: According to the spec https://datatracker.ietf.org/doc/html/rfc7662#section-2.1

> token_type_hint
      OPTIONAL.  A hint about the type of the token submitted for
      introspection.  The protected resource MAY pass this parameter to
      help the authorization server optimize the token lookup.  If the
      server is unable to locate the token using the given hint, it MUST
      extend its search across all of its supported token types.  An
      authorization server MAY ignore this parameter, particularly if it
      is able to detect the token type automatically.  Values for this
      field are defined in the "OAuth Token Type Hints" registry defined
      in OAuth Token Revocation [[RFC7009](https://datatracker.ietf.org/doc/html/rfc7009)].

So IMO when token_type_hint is missing we have to test access_token and then refresh_token if access_token is unkown. If token_type_hint is provided we have to test first the type but fallback to the other one if the one provided in parameter is unknown... :thinking: wdyt ?

lgw-gravitee: yeap, you're right. Ill apply the change
(I need to read rfc more carefully next time)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -126,9 +126,24 @@ public Maybe<Token> getRefreshToken(String refreshToken, Client client) {
...
+    }
+
+    @Override
+    public Maybe<Token> introspect(String token, TokenTypeHint hint) {
+        switch (hint) {
+            case REFRESH_TOKEN:
+                return introspectionTokenFacade.introspectRefreshToken(token)
+                        .map(this::convertRefreshToken)
+                        .switchIfEmpty(introspectionTokenFacade.introspectAccessToken(token).map(this::convertAccessToken));
+            case ID_TOKEN:
```
leleueri, line 142: **thought**: is it useful to process ID_TOKEN ? IMO it can be the default behaviour by returning Maybe.empty() because introspect is not about id_token and anyway this token is currently not persisted in the DB so we can save access to the storage layer.

lgw-gravitee: right. I fixed that

leleueri (approved):

PR #4959, opened 2024-10-07
fix: PEN-11 json cfg validator

lgw-gravitee:


podlesrafal (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/jsonstring/JsonString.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.METHOD, ElementType.FIELD, ElementType.ANNOTATION_TYPE, ElementType.CONSTRUCTOR, ElementType.PARAMETER, ElementType.TYPE_USE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Constraint(validatedBy = JsonStringValidator.class)
+public @interface JsonString {
```
leleueri, line 31: question: is it possible to add a param to specify which field is invalid ? We only have configuration here but this annotation maybe useful somewhere else. wdyt ?

leleueri (commented):
In addition of the question, maybe we should apply this to all plugins.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/jsonstring/JsonString.java
lgw-gravitee: you can set custom error message 
`    @JsonString(message = "configuration must be json form")
`

but more general solution would be in ValidactionExceptionMapper where propertyPath can be retrieved and setup in error message

leleueri (approved):

PR #4964, opened 2024-10-08
fix: AM-4068 with 'includeLinkedIdentities'=true the mongo db query uses more…

lgw-gravitee:
… general index for referenceType and referenceId only - it basically makes a collection scan

query:
image
plan:
image

when query is extended with OR...identities_username=... then more general index is fired which makes basically collection scan.


podlesrafal (approved):

mregulski-gravitee (approved):

leleueri (commented):
Do we have to apply the kin dof indexes on R2DBC implementation ?

lgw-gravitee:
4.1.23-update-user-identities-table.yml was already added
image

leleueri (approved):

PR #4966, opened 2024-10-08
fix: added hash of error to the session

podlesrafal:
To avoid manipulating error message by changing the url, I introduced hash of error message that is stored in session. 
Fixes: https://gravitee.atlassian.net/browse/PEN-10

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/HashUtil.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/HashUtil.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.common.utils;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.HexFormat;
+
+@Slf4j
+public class HashUtil {
+    public static String generateSHA256(String input) {
+        try {
+            MessageDigest digest = MessageDigest.getInstance("SHA-256");
+            byte[] encodedHash = digest.digest(input.getBytes(StandardCharsets.UTF_8));
+            return HexFormat.of().formatHex(encodedHash);
+        } catch (NoSuchAlgorithmException e) {
+            log.debug("Cannot generate hash", e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static boolean compare(String hash, String rawString) {
+        String toCompare = generateSHA256(rawString);
+        return hash.equals(toCompare);
+    }
+}
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
@@ -23,9 +23,11 @@ import io.gravitee.am.model.oidc.Client;
 import io.gravitee.am.model.safe.ClientProperties;
 import io.gravitee.am.model.safe.DomainProperties;
 import io.gravitee.am.model.safe.UserProperties;
+import io.vertx.rxjava3.core.MultiMap;
 import io.vertx.rxjava3.ext.web.RoutingContext;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
+import org.jsoup.internal.StringUtil;
 
 import java.util.HashMap;
 import java.util.Map;
@@ -33,8 +35,13 @@ import java.util.Optional;
 
 import static io.gravitee.am.common.utils.ConstantKeys.CLIENT_CONTEXT_KEY;
 import static io.gravitee.am.common.utils.ConstantKeys.DOMAIN_CONTEXT_KEY;
+import static io.gravitee.am.common.utils.ConstantKeys.ERROR_CODE_PARAM_KEY;
+import static io.gravitee.am.common.utils.ConstantKeys.ERROR_DESCRIPTION_PARAM_KEY;
+import static io.gravitee.am.common.utils.ConstantKeys.ERROR_HASH;
+import static io.gravitee.am.common.utils.ConstantKeys.ERROR_PARAM_KEY;
 import static io.gravitee.am.common.utils.ConstantKeys.PARAM_CONTEXT_KEY;
 import static io.gravitee.am.common.utils.ConstantKeys.REQUEST_CONTEXT_KEY;
+import static io.gravitee.am.common.utils.ConstantKeys.SERVER_ERROR;
 import static io.gravitee.am.common.utils.ConstantKeys.USER_CONTEXT_KEY;
 
 /**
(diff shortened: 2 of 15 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/HashUtil.java
```diff
@@ -0,0 +1,50 @@
...
+
+            StringBuilder hexString = new StringBuilder();
+            for (byte b : encodedHash) {
+                String hex = Integer.toHexString(0xff & b);
+                if (hex.length() == 1) {
+                    hexString.append('0');
+                }
+                hexString.append(hex);
+            }
+            return hexString.toString();
```
leleueri, line 39: Question: why not using the a utility method brings by one of our dependencies or using the std Base64 encoder ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
```diff
@@ -59,11 +65,38 @@ public static Map<String, Object> generateData(RoutingContext context, Domain do
...
+                }
+                if (queryParams.get(ERROR_DESCRIPTION_PARAM_KEY) != null) {
+                    data.put(ERROR_DESCRIPTION_PARAM_KEY, queryParams.get(ERROR_DESCRIPTION_PARAM_KEY));
+                    error.append("$");
+                    error.append(queryParams.get(ERROR_DESCRIPTION_PARAM_KEY));
+                }
+                if (!StringUtil.isBlank(error.toString()) && !HashUtil.compare(errorHash, error.toString())) {
+                    data.put(ERROR_PARAM_KEY, "Unknown error");
+                    data.put(ERROR_CODE_PARAM_KEY, null);
+                    data.put(ERROR_DESCRIPTION_PARAM_KEY, null);
```
leleueri, line 95:
```suggestion
                data.put(ERROR_PARAM_KEY, SERVER_ERROR);
                data.put(ERROR_CODE_PARAM_KEY, null);
                data.put(ERROR_DESCRIPTION_PARAM_KEY, "Unknown error occurred");
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/error/AbstractErrorHandler.java
```diff
@@ -66,9 +69,19 @@ public final void handle(RoutingContext routingContext) {
...
                 parameters.set(ConstantKeys.ERROR_DESCRIPTION_PARAM_KEY, "Unexpected error occurred");
 
+                String toHash = SERVER_ERROR + "$" + "Unexpected error occurred";
+                String hash = HashUtil.generateSHA256(toHash);
+                if (routingContext.session() != null) {
+                    if (!routingContext.session().isDestroyed()) {
+                        routingContext.session().put(ERROR_HASH, hash);
+                    } else {
+                        routingContext.session().put(ERROR_HASH, hash);
+                    }
```
leleueri, line 82: question: Are you sure about this one ? either the condition is not useful and you can add the hash everytime or a destoyed session can't receive a new value without an initialization.

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/HashUtil.java
```diff
@@ -0,0 +1,50 @@
...
+        } catch (NoSuchAlgorithmException e) {
+            log.debug("Cannot generate hash. Using plain string", e);
+            return input;
+        }
```
mregulski-gravitee, line 43: I know this method is probably only going to be used for this one case, but this catch clause makes it misleading - it might in fact return a string that's NOT the sha256 hash of the input.

However, given all compliant JVM implementations MUST have a SHA-256 implementation, we can be pretty confident this exception will never be thrown - and if it **gets** thrown, something is bery very broken.
IMO it means it's better to just rethrow the NoSuchAlgorithmException as a RuntimeException so that we stop compiler from complaining without introducing incorrect code paths.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/error/ErrorHandler.java
```diff
@@ -97,29 +99,41 @@ private void handleException(RoutingContext routingContext, String errorCode, St
...
-            context.session().destroy();
+    private void doRedirect(RoutingContext context, String url, String errors) {
+
+        if (context.session() != null) {
+            if (destroySession) {
+                context.session().destroy();
+            }
+            if (errors != null) {
+                context.session().put(ERROR_HASH, errors);
+            }
```
mregulski-gravitee, line 136: If `destroySession` is `true`, and there's an error this code will destroy the session (clearing all the data), and then put the error hash in in (which essentially starts a new session, with a new session id; see `io.vertx.ext.web.sstoreio.vertx.ext.web.sstore.AbstractSession` `#put()`, `#data()`)

Is that the behaviour we want?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/error/AbstractErrorHandler.java
podlesrafal: when the session is destroyed and there is .put() method called, the session will be recreated.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/error/ErrorHandler.java
podlesrafal: Yes, we kind of cleaning up session and then recreate only with error hash.

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/error/ErrorHandler.java
```diff
@@ -97,29 +99,41 @@ private void handleException(RoutingContext routingContext, String errorCode, St
...
             logger.error("Unable to handle root error response", e);
-            doRedirect(routingContext, errorPageURL);
+            doRedirect(routingContext, errorPageURL, null);
         }
     }
 
-    private void doRedirect(RoutingContext context, String url) {
-        if (destroySession && context.session() != null) {
-            context.session().destroy();
+    private void doRedirect(RoutingContext context, String url, String errors) {
```
leleueri, line 128:
```suggestion
    private void doRedirect(RoutingContext context, String url, String errorHash) {
```

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/login/LoginCallbackFailureHandler.java
```diff
@@ -283,7 +285,15 @@ private void redirectToLoginPage(MultiMap originalParams,
         }
         params.set(Parameters.CLIENT_ID, client.getClientId());
         params.set(ConstantKeys.ERROR_PARAM_KEY, "social_authentication_failed");
-        params.set(ConstantKeys.ERROR_DESCRIPTION_PARAM_KEY, encodeURIComponent(throwable.getCause() != null ? throwable.getCause().getMessage() : throwable.getMessage()));
+        String errorDescription = (throwable.getCause() != null ? throwable.getCause().getMessage() : throwable.getMessage());
+        params.set(ConstantKeys.ERROR_DESCRIPTION_PARAM_KEY, errorDescription);
```
mregulski-gravitee, line 289: `errorDescription` used to be URI-encoded via `encodeURIComponent`. We do use it as a URI param, so I think we should keep encoding it.

mregulski-gravitee (approved):

PR #4996, opened 2024-10-11
fix: helm charts should be used in branch version

lgw-gravitee:
fixes problem with HelmCharts in v4.5 used across all maint version


leleueri (approved):

podlesrafal:
Shouldnt be apply only on 4.2, 4.3, 4.4 versions?

podlesrafal (approved):

leleueri:
> Shouldnt be apply only on 4.2, 4.3, 4.4 versions?

no because we also need to avoid this issue in the future so 4.5 & master is fine (and 4.1.x also as we want to release the latest version today :) )

leleueri (approved):

PR #5007, opened 2024-10-15
fix: http status code on rate limit exception

podlesrafal:
fixes https://gravitee.atlassian.net/browse/AM-4101


lgw-gravitee (approved):

leleueri (commented):
Should also be apply on 4.3.x, 4.4.x, 4.5.x and master

leleueri (commented):
suggestion: we should have a jest test in place to check the MFA rate limite, maybe you can add an assertion on the status ?

leleueri (approved):

PR #5008, opened 2024-10-15
feat(mapi): bulk create users (domain & org)

mregulski-gravitee:
AM-3573

This PR introduces new bulk endpoints for creating domain & organisation users.
It's implemented using a new `BulkRequest` class that provides a unified way of representing and handling bulk execution of existing operations.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/bulk/BulkRequest.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/bulk/BulkRequest.java
@@ -15,33 +15,121 @@
  */
 package io.gravitee.am.management.handlers.management.api.bulk;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.gravitee.am.common.utils.Indexed;
+import io.gravitee.am.model.Acl;
 import io.reactivex.rxjava3.core.Single;
 import jakarta.ws.rs.core.Response;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.experimental.Accessors;
 import org.springframework.util.CollectionUtils;
 
 import java.util.List;
 import java.util.function.Function;
 
-public record BulkRequest<T>(List<T> items) {
+/**
+ * Represents a create, update or delete operation on multiple entities.
+ *
+ * @param <T> type of item included
+ */
+@RequiredArgsConstructor
+@Getter
+@Accessors(fluent = true)
+public class BulkRequest<T> {
+    @RequiredArgsConstructor
+    @Getter
+    @Accessors(fluent = true)
+    public enum Action {
+        CREATE(Acl.CREATE), UPDATE(Acl.UPDATE), DELETE(Acl.DELETE);
+        private final Acl requiredAcl;
+    }
+
+    @JsonProperty("action")
+    private final Action action;
+    @JsonProperty("items")
+    private final List<T> items;
+
+    protected BulkRequest(Action action) {
+        this(action, List.of());
+    }
 
     /**
      * Apply the processor to each individual item in this request. There's no guarantees about ordering or concurrency
-     * of processing; the response is guaranteed to have results of processing individual items in the input order
+     * of processing; the response is guaranteed to have results of processing individual items in the input order.
      */
-    public Single<Response> processOneByOne(Function<T, Single<BulkOperationResult>> processor) {
-        if (CollectionUtils.isEmpty(items)) {
+    public final <R> Single<Response> processOneByOne(Function<T, Single<BulkOperationResult<R>>> processor) {
+        if (CollectionUtils.isEmpty(items())) {
             return Single.just(Response.noContent().build());
         }
-        return Indexed.toIndexedFlowable(items)
+        return Indexed.toIndexedFlowable(items())
                 .flatMapSingle(indexed -> processor.apply(indexed.value())
                         .map(result -> result.withIndex(indexed.index())))
                 .toList()
                 .map(this::makeResponse);
     }
 
-    private Response makeResponse(List<BulkOperationResult> bulkOperationResults) {
-        var bulkResponse = new BulkResponse(bulkOperationResults);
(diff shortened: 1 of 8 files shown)


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/bulk/BulkRequest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import io.gravitee.am.common.utils.Indexed;
+import io.reactivex.rxjava3.core.Single;
+import jakarta.ws.rs.core.Response;
+import org.springframework.util.CollectionUtils;
+
+import java.util.List;
+import java.util.function.Function;
+
+public record BulkRequest<T>(List<T> items) {
```
leleueri, line 26: **question**: wondering if we should have a operation type in addition of the list of items. :thinking: to be able to make the distinction between creation or update bulk request.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/UsersResource.java
```diff
@@ -148,6 +145,46 @@ public void create(
...
+    @ApiResponse(responseCode = "201", description = "All users successfully created",
+            content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = BulkResponse.class)))
+    @ApiResponse(responseCode = "500", description = "Internal server error")
+    public void bulkCreate(
+            @PathParam("organizationId") String organizationId,
+            @PathParam("environmentId") String environmentId,
+            @PathParam("domain") String domainId,
+            @Parameter(name = "user", required = true)
+            @Valid @NotNull final BulkRequest<NewUser> newUsers,
```
leleueri, line 172: question: should we really use a NewUser ? I have some doubt as IMO the update user using bulk should also be done on this bulk endpoint using a POST method.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/users/UsersResource.java
```diff
@@ -138,6 +138,36 @@ public void create(
...
+            summary = "Create a platform user or Service Account",
+            description = "User must have the ORGANIZATION_USER[READ] permission on the specified organization")
+
+    @ApiResponse(responseCode = "201", description = "Users or Service Accounts successfully created",
+            content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = NewOrganizationUser.class)))
+    @ApiResponse(responseCode = "500", description = "Internal server error")
+    public void create(
+            @PathParam("organizationId") String organizationId,
+            @Valid @NotNull final BulkRequest<NewOrganizationUser> newOrganizationUsers,
```
leleueri, line 156: question: should we really use a NewUser ? I have some doubt as IMO the update user using bulk should also be done on this bulk endpoint using a POST method.

gravitee-am-management-api/gravitee-am-management-api-rest/src/test/java/io/gravitee/am/management/handlers/management/api/bulk/BulkRequestTest.java
```diff
@@ -0,0 +1,64 @@
...
+                .awaitDone(MAX_DELAY_MS * 5, TimeUnit.MILLISECONDS)
+                .assertComplete()
+                .assertValueCount(1)
+                .values()
+                .get(0)
+                .getResults();
+
+        // even with out-of-order processing, results should be returned in the input order
+        AssertionsForInterfaceTypes.assertThat(actualResults).containsExactlyElementsOf(expectedResults);
+
```
lgw-gravitee, line 56: it may be feasible to user TestScheduler where you can control timeflow :P

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/UsersResource.java
mregulski-gravitee: I've asked about it either on slack or on daily, and the answer was that there will be _separate_ endpoints for different operations (create/update/delete). If not, we will need to discuss another structure

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/users/UsersResource.java
mregulski-gravitee: (see answer in domain's UsersResource)

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/bulk/BulkRequest.java
mregulski-gravitee: What I understood from discussions on slack/daily was there's going to be separate endpoints for different operations. So a "bulk create" endpoint would accept a BulkRequest<NewThing> and process accordingly, and a "bulk update" endpoint would accept a BulkRequest<UpdatedThing>.

If that's not the case, we'll need to change this class to force in-order execution of each "item".

leleueri: yes but which HTTP method do you plan to use ? With a step back single bulk action on UserResource maybe better but this is a discussion to have will all team member. 

But I don't understand why we'll need to change this class to force in-order execution of each "item".
I was thinking to the something like:

```
POST /organisation/DEFAULT/environment/DEFAULT/domain/domainId/users/bulk
Content-Type: application /json

{
  "action": "create", 
  "items": [ {"username":...} , ...]
}
```
or
```
POST /organisation/DEFAULT/environment/DEFAULT/domain/domainId/users/bulk
Content-Type: application /json

{
  "action": "update",
  "items": [ {"id":...} , ...]
}
```
or
```
POST /organisation/DEFAULT/environment/DEFAULT/domain/domainId/users/bulk
Content-Type: application /json

{
  "action": "delete",
  "items": [ {"id":...} , ...]
}
```

mregulski-gravitee: oh, so still one action per request, just with a single endpoint handling each action? That makes sense, I'll update the request class accordingly

lgw-gravitee (approved):

PR #5037, opened 2024-10-18
fix: display server error when session is empty.

podlesrafal:
Fixes PEN-10

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
@@ -72,32 +72,41 @@ public class ThymeleafDataHelper {
         }
         data.put(PARAM_CONTEXT_KEY, parameters);
         if (context.session() != null && context.getDelegate() != null) {
+            MultiMap queryParams = context.queryParams();
             String errorHash = context.session().get(ERROR_HASH);
+            String error = queryParams.get(ERROR_PARAM_KEY);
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
```diff
@@ -72,34 +72,43 @@ public static Map<String, Object> generateData(RoutingContext context, Domain do
...
-                if (!StringUtil.isBlank(error.toString()) && !HashUtil.compare(errorHash, error.toString())) {
-                    data.put(ERROR_PARAM_KEY, SERVER_ERROR);
-                    data.put(ERROR_CODE_PARAM_KEY, null);
-                    data.put(ERROR_DESCRIPTION_PARAM_KEY, "Unknown error occurred");
+                if (!StringUtil.isBlank(errorSB.toString()) && !HashUtil.compare(errorHash, errorSB.toString())) {
+                    setServerError(data);
                 }
+            } else if (error != null || errorCode != null || errorDescription != null) {
+                setServerError(data);
             }
```
leleueri, line 99: question: is it something we want to do even if the session is null ? Here the condition to call the setServerError method is the session existance but what is the session is null ? :thinking:

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/utils/ThymeleafDataHelper.java
leleueri: discussed this point internally, seems ok

PR #5041, opened 2024-10-18
chore(mergify): declare 4.6.x in mergify config and remove 4.0.x

leleueri:


leleueri:
Mergify configuration, no need to validate all IT as java code doesn't change (this is a flaky).

PR #5042, opened 2024-10-18
fix: op claims in additional information

podlesrafal:
Fixes AM-4125


leleueri (changes requested):
please add a small test so if we introduce this filter in the future we will be warned

leleueri (approved):

mregulski-gravitee (approved):

PR #5072, opened 2024-10-24
feat:AM-3573 customizable max request size and operation count for scim bulk…

lgw-gravitee:
… operations

fixes AM-3574

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/resources/bulk/BulkEndpointConfiguration.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/resources/bulk/BulkEndpointConfiguration.java
@@ -1,0 +1,23 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.scim.resources.bulk;
+
+import lombok.Builder;
+
+@Builder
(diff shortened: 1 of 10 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/service/impl/BulkServiceImpl.java
```diff
@@ -68,56 +70,14 @@ public class BulkServiceImpl implements BulkService {
...
-                                        io.gravitee.am.gateway.handler.scim.model.Error error = new Error();
-                                        error.setScimType("invalidSyntax"); // should not happen
-                                        validOperation.setResponse(error);
-                                        return Single.just(validOperation.asResponse());
-
-                                }
-                            })
-                            .onErrorResumeNext(ex-> {
+                .concatMapMaybe(operation -> Single.just(operation)
+                            .filter(x -> failOnErrors == null || failOnErrors <= 0 || failuresCounter.get() < failOnErrors)
```
mregulski-gravitee, line 77: should we allow negative values for `failOnErrors`?
I think it would be better to change the `BulkRequest#getFailOnErrors()` to always return a non-negative int (for null, we could default to `Integer.MAX_VALUE`)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/resources/bulk/BulkEndpointConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.scim.resources.bulk;
+
+import lombok.Builder;
+import lombok.Value;
+
+@Builder
+@Value
+public class BulkEndpointConfiguration {
```
mregulski-gravitee, line 23: why not make this a `@Builder public record`?

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -301,6 +301,10 @@ handlers:
   oauth2:
     response:
       strict: false
+  scim:
```
mregulski-gravitee, line 304: is this exposed in the helm chart as well?

lgw-gravitee: The **handlers** section is fully configurable in Helm's values YAML. Therefore, a **scim** section can be added there, resulting in its presence in the gravitee.yaml file.

mregulski-gravitee (approved):

PR #5085, opened 2024-11-04
fix: update bulk user without id

podlesrafal:
Fixes AM-4277


leleueri (changes requested):
Can you please add a test ?

PR #5086, opened 2024-11-04
chore(test): added max size bulk test

podlesrafal:
AM-4221


leleueri (changes requested):
**TODO**: I don't see test cases in the mAPI spec file, can you please add them ?

leleueri (approved):

PR #5088, opened 2024-11-05
fix: prevent remove default reporter

podlesrafal:
fixes AM-3333

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
@@ -36,6 +36,7 @@ import io.gravitee.am.service.EventService;
 import io.gravitee.am.service.ReporterService;
 import io.gravitee.am.service.exception.AbstractManagementException;
 import io.gravitee.am.service.exception.ReporterConfigurationException;
+import io.gravitee.am.service.exception.ReporterDeleteException;
 import io.gravitee.am.service.exception.ReporterNotFoundException;
 import io.gravitee.am.service.exception.TechnicalManagementException;
 import io.gravitee.am.service.model.NewReporter;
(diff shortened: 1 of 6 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ReporterServiceImpl.java
```diff
@@ -238,11 +238,15 @@ public Single<Reporter> update(Reference reference, String reporterId, UpdateRep
...
 
     @Override
-    public Completable delete(String reporterId, User principal) {
+    public Completable delete(String reporterId, User principal, boolean internal) {
         LOGGER.debug("Delete reporter {}", reporterId);
         return reporterRepository.findById(reporterId)
                 .switchIfEmpty(Maybe.error(new ReporterNotFoundException(reporterId)))
                 .flatMapCompletable(reporter -> {
+                    // cannot remove system reporter
+                    if (reporter.isSystem() && !internal) {
```
leleueri, line 247: **question**: why introduce this internal param as we have the "isSystem" attribute linked to the reporter ?

podlesrafal: The `internal` parameter is set to true when we remove domain. Otherwise there is an error on removing domain because it cannot use default reporter

podlesrafal: internal stands for "call from internal method, not from api"

leleueri: ok, in that case instead of adding the internal param on the `delete(repoter, authUser)` method, maybe we can keep the interface signature unchanged and the service implementation could call a private method with this additional parameter (with JavaDoc explaining the purpose of this param)) so the method for public delete call the private method with false and the "internal" delete method call the private method with "true". Even if you don't want to implement my proposal, a javadoc should be provided to explain the purpose of this param

podlesrafal: I reorganised this a bit and added javadoc

PR #5096, opened 2024-11-06
fix: payload errors for mapi bulk

podlesrafal:
Fixes: AM-4304.
Feature will be testes using Jest test: #5086


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/users/OrganizationUsersResource.java
```diff
@@ -164,10 +171,19 @@ public void create(
                     schema = @Schema(implementation = BulkResponse.class)))
     @ApiResponse(responseCode = "500", description = "Internal server error")
     public void handleBulkOperation(
+            @HeaderParam(HttpHeaders.CONTENT_LENGTH) Integer contentLength,
             @PathParam("organizationId") String organizationId,
```
lgw-gravitee, line 175: Im wondering, its just a header right? can be setup to any value?

mregulski-gravitee: There's no guarantee the client sets Content-Length to a legitimate value, the validation needs to look at the actual payload.

Doing this validation in this handler doesn't really make much sense - `handleBulkOperation` is called after the request body is fully read, parsed as json and put in memory. So what does a limit on **payload size** protect against?
The limit on number of items in a single request still makes sense, as it protects against generating a massive load against the database.

podlesrafal: I had the same feeling. Checking it when the request was already processed and object created in this place doesn't make sens. I couldn't possibility to limit max size of request only for one endpoint. We can set it globally but im not sure about this solution.

podlesrafal: Changed the checking payload size to filter. @lgw-gravitee @mregulski-gravitee you can review

mregulski-gravitee (approved):

lgw-gravitee (approved):

PR #5098, opened 2024-11-06
fix: add DOMAIN_SETTINGS:DELETE permission to DOMAIN_PRIMARY_OWNER

mregulski-gravitee:

AM-2749


mregulski-gravitee:
@leleueri should we also add `CREATE` and `LIST`? Currently we only ever check READ or UPDATE.

leleueri (approved):

leleueri:
> @leleueri should we also add `CREATE` and `LIST`? Currently we only ever check READ or UPDATE.

Regarding DOMAIN_SETTINGS persmission, I don't think it is useful .

PR #5099, opened 2024-11-07
AM-2749: missing domain primary owner permission

mregulski-gravitee:

AM-2749

updates the upgraded so that #5098 gets applied on existing installations


leleueri (approved):

leleueri (commented):
Suggestion: to ease the support, do you think the VERSION should be set to the AM version (4.6.0 here) ?

leleueri:
> Suggestion: to ease the support, do you think the VERSION should be set to the AM version (4.6.0 here) ?

Ignore this comment (I can't remove it).  With a step back, it is simpler to have an increment.

PR #5109, opened 2024-11-13
fix(helm): remove probes from upgrader manifest

leleueri:
as upgraders are executed before listening the HTTP port

 and we are not able to evaluate the duration of an upgrade

 as consequence we decided to remove probes for upgraders

 we also cleanup the cockpit/cloud settings as it is useless to connect to cockpit during upgrade

Default value for the upgrader fwk has been removed to always execute the upgraders during mAPI startup (legacy behaviour)

fixes AM-4313

### DIFF
--- a/helm/templates/api/api-deployment.yaml
+++ b/helm/templates/api/api-deployment.yaml
@@ -115,8 +115,10 @@ spec:
               containerPort: {{ .Values.api.http.services.core.http.port | default "18093" }}
             {{- end }}
           env:
+            {{- if .Values.api.upgrader }}
             - name: upgrade.mode
               value: "false"
+            {{- end }}
             {{- if $plugins }}
             - name: GRAVITEE_PLUGINS_PATH_0
               value: '${gravitee.home}/plugins'
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -244,7 +244,11 @@ applications:
 
 api:
   enabled: true
-  upgrader: false
+  ## upgrader can be enabled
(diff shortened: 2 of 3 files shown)


lgw-gravitee (approved):

helm/templates/api/api-deployment.yaml
```diff
@@ -115,8 +115,10 @@ spec:
               containerPort: {{ .Values.api.http.services.core.http.port | default "18093" }}
             {{- end }}
           env:
-            - name: upgrade.mode
+            {{- if .Values.api.upgrader }}
+            - name: "upgrade.mode"
               value: "false"
+            {{- end }}
```
passionne, line 121:
```suggestion
            {{- if .Values.api.upgrader }}
            - name: "upgrade.mode"
              value: {{ .Values.api.upgrader }}
            {{- end }}
```
This means that the upgrader setup can not be true :thinking:

helm/values.yaml
```diff
@@ -244,7 +244,6 @@ applications:
 
 api:
   enabled: true
-  upgrader: false
```
passionne, line 247: Maybe keep it as a comment to keep default behaviour, but still communicate about this option.

helm/templates/api/api-deployment.yaml
leleueri: on the mAPI container, this value should be either false (in case of the job is enabled) or missing to always execute it.

passionne (approved):

PR #5126, opened 2024-11-15
fix: create index on credentials using createdAt and username

leleueri:
fixes AM-4339


lgw-gravitee (commented):
would index rt1ri1un1 be removed automacilly?

lgw-gravitee (approved):

PR #5153, opened 2024-11-20
AM-4336 certs validation

lgw-gravitee:


gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-certificate/src/main/java/io/gravitee/am/plugins/certificate/plugin/CertificatePluginHandler.java
```diff
@@ -31,11 +35,31 @@ public class CertificatePluginHandler extends AmPluginHandler<Certificate<?, ?>>
...
     public boolean canHandle(Plugin plugin) {
         return type().equalsIgnoreCase(plugin.type());
     }
 
+    @Override
+    protected void handle(Plugin plugin, Class<?> pluginClass) {
+        super.handle(plugin, pluginClass);
+        if (pluginManager.findById(plugin.id()) != null) {
+            registerValidator(plugin);
+        }
```
leleueri, line 51: **Thought**: I would move it to the AmPluginHandler (with the registerValidator) so that all plugin can take benefit of this validator. (obviously we will still have to adapt the other plugin beans)

leleueri (changes requested):
Unit tests are failing

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-certificate/src/main/java/io/gravitee/am/plugins/certificate/plugin/CertificatePluginHandler.java
lgw-gravitee: I would handle it in https://gravitee.atlassian.net/browse/AM-3373. For now, its strict "Certificate" feature.
WDYT?

leleueri (approved):

PR #5154, opened 2024-11-20
fix: evaluate session state before FORM rendering and submission

leleueri:
a new attribute has been added to the session to determine if the session is ongoing or not

 if it is actions related to a form are executed as usual

 if it is not a redirect to authorization endpoint is perform to follow the auth flow

 this is to avoid issues when a user is fully authenticated but use the back arraow on the browser

 this solution is not perfect as a back click before the end of the auth flow

 will not skip the page rendering

fixes AM-4342

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/BypassDirectRequestHandler.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/BypassDirectRequestHandler.java
@@ -1,0 +1,71 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.gravitee.am.gateway.handler.root.resources.handler;
+
+
+import io.gravitee.am.common.utils.ConstantKeys;
+import io.vertx.core.Handler;
+import io.vertx.rxjava3.ext.web.RoutingContext;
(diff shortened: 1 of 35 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/BypassDirectRequestHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    public BypassDirectRequestHandler(boolean skip) {
+        this.skipHandlerExecution = skip;
+    }
+
+    @Override
+    public void handle(RoutingContext routingContext) {
+        if (!skipHandlerExecution && (routingContext.session() == null ||
+                !hasText(routingContext.session().get(ConstantKeys.SESSION_KEY_AUTH_FLOW_STATE)) ||
+                !ConstantKeys.SESSION_KEY_AUTH_FLOW_STATE_ONGOING.equalsIgnoreCase(routingContext.session().get(ConstantKeys.SESSION_KEY_AUTH_FLOW_STATE)))) {
+            // SESSION_KEY_AUTH_FLOW_STATE is used by the authorization endpoint to flag an ongoing authentication
```
lgw-gravitee, line 52: It's a little bit blurry. How about extracting part of it into something like `checkAuthFlowState(ctx)`?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/scope/impl/ScopeManagerImpl.java
```diff
@@ -75,7 +79,7 @@ public void afterPropertiesSet() {
 
         logger.info("Register event listener for scopes events for domain {}", domain.getName());
         eventManager.subscribeForEvents(this, ScopeEvent.class, domain.getId());
-
+        this.alwaysProvideEnhancedScopes = OIDC_ALWAYS_ENHANCE_SCOPE.from(environment);
     }
```
lgw-gravitee, line 83: **question** What is the reason behind replacing @Value with Environment? Here, I believe we're adding another dependency?

leleueri: My point was to centralized the "legacy" setings to easily found them. Environment and @Value are both Spring element so not really new dep (we are still dependant to spring)

lgw-gravitee (approved):

PR #5179, opened 2024-11-26
Feat am 4384 new dockerfiles

leleueri:
fixes AM-3956

the open-jdk need to be replaced by non-deprecated base image


leleueri:
fixed by https://github.com/gravitee-io/gravitee-access-management/pull/5180

PR #5182, opened 2024-12-02
feat: set latPasswordResetForNewUser

podlesrafal:
AM-4352

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
@@ -26,12 +26,15 @@ import io.gravitee.am.gateway.handler.scim.model.Meta;
 import io.gravitee.am.gateway.handler.scim.model.Name;
 import io.gravitee.am.gateway.handler.scim.model.User;
 import io.gravitee.am.identityprovider.api.DefaultUser;
+import io.gravitee.am.service.exception.UserInvalidException;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.Instant;
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -48,6 +51,8 @@ import static java.util.stream.Collectors.toList;
  */
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class UserMapper {
+
+    public static final String LAST_PASSWORD_RESET_KEY = "lastPasswordReset";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(UserMapper.class);
 
@@ -193,6 +198,20 @@ public class UserMapper {
 
         // set additional information
         if (scimUser instanceof GraviteeUser graviteeUser && graviteeUser.getAdditionalInformation() != null) {
+            var lastPasswordReset = graviteeUser.getAdditionalInformation().get(LAST_PASSWORD_RESET_KEY);
+            if (lastPasswordReset != null) {
+                if (lastPasswordReset instanceof String) {
+                    try {
+                        user.setLastPasswordReset(Date.from(Instant.parse((String) lastPasswordReset)));
+                    } catch (Exception e) {
+                        LOGGER.error("Cannot parse lastPasswordReset. Be sure it is in ISO 8601 format.", e);
+                        throw new UserInvalidException("Unable to parse lastPasswordReset date. Be sure it is in ISO 8601 format.", e);
+                    }
+                } else {
+                    LOGGER.error("lastPasswordReset must be in ISO 8601 format");
+                    throw new UserInvalidException("Unable to parse lastPasswordReset. lastPasswordReset must be in ISO 8601 format.");
+                }
+            }
             additionalInformation.putAll(graviteeUser.getAdditionalInformation());
         }
         user.setAdditionalInformation(additionalInformation);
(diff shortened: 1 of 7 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/model/User.java
```diff
@@ -286,179 +290,9 @@ public class User extends Resource {
...
-        this.x509Certificates = x509Certificates;
-    }
-
-    public String getSource() {
-        return source;
-    }
+    /**
+     * Extended custom AM source field. Represents last password update.
+     */
+    private Date lastPasswordReset;
```
leleueri, line 296: **question**: as it is not a standard SCIM field, can you please ask to product team, if it is OK to have it in this "default" SCIM user model or do we have instead to move it in the GraviteeUser model ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
```diff
@@ -193,6 +197,10 @@ public static io.gravitee.am.model.User convert(User scimUser) {
 
         // set additional information
         if (scimUser instanceof GraviteeUser graviteeUser && graviteeUser.getAdditionalInformation() != null) {
+            var lastPasswordReset = graviteeUser.getAdditionalInformation().get(LAST_PASSWORD_RESET_KEY);
+            if (lastPasswordReset instanceof Long) {
+                user.setLastPasswordReset(Date.from(Instant.ofEpochMilli((Long) lastPasswordReset)));
```
leleueri, line 202: **question:** should we manage Long or should we manage String format as it is the case the Metadata https://datatracker.ietf.org/doc/html/rfc7643. I'm ok with the long as it is the format used in the MAPI but I raise this point as the dates defiend for the metadata are using String ...

podlesrafal: @tcompiegne I think this question is for you :)

leleueri: IMO we should be align of the same format across the SCIM API (so we should go with the String)... 
Long is more handy than string but having two date formats inside the same API is confusing IMO.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/main/java/io/gravitee/am/gateway/handler/scim/mapper/UserMapper.java
```diff
@@ -193,6 +198,18 @@ public static io.gravitee.am.model.User convert(User scimUser) {
...
+                try {
+                    user.setLastPasswordReset(Date.from(Instant.parse((String) lastPasswordReset)));
+                } catch (Exception e) {
+                    LOGGER.error("Cannot parse lastPasswordReset. Be sure it is in ISO 8601 format.", e);
+                    throw new UserInvalidException("Unable to parse lastPasswordReset date. Be sure it is in ISO 8601 format.", e);
+                }
+            } else {
+                LOGGER.error("lastPasswordReset must be in ISO 8601 format");
+                throw new UserInvalidException("Unable to parse lastPasswordReset. lastPasswordReset must be in ISO 8601 format.");
+            }
```
leleueri, line 212: **issue**: lastResetPassword is optional, I'm afraid that if the value is null you will throw an exception.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/UserServiceTest.java
```diff
@@ -730,4 +730,37 @@ public void shouldDeleteUser_noExternalProvider() {
...
+    public void shouldCreateUser_with_lastPasswordReset() {
+
+        Date lastPasswordResetDate = new Date();
+
+        GraviteeUser newUser = mock(GraviteeUser.class);
+        when(newUser.getSource()).thenReturn("unknown-idp");
+        when(newUser.getUserName()).thenReturn("username");
+        when(newUser.getPassword()).thenReturn(null);
+        Map<String, Object> ai = new HashMap<>();
+        ai.put("lastPasswordReset", lastPasswordResetDate);
```
leleueri, line 744: question: Should we use Instant instead of Date and use the Instant.toString method to have the right format (don't know if the toString is using the format used in the SCIM spec) ?

leleueri (changes requested):

leleueri (approved):

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/UserServiceTest.java
```diff
@@ -844,4 +848,87 @@ public void shouldThrowAnExceptionAndReportAuditIfPasswordIsInvalidOnUpdateUser(
...
+        when(userRepository.create(newUserDefinition.capture())).thenReturn(Single.just(user));
+
+        TestObserver<User> testObserver = userService.create(newUser, null, "/", null, new Client()).test();
+        testObserver.assertNoErrors();
+        testObserver.assertComplete();
+
+        testObserver.assertValue(u -> ((GraviteeUser) u).getAdditionalInformation().get("lastPasswordReset") != null);
+        assertFalse(newUserDefinition.getValue().isInternal());
+        assertTrue(newUserDefinition.getValue().isEnabled());
+    }
```
leleueri, line 883: duplicate with the on just after (same name)? but this one has invalid date format

PR #5185, opened 2024-12-03
AM-3901: make response_mode configurable for OIDC provider

mregulski-gravitee:

AM-3901

This PR allows configuring the `response_mode` (`query` or `fragment`) to be used by the OIDC provider
It also fixes a missing CSP nonce in the login callback page.


gravitee-am-common/src/main/java/io/gravitee/am/common/web/URLEncodedUtils.java
leleueri, line 1: **issue**: moving this class will break External IDP modules (for example [franceconnect](https://github.com/gravitee-io/gravitee-am-identityprovider-franceconnect/blob/main/src/main/java/com/graviteesource/am/identityprovider/franceconnect/authentication/FranceConnectAuthenticationProvider.java#L221))

gravitee-am-identityprovider/gravitee-am-identityprovider-oauth2-generic/src/main/resources/schemas/schema-form.json
```diff
@@ -69,6 +69,17 @@
       ],
       "default": "code"
     },
+    "responseMode": {
+      "type": "string",
+      "title": "OpenID response mode",
+      "description": "Specifies how the OIDC provider should return the authorization params to AM.</br>Leave default to pick automatically based on the response_type",
+      "enum": [
+        "default",
```
leleueri, line 77: question: why do we need "default" as possible value ?

gravitee-am-common/src/main/java/io/gravitee/am/common/web/NameValuePair.java
leleueri, line 1: **issue**: moving this class will break External IDP modules (for example franceconnect)

leleueri (changes requested):

gravitee-am-identityprovider/gravitee-am-identityprovider-facebook/src/main/java/io/gravitee/am/identityprovider/facebook/FacebookIdentityProviderConfiguration.java
```diff
@@ -78,8 +79,9 @@ public String getCodeParameter() {
         return CODE_PARAMETER;
     }
 
-    public String getResponseType() {
-        return RESPONSE_TYPE;
```
leleueri, line 82: **todo**: RESPONSE_TYPE declaration becomes useless at first look

gravitee-am-identityprovider/gravitee-am-identityprovider-github/src/main/java/io/gravitee/am/identityprovider/github/GithubIdentityProviderConfiguration.java
```diff
@@ -78,8 +79,12 @@ public String getCodeParameter() {
         return codeParameter;
     }
 
+    public ProviderResponseType getProviderResponseType() {
+        return ProviderResponseType.CODE;
+    }
+
     public String getResponseType() {
-        return responseType;
```
leleueri, line 82: **todo**: responseType declaration becomes useless at first look

gravitee-am-identityprovider/gravitee-am-identityprovider-facebook/src/main/java/io/gravitee/am/identityprovider/facebook/FacebookIdentityProviderConfiguration.java
```diff
@@ -15,21 +15,23 @@
...
 
-    private static final String USER_AUTHORIZATION_URI = "https://www.facebook.com/v8.0/dialog/oauth";
-    private static final String ACCESS_TOKEN_URI = "https://graph.facebook.com/v8.0/oauth/access_token";
-    private static final String USER_PROFILE_URI = "https://graph.facebook.com/v8.0/me";
-    private static final String CODE_PARAMETER = "code";
-    private static final String RESPONSE_TYPE = "code";
+    private String userAuthorizationUri = "https://www.facebook.com/v8.0/dialog/oauth";
+    private String accessTokenUri = "https://graph.facebook.com/v8.0/oauth/access_token";
+    private String userProfileUri = "https://graph.facebook.com/v8.0/me";
+    private String codeParameter = "code";
```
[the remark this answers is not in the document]
mregulski-gravitee: i've made them non-final to allow creating a test with wiremock in the future, like we have for the github provider (same for linkedin provider)

gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/authentication/AbstractOpenIDConnectAuthenticationProvider.java
```diff
@@ -179,67 +181,66 @@ private UriBuilder prepareSignInUrl(String redirectUri) {
             builder.addParameter(io.gravitee.am.common.oidc.Parameters.NONCE, SecureRandomString.generate());
         }
 
+        // using response_mode parameter with the default value for selected resposne_type is not recommended
```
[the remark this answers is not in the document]
mregulski-gravitee: * regarding the suggestion: it should have been "==", not "!=" (the idea is that we include the `response_mode` in the sign in url only if it's different than the default for the selected `response_type`)
* as for the question: I feel like 99.9% of cases, users should be fine with `default`. If not, the most likely case is that they're working around some bug or "uniqueness" of the upstream provider, because the AM is what handles he response, and we can handle both types just fine. This leads me to think it's better to leave a possiblity of "unsafe" config to get broader compatiblity. It might be worth displaying some kind of warning in the UI though.

leleueri (changes requested):
I made some test and the following case is not working.

locally, I create an OIDC provider with auth flow set to code and response mode set to fragment (this provider target an app on AM master env). The application on master env accept the fragment response mode so everything in term of config is OK (at first look)
but when I initiate the login flow on my local domain, when I clisk on the OIDC button, I'm redirected to the login page with "unsupported_response_mode" probably because the response mode is in UpperCase 


`https://am.master/domain-eric-v2/oauth/authorize?client_id=test-am-3901&response_type=code&response_mode=FRAGMENT&redirect_uri=http://localhost:8092/domain/login/callback&state=eyJr...`

gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
```diff
@@ -75,6 +75,7 @@ public interface ConstantKeys {
     String _CSRF = "_csrf";
     String SOCIAL_PROVIDER_CONTEXT_KEY = "socialProviders";
     String INTERNAL_PROVIDER_CONTEXT_KEY = "identityProviders";
+    String URL_HASH_PARAMETER = "urlhash";
```
leleueri, line 78:
```suggestion
    String URL_HASH_PARAMETER = "urlHash";
```
The constant you moved here had a capital H

leleueri: There is probably something wrong with the JEST test as this error prevent me to authenticate. After the authentication on the OIDC IDP domain, the callback fails because the urlHash param can't be extracted (due to the syntax error).

gravitee-am-test/specs/gateway/oidc-idp/common.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+    return await performGet(oidcProviderLoginUrl).expect(302);
+  };
+
+  const submitFragmentForm = async (response: BasicResponse) => {
+    const dom = cheerio.load(response.text);
+    const action = dom('#callbackForm').attr('action');
+    const url = action.startsWith('http:') ? action : `${process.env.AM_GATEWAY_URL}/${clientDomain.hrid}/login/` + action;
+    const headers = response.header['set-cookie'] ? { Cookie: response.header['set-cookie'] } : {};
+    return performFormPost(url, '', { urlhash: new URL(response.request.url).hash }, headers);
```
leleueri, line 153:
```suggestion
    return performFormPost(url, '', { urlHash: new URL(response.request.url).hash }, headers);
```

leleueri (changes requested):

gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
mregulski-gravitee: That's inconsistent with the default template that submits this parameter (`login_callback.html`)... Has the 'token' / 'id token' flow ever actually worked?

leleueri: Yes it works because the form post the parameters using the fields name which is initialiezd with the capital H.

mregulski-gravitee:
> Change has been made on urlHash into the java code but not in the JEST test. Surprisingly, the jests tests pass 🤔 it looks suspicious isn't it ?

I've looked into it, `HttpServerRequest#getParam()` appears to be case-insensitive:
image

leleueri (approved):

PR #5193, opened 2024-12-05
fix: correct color of badge status

podlesrafal:
change needed due to change type of Status.java from interface to enum.
Fixes AM-4435


leleueri (changes requested):
Please adapt also the User History page

leleueri (approved):

PR #5194, opened 2024-12-05
fix: remove tar file in the stages image to save space and remove the c…

leleueri:
…hgrp as the copy is doing ti

fixes AM-4370


lgw-gravitee (approved):

leleueri:
CI has been skipped to save credits as those files are not evaluated

PR #5235, opened 2024-12-10
chore: update ldap idp version

podlesrafal:
Fixes: AM-4510


leleueri (commented):
need to be applied on 4.1.x (label added)

PR #5267, opened 2024-12-17
AM-4048

mregulski-gravitee:
AM-4048

return a 404 instead of a 500 when attempting to revoke a non-existing account access token


leleueri (changes requested):
Looks like this PR is a mix of AM-3901and Am-4048

PR #5301, opened 2025-01-02
feat: Allow IPv4 or IPv6 configuration

francisko0:

https://github.com/gravitee-io/issues/issues/10247

## :pencil2: 

The option -Djava.net.preferIPv4Stack=true , block a deployment on an IPv6 cluster.
We can always add this option in the configuration JAVA_OPTS


leleueri:
Thank for the contribution and sorry for the delay to review and merge the PR.

PR #5302, opened 2025-01-02
AM-4591  introducing dataplane modules

lgw-gravitee:
fixes AM-4591

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/pom.xml
+++ b/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/pom.xml
@@ -432,6 +432,13 @@
             <type>zip</type>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>io.gravitee.am.dataplane</groupId>
+            <artifactId>gravitee-am-dataplane-mongodb</artifactId>
+            <version>${project.version}</version>
+            <type>zip</type>
+            <scope>runtime</scope>
+        </dependency>
     </dependencies>
 
     <build>
@@ -451,6 +458,12 @@
                         <configuration>
                             <outputDirectory>${basedir}/target/staging</outputDirectory>
                             <artifactItems>
+                                <artifactItem>
+                                    <groupId>io.gravitee.am.dataplane</groupId>
+                                    <artifactId>gravitee-am-dataplane-mongodb</artifactId>
+                                    <version>${project.version}</version>
+                                    <type>zip</type>
+                                </artifactItem>
                                 <!-- Identity providers -->
                                 <artifactItem>
                                     <groupId>io.gravitee.am.identityprovider</groupId>
--- a/gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlanePluginManager.java
+++ b/gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlanePluginManager.java
@@ -1,0 +1,72 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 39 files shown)


gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlanePluginManager.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private final PluginContextFactory pluginContextFactory;
+
+    public Optional<DataPlaneProvider> create(DataPlaneDescription dataPlanDescription) {
+        log.debug("Looking for a data plan provider for [{}]", dataPlanDescription.type());
+        var dataPlan = ofNullable(get(PLUGIN_ID_PREFIX + dataPlanDescription.type())).orElseGet(() -> {
+            throw new IllegalStateException("No data plan provider is registered for type " + dataPlanDescription.type());
+        });
+        if (dataPlan.provider() == null) {
+            return null;
```
leleueri, line 49: **thought**: IMO we should return Optional.empty here isn't it ?

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/SingleDataPlaneLoader.java
```diff
@@ -0,0 +1,34 @@
...
+    private static final String DATA_PLANE_KEY = "dataPlane";
+    private static final String DATA_PLANE_REPOSITORY_BASE_KEY = "repositories.gateway";
+    private static final String DATA_PLANE_TYPE_KEY = DATA_PLANE_REPOSITORY_BASE_KEY + ".type";
+
+    @Override
+    protected void register() {
+        var dataPlaneId = configuration.getProperty(DATA_PLANE_KEY, String.class);
+        String dataPlaneType = configuration.getProperty(DATA_PLANE_TYPE_KEY, String.class, "mongodb");
+        var description = new DataPlaneDescription(dataPlaneId, dataPlaneId, dataPlaneType, DATA_PLANE_REPOSITORY_BASE_KEY);
+        create(description);
```
leleueri, line 32: **discussion**: I was thinking we discuss the Gateway specificity. Initially I was thinking about using gateway scope as source of truth for the DataPlane plugin on Gateway side but during refinement we said it should be better to have the same "dataPlanes" map into the gravitee.yaml but with a single entry to be aligned with the mAPI. Don't said the implementation you proposed is not good, just want to highlight to point. Maybe something to discuss wit the team & archi team. If we merge with the implementation you propose, please open a JIRA issue to track the point (with target/fixed version 4.7.0 as we have to decide before the release)

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlanePluginManager.java
lgw-gravitee: ups. Thanks.

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/SingleDataPlaneLoader.java
lgw-gravitee: I changed to dataPlane.

leleueri (commented):
code looks OK but all the integration tests are failing;

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/pom.xml
```diff
@@ -451,6 +458,12 @@
                         <configuration>
                             <outputDirectory>${basedir}/target/staging</outputDirectory>
                             <artifactItems>
+                                <dependency>
```
podlesrafal, line 461: shouldnt be there artifactItem?

lgw-gravitee: nice one. it could be a cause of all my problems

leleueri (approved):

PR #5303, opened 2025-01-03
feat: assign dataPlaneId to Domain

podlesrafal:
AM-4165


gravitee-am-ui/src/app/domain/creation/domain-creation.component.html
```diff
@@ -27,21 +30,42 @@ <h1>New domain</h1>
...
+            @if (oneDataPlane) {
+              <mat-option [value]="dataPlanes[0].id"> {{ dataPlanes[0].name }}</mat-option>
+            } @else {
+              @for (dp of dataPlanes; track dp) {
+                <mat-option [value]="dp.id">{{ dp.name }}</mat-option>
+              }
+            }
+          </mat-select>
+          <mat-hint> Data plane to be used with this domain. </mat-hint>
+        </mat-form-field>
```
leleueri, line 60: **TODO**: in addition of been disabled, the field has to be hidden if there is only one DP.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/DataPlanesResource.java
```diff
@@ -0,0 +1,72 @@
...
+                    content = @Content(mediaType = "application/json",
+                            array = @ArraySchema(schema = @Schema(implementation = DataPlane.class)))),
+            @ApiResponse(responseCode = "500", description = "Internal server error")})
+    public void list(
+            @PathParam("organizationId") String organizationId,
+            @PathParam("environmentId") String environmentId,
+            @Suspended final AsyncResponse response) {
+
+        checkAnyPermission(organizationId, environmentId, Permission.DOMAIN, Acl.CREATE)
+                .andThen(Maybe.just(multiDataPlaneLoader.getDataPlanes().stream()))
```
lgw-gravitee, line 68: why stream?
also its never empty, so you may omit Maybe.just I guess.

PR #5311, opened 2025-01-08
fix: allow to create user with whitespace by default

podlesrafal:
fixes AM-4602

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -340,7 +340,7 @@ user:
         pattern: ^[^±!£$%^&*§¡€¢¶•ªº«»\\/<>?|=]{0,100}$
   username:
     policy:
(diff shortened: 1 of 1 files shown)


podlesrafal:
@leleueri We can had inconsistence in the default username validation. I just aligned to default one that is defined in https://github.com/gravitee-io/gravitee-access-management/blob/8e7ae0dc88a96b9f15898fb263c8d3f01b5b33b3/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/user/UserValidatorImpl.java#L46 and in gravitee.yml for gateway. If you think that we should not allow white space in username, I will change UserValidatorImpl as well.

leleueri:
> @leleueri We can had inconsistence in the default username validation. I just aligned to default one that is defined in
> 
> https://github.com/gravitee-io/gravitee-access-management/blob/8e7ae0dc88a96b9f15898fb263c8d3f01b5b33b3/gravitee-am-service/src/main/java/io/gravitee/am/service/validators/user/UserValidatorImpl.java#L46
> 
> and in gravitee.yml for gateway. If you think that we should not allow white space in username, I will change UserValidatorImpl as well.

:thinking: 
I didn't checked the GW regex but indeed the space is missing on GW side... so better to align the mAPI config.

leleueri (approved):

PR #5312, opened 2025-01-08
fix: correct display drop down list in table

podlesrafal:
fixes AM-4600


leleueri (approved):
This should also be applied on 4.3.x

PR #5347, opened 2025-01-10
fix: tokens should be removed when user is disabled

lgw-gravitee:
fixes AM-4062


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/AbstractUserService.java
```diff
@@ -156,12 +160,13 @@ private Single<UpdateUser> updateWithUserProvider(UpdateUser updateUser, User
     }
 
     @Override
-    public Single<User> updateStatus(ReferenceType referenceType, String referenceId, String id, boolean status, io.
+    public Single<User> updateStatus(ReferenceType referenceType, String referenceId, String userId, boolean status, io.
             gravitee.am.identityprovider.api.User principal) {
-        return getUserService().findById(referenceType, referenceId, id)
+        return getUserService().findById(referenceType, referenceId, userId)
                 .flatMap(user -> {
+                    Completable removeTokens = status ? Completable.complete() : tokenService.deleteByUserId(userId);
```
leleueri, line 167: **TODO**: in addition of the status, could be nice to check also the referenceType as the user maybe an Organization user. So no need to revoke the tokens if refType == ORGANIZATION. Also may I ask you to add few unit test to validate the behaviour ?

leleueri (changes requested):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/AbstractUserService.java
lgw-gravitee: I put this logic into to separate classes: OrganizationUserService and UserService

lgw-gravitee: In regards tests - I will add some jests.

leleueri (approved):

PR #5349, opened 2025-01-13
fix: ciba device notifier header ui and EL support

podlesrafal:
fixes: AM-4596


gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-api/pom.xml
```diff
@@ -43,7 +43,13 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>io.gravitee.am.repository</groupId>
+            <artifactId>gravitee-am-repository-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
+        </dependency>
```
leleueri, line 51: **question**: why adding this dependency ? there is no import related to repository package :thinking:

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-http/src/main/java/io/gravitee/am/authdevice/notifier/http/provider/HttpAuthenticationDeviceNotifierProvider.java
```diff
@@ -69,16 +74,30 @@ public class HttpAuthenticationDeviceNotifierProvider implements AuthenticationD
 
     @Override
     public Single<ADNotificationResponse> notify(ADNotificationRequest request) {
+
+        AuthenticationContext authenticationContext = new SimpleAuthenticationContext(request.getContext().get(CIBA_AUTH_REQUEST_KEY));
+        //removing context from request to not duplicate.
+        request.setContext(null);
+        authenticationContext.setAttribute("cibaRequest", request);
+        TemplateEngine templateEngine = authenticationContext.getTemplateEngine();
```
leleueri, line 82: **TODO**: no need to use SimpleAuthContxt. I was more thinking to provide a `getTemplateEngine` in the ADNotificationRequest so all teh future plugins will have access to the same Engine. 

Something like:

```
private TemplateEngine templateEngine;

public TemplateEngine getTemplateEngine() {
if (templateEngine == null) {
            templateEngine = TemplateEngine.templateEngine();
            TemplateContext templateContext = templateEngine.getTemplateContext();
            // maybe evaluableRequest (and some other classes can be move in gravitee-am-common or something like this
            templateContext.setVariable("request", new EvaluableRequest(new VertxHttpServerRequest(routingContext.getRequest())));
            templateContext.setVariable("context", new EvaluableExecutionContext(this.routingContext.getData()));
        }
        return templateEngine;
  }
```

In addition, looks like the ADNotificationResponse & ADNotificationRequest are dupliacted. One instance is defined into the package `io.gravitee.am.gateway.handler.ciba.ad.model`  the other into `io.gravitee.am.authdevice.notifier.api.model`. I don't remember if there is a reason, but looks like it is useless and only the definition of `io.gravitee.am.authdevice.notifier.api.model`should be used.

leleueri (changes requested):

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-api/pom.xml
leleueri: You resolve this comment but I do not see answer and I'm still thinking the dependency is not useful.

leleueri (approved):

PR #5350, opened 2025-01-13
chore(dataplane): use Credential repository coming from the DataPlane…

leleueri:
… plugin

Fixes AM-4610

related-to AM-4569


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/service/impl/CredentialGatewayServiceImpl.java
```diff
@@ -0,0 +1,258 @@
...
+ * @author GraviteeSource Team
+ */
+@Component
+@Slf4j
+public class CredentialGatewayServiceImpl implements CredentialGatewayService {
+
+    @Lazy
+    @Autowired
+    private DataPlaneRegistry dataPlaneRegistry;
+
```
lgw-gravitee, line 58: @wonders Im wondering, since  we have single dataplane for the Gateway, can't we inject here CredentialRepository?

lgw-gravitee (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/service/impl/CredentialGatewayServiceImpl.java
leleueri: good point, I was also thinking the GW side may be simplified. Will try to do something during the task about users.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/service/impl/CredentialGatewayServiceImpl.java
```diff
@@ -0,0 +1,258 @@
...
+                            log.error("An error occurs while trying to delete credential: {}", id, ex);
+                            return Completable.error(new TechnicalManagementException(
+                                    String.format("An error occurs while trying to delete credential: %s", id), ex));
+                        })
+                );
+    }
+
+    @Override
+    public Completable deleteByUserId(Domain domain, String userId) {
+        log.debug("Delete credentials by {} {} and user id: {}", domain.getId(), userId);
```
podlesrafal, line 231: you have 3 placeholders and only 2 parameters. same in log.error and for deleteByDomain

PR #5362, opened 2025-01-15
AM-4608 : Use DataPlane for GroupRepository

leleueri:
This PR make the disctinciton between Organization level and Domain level groups by creating 2 services.
As they are using the same data structure They are using the same collection named "groups" in database, this is fine to avoid data migration for legacy deployement.

This PR also include some refactoring


gravitee-am-service/src/main/java/io/gravitee/am/service/OrganizationGroupService.java
```diff
@@ -32,60 +32,29 @@
  * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
  * @author GraviteeSource Team
  */
-public interface GroupService {
+public interface OrganizationGroupService {
 
     Single<Page<Group>> findAll(ReferenceType referenceType, String referenceId, int page, int size);
```
lgw-gravitee, line 38: @question does it make sense to keep this method? I mean, it should be always Organization referenceType

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/MembershipServiceImpl.java
```diff
@@ -359,7 +359,7 @@ private Completable checkMember(String organizationId, Membership membership) {
             return orgUserService.findById(ReferenceType.ORGANIZATION, organizationId, membership.getMemberId())
                     .ignoreElement();
         } else {
-            return groupService.findById(ReferenceType.ORGANIZATION, organizationId, membership.getMemberId())
+            return orgGroupService.findById(ReferenceType.ORGANIZATION, organizationId, membership.getMemberId())
                     .ignoreElement();
```
lgw-gravitee, line 363: The example of previous comment

gravitee-am-service/src/main/java/io/gravitee/am/service/OrganizationGroupService.java
leleueri: true, I will change it

lgw-gravitee (approved):

PR #5466, opened 2025-02-11
chore(openapi): Update Self-Account open api spec

lgw-gravitee:


docs/self-service-account-api-descriptor.yml
leleueri, line 1: suggestion: maybe line 182 we can add into the response description that the Security attribute of the EnrolledFactor is not provided

leleueri (approved):

PR #5472, opened 2025-02-13
fix: use the tid as seed to generate the moving factor

leleueri:
this change allows to generate a new challenge even during the enrollment phase
    
      Since the code is renewed on each challenge call, the Jest test has to be rework for
    
      email factor otherwise the code read before the verification method is invalidated
    
      due to the second call on challenge endpoint made to get the XSRF token

fixes AM-4734


podlesrafal:
Could you add some tests?

podlesrafal (approved):

PR #5473, opened 2025-02-14
Am 4599 dataplaneid in events

podlesrafal:
feat: AM-4599


gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/spring/SpringEventRepository.java
```diff
@@ -30,4 +30,7 @@
 public interface SpringEventRepository extends RxJava3CrudRepository<JdbcEvent, String> {
     @Query("select * from events e where e.updated_at >= :from and e.updated_at <= :to")
     Flowable<JdbcEvent> findByTimeFrame(@Param("from") LocalDateTime from, @Param("to") LocalDateTime to);
+
+    @Query("select * from events e where e.updated_at >= :from and e.updated_at <= :to and e.data_plane_id = :dataPlaneId")
```
leleueri, line 34: issue: IMO we should adapt this query to search by dataplaneID or by null dataplaneID. Some event sent by the mAPI may not have those informations (org level sync event for example)

leleueri (changes requested):
TODO: All the events related to a domain action should contains the dataplaneId/envId. So there is some missing changes to be able to provide this infos.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EventServiceImpl.java
```diff
@@ -43,12 +47,37 @@ public class EventServiceImpl implements EventService {
...
 
+    @Lazy
+    @Autowired
+    private DomainRepository domainRepository;
+
     @Override
     public Single<Event> create(Event event) {
+        if (event.getPayload().getReferenceType().equals(ReferenceType.DOMAIN)) {
+            return domainRepository.findById(event.getPayload().getReferenceId())
+                    .flatMapSingle(domain -> {
```
leleueri, line 58: **suggestion**: I would deprecate this method as I'm not a big fan of doing a search here as most of the time the domain should be available (obviously, the methods signatures on which the eventService.create method is called will have to change to take the Domain object instead of the domainId). Instead I think a method like send(Event) and sendDomain(Event) can be implemented. 

* Send(Event) for the events where domain is not involved (very few)
* sendDomain(Event) where event is specific to a domain so this method before sending the event can check that the dataplaneId and envId are present (or the domain can be pass as param so this method set the values)

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcEventRepository.java
```diff
@@ -122,6 +136,7 @@ public Single<Event> update(Event item) {
         update = databaseDialectHelper.addJsonField(update, COL_PAYLOAD, item.getPayload());
         update = addQuotedField(update, COL_CREATED_AT, dateConverter.convertTo(item.getCreatedAt(), null), LocalDateTime.class);
         update = addQuotedField(update, COL_UPDATED_AT, dateConverter.convertTo(item.getUpdatedAt(), null), LocalDateTime.class);
+        update = addQuotedField(update, COL_DATA_PLANE_ID, item.getDataPlaneId(), String.class);
```
leleueri, line 139: **todo**: environmentId is missing

gravitee-am-repository/gravitee-am-repository-tests/src/test/java/io/gravitee/am/repository/management/api/EventRepositoryTest.java
```diff
@@ -74,6 +74,72 @@ public void testFindByTimeFrame() {
...
+        TestSubscriber<Event> testSubscriber = eventRepository.findByTimeFrameAndDataPlaneId(from, to, "default").test();
+        testSubscriber.awaitDone(10, TimeUnit.SECONDS);
+
+        testSubscriber.assertComplete();
+        testSubscriber.assertNoErrors();
+        testSubscriber.assertValueCount(2);
+    }
+
     @Test
     public void testFindById() {
```
leleueri, line 144: **nickpick**: I wish you update this test to test the fact that the dataPlaneId and the EnvId are persisted and read successfully

gravitee-am-service/src/main/java/io/gravitee/am/service/EmailTemplateService.java
```diff
@@ -48,35 +49,35 @@ public interface EmailTemplateService {
...
+    Single<Email> create(Domain domain, NewEmail newEmail, User principal);
 
-    Single<Email> create(String domain, NewEmail newEmail, User principal);
+//    Single<Email> create(String domain, NewEmail newEmail, User principal);
 
-    Single<Email> create(ReferenceType referenceType, String referenceId, String client, NewEmail newEmail, User principal);
+    Single<Email> create(Domain domain, String client, NewEmail newEmail, User principal);
 
-    Single<Email> create(String domain, String client, NewEmail newEmail, User principal);
+//    Single<Email> create(String domain, String client, NewEmail newEmail, User principal);
```
leleueri, line 60: **todo**: if not useful anymore, can we remove it?

leleueri (approved):

PR #5491, opened 2025-02-17
fix: changed dictionary provider for email service

podlesrafal:
About fix: sometimes loading dictionary is on different thread then getting it and it causes an issue when sending an email. Changed to DomainBasedDictionaryProvider.
fixes: AM-4764


leleueri (approved):
LGTM. Rember to apply the labels to cherry pick this fix on maint versions

PR #5498, opened 2025-02-18
fix: loop redirect on invalid form /resetPassword

podlesrafal:
Fixes AM-4758

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
@@ -600,8 +600,7 @@ public class RootProvider extends AbstractProtocolProvider {
 
         // Forgot password route
         final var forgotPasswordAccessHandler = new ForgotPasswordAccessHandler(domain);
-        final var resetPasswordFailureHandler = new ErrorHandler(PATH_RESET_PASSWORD);
-        rootRouter.route(GET, PATH_FORGOT_PASSWORD)
+       rootRouter.route(GET, PATH_FORGOT_PASSWORD)
                 .handler(clientRequestParseHandler)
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
```diff
@@ -899,5 +896,7 @@ private void errorHandler(Router router) {
         router.route(PATH_MFA_RECOVERY_CODE).failureHandler(errorHandler);
 
         router.route(PATH_CONFIRM_REGISTRATION).failureHandler(errorHandler);
+
+        router.route(PATH_RESET_PASSWORD).failureHandler(errorHandler);
```
leleueri, line 900: **issue**: I don't think the fix is valid here, we indeed remove the loop in case of invalid Tempalte but if the template is valid and if the resetPassword is failing, then the user will be redirect to the default error page instead of the reset password one.

podlesrafal: In ResetPasswordSubmissionEndpoint: https://github.com/gravitee-io/gravitee-access-management/blob/2b80a54f4e3bbd19aaa5399629630d71e9275669/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/user/password/ResetPasswordSubmissionEndpoint.java#L75 is handling exception from reset password. If there is exception then the redirection is to /resetPassword endpoint with error in params. The general errorHandler will handle all other errors like invalid html that are not related with resetPassword method from Service

leleueri: made a quick test. looks ok

leleueri (approved):

PR #5499, opened 2025-02-19
feat: moved purge services to gateway

lgw-gravitee:
fixes: AM-4567


gravitee-am-gateway/gravitee-am-gateway-services/gravitee-am-gateway-services-purge/src/main/java/io/gravitee/am/gateway/services/purge/ScheduledPurgeService.java
```diff
@@ -51,7 +51,7 @@ public class ScheduledPurgeService extends AbstractService implements Runnable {
     @Value("${services.purge.enabled:true}")
     private boolean enabled;
 
-    @Value("${management.type:mongodb}")
+    @Value("${repositories.management.type:mongodb}")
```
leleueri, line 54:
```suggestion
    @Value("${repositories.gateway.type:mongodb}")
```
Now the purge manager is on the GW side, better to use a repository type linked to the Gateway (gateway or oauth2) wdyt ?

leleueri: also maybe we should get the two types (for gateway and oauth2) and in the `doStart` method  test if one of them is `jdbc`

leleueri (changes requested):

leleueri (approved):

PR #5507, opened 2025-02-21
chore(dataplane): remove reference of Token and MFA RateLimit & Verif…

leleueri:
…yAttempt repositories from the management API

 now mAPI send an event to request token deletion or MFA limites cleanup

fixes AM-4571


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -534,11 +534,11 @@ liquibase:
...
-#mfa_rate:
-#  enabled: true
-#  limit: 5
-#  timePeriod: 15
-#  timeUnit: Minutes
+mfa_rate:
+  enabled: true
+  limit: 5
+  timePeriod: 1
+  timeUnit: Minutes
```
lgw-gravitee, line 541: shouldn't we leave it as //commented?

lgw-gravitee (approved):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
leleueri: yes good catch, I missed this one

PR #5550, opened 2025-02-25
chore: provide default Runner to start AM within IDEA usign PSQL or M…

leleueri:
…ongo

Those runners
* doesn't have license
* are targeting the target/distribution as gravitee.home so make plugin command becomes useless, we only need to do a maven install.


lgw-gravitee (approved):

leleueri:
don't wait the ci as it is only config for IDE

PR #5565, opened 2025-02-27
feat: introducing upgrader framework to gateway

lgw-gravitee:
fixes: AM-4577


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -495,7 +496,7 @@ ds:
 
 # For AM GATEWAY, liquibase updates are disabled by default.
 liquibase:
-  enabled: false
+  enabled: true
```
leleueri, line 499: **issue**: keep false here please. Should be true only for deployement where the GW/AOuth2 scopes are using a DB different from Managment scope

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_7_0/4.7.0-12-dataplane-upgrader-table.yml
leleueri, line 1: **thought**: Wondering if we should also have the same table creation in the liquibase scripts bind to repositories :thinking: As this table will used by the DP plugin  but also by the GW to process the upgraders, IMO both script should be able to create this table. Otherwise it may cause some issue when the HTTP connector will come in as in this case the DP will be executed on the GW side but after the repostiory startup and I'm afraid the repository loading will fail as the table will not exist... :thinking:

leleueri (changes requested):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
lgw-gravitee: I added it by mistake, thanks

gravitee-am-gateway/gravitee-am-gateway-service/src/main/java/io/gravitee/am/gateway/service/upgraders/DummyUpgrader.java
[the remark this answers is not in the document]
lgw-gravitee: nope, I can remove it

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_7_0/4.7.0-12-dataplane-upgrader-table.yml
lgw-gravitee: good point

PR #5580, opened 2025-03-06
AM-4842 fix: redirect_uri validation

lgw-gravitee:
fixes AM-4842


leleueri (approved):

leleueri (changes requested):
need to be done also on return_url validator

leleueri (approved):

PR #5600, opened 2025-03-11
chore(mergify): update config to manage 4.7.x and remove EOL branches

leleueri:


lgw-gravitee (approved):

leleueri:
skip CI to save credit as it is only merfigy config

PR #5605, opened 2025-03-12
fix: warn when domain not found on event sync

podlesrafal:
fixes: AM-4853

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EventServiceImpl.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EventServiceImpl.java
@@ -22,9 +22,7 @@ import io.gravitee.am.repository.management.api.DomainRepository;
 import io.gravitee.am.repository.management.api.EventRepository;
 import io.gravitee.am.service.EventService;
 import io.gravitee.am.service.exception.AbstractManagementException;
-import io.gravitee.am.service.exception.DomainNotFoundException;
 import io.gravitee.am.service.exception.TechnicalManagementException;
-import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EventServiceImpl.java
```diff
@@ -65,8 +63,10 @@ public Single<Event> create(Event event) {
                         event.setEnvironmentId(domain.getReferenceId());
                         return eventCreation(event);
                     })
-                    .switchIfEmpty(Maybe.error(new DomainNotFoundException(event.getPayload().getReferenceId())))
-                    .toSingle();
+                    .switchIfEmpty(Single.fromCallable(() -> {
+                        LOGGER.warn("Domain not found for referenceId: {}", event.getPayload().getReferenceId());
+                        return event;
+                    }).flatMap(this::eventCreation));
```
leleueri, line 69: **issue**: eventCreation will be called twice when the domain exists. please add verification on the Unit test to ensure we publish the event only once in both case

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EventServiceImpl.java
leleueri: ah ok, I missed parentheses.
It should work indeed but IMO better to have a single call to eventCreation.

Maybe something like this
```
domainRepository.findById(event.getPayload().getReferenceId())
                    .map(domain -> {
                        event.setDataPlaneId(domain.getDataPlaneId());
                        event.setEnvironmentId(domain.getReferenceId());
                        return event;
                    })
                    .switchIfEmpty(Single.fromCallable(() -> {
                        LOGGER.warn("Domain not found for referenceId: {}", event.getPayload().getReferenceId());
                        return event;
                    }).flatMap(this::eventCreation));
```

leleueri (approved):
I made a suggestion, I let you decide if it is a good one.

PR #5606, opened 2025-03-12
fix: unable to select groups in group mapper in organization

podlesrafal:
Fixes: AM-4841
This is fixing only the UI issue. Handling group mapper in organization will be handled in this task: AM-4923


leleueri (approved):
LGTM, should it beapply on 4.6 & 4.5 too ?

PR #5612, opened 2025-03-13
chore(dataplane): provide migration script for all backends

leleueri:
this PR update the OpenAPI specification and provide generic scripts to handle the data migration for each backend

### DIFF
--- a/docs/upgrades/4.7.0/postgres-migration.sh
+++ b/docs/upgrades/4.7.0/postgres-migration.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+#
+# Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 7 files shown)


docs/upgrades/4.7.0/postgres-migration.sh
```diff
@@ -0,0 +1,82 @@
...
+else
+  echo "Destination database already exists."
+fi
+
+# Restore the backup for each table to the destination database
+for TABLE in "${TABLES[@]}"; do
+  echo "Restoring table: $TABLE to the destination database..."
+  export PGPASSWORD=$DEST_DB_PASSWORD
+  pg_restore -h $DEST_DB_HOST -p $DEST_DB_PORT -U $DEST_DB_USER -d $DEST_DB_NAME -v /tmp/${TABLE}_backup.dump
+
```
lgw-gravitee, line 67: I put it here but it applies to all databases. 
Im worried about transcation bottleneck in larger scale
WDYT about adding --no-transcation?

lgw-gravitee (approved):
Approved and put a single comment.

PR #5620, opened 2025-03-17
AM-4563 feat: added dataplanes to helm scripts

lgw-gravitee:
solves AM-4563


helm/templates/gateway/_gateway_configmap-repositories.tpl
```diff
@@ -0,0 +1,34 @@
...
+  oauth2:
+{{- if or (eq .Values.oauth2.type "mongodb") (kindIs "invalid" .Values.oauth2.type) }}
+    type: mongodb
+{{- $defaultMongo | toYaml | nindent 4 }}
+{{- else if (eq .Values.oauth2.type "jdbc") }}
+    type: jdbc
+{{- $defaultJdbc | toYaml | nindent 4 }}
+{{- end }}
+  gateway:
+    dataPlaneId: {{ .Values.gateway.dataPlaneId }}
```
leleueri, line 25:
```suggestion
    dataPlane:
      Id: {{ .Values.gateway.dataPlaneId }}
```

Wondering if we have to introduce a variable gateway.dataPlaneId in the values.yaml :thinking: 
* If the admin is deploying using DP mode, the repository will use the new repositories section in the value and the admin will have to specify the dataPlane Id inside this section.
* If the admin is NOT deploying using DP mode, then that means we can define "default" as static value (or not define the field in this case, AM will use the default value)

helm/values.yaml
```diff
@@ -618,6 +618,7 @@ gateway:
   enabled: true
   name: gateway
   type: mongodb
+  dataPlaneId: default
```
leleueri, line 621: Wondering if we have to introduce a variable gateway.dataPlaneId in the values.yaml 🤔

leleueri (changes requested):

leleueri (commented):
**TODO**: In addition of the comment made yesterday, I something is missing. As for the MAPI, when the upgrade.mode is enabled we should also trigger a pre-upgrade jb for the GW so the upgrader will be executed first then the rolling update of the GW will be done.

helm/templates/gateway/_gateway_configmap-repositories.tpl
lgw-gravitee: agree

helm/values.yaml
```diff
@@ -616,6 +616,11 @@ api:
 
 gateway:
   enabled: true
+  ## upgrader can be enabled
+  ## this is missing by default to be sure that upgrades are managed on mAPI startup
+  ## * missing upgrader key or upgrader = false will not execute the upgrader job but will execute the upgrade on MAPI startup
+  ## * upgrader = true will execute the upgrader job and the upgrader will not be executed on MAPI startup
```
leleueri, line 622:
```suggestion
  ## upgrader can be enabled
  ## this is missing by default to be sure that upgrades are managed on gateway startup
  ## * missing upgrader key or upgrader = false will not execute the upgrader job but will execute the upgrade on gateway startup
  ## * upgrader = true will execute the upgrader job and the upgrader will not be executed on gateway startup
```

leleueri (approved):
Minor change (typo due to copy/past). No need to submit another review

PR #5634, opened 2025-03-18
fix: exception error on mongo index creation

podlesrafal:
The problem with Timeout Exception was because the subscriber didn't finish. I've changed to blockingSubscribe and now all operation of creation index is finished.

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoUtils.java
+++ b/gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoUtils.java
@@ -28,6 +28,7 @@ import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
 import io.reactivex.rxjava3.functions.Function;
 import io.reactivex.rxjava3.functions.Predicate;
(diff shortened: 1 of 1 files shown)


leleueri (commented):
question: do we have to do the same on Reporters ? also if I remember well we were usign blockingGet in the past... could be nice to wheck why we come to a subscribe to avoid regression if any.

podlesrafal:
OMG, history of this peace of code is so long. It was changed from subscribe to blocking operation many times :D 
This is the history:
- **BlockingGet -> subscribe:** 55d63e83 fix: do not perform blockingGet when we create indexes in repository layer
- **subscribe -> blockingSubscribe:** 37638f57 fix: TimeException on MongoDB due to too many index creations in parallel
- **blockingSubscribe -> blockingAwait:** 5b1bc9bc fix: avoid infinite blocking call durint Indexes creation
- **blockingAwait -> subscribe:** 18df9bcb fix: use subscribe instead of blocking await to allow the process to start on kubernetes (this commit doesn't fix the timeout issue)
NOW: 
- **subscribe -> blockingSubscribe with schedulers** 
To sum up, the subscribe generate error but it works. 
My latest option uses blockingSubscribe with schedulers, so it run the index creation on separate thread to not block main thread. I think this can resolve the issue. It works for me locally with fresh DB.
@leleueri ☝️

leleueri (approved):

PR #5637, opened 2025-03-19
fix: Login enumeration due to auth duration

podlesrafal:
Added fake password encoding to prelong user authentication when username not found.
Tests (user exist):
Screenshot 2025-03-19 at 12 16 12
Screenshot 2025-03-19 at 12 16 26
Fake password is "password" in bcrypt. Had to use bcrypt format because Bcrypt password encoder checks the format.


leleueri (commented):
Question: should it be done inside the IDP or outside when all IDP send a UsernameNotFoundException? if we are doing it in each IDP on GW side the login step may become very long according to the number of IDP configured.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/provider/security/ManagementAuthenticationProvider.java
```diff
@@ -148,10 +162,18 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+        if (userException == userNotFoundException) {
+            //Didn't find user in any IDP (Mongo or JDBC), so no password encoding has been proceeded.
+            doFakePasswordEncoding();
+        }
         throw new BadCredentialsException("No user found for providers " + StringUtils.collectionToDelimitedString(identities, ","));
     }
 
+    private void doFakePasswordEncoding(){
+        //PEN-21 Encoding password takes a while. To ensure execution time the same for not existing user, introduced fake password checking.
+        bCryptPasswordEncoder.matches(UUID.randomUUID().toString(), "$2a$10$hdjt9YGrSudbIljTqAtcW.KOxNJscq00Nxv088wPy6GDKXCJe0aCm");
```
leleueri, line 174: question: we discussed it in meeting so I understand why the UUID is used but should we align the behaviour with the GW which is evaluating the password provided by the user. Maybe it is more realistic to test the pwd provided by the user here too ? :thinking:

leleueri (commented):
CI is failing on unit test execution

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/provider/security/ManagementAuthenticationProvider.java
```diff
@@ -148,10 +162,18 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+        if (userException == userNotFoundException) {
+            //Didn't find user in any IDP (Mongo or JDBC), so no password encoding has been proceeded.
+            doFakePasswordEncoding(authentication.getCredentials().toString());
+        }
         throw new BadCredentialsException("No user found for providers " + StringUtils.collectionToDelimitedString(identities, ","));
     }
 
+    private void doFakePasswordEncoding(String password){
+        //PEN-21 Encoding password takes a while. To ensure execution time the same for not existing user, introduced fake password checking.
+        bCryptPasswordEncoder.matches(UUID.randomUUID().toString(), "$2a$10$hdjt9YGrSudbIljTqAtcW.KOxNJscq00Nxv088wPy6GDKXCJe0aCm");
```
leleueri, line 174: issue: the password is not used, UUID is still in place.

leleueri (changes requested):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/authentication/provider/security/ManagementAuthenticationProvider.java
podlesrafal: 🤦  fixed

leleueri (approved):

PR #5712, opened 2025-03-27
AM-4913 fix: added bidirectional relation for saml certificate

lgw-gravitee:
fixes AM-4913


lgw-gravitee:
bug introduced in https://gravitee.atlassian.net/browse/AM-3256

leleueri (approved):

PR #5717, opened 2025-03-28
fix: provide client secret on update/patch DCR response

leleueri:
only when the clientSecret is not hashed

fixes AM-4849


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/clientregistration/impl/DynamicClientRegistrationServiceImpl.java
```diff
@@ -139,15 +144,36 @@ public Single<Client> patch(Client toPatch, DynamicClientRegistrationRequest req
...
+
+    private Client assignSecret(Client app) {
+        // with client secret rotation, should we return the first secret of the list or maybe the oldest one ?
+        final var noneSettingsId = ofNullable(app.getSecretSettings())
+                .map(settings -> settings.stream()
+                        .filter(s -> SecretHashAlgorithm.NONE.name().equals(s.getAlgorithm()))
+                        .map(ApplicationSecretSettings::getId)
+                        .collect(Collectors.toSet()))
+                .orElse(Set.of());
+
```
lgw-gravitee, line 159: suggestion: fewer indention always better :P
```
ofNullable(app.getSecretSettings())
                .stream()
                .flatMap(List::stream)
                .filter(s -> SecretHashAlgorithm.NONE.name().equals(s.getAlgorithm()))
                .map(ApplicationSecretSettings::getId)
                .collect(Collectors.toSet());
```

wdyt?

lgw-gravitee (approved):

PR #5765, opened 2025-04-02
AM-4825 error message

lgw-gravitee:
fixes AM-4825


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/builder/AuthenticationAuditBuilder.java
```diff
@@ -71,4 +75,14 @@ private String getDisplayName(User user) {
 
         return displayName;
     }
+
+    @Override
+    public Audit build(ObjectMapper mapper) {
+        if (throwable instanceof BadCredentialsException) {
+            // removes username from audit log
+            throwable(new BadCredentialsException(throwable.getMessage()));
```
leleueri, line 83: **quetion**: as we create a new Exception using the message of the current one, how it can remove the "username" ?

podlesrafal (approved):

PR #5792, opened 2025-04-11
fix: when CAS redirects to callback we should omit LoginCallbackOpenI…

lgw-gravitee:
…DConnectFlowHandler step

fixes AM-4984


podlesrafal (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/login/LoginCallbackOpenIDConnectFlowHandler.java
```diff
@@ -76,6 +76,13 @@ public void handle(RoutingContext context) {
                 context.next();
                 return;
             }
+
+            // if CAS callback is used continue
+            if (request.getParam(Parameters.TICKET) != null && request.getParam(Parameters.STATE) != null) {
```
tcompiegne, line 81: Do we still have both TICKET and STATE parameters at the same time ? or can we have only the TICKET parameter ?

lgw-gravitee: right. I fixed that.

tcompiegne (approved):

PR #5809, opened 2025-04-25
AM-4922 fix: shrunk AM bundle size

lgw-gravitee:
fixes [AM-4922]

dataplane JDBC plugin shrunk by ~20MB
dataplane mongodb plugin shrunk by ~20MB

it will save in full-am-bundle ~80MB


leleueri (commented):
**question**: how this change can work we are marking liquibase as provided but we are not specifying the liquibase dependency in the mAPI / GW side... :thinking: ? If this dep is marked as "provied" it should be missing from the plugin bundle and the startup should fail... wdyt ?

leleueri:
> **question**: how this change can work we are marking liquibase as provided but we are not specifying the liquibase dependency in the mAPI / GW side... 🤔 ? If this dep is marked as "provied" it should be missing from the plugin bundle and the startup should fail... wdyt ?

I have the answer to my question. Liquibase is part of the repository plugin and DataPlane plugin is created by inheritin the repository context (to be able to mutualize the connection pool). However, with an AM installation using Mongo if we want to create a dataplane using jdbc (in theory it is possible) it will fail as the repository layer will be based on mongo so liquibase will be out of the scope.

Maybe we should declare liquibase as a dependency for the "gravitee-am-repository-jdbc-api" and make sure this depend is exclude from the plugins (repository-jdbc/reporter-jdbc/identityprovider-jdbc & dataplane-jdbc)

wdyt ?

PR #5877, opened 2025-05-14
fix: error message on client secret creation

podlesrafal:
fix: AM-5103


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationSecretServiceImpl.java
```diff
@@ -76,7 +76,7 @@ public Single<ClientSecret> create(Domain domain, Application application, NewCl
         List<ClientSecret> secrets = application.getSecrets();
         String newSecretName = newClientSecret.getName().trim();
         if (secrets.stream().map(ClientSecret::getName).anyMatch(n -> n.equals(newSecretName))) {
-            return Single.error(() -> new ClientSecretInvalidException(String.format("Secret with description %s already exists", newClientSecret)));
+            return Single.error(() -> new ClientSecretInvalidException(String.format("Secret with description %s already exists", newClientSecret.getName().trim())));
```
leleueri, line 79: todo: you are already doing newClientSecret.getName().trim() in line 77. you should be able to reuse it

PR #5906, opened 2025-05-20
feat: Redirect URI EL evaluation

lgw-gravitee:
solves AM-4927


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -787,6 +787,14 @@ private Single<Application> validateRedirectUris(Application application, boolea
                                 return Single.error(new InvalidRedirectUriException("redirect_uri : " + redirectUri + IS_MALFORMED));
                             }
                         }
+                        if(domain.isRedirectUriExpressionLanguageEnabled()){
+                            long uriWithELCount = oAuthSettings.getRedirectUris().stream()
+                                    .filter(uri -> new RedirectUri(uri).containsEL())
+                                    .count();
+                            if(uriWithELCount > 1) {
+                                return Single.error(new InvalidRedirectUriException("There can be only a single redirect_uri with EL"));
```
leleueri, line 795: this is not true, we can have multiple redirect URi with EL if the URI have different base URL.
ex:

* This is not valid as we do not know which one to pick
http://host1/path?query={#xxx}
http://host1/path?query={#yyy}

* This is valid as we can discrimiate
http://host1/path/one?query={#xxx}
http://host1/path/two?query={#yyy}
http://host2/path/two?query={#zzz}

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectELUriValidationHandler.java
```diff
@@ -0,0 +1,90 @@
...
+        final Client client = routingContext.get(ConstantKeys.CLIENT_CONTEXT_KEY);
+        List<String> redirectUris = client.getRedirectUris();
+        if(redirectUris == null || redirectUris.isEmpty()){
+            return List.of();
+        }
+        ExecutionContext executionContext = prepareContext(client, routingContext);
+        return Flowable.fromIterable(redirectUris)
+                .flatMapSingle(uri -> processQueryParams(uri, executionContext))
+                .toList()
+                .blockingGet();
```
leleueri, line 62: todo: please avoid blockingGet

leleueri (changes requested):

leleueri (commented):
I don't see where you are using the EL during the callback redirect

leleueri:
I found another issue.

Here is the callback I defined `https://callback?username={#context.attributes['user'].username}`, there is not trailing slash at the end.

In the overview page, the example contains an encoded trailing slash 
`http://localhost:8092/domain/oauth/authorize?client_id=app&response_type=code&redirect_uri=https%3A%2F%2Fcallback%2F`

Due to this trailing slach, the callback url doesn't contains the username parameter. 

If I remove the trailing slash from the example, the redirect is working fine and the username parameter is present.

leleueri (changes requested):
Please fix the two issues I found before merging

gravitee-am-service/src/main/java/io/gravitee/am/service/utils/EvaluableRedirectUri.java
```diff
@@ -0,0 +1,86 @@
...
+                .build()
+                .getQuery();
+        if(query == null){
+            return Maybe.empty();
+        } else {
+            return ctx
+                    .getTemplateEngine()
+                    .eval(query, String.class)
+                    .doOnError(err -> log.error(err.getMessage()))
+                    .onErrorResumeNext(err -> Maybe.error(new ELParamParsingException("EL evaluation exception")));
```
leleueri, line 52:
```suggestion
                    .onErrorResumeNext(err -> Maybe.error(new ELParamParsingException("query parameter processing exception")));
```

leleueri (approved):

podlesrafal (approved):

PR #5930, opened 2025-05-28
chore: update mergify config to manage 4.8.x

leleueri:


leleueri:
by pass the integration test as only mergify settings has change

PR #5933, opened 2025-05-29
fix: r2dbc-mssql in 1.0.1+ version problem with cursoredExecution

lgw-gravitee:
fixes AM-5106

added preferCursoredExecution boolean property to mssql connection option
also made tests to work with preferCursoredExecution=false

Changes between 1.0.0 and 1.0.1 -> https://github.com/r2dbc/r2dbc-mssql/compare/v1.0.0.RELEASE...v1.0.1.RELEASE

In MssqlConnectionConfiguration they changed
image

The DefaultCursorPreference returns true, when the query is "SELECT"

The documenations states:

_Whether to prefer cursors or direct execution for queries. Uses by default direct. Cursors require more round-trips but are more backpressure-friendly. Defaults to direct execution. Can be boolean or a Predicate<String> accepting the SQL query. (Optional)_

It seems like the feature is broken, I created an issue on r2dbc-mssql repo

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-jdbc-api/src/main/java/io/gravitee/am/repository/jdbc/provider/impl/ConnectionFactoryProvider.java
+++ b/gravitee-am-repository/gravitee-am-repository-jdbc-api/src/main/java/io/gravitee/am/repository/jdbc/provider/impl/ConnectionFactoryProvider.java
@@ -54,6 +54,7 @@ import static io.r2dbc.spi.ConnectionFactoryOptions.USER;
 public class ConnectionFactoryProvider {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConnectionFactoryProvider.class);
     public static final String TAG_SOURCE = "pool";
+    public static final String TAG_PREFER_CURSORED_EXECUTION = "preferCursoredExecution";
     public static final String TAG_DRIVER = "r2dbc_driver";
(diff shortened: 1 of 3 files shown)


gravitee-am-repository/gravitee-am-repository-jdbc-api/src/main/java/io/gravitee/am/repository/jdbc/provider/impl/ConnectionFactoryProvider.java
```diff
@@ -178,7 +179,12 @@ public ConnectionFactory factory() {
 
             builder = TlsOptionsHelper.setSSLOptions(builder, environment, prefix, driver);
 
-            connectionPool = (ConnectionPool)ConnectionFactories.get(builder.build());
+            final String preferCursorExecution = environment.getProperty(prefix + "preferCursorExecution");
```
leleueri, line 182: should we consider having a default value here set to false so it will works without issue if we upgrade the driver ? Also, maybe it could be nice to have the dependency updated to the 1.0.2 so the test container will pass with the latest driver version. wdyt ?

lgw-gravitee: We agreed to leave tests with 1.0.0 as the 1.0.2 seems to be buggy

PR #5941, opened 2025-06-02
fix: misaligned groovy script

podlesrafal:
The issue was in yarn.lock. The most probably this is because renovate introduced this regression by upgrading codemirror version and not all codemirror versions ware aligned.


leleueri (commented):
did you also check that it is fixed for the HTTP Callout policy ?

PR #5947, opened 2025-06-02
feat(idp): add support for PBKDF2 password encoder

tcompiegne:
closes AM-5185


leleueri (commented):
Question: should we manage it as encoder available in the Gravitee.yaml for default idp ? In the DefaultIdentityProviderService, there is a list of "supported" encoder but this PBKDF2 is not part of it.

leleueri (approved):

PR #5985, opened 2025-06-05
chore: angular 19 update

podlesrafal:
Explanation why we need standalone:false in every component: https://blog.angular.dev/the-future-is-standalone-475d7edbc706


leleueri (approved):
Does it make sens to open a technical task to move our component in standalone mode ?

PR #5987, opened 2025-06-06
AM-5196 sanitize x headers

lgw-gravitee:
fixes AM-5196


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
```diff
@@ -326,6 +326,8 @@ protected void doStart() throws Exception {
         // create the root router
         final Router rootRouter = Router.router(vertx);
 
+        sanitizeXForwardedHeaders(rootRouter);
```
leleueri, line 329: question: instead of adding it in each Provider, is it possible to add the handler during domain creation so if we are adding new protocol this sanitization will not be missed ?

PR #5996, opened 2025-06-09
fix: detailed audit logs for organization role

podlesrafal:
What was the issue:
When user with ORGANIZATION_* role had assigned DOMAIN_AUDIT, the audit log message was filtered.
When user with DOMAIN_* role had assigned DOMAIN_AUDIT, the log message was not filtered.
The issue was that reference was to Domain not Organization. With this change, only organization based role can see full audit message.
Fixes: AM-5226


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/AuditsResource.java
```diff
@@ -91,7 +91,7 @@ public void list(
 
         checkAnyPermission(organizationId, environmentId, domain, Permission.DOMAIN_AUDIT, Acl.LIST)
                 .andThen(auditService.search(domain, queryBuilder.build(), param.getPage(), param.getSize())
-                .flatMap(auditPage -> hasPermission(authenticatedUser, ReferenceType.DOMAIN, domain, Permission.DOMAIN_AUDIT, Acl.READ)
+                .flatMap(auditPage -> hasPermission(authenticatedUser, ReferenceType.ORGANIZATION, organizationId, Permission.DOMAIN_AUDIT, Acl.READ)
```
leleueri, line 94: question: Don't you think we should evaluate the hasPermission on domain,organization but also environement and apply a OR operator on them ? Here you only allow the action for orgniaztion user. wdyt ?

podlesrafal: I was lost how it should work but then I found this comment: `The LIST acl allows to retrieve list of entities with only basic information while READ acl allows access to all information about a specified entity.` I will change to what you suggested.

leleueri (approved):

PR #6007, opened 2025-06-10
fix: allow to enable proxy from plugin

podlesrafal:
Fixes: AM-5053
This relates to https://github.com/gravitee-io/gravitee-am-resource-http-factor/pull/63.
What was the issue: when useSystemProxy was not checked then the proxy was used even if the host was in excluded hosts because it passed null as a url. With this change the system proxy is used when checkbox is selected and url is not in excluded hosts.


leleueri (commented):
Not sure about the fix, I would instead update the plugin to always call the createWebClient taking the URL, and let the isExludedHost evaluate if the proxy setting has to be set. wdyt ?

podlesrafal:
When we provide every time url, then the checkbox "Use System Proxy" is useless, because it will enable proxy based on config.

leleueri (approved):

PR #6091, opened 2025-06-25
chore: account linking jest tests

podlesrafal:
Tests for AM-5281

### DIFF
--- a/gravitee-am-test/api/commands/gateway/login-commands.ts
+++ b/gravitee-am-test/api/commands/gateway/login-commands.ts
@@ -36,8 +36,17 @@ export const initiateLoginFlow = async (clientId, openIdConfiguration, domain, r
   return authResponse;
 };
 
-export const login = async (authResponse, userName, clientId, password = 'SomeP@ssw0rd', rememberMe = false) => {
+export const login = async (authResponse, userName, clientId, password = 'SomeP@ssw0rd', rememberMe = false, redirectOidc = false) => {
+  if (redirectOidc) {
+    //This is for testing OIDC as a provider
+    const loginResp = await performGet(authResponse.headers['location'], '', { Cookie: authResponse.headers['set-cookie'] }).expect(302);
+    authResponse = await performGet(loginResp.headers['location'], '', { Cookie: loginResp.headers['set-cookie'] }).expect(302);
+  }
   const loginResult = await extractXsrfTokenAndActionResponse(authResponse);
+  if (redirectOidc) {
+    const parsedUrl = new URL(loginResult.action);
+    clientId = parsedUrl.searchParams.get('client_id');
+  }
   return await performFormPost(
     loginResult.action,
     '',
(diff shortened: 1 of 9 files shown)


gravitee-am-test/api/commands/gateway/login-commands.ts
```diff
@@ -80,64 +89,92 @@ export const postConsent = async (consent) => {
...
     }).expect(200);
 
     const authorize2 = await postConsent(consent);
     expect(authorize2.headers['location']).toBeDefined();
-    expect(authorize2.headers['location']).toContain(`${process.env.AM_GATEWAY_URL}/${domain.hrid}/oauth/authorize`);
+    expect(authorize2.headers['location']).toContain(`/oauth/authorize`);
 
     const tokenResponse = await performGet(authorize2.headers['location'], '', {
-      Cookie: authorize.headers['set-cookie'],
+      Cookie: authorize2.headers['set-cookie'],
```
leleueri, line 155: **comment**: wondering if this fix could be one of the responsible of some flaky test 🤔

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+      {
+        identityProviders: [{ identity: localIdp.values().next().value.id, priority: -1 }],
+      },
+      app.id,
+    ),
+  );
+  domainUser = getUser('domain-user');
+  // Create the same user
+  await createUser(domainAccLinking.id, accessToken, domainUser);
+  await delay(6000);
```
leleueri, line 191:
```suggestion
  await waitForDomainSync();
```

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+          exitIfNoAccount: false,
+          exitIfMultipleAccount: false,
+          userAttributes: [{ name: 'email', value: 'test@test.fr' }],
+        }),
+        enabled: true,
+        condition: '',
+      },
+    ]);
+    await updateDomainFlows(domainAccLinking.id, accessToken, flows);
+    await delay(6000);
```
leleueri, line 233:
```suggestion
    await waitForDomaiSync();
```

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+      domainAccLinkingConfiguration,
+      domainAccLinking,
+      'https://test.com',
+      'code',
+      true,
+    );
+    expect(user1TokenResponse.headers['location']).toContain('test.com?code=');
+    cookies = user1TokenResponse.headers['set-cookie'];
+    const allUsers = await getAllUsers(domainAccLinking.id, accessToken);
+    expect(allUsers.data.length).toBe(1);
```
leleueri, line 248: **suggestion**: Maybe we can check also that the user has an entry in the identities attributes wdyt ?

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+  });
+
+  it('Should login to second app with local idp', async () => {
+    const authResponse = await performGet(
+      domainAccLinkingConfiguration.authorization_endpoint,
+      '?response_type=code&client_id=app-local&redirect_uri=https://test.com',
```
leleueri, line 254: suggestion: we can use the object here to provide the clientId instead of hardcoded value so it will be more simple the change value if we need to.

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
podlesrafal: wait for domain sync waites for 10s. IMHO I dont think we need 10s as sync between MAPI and GW is every 5s.

gravitee-am-test/api/commands/gateway/login-commands.ts
podlesrafal: We will see :)

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
leleueri: waitForDomainSync was initially 6s but we had troubles that's why I set it to 10.
I would prefer to use the domainSync domain even if it is waiting 4s more at least we clearly highlight why we are waiting and if we succeed to provide internal API to have the sync state, we have only method to update

gravitee-am-test/specs/gateway/account-link.jest.spec.ts
leleueri: waitForDomainSync was initially 6s but we had troubles that's why I set it to 10.
I would prefer to use the domainSync domain even if it is waiting 4s more at least we clearly highlight why we are waiting and if we succeed to provide internal API to have the sync state, we have only method to update

PR #6168, opened 2025-07-09
fix: error handling on reporter insert

podlesrafal:
It will be applied only on 4.4.x as 4.5.x has this fix
Fixes: AM-5366


gravitee-am-reporter/gravitee-am-reporter-mongodb/src/main/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporter.java
```diff
@@ -314,7 +314,12 @@ private Flowable<BulkWriteResult> bulk(List<Audit> audits) {
             return Flowable.empty();
         }
 
-        return Flowable.fromPublisher(reportableCollection.bulkWrite(this.convert(audits)));
+        return Flowable.fromPublisher(reportableCollection.bulkWrite(this.convert(audits)))
+                .retry(10)
+                .onErrorResumeNext(throwable -> {
+                    logger.error("Reporter couldn't write to MongoDB after 10 attempts.", throwable);
+                    return Flowable.empty();
+                });
```
lgw-gravitee, line 322: image

I can see in method above there is a retry after this Publisher. Why isn't it working?

PR #6175, opened 2025-07-14
AM-5239 feat: mfa enroll flow

lgw-gravitee:
closes AM-5239


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/FlowEntity.java
```diff
@@ -142,29 +142,18 @@ private static String getIcon(Type type) {
...
-            default:
-                return "communication:shield-thunder";
-        }
+        return switch (type) {
+            case ROOT -> "home:earth";
+            case LOGIN_IDENTIFIER -> "code:right-circle";
+            case LOGIN -> "finance:file";
+            case CONSENT -> "general:shield-check";
+            case REGISTER -> "communication:shield-user";
+            case RESET_PASSWORD, MFA_ENROLLMENT -> "finance:protected-file";
```
leleueri, line 151: nickpick: no need to chance it now but I think we should ask to Product team which icon they want https://components.gravitee.io/?path=/docs/atoms-gv-icon--all. could be nice to have one specific for this action

lgw-gravitee: we agreed on shield

PR #6177, opened 2025-07-14
fix: min validation on users get

podlesrafal:
Fixes AM-5349


leleueri (changes requested):
TODO: please manage also this @Min on the OrganizationUsersResource

PR #6182, opened 2025-07-15
feat: extension grant jwks based

gamachec:


leleueri:
Thanks for your contribution proposal @gamachec 
We based our PR https://github.com/gravitee-io/gravitee-access-management/pull/6329 on your change to avoid the creation of another plugin. 

We have to improve our CI to be able to execute some non regression tests on contributor PR so next time we will be able to ask you for the change and you name will be part of the contributor list.

Thanks again for this proposal.

PR #6183, opened 2025-07-15
fix: make JWKSet more resilient

podlesrafal:
I've tested this fix and in current implementation the flow is:
AheadCaching
RateLimit
OutageTolerance
Retry

How to tests:
The best way is to put some random delay on /.wel-known/jwks.json endpoint.


gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/jwt/jwks/remote/RemoteJWKSourceResolver.java
```diff
@@ -35,9 +38,22 @@ public RemoteJWKSourceResolver(String url) {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public JWKSource<C> resolve() {
         try {
-            return new RemoteJWKSet(new URL(url));
+            return (JWKSource<C>) JWKSourceBuilder.create(new URL(url))
+                    .cacheForever()
```
leleueri, line 45: **issue**: don't think cacheForever is good thing as the JWKS value may change. IMO better to have caching with TTL

gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/jwt/jwks/remote/RemoteJWKSourceResolver.java
```diff
@@ -35,9 +38,22 @@ public RemoteJWKSourceResolver(String url) {
...
+                    .outageTolerant(true)
+                    .retrying(true)
+                    .healthReporting((healthReport) -> {
+                        if (HealthStatus.HEALTHY.equals(healthReport.getHealthStatus())) {
+                            log.debug("JWK Source healthy at {}: {}", healthReport.getTimestamp(), healthReport.getHealthStatus());
+                            return;
+                        }
+                        log.error("JWK Source degraded at {}: ", healthReport.getTimestamp(), healthReport.getException());
```
leleueri, line 53: **question:** should it be warn instead of error ?

gravitee-am-identityprovider/gravitee-am-identityprovider-common/src/main/java/io/gravitee/am/identityprovider/common/oauth2/jwt/jwks/remote/RemoteJWKSourceResolver.java
```diff
@@ -35,9 +38,22 @@ public RemoteJWKSourceResolver(String url) {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public JWKSource<C> resolve() {
         try {
-            return new RemoteJWKSet(new URL(url));
+            return (JWKSource<C>) JWKSourceBuilder.create(new URL(url))
```
leleueri, line 44: **issue**: as you change the way to create the resolver, you are not reading the connection timeout and readTimeout from the JVM properties so with this implementation you are using the default 500ms. Better to use the create method which is taking the ResourceRetriever as parameter so you can specify the timeout values. By default we should set 5s as connectionTimeout and 10s as readTimeout and also make this values configurable based on the gravitee.yaml (maybe we should use the httpClient section to take the value wdyt?)

leleueri (changes requested):

PR #6201, opened 2025-07-21
fix: indexes for devices

podlesrafal:
Fixed indexes for devices collection
fixes AM-5379

PR #6207, opened 2025-07-21
fix: device indexes

podlesrafal:
Fixes indexes for device collection
Closes AM-5379


gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_7_0/4.7.10-device-indexes.yml
```diff
@@ -0,0 +1,43 @@
...
+                  name: expires_at
+            indexName: idx_devices_domain_user_expires
+            tableName: devices
+            unique: false
+        - createIndex:
+            columns:
+              - column:
+                  name: reference_id
+              - column:
+                  name: reference_type
```
lgw-gravitee, line 24: reference_type is more general and should be before reference_id

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_7_0/4.7.10-device-indexes.yml
```diff
@@ -0,0 +1,43 @@
...
+              - column:
+                  name: client
+              - column:
+                  name: user_id
+              - column:
+                  name: device_identifier_id
+              - column:
+                  name: device_id
+              - column:
+                  name: expires_at
```
lgw-gravitee, line 34: beside that, do we need such enormous index?

leleueri (changes requested):

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_7_0/4.7.10-device-indexes.yml
podlesrafal: I dont agree. More selective fields should be first. We have only few types, but we can have many ids of references. So first field in index must be more selective which is referenceId.

leleueri (changes requested):
see https://github.com/gravitee-io/gravitee-access-management/pull/6201

leleueri (approved):

PR #6208, opened 2025-07-22
fix: map dynamic roles when account linking is used.

podlesrafal:
Closes: AM-5385


leleueri (commented):
also mergify tags are missing

PR #6214, opened 2025-07-23
AM-5376 duplicate insert mongo audit

lgw-gravitee:
closes AM-5318


leleueri (commented):
LGTM but we need to move the retry in the flatMap also for JDBC reporter. wdyt?

leleueri (approved):

PR #6220, opened 2025-07-24
fix: default constructor for RemoteJWKSourceResolver

podlesrafal:
Added constructor only with url to provide compatibility with plugins
Fixes: AM-5412


leleueri (approved):
Tested locally, FC is working

PR #6244, opened 2025-07-29
fix: AM-5434 purge service

lgw-gravitee:


gravitee-am-management-api/gravitee-am-management-api-services/gravitee-am-management-api-services-purge/src/main/java/io/gravitee/am/management/services/purge/ScheduledPurgeService.java
```diff
@@ -63,16 +61,20 @@ public class ScheduledPurgeService extends AbstractService implements Runnable {
...
+        String type = environment.getProperty("repositories.management.type", "mongodb");
+        if (enabled && "jdbc".equalsIgnoreCase(type)) { // even if enabled, useless for Mongo implementation
             super.doStart();
             logger.info("Purge service has been initialized with cron [{}]", cronTrigger);
             // Sync must start only when doStart() is invoked, that's the reason why we are not
             // using @Scheduled annotation on doSync() method.
             scheduler.schedule(this, new CronTrigger(cronTrigger));
         } else {
-            logger.warn("Purge service has been disabled");
+            logger.warn("Purge service has been disabled, enabled={}, type={}", enabled, type);
```
leleueri, line 77: should we keep a warn level here 🤔  IMO this is more an INFO wdyt?

PR #6245, opened 2025-07-29
chore: update the CertificateProvider interface to provide jwtBuilder…

leleueri:
… and jwtParser

 this update is required to improve preformance using CloudHSM plugin

 thanks to this change the plugin is able to provide a builder or parser backed by a Pool of Signature object

 this pool allow to avoid some network call to initialize the signature and provide a control on the concurrent actio on the HSM

fixes AM-5435

fixes gravitee-io/issues#10704

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-certificate/src/main/java/io/gravitee/am/gateway/certificate/impl/CertificateProviderManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-certificate/src/main/java/io/gravitee/am/gateway/certificate/impl/CertificateProviderManagerImpl.java
@@ -83,12 +83,12 @@ public class CertificateProviderManagerImpl implements CertificateProviderManage
             if (keyValue instanceof KeyPair) {
                 PrivateKey privateKey = ((KeyPair) keyValue).getPrivate();
                 PublicKey publicKey = ((KeyPair) keyValue).getPublic();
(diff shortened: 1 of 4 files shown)


gravitee-am-gateway/gravitee-am-gateway-certificate/src/main/java/io/gravitee/am/gateway/certificate/impl/CertificateProviderManagerImpl.java
```diff
@@ -78,6 +78,12 @@ public CertificateProvider create(io.gravitee.am.certificate.api.CertificateProv
         // create certificate provider
         CertificateProvider certificateProvider = new CertificateProvider(provider);
         try {
+            if (provider.canProvideJwtProcessors()) {
+                certificateProvider.setJwtBuilder(provider.jwtBuilder());
+                certificateProvider.setJwtParser(provider.jwtParser());
+                return certificateProvider;
+            }
```
lgw-gravitee, line 85: how about using Optional instead? 
if present then set

thanks to it we remove null return methods:P

lgw-gravitee: and canProvideJwtProcessors

leleueri: indeed

PR #6249, opened 2025-07-30
fix: AM-5438 update User in RoutingContext after successful challange 

lgw-gravitee:
fixes AM-5438


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengePostEndpoint.java
```diff
@@ -272,6 +272,7 @@ private Handler<AsyncResult<Void>> verifyHandler(RoutingContext routingContext,
                         return;
                     }
 
+                    routingContext.put(ConstantKeys.USER_CONTEXT_KEY, fh.result());
```
leleueri, line 275: can you please open another ticket to provide as context attribute the fact that the user is currently enrolling a factor ? We have this boolean here, I guess we can have it also on the PRE MFA CHALLENGE flow. It can be useful to have such information to create policy condition. (maybe it is available already in the session and accessible from the EL, worth to check)

leleueri (approved):

PR #6279, opened 2025-08-05
fix: AM-5233 global kafka reporter

lgw-gravitee:
fixes AM-5233


gravitee-am-model/src/main/java/io/gravitee/am/model/common/event/EventKey.java
```diff
@@ -0,0 +1,59 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
```
wajb, line 2: Is this a standard disclaimer and do we need to stick with 2015? Thanks.

lgw-gravitee: yes, its standard and it's added by Maven plugin `mvn license:format`
no build will pass without strict match.

PR #6280, opened 2025-08-05
fix: AM-5436 added missing mfa* options for custom flows

lgw-gravitee:
fixes AM-5436

### DIFF
--- a/gravitee-am-service/src/main/resources/flow/am-schema.json
+++ b/gravitee-am-service/src/main/resources/flow/am-schema.json
@@ -21,7 +21,9 @@
         "reset_password",
         "token",
         "connect",
-        "webauthn_register"
+        "webauthn_register",
+        "mfa_challenge",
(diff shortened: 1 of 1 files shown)


gravitee-am-service/src/main/resources/flow/am-schema.json
```diff
@@ -35,7 +37,9 @@
           "reset_password": "RESET PASSWORD",
           "registration_confirmation": "REGISTRATION CONFIRMATION",
           "token": "TOKEN",
-          "webauthn_register": "WEBAUTHN_REGISTER"
+          "webauthn_register": "WEBAUTHN_REGISTER",
```
podlesrafal, line 40: Maybe use without `_` like MFA CHALLENGE and MFA ENROLLMENT? Webauthn you can also change :)

leleueri (approved):

PR #6282, opened 2025-08-05
fix: remove automatic assignment of default values for LDAP configura…

sclarkamlab:
fixes: AM-5374

The changes introduced in this PR removes the re-application of default values to an existing LDAP configuration. When the initial configuration is created in the UI, the default values are applied correctly. Then when the user clears the value for any of the fields with a default value, the change is persisted to the database and the default value is no longer applied.

## :memo: Test scenarios 
Follow reproduction steps outlined in the ticket [AM-5374](https://gravitee.atlassian.net/browse/AM-5374)

### Additional Testing
- Create a new LDAP entry, keep the default value for `Group search base`
- Save the LDAP provider
- Edit the record
- Verify the default value was saved (`ou=applications`)
- Edit the Group search base, by removing the value entirely
- Save the record
- Reload the page and verify the value is now empty

[AM-5374]: https://gravitee.atlassian.net/browse/AM-5374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-ui/src/app/domain/settings/providers/provider/settings/settings.component.ts
+++ b/gravitee-am-ui/src/app/domain/settings/providers/provider/settings/settings.component.ts
@@ -91,15 +91,6 @@ export class ProviderSettingsComponent implements OnInit {
       .pipe(map((schema) => enrichOIDCFormWithCerts(schema, this.certificates)))
       .subscribe((data) => {
         this.providerSchema = data;
-        if (data) {
-          // handle default null values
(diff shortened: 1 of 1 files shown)


leleueri (approved):
Looks OK.
The LDAP issue describe in the ticket is fixed and a custom mongo IDP without the "CRUD" field in the JSON config has the default value set when it is loaded

PR #6284, opened 2025-08-06
[AM-4985] Add support for escaping regex special characters in FilterCriteriaParser

sclarkamlab:
fixes: AM-4985 - [Link](https://gravitee.atlassian.net/browse/AM-4985)

This pull request improves the handling and testing of regex special characters in filter criteria parsing for MongoDB queries. The changes ensure that when using regex-based operators, special characters in filter values are properly escaped, preventing malformed queries and potential security issues. Additionally, comprehensive parameterised tests are added to verify correct escaping for all supported regex characters.

**Regex handling improvements:**

* Added a list of regex operators (`co`, `sw`, `ew`) and a list of allowed regex special characters to `FilterCriteriaParser`, and implemented logic to escape these characters in filter values when a regex operator is used. 

**Testing enhancements:**

* Introduced parameterised tests in `FilterCriteriaParserTest` to validate that all allowed regex special characters are correctly escaped in filter values for regex operators.

## :memo: Test scenarios 
Follow the steps outlined in the ticket for testing, straight forward to set up and verify.

Alternatively:
 - Create two users with email addresses:
 - `bug.admin@bug.com`
 - `bug+admin@bug.com`
 
Navigate to the user management page: https://example.com/environments/default/domains/my-domain/settings/users
Search for these users using the SCIM query language:

- `email ew "+admin@bug.com"` -> Should return only 1 record
- `email ew ".admin@bug.com"` -> Should return only 1 record
- `email ew "@bug.com"` -> Should return both records

## 📓  Notes
Although the ticket specifically calls out a set of characters to support, some of these are explicitly restricted by the current implementation, namely: `;`, `{`, `}`, `$` and `^`

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/common/FilterCriteriaParser.java
+++ b/gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/common/FilterCriteriaParser.java
@@ -47,9 +47,31 @@ public final class FilterCriteriaParser {
             add("{");
             add("}");
             add("$");
+            add("^");
         }
     };
     public static final String FOUND_IN_THE_THE_SEARCH_QUERY = "] found in the the search query";
+    private static final List<String> REGEX_OPERATORS = new ArrayList<>(){
+        {
+            add("co");
+            add("sw");
+            add("ew");
+        }
+    };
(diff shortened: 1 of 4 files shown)


leleueri (commented):
TODO: can you please add a test in the gravitee-am-repository-test module so can assert that both implementation Mongo & R2DBC can manage the same request?

leleueri (approved):

leleueri:
I didn't test it as the Unit test prove to behaviour

PR #6288, opened 2025-08-07
fix: limit the number of threads available to perform blocking token …

leleueri:
…signatures

 this is related to HSM plugin which is doing blocking call in a native library

 in this case the worker thread pool of VertX is not big enough

fixes AM-5482


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/jwt/impl/JWTServiceImpl.java
```diff
@@ -55,6 +56,9 @@ public class JWTServiceImpl implements JWTService {
     @Autowired
     private ObjectMapper objectMapper;
 
+    @Autowired
+    private IoExecutor executor;
+
```
lgw-gravitee, line 61: According to following comment I would put here qualifier with specific bean name "jwtSignerExectuor"

lgw-gravitee (approved):

PR #6298, opened 2025-08-08
feat:purge events after retention time

FuYu3699:
open this as a draft for preliminary review, still need to add tests and local testing.

[JIRA AM-5431](https://gravitee.atlassian.net/browse/AM-5431)

This PR add events records purge logic with default retention time set to 90 days. This retention time is configurable, but this may not be needed expose to customer unless needed.

## :memo: Test scenarios 
MongoDB:
Setup API to use MongoDB
Modify purge configuration in api [gravitee.yml](https://github.com/gravitee-io/gravitee-access-management/blob/ee3258e2e09d05955a97b607c65061ed442a0e8c/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml#L148) to the following:
```
  # Data purge service (JDBC only).
  purge:
    enabled: true                               # enable the JDBC purge task (default: true)
    cron: 0 * * * * *                          # configure the frequency (default: every day at 11 PM)
    #exclude: login_attemps, refresh_token      # coma separated list of table to exclude from the purge process
    events:
      retention:
        days: 2                                # event retention period in days (default: 90 days)
```

Rebuild and launch management api, and wait for a minute
Check events table in MongoDB, any events older than 2days should be purged.

JDBC:
Setup environment to use RDBMS backend:
Modify purge configuration in gateway [gravitee.yml](https://github.com/gravitee-io/gravitee-access-management/blob/ee3258e2e09d05955a97b607c65061ed442a0e8c/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml#L194), and add the 
following in the `services` section:
```
  # Data purge service (JDBC only).
  purge:
    enabled: true                               # enable the JDBC purge task (default: true)
    cron: 0 * * * * *                          # configure the frequency (default: every day at 11 PM)
    #exclude: login_attemps, refresh_token      # coma separated list of table to exclude from the purge process
    events:
      retention:
        days: 2                                # event retention period in days (default: 90 days)
```

Launch Gateway and wait for a minute to check the events database, any events older than 2 days should be purged.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -161,6 +161,15 @@ services:
     prometheus:
       enabled: true
 
+  # Data purge service (JDBC only).
+  # purge:
+    # enabled: true                               # enable the JDBC purge task (default: true)
+    # cron: 0 0 23 * * *                          # configure the frequency (default: every day at 11 PM)
+    # exclude: login_attempts, refresh_token      # comma separated list of table to exclude from the purge process
+    # events:                                     # event retention period in days (default: 90 days)
+    #   retention:
+    #     days: 90
+
   # Gateway monitoring service.
   # This service retrieves metrics like os / process / jvm metrics.
   monitoring:
--- a/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -144,11 +144,11 @@ services:
     prometheus:
       enabled: true
 
-  # Data purge service (JDBC only).
-  purge:
-    enabled: true                               # enable the JDBC purge task (default: true)
-    cron: 0 0 23 * * *                          # configure the frequency (default: every day at 11 PM)
-    #exclude: login_attemps, refresh_token      # coma separated list of table to exclude from the purge process
+  # Data purge.
+  # purge:
+  #   events:                                     # event retention period in days (default: 90 days) for MongoDB
+  #     retention:
+  #       days: 90
 
   # platform notifier service
   notifier:


gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/MongoEventRepository.java
```diff
@@ -65,12 +69,21 @@ public class MongoEventRepository extends AbstractManagementMongoRepository impl
...
+    @Value("${services.purge.events.retention.days:90}")
+    private int eventsRetentionDays;
+
     @PostConstruct
     public void init() {
         eventsCollection = mongoOperations.getCollection("events", EventMongo.class);
         super.init(eventsCollection);
         super.createIndex(eventsCollection, Map.of(new Document(FIELD_UPDATED_AT, 1), new IndexOptions().name("u1")));
         super.createIndex(eventsCollection, Map.of(new Document(FIELD_UPDATED_AT, 1).append(DATA_PLANE_ID, 1), new IndexOptions().name("u1dp1")));
+        super.createIndex(eventsCollection, Map.of(new Document(FIELD_CREATED_AT, 1), new IndexOptions().name("e1").expireAfter((long) eventsRetentionDays, TimeUnit.DAYS)));
```
leleueri, line 86: **question**: what is the behaviour if you restart the process after updating the retention days? Do the index is updated, is there an error or it has no effect?

FuYu3699: I will be no effect, rerun the service won't update the index if it has been created. 
if someone want to update it, he can remove the index manually, and let the service to recreate the index with the new TTL.

FuYu3699: what would be the expectation, do we want to update the new value?

FuYu3699:
> > I don't feel that we should doc this as the original intention is to hardcode the rentention time to be 3 or 4 months (for both MongoDB and JDBC scenarios). I added this event retention days that offers the convenience for testing. it is a bit awkward production wise since in MongoDB case, it will use this setting regardless the purge is enabled, but for JDBC case the enabled has to be true.
> > > Actually one other question: do the docs/comments in the `gravitee.yml` need to be updated since some of the settings now apply to mongodb as well as jdbc and it no longer all relates to the purge service?
> > > `# Data purge service (JDBC only).`
> > > Relatedly, given the hierarchy of settings, a user might expect events in mongodb to never purge when `services.purge.enabled` is false. Should we respect that?
> 
> Thanks for the explanation. This is fine if it's an undocumented setting, but the task does say the setting should be defined, which does imply documented IMO. 🤔

I documented the configuration changes in the yaml files.

wajb (commented):
Just a couple of points about the yml docs.

I validated the behaviour locally with both mongodb and postgres repositories.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -161,6 +161,15 @@ services:
     prometheus:
       enabled: true
 
+  # Data purge service (JDBC only).
```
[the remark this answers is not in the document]
wajb: Or does mongodb work differently and not use these settings on the gateway side? 🤔

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -161,6 +161,15 @@ services:
...
+  # purge:
+    # enabled: true                               # enable the JDBC purge task (default: true)
+    # cron: 0 0 23 * * *                          # configure the frequency (default: every day at 11 PM)
+    # exclude: login_attemps, refresh_token      # coma separated list of table to exclude from the purge process
```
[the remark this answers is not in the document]
FuYu3699: yeah, I saw them, didn't touch, but yeah, I can get them corrected. :)

wajb (approved):
LGTM. Be careful of commit prefixes (and I believe we're meant to put the JIRA ID in the body - see [guidelines](https://gravitee.slab.com/posts/commit-conventions-sqc1f3zj)).

PR #6340, opened 2025-08-18
feat: enhance file retention logic with debug logging and improved cleanup

sclarkamlab:
fixes: AM-5504

Trigger the clean up process when the service first starts up to clean up any existing files left over from the service running previously. Also added in some debug logging for future investigations.

--- 

The log cleanup was working as designed, but it only ran during the daily rollover at midnight.
This meant that if the service was restarted during the day, old log files weren’t removed until the next midnight, which gave the impression cleanup wasn’t working.

We’ve updated the logic so that cleanup also runs immediately on startup (when retention is configured), in addition to the midnight rollover. This ensures old files are purged right away after a restart.

## :memo: Test scenarios 
Follow steps in the ticket, key indicator here is that the clean up process is now triggered on start up as well as the usual daily schedule of midnight.

### DIFF
--- a/gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
+++ b/gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
@@ -116,11 +116,15 @@ public class VertxFileWriter<T extends ReportEntry> {
     }
 
     public Future<Void> initialize() {
+        LOGGER.debug("Initializing file reporter, retainDays: {}", retainDays);
         // Calculate Today's Midnight, based on Configured TimeZone (will be in past, even if by a few milliseconds)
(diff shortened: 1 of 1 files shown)


wajb (approved):
Validated change in behaviour compared to master: using `touch -t` I was able to artificially age log files in both Management API and Gateway, then confirm they were removed according to the retention period in both a debugging session and logs.

```
09:46:34.869 [Thread-9] [] DEBUG i.g.a.r.file.vertx.VertxFileWriter - Initializing file reporter, retainDays: 1
09:46:34.869 [Thread-9] [] DEBUG i.g.a.r.file.vertx.VertxFileWriter - Scheduling next rollover for 2025-08-23T00:00+01:00[Europe/London]
09:46:49.440 [Thread-9] [] DEBUG i.g.a.r.file.vertx.VertxFileWriter - Checking file [70ec6fe0-52aa-4560-ac6f-e052aa3560e4-2025_08_22.json]
09:47:10.903 [Thread-9] [] DEBUG i.g.a.r.file.vertx.VertxFileWriter - File [70ec6fe0-52aa-4560-ac6f-e052aa3560e4-2025_08_22.json] should be removed
```

PR #6365, opened 2025-08-20
fix: AM-2741 remove ineffectual maxPoolSize settings

wajb:
[AM-2741](https://gravitee.atlassian.net/browse/AM-2741)

Reverts addition of `maxPoolSize` settings that were introduced in https://github.com/gravitee-io/gravitee-access-management/pull/6333. This restores the hardcoded default of 10, which applies only to a few Management API service usages (reCAPTCHA verification and newsletter subscriptions). The settings were confusing and ineffectual in the context of the feature request, since Identity Providers are configurable with their own `HTTP Client max pool size` settings. These settings are now respected for both HTTP/1.1 and HTTP/2 (if the latter is enabled).

The net new `gravitee.yml` settings and their defaults (in both Management API and Gateway) are:
```yml
httpClient:
  http2:
    enabled: true
    connectionWindowSize: 65535
    keepAliveTimeout: 60 # in seconds
```

## :memo: Test scenarios 
As previously, for the primary IdP use case, steps similar to https://gravitee.atlassian.net/browse/AM-4040 can be followed. We can then observe logs according to configuration provided in the gateway's `gravitee.yml` (requires DEBUG level logging in `logback.xml`). Either of the following will be logged at startup, depending on the HTTP/2 enabled setting:

`DEBUG i.g.am.service.http.WebClientBuilder - HTTP/2 enabled with ALPN protocol negotiation`
`DEBUG i.g.am.service.http.WebClientBuilder - HTTP/1.1 enforced`

It ought to be possible to inspect the usage of the connection window size and keepalive values by capturing the SETTINGS frame parameters `SETTINGS_INITIAL_WINDOW_SIZE` and `SETTINGS_KEEPALIVE_TIMEOUT`, respectively, in TCP traffic. But note this requires a https-served IdP (_domain_ when using AM) and a tool like wireshark. (This is arguably only testing the Vert.x library, however, since debugging confirms they are configured in the HTTP client 😁.) 

n/a

n/a

@gravitee-io/am


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -405,7 +405,6 @@ gateway:
 #httpClient:
 #  timeout: 10000 # in milliseconds
 #  readTimeout: 5000 # in milliseconds
-#  maxPoolSize: 10
```
FuYu3699, line 408: would this still be needed for http scenario?

FuYu3699: thanks. what does the hardcoded pool size do then, if it doesn't affect anything, shall we consider to remove it completely?

FuYu3699 (approved):
I haven't tested this change locally, but the code changes LGTM.

PR #6397, opened 2025-08-28
Am 5484 filereporter creates newfile

FuYu3699:
https://gravitee.atlassian.net/browse/AM-5484

The filename pattern matching doesn’t work because the filename is overwritten with the actual date after initialisation. The class only creates a new log file during each service startup.

Added a unit test to simulate the use case and introduced the template filename that won't get overwritten when rollover happens.

## :memo: Test scenarios 

follow the Jira ticket to find testing steps.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/impl/UserAuthenticationServiceImpl.java
@@ -15,6 +15,7 @@
  */
 package io.gravitee.am.gateway.handler.common.auth.user.impl;
 
+import com.google.common.base.Strings;
 import io.gravitee.am.common.audit.EventType;
 import io.gravitee.am.common.exception.authentication.AccountDisabledException;
 import io.gravitee.am.common.exception.authentication.AccountEnforcePasswordException;
@@ -82,6 +83,7 @@ import static java.util.Optional.ofNullable;
 public class UserAuthenticationServiceImpl implements UserAuthenticationService {
 
     private static final String SOURCE_FIELD = "source";
+    private static final String LAST_IDENTITY_FIELD = "last_identity";
     private final Logger LOGGER = LoggerFactory.getLogger(this.getClass());
 
     @Autowired
@@ -201,6 +203,7 @@ public class UserAuthenticationServiceImpl implements UserAuthenticationService 
                             // retrieve information from the idp user and update the user
                             Map<String, Object> additionalInformation = idpUser.getAdditionalInformation() == null ? new HashMap<>() : new HashMap<>(idpUser.getAdditionalInformation());
                             additionalInformation.put(SOURCE_FIELD, user.getSource());
+                            additionalInformation.put(LAST_IDENTITY_FIELD, user.getLastIdentityUsed());
                             additionalInformation.put(Parameters.CLIENT_ID, user.getClient());
                             ((DefaultUser) idpUser).setAdditionalInformation(additionalInformation);
                             final User preConnectedUser = create0(idpUser, false);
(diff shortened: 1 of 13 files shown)


gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
```diff
@@ -146,14 +149,16 @@ private Future<Void> setFile(ZonedDateTime now) {
                     return promise.future();
                 }
 
-                String simpleFilename = file.getName();
-                int datePattern = simpleFilename.toLowerCase(Locale.ENGLISH).indexOf(YYYY_MM_DD);
+                // Use the template filename for pattern matching during rollover
+                // This ensures we can always find and replace the date pattern, even after
+                // the filename has been processed with an actual date
+                int datePattern = templateFilename.toLowerCase(Locale.ENGLISH).indexOf(YYYY_MM_DD);
```
wajb, line 155: suggestion: this could be made more readable by using string replacement. I think this should work whether or not the template has a date placeholder:

```java
filename = dir.getAbsolutePath() + File.separatorChar + templateFilename.replaceFirst("(?i)" + YYYY_MM_DD, fileDateFormat.format(new Date(now.toInstant().toEpochMilli())));
```

gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
```diff
@@ -189,6 +194,16 @@ private Future<Void> setFile(ZonedDateTime now) {
         return promise.future();
     }
 
+    // Visible for testing only.
+    String getCurrentFilename() {
+        return filename;
```
wajb, line 199: question: would it be possible to scan the directory rather than inspecting this field?

gravitee-am-reporter/gravitee-am-reporter-file/src/test/java/io/gravitee/am/reporter/file/vertx/VertxFileWriterTest.java
```diff
@@ -193,4 +197,66 @@ public void shouldRemoveOldFiles() throws IOException {
...
+                    mock(Formatter.class),
+                    templateFilename,
+                    1L);
+
+            AtomicReference<String> firstFilename = new AtomicReference<>();
+            AtomicReference<String> secondFilename = new AtomicReference<>();
+            AtomicReference<Boolean> testCompleted = new AtomicReference<>(false);
+
+            vertxFileWriter.initialize().onComplete(initResult -> {
+                if (initResult.succeeded()) {
```
wajb, line 220: minor: could this be an assertion rather than a condition? The `else` could be removed either way.

gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
```diff
@@ -189,6 +194,16 @@ private Future<Void> setFile(ZonedDateTime now) {
...
+    }
+
+    // Visible for testing only.
+    Future<Void> setFileForTest(ZonedDateTime dateTime) {
```
wajb, line 203: thought: the convention for other methods is to use `protected`. In java this appears to mean they're accessible in the same package. If you just made `setFile` protected you wouldn't need this method.

wajb (commented):
Added some comments but have not retested.

gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
FuYu3699: nice one, updated

gravitee-am-reporter/gravitee-am-reporter-file/src/main/java/io/gravitee/am/reporter/file/vertx/VertxFileWriter.java
FuYu3699: updated

wajb (approved):
Retested and confirmed resolved by debugging new logic during rollover. Note that the steps outlined in JIRA cannot be used to directly verify the fix since rollover still needs to occur at midnight.

PR #6422, opened 2025-09-03
AM-5510 update delete statement to support mssql

FuYu3699:

https://gravitee.atlassian.net/browse/AM-5510

**MsSQL server doesn’t support delete statement with alias which can cause the following error:**
```
delete from account_access_tokens a where a.user_id = 'user-1' and a.reference_id = 'account-123' and a.reference_type = 'ACCOUNT'

SQL Error [102] [S0001]: Incorrect syntax near 'a'.
```

**updating the statement to the following can fix this:**
```
delete from account_access_tokens where user_id = 'user-1' and reference_id = 'account-123' and reference_type = 'ACCOUNT'

Updated Rows	1
Execute time	0.111s
```

update the statement to not use alias.


leleueri (changes requested):
todo: may I ask you to create a test for this deleteByUserId in the gravitee-am-repositories-tests module ? Looks like the deleteByUser is  missing so it will help to avoid regression in the future.

leleueri (approved):

PR #6444, opened 2025-09-05
fix: Enhance idp plugin redeployment to avoid downtime

podlesrafal:
AM-5531

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
@@ -35,13 +35,16 @@ import io.gravitee.common.service.AbstractService;
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
+import io.reactivex.rxjava3.schedulers.Schedulers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import java.io.IOException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.ReentrantLock;
 
 import static io.gravitee.am.identityprovider.api.common.IdentityProviderConfigurationUtils.extractCertificateId;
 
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
```diff
@@ -171,47 +179,151 @@ private Single<IdentityProvider> updateAuthenticationProvider(IdentityProvider i
...
             return forceUpdateAuthenticationProvider(identityProvider);
         } else {
-            logger.debug("\tIdentity provider already initialized: {} [{}]", identityProvider.getName(), identityProvider.getType());
+            logger.debug("Identity provider already initialized: {} [{}]", identityProvider.getName(), identityProvider.getType());
             return Single.just(identityProvider);
         }
     }
 
     private Single<IdentityProvider> forceUpdateAuthenticationProvider(IdentityProvider identityProvider) {
         return Single.fromCallable(() -> {
```
sclarkamlab, line 188: Can this be converted to Single.just, since it does not actually do any async work and should simplify the code a bit?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
```diff
@@ -171,47 +179,151 @@ private Single<IdentityProvider> updateAuthenticationProvider(IdentityProvider i
...
 
     private void clearProviders() {
         providers.keySet().forEach(this::clearProvider);
     }
 
+    private Single<Providers> createNewProviders(IdentityProvider identityProvider) throws IOException {
+        var authProviderConfig = new AuthenticationProviderConfiguration(identityProvider, certificateManager);
+        try (var authenticationProvider = identityProviderPluginManager.create(authProviderConfig)) {
+
+            if (authenticationProvider != null) {
```
sclarkamlab, line 211: Do we generally prefer to return early?

```
if (authenticationProvider == null) {
    return Single.error....
}
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
```diff
@@ -171,47 +179,150 @@ private Single<IdentityProvider> updateAuthenticationProvider(IdentityProvider i
...
     private void clearProviders() {
         providers.keySet().forEach(this::clearProvider);
     }
 
+    private Single<Providers> createNewProviders(IdentityProvider identityProvider) throws IOException {
+        var authProviderConfig = new AuthenticationProviderConfiguration(identityProvider, certificateManager);
+        try (var authenticationProvider = identityProviderPluginManager.create(authProviderConfig)) {
+
+            if (authenticationProvider == null) {
+                return Single.error(new RuntimeException("Failed to create authentication provider for " + identityProvider.getName()));
```
wajb, line 211: thought: this error is only thrown for redeployments and never initial deployments. Would an extraction of a single factory method help with consistency, or is this difference intentional? (The name `createNewProviders` doesn't convey that it's only used for redeployments, so moving logic a little could help with readability too IMO.)

PR #6583, opened 2025-09-29
chore(cicd): AM-5702 the maint branch should be created only on release-version

lgw-gravitee:
closes AM-5702

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -98,6 +98,10 @@ parameters:
     type: boolean
     default: false
     description: "Is this a hotfix?"
+  create_maintenance_version_branch:
+    type: boolean
+    default: false
+    description: "Creates maintenance version branch"
 
 commands:
   create_docker_context:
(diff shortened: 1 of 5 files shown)


.circleci/workflows.yml
leleueri, line 1997: the parameter is not provided here.

release/steps/am-release-alpha-pipeline.mjs
```diff
@@ -15,7 +15,8 @@ export class ReleaseAlphaVersion extends AbstractStepRelease {
             gio_action: 'release-alpha-version',
             rc_requested: true,
             tag_latest: false,
-            tag_latest_support: false
+            tag_latest_support: false,
+            create_maintenance_version_branch: false
```
leleueri, line 19: todo: we need a parameter here to specify if we want to crearte a branch or not. During code freeze, we need to create a branch.

leleueri (changes requested):

PR #6587, opened 2025-09-29
fix: do not propagate the IDP exception when the user aborted the aut…

leleueri:
…hentication

fixes AM-5491

# Description

The purpose of this PR is to not display the error message on the Login page when a user abort the login with FranceConnect. The abort button on FC callback AM with an "access_denied" error and error _description "user auth aborted", as there is an error parameter AM propagate a generic error on the Login page but in this specific case the has been interrupted by the user.

This PR doesn't implement the ideal solution where the IDP Plugin is able to interpret the error as this solution will required changes in the IdenttyProvider interface and brings a BreakingChange for each IDP plugin.

A improvement will have to be manage in a future feature version to manage this case in a clean way.

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
@@ -17,6 +17,7 @@ package io.gravitee.am.gateway.handler.root.resources.auth.provider;
 
 import io.gravitee.am.common.exception.authentication.BadCredentialsException;
 import io.gravitee.am.common.exception.authentication.LoginCallbackFailedException;
+import io.gravitee.am.common.exception.authentication.UserAuthenticationAbortedException;
 import io.gravitee.am.common.jwt.Claims;
 import io.gravitee.am.common.oauth2.Parameters;
 import io.gravitee.am.gateway.handler.common.auth.AuthenticationDetails;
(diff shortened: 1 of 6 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
wajb, line 184: question: should we prevent logging an error when the exception is a `UserAuthenticationAbortedException`, or is this ok? There is some logic in `LoginCallbackFailureHandler` that logs at debug level for `AuthenticationException`s but I see that's not the case in this class.

```
14:39:48.216 [Thread-18] [] ERROR i.g.a.g.h.r.r.a.p.SocialAuthenticationProvider - Unable to authenticate social provider
io.gravitee.am.common.exception.authentication.UserAuthenticationAbortedException: User auth aborted
	at io.gravitee.am.gateway.handler.root.resources.auth.provider.SocialAuthenticationProvider.lambda$authenticate$0(SocialAuthenticationProvider.java:132)
```

wajb (approved):

wajb:
Tested successfully on 4.4 with steps outlined in JIRA.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/SocialAuthenticationProvider.java
leleueri: you are right should be in debug for this case IMO

PR #6646, opened 2025-10-06
fix: use tokenRequest HttpResponse instead of generate a new one

leleueri:
fixes AM-5744

# description

This is a minor optimization to avoid useless object creation.


sclarkamlab:
Jest tests are passing so should be all good.

PR #6651, opened 2025-10-07
AM-5459: set datasource as highest precedence in mongo provider

sclarkamlab:
AM-5459

This pull request updates the logic in the `MongoAbstractProvider` class to ensure that datasource configuration is prioritized when building the MongoDB client wrapper. The main change is to check for the presence of a datasource before any other client selection logic.

Client selection logic improvements:

* The check for `shouldUseDatasource()` and the corresponding call to `configureDatasourceClient()` have been moved to the beginning of the `buildClientWrapper` method, ensuring that datasource configuration takes precedence over other client options.
* The previous redundant check for `shouldUseDatasource()` at the end of the method has been removed to prevent duplicate logic and streamline the client selection process.

## :memo: Test scenarios 
Same as AM-5459

n/a


FuYu3699 (approved):

wajb (commented):
Looks good (untested).

PR #7392, opened 2026-02-24
feat: add application type filtering to Applications overview (AM-6110)

Johngren:
Allow administrators to filter applications by type (Web, Single-Page App, Native, Agentic Application, Backend to Backend, Resource Server) using a multi-select dropdown in the Applications list. The filter works in conjunction with the existing search functionality and is persisted in URL query parameters for bookmarkability.

Backend: Added type filtering support across the full stack — repository interfaces (MongoDB and JDBC), service layer, and REST API with a new repeatable `type` query parameter. OpenAPI spec updated accordingly.

Frontend: Added Application Type multi-select dropdown, a Clear Filters button (resets both search and type filter), and a loading guard to prevent UI flicker when clearing filters from a no-results state.

## :memo: Test scenarios 

Please see ticket https://gravitee.atlassian.net/browse/AM-6110

image

### DIFF
--- a/docs/mapi/openapi.yaml
+++ b/docs/mapi/openapi.yaml
@@ -1042,6 +1042,26 @@ paths:
         in: query
         schema:
           type: string
+      - name: type
+        in: query
+        description: >-
+          Filter applications by type. Multiple values can be provided
+          to match any of the specified types (OR logic). When omitted,
+          applications of all types are returned.
+        required: false
+        schema:
+          type: array
+          items:
+            type: string
+            enum:
+            - WEB
+            - NATIVE
+            - BROWSER
+            - SERVICE
+            - RESOURCE_SERVER
+            - AGENT
+        style: form
+        explode: true
       responses:
         "200":
           description: List registered applications for a security domain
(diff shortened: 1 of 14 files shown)


docs/mapi/openapi.yaml
```diff
@@ -1042,6 +1042,26 @@ paths:
         in: query
         schema:
           type: string
+      - name: type
+        in: query
+        description: >-
```
wajb: suggestion: if this was added manually it will get overwritten the next time this spec is generated. I believe you can provide descriptions in the context of endpoint parameters using the `@Parameter(description = "...")` annotation. (This would be needed in addition to `@QueryParam("type")`.

gravitee-am-ui/src/app/domain/applications/applications.component.ts
```diff
@@ -27,16 +27,30 @@ import { ApplicationService } from '../../services/application.service';
...
+  @ViewChild('searchInput') searchInput: ElementRef;
+
   applications: any[];
   private searchValue: string;
+  private isLoading = false;
   domainId: string;
   page: any = {};
+  selectedTypes: string[] = [];
+
+  applicationTypes = [
```
wajb: suggestion: it would be nice to see this list abstracted actually. There is a similar one in `ApplicationCreationStep1Component` but also elsewhere. NBD for this improvement though.

PR #7721, opened 2026-04-10
AM-4872 client id plus

sclarkamlab:

AM-4872

### 1. Support `+` character in client_id for OAuth2 authentication

When a `client_id` contains a literal `+` character, authentication fails because `URLDecoder.decode()` treats `+` as a space (per `application/x-www-form-urlencoded` rules). The decoded client_id no longer matches the stored value.

**Root cause:** `ClientAuthHandlerImpl.resolveClient()` called `decodeURIComponent()` on every client_id regardless of source (introduced by #2403 for an unrelated UI flow). For query params (already URL-decoded by Vert.x `getParam()`) and JWT assertions (parsed by Nimbus), this was a double-decode that broke any percent-encoded character, not just `+`.

**Fix:**

- **`ClientAuthHandlerImpl.resolveClient()`**: removed the blanket `decodeURIComponent()` call. The lookup now uses the raw value from `parseClientId()`.
- **`ClientAuthHandlerImpl.parseClientId()`**: percent-decoding is applied inside the Basic auth branch only, the one path where decoding is genuinely needed per RFC 6749 §2.3.1 (Base64-decoding does not URL-decode).
- **`ClientBasicAuthProvider.urlDecode()`**: percent-decoding (preserving literal `+`) for credential comparison after lookup.

The `+` to `%2B` preservation is kept because most clients do not actually form-url-encode their credentials before Base64-encoding, so `+` is almost always literal in practice.

`login_hint` is unaffected, it flows through a separate code path (`UsernameHelper.escapeUsernameParam`) which already has its own `+` handling (#7889 & #10469).

**`client_secret_post` note:** For POST-based auth, `+` in form bodies means space per the form-urlencoded spec. Clients must send `%2B` to represent a literal `+`. This is correct behaviour and no server-side fix is needed, Vert.x decodes `%2B` to `+` before our code sees it.

### 2. Display user profile fields with falsy values

`additionalInformation` fields with falsy values (`false`, `0`, `""`) were hidden in the user profile UI because the `*ngIf` used a truthy check. Changed to `!= null` (loose equality) to show all values except `null`/`undefined`.

## :memo: Test scenarios

**Unit tests** (`ClientBasicAuthProviderTest.java`):
- Basic auth with `+` in both client_id and client_secret

**Integration tests** (`client-secrets.jest.spec.ts`):
- Create app with `+` in client_id, authenticate via `client_secret_basic`
- Create app with `+` in client_id, authenticate via `client_secret_post` (percent-encoded `%2B`)
- Domain secret settings reset before tests to avoid expiry side effects

**Manual verification:**
- Local gateway tested with curl, confirmed `client_secret_basic` with literal `+` in Base64 payload succeeds
- Confirmed `client_secret_post` with `%2B` encoding succeeds
- Confirmed `client_secret_post` with raw `+` correctly fails (spec-compliant: `+` means space in form bodies)

## :computer: Screenshots

N/A


leleueri (commented):
Are we confident about this change ? 

What is the impact for the login_hint query parameter which may contain an email address alias like my.address+alias@gmail.com ? We had such kind of issue in the past (maybe not fixed in the right way) and I'm afraid that this change may break something

wajb:
> Are we confident about this change ?
> 
> What is the impact for the login_hint query parameter which may contain an email address alias like [my.address+alias@gmail.com](mailto:my.address+alias@gmail.com) ? We had such kind of issue in the past (maybe not fixed in the right way) and I'm afraid that this change may break something

I agree this approach appears dangerous. The example `curl` should work when any client ID value is passed with proper encoding (e.g. `curl --data-urlencode 'client_id=my-app+test'`). If the problem affects only MongoDB, it's conceivable there's an issue in the repo layer.

sclarkamlab:
> Are we confident about this change ?
> 
> What is the impact for the login_hint query parameter which may contain an email address alias like [my.address+alias@gmail.com](mailto:my.address+alias@gmail.com) ? We had such kind of issue in the past (maybe not fixed in the right way) and I'm afraid that this change may break something

login_hint is not affected — completely separate code path.

- client_id flows through ClientAuthHandlerImpl.resolveClient() →  parseClientId() (token endpoint) — where this fix applies
- login_hint flows through UsernameHelper.escapeUsernameParam() (login/redirect endpoints) — which already has its own + handling for email aliases (#7889 & #10469)

These paths never intersect. Verified live against a local gateway with + in client_id for both client_secret_basic and client_secret_post.

wajb:
> > Are we confident about this change ?
> > What is the impact for the login_hint query parameter which may contain an email address alias like [my.address+alias@gmail.com](mailto:my.address+alias@gmail.com) ? We had such kind of issue in the past (maybe not fixed in the right way) and I'm afraid that this change may break something
> 
> login_hint is not affected — completely separate code path.
> 
> * client_id flows through ClientAuthHandlerImpl.resolveClient() →  parseClientId() (token endpoint) — where this fix applies
> * login_hint flows through UsernameHelper.escapeUsernameParam() (login/redirect endpoints) — which already has its own + handling for email aliases (#7889 & #10469)
> 
> These paths never intersect. Verified live against a local gateway with + in client_id for both client_secret_basic and client_secret_post.

I think the issue might be a little broader and relates to the `decodeURIComponent()` applied to the client_id that was introduced some time ago (https://github.com/gravitee-io/gravitee-access-management/pull/2403). I'd say it's a bug that calling the token endpoint is doubly decoding the client_id. Making a targeted exception for the `+` character doesn't solve the general problem unfortunately.

sclarkamlab:
Good point, you're right that the `+` patch was treating a symptom. I've pushed a refactor in 7a572be that addresses the root cause.

**Before:** `resolveClient()` called `decodeURIComponent()` on every `client_id` regardless of source. For query params (already decoded by Vert.x `getParam()`) and JWT assertions (parsed by Nimbus), this was a double-decode, breaking any percent-encoded character, not just `+`.

**After:** Removed the blanket decode from `resolveClient()`. Moved percent-decoding into `parseClientId()` for the Basic auth branch only, the one path where decoding is genuinely needed per RFC 6749 §2.3.1 (Base64-decoding doesn't URL-decode).

The `+` to `%2B` preservation is kept inside that Basic auth branch since most clients don't actually form-url-encode their credentials before Base64-encoding (so `+` is almost always literal in practice).

Net result:
- Query param path: no double-decode (fixes the general percent-encoding issue you flagged)
- JWT assertion path: no double-decode
- Basic auth path: decoded once, with literal `+` preserved

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
```diff
@@ -427,6 +427,82 @@ describe('Usage expired secrets', () => {
...
+      name: 'plus-client',
+      type: 'SERVICE',
+      clientId: 'my+client+id',
+      redirectUris: ['https://callback'],
+    });
+    expect(plusApp).toBeDefined();
+    expect(plusApp.settings.oauth.clientId).toEqual('my+client+id');
+    plusAppSecret = plusApp.settings.oauth.clientSecret;
+    expect(plusAppSecret).toBeDefined();
+    await delay(6000);
```
lgw-gravitee: q: is it needed? its in the end of a test

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
```diff
@@ -427,6 +427,82 @@ describe('Usage expired secrets', () => {
...
+  it('should reset domain secret settings before + tests', async () => {
+    // Previous tests leave secretSettings with a 5s expiry, which would cause
+    // secrets created here to expire before we can authenticate with them.
+    domain = await patchDomain(domain.id, accessToken, {
+      secretSettings: {
+        enabled: false,
+        expiryTimeSeconds: 0,
+      },
+    });
+  });
```
lgw-gravitee: this test does nothing - no assertions

lgw-gravitee: you can add innerBlock beforeAll()

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
```diff
@@ -427,6 +427,82 @@ describe('Usage expired secrets', () => {
...
+    });
+    const postSecret = postApp.settings.oauth.clientSecret;
+    await patchApplication(domain.id, accessToken, {
+      settings: {
+        oauth: {
+          tokenEndpointAuthMethod: 'client_secret_post',
+        },
+      },
+    }, postApp.id);
+    await delay(6000);
```
lgw-gravitee: q: can we use DomainState poller here?

lgw-gravitee (dismissed):

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
```diff
@@ -427,6 +427,82 @@ describe('Usage expired secrets', () => {
...
+      redirectUris: ['https://callback'],
+    });
+    expect(plusApp).toBeDefined();
+    expect(plusApp.settings.oauth.clientId).toEqual('my+client+id');
+    plusAppSecret = plusApp.settings.oauth.clientSecret;
+    expect(plusAppSecret).toBeDefined();
+    await delay(6000);
+  });
+
+  it('should authenticate with Basic auth when client_id contains +', async () => {
```
wajb: suggestion: can we add a test case where the client id and secret are uri-encoded (the proper OAuth 2.0 standard) for the same credentials?

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
```diff
@@ -427,6 +427,82 @@ describe('Usage expired secrets', () => {
   });
 });
 
+describe('AM-4872: client_id containing + character', () => {
```
wajb: thought: these are really covering gateway logic and might be more appropriate in `gravitee-am-test/specs/gateway`, or split if we want to explicitly test application CRUD too.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
```diff
@@ -211,6 +211,10 @@ private void parseClientId(HttpServerRequest request, Handler<AsyncResult<String
                 } else {
                     clientId = clientAuthentication;
                 }
+                // RFC 6749 §2.3.1: credentials in Basic auth are form-url-encoded.
+                // Percent-decode only, preserving literal '+' (most clients do not
+                // form-encode before Base64-encoding, so '+' is almost always literal).
```
wajb: question: where does this assertion come from? It seems odd that most clients would not be OAuth compliant :/

We're url-decoding the secret too so there must be some assumption that these are generally encoded I would have thought 🤔

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
```diff
@@ -211,6 +211,10 @@ private void parseClientId(HttpServerRequest request, Handler<AsyncResult<String
...
+                clientId = decodeURIComponent(clientId.replace("+", "%2B"));
```
wajb: suggestion: if we reuse the `ClientBasicAuthProvider.urlDecode` implementation we can help with consistency and maintainability, and benefit from the fallback handling logic here too.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: Claude adding over the top comments, removed it now

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: Done

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
sclarkamlab: We dont have domain state in 4.8 version

gravitee-am-test/specs/management/client-secrets.jest.spec.ts
sclarkamlab: moved to a different spec

wajb (approved):

PR #7745, opened 2026-04-16
Agent Identity

sclarkamlab:

[GMA-47](https://gravitee.atlassian.net/browse/GMA-47) — `USER_EMBEDDED` persona foundation
[GMA-48](https://gravitee.atlassian.net/browse/GMA-48) — `HOSTED_DELEGATED` persona + JWKS + `jwt-bearer`
[GMA-49](https://gravitee.atlassian.net/browse/GMA-49) — `AUTONOMOUS` persona + token-exchange `act` chain

## :jigsaw: Data model at a glance

A **Blueprint** is a normal `Application` row with `type=AGENT` and `agentSettings` populated — there is no separate Blueprint entity. It is the trust anchor for N ephemeral **agent instances** that are **not persisted** anywhere in AM; they exist only inside the signed JWT client assertion they present. At request time the gateway builds a **runtime `Client`** by cloning the blueprint (its `clientId` is preserved for downstream `act` claim emission and audit) and recording the assertion's `sub` on a transient `agentInstanceId` field.

```mermaid
erDiagram
    APPLICATION ||--o| AGENT_SETTINGS : "has (when type=AGENT)"
    APPLICATION ||--o| OAUTH_SETTINGS : "has"
    OAUTH_SETTINGS ||--o{ JWK : "holds N public keys"
    APPLICATION ||--o{ AGENT_INSTANCE : "trusts (not persisted)"
    AGENT_INSTANCE ||--|| JWK : "holds private pair of"
    RUNTIME_CLIENT }o--|| APPLICATION : "cloned from (request-scoped)"
    RUNTIME_CLIENT }o--o| AGENT_INSTANCE : "represents (for agent flows)"

    APPLICATION {
        string id PK
        string clientId "stable, pre-registered"
        string type "STANDARD | AGENT | ..."
        bool persisted "yes"
    }
    AGENT_SETTINGS {
        enum agentType "USER_EMBEDDED | HOSTED_DELEGATED | AUTONOMOUS"
        int maxPublicKeysPerWorkload
    }
    OAUTH_SETTINGS {
        string tokenEndpointAuthMethod
        json jwks "reused for private_key_jwt AND agent assertions"
    }
    JWK {
        string kid PK "must be unique"
        string kty "must be asymmetric"
    }
    AGENT_INSTANCE {
        string subClaim "opaque, operator-chosen"
        string kid "matches a JWK in blueprint"
        bool persisted "NO - exists only in signed JWT"
    }
    RUNTIME_CLIENT {
        string clientId "= blueprint clientId (used for act claim)"
        string agentInstanceId "= sub of assertion (transient)"
        bool persisted "NO - request-scoped clone"
    }
```

Key points that tend to come up in review:

- **Blueprint ≠ Template.** Templates are creation-time factories; blueprints are live runtime clients authenticating N instances. They share "protected, governance-bearing" semantics but are otherwise unrelated. Captured in the new `AgentType` javadoc.
- **No workload identity provisioning.** AM never stores per-instance rows. The blueprint's JWKS is the complete set of trusted instances; removing a `kid` revokes every instance holding that key.
- **JWKS is shared with `private_key_jwt`** — there is no parallel "agent keys" store.
- **Agent instance identity = `sub` claim** on the signed assertion, carried through to the issued access token's `client_id`, with the blueprint id placed in the `act.sub` claim (RFC 8693).

Delivers the Blueprint Agent Identity foundation across all three personas (GMA-47/48/49).

**Model & persistence**
- New `AgentSettings` and `AgentType` (`USER_EMBEDDED`, `HOSTED_DELEGATED`, `AUTONOMOUS`) on `ApplicationAdvancedSettings`, plus an `agentIdentityMode` flag.
- MongoDB persistence mappings (`AgentSettingsMongo`, `ApplicationAdvancedSettingsMongo`, `ApplicationSettingsMongo`).
- JDBC persistence works via existing `applications.settings` JSON column — no schema change for the main row; new fields round-trip through Jackson.
- New **`agent_applications` side table** (JDBC) holding `application_id` (PK) + `agent_type`, with indexes on both columns. Populated on application create/update when `agentIdentityMode=true`, cleaned on delete. Used to efficiently filter agent-flagged applications without scanning the JSON settings blob. MongoDB uses an equivalent `$match` on `settings.advanced.agentIdentityMode`. Liquibase changelog: `4.12.0-agent-applications-table.yml`.
- `PatchApplicationSettings` / `PatchApplicationAdvancedSettings` / `NewApplication` updated.

**Management API**
- New read-only endpoint **`GET /organizations/{org}/environments/{env}/domains/{domain}/applications/agents`** — lists applications with `agentIdentityMode=true` backed by the `agent_applications` side table. Supports `page`, `size`, `q` (name search). Returns the full `Application` payload so the UI can render agent-specific projection fields (`settings.advanced.agentIdentityMode`, `settings.agent.agentType`, `settings.oauth.clientId`). Permission-gated on `APPLICATION[LIST]` + per-app `APPLICATION[READ]`.

**Wizard & UI**
- Agent tile with persona selector in the New Application wizard (`step1`, `step2`).
- Dedicated Agent Identity settings screen (`agent.component.*`) exposing `agentType` (read-only) and agent-specific settings. Grant types / token TTLs / JWKS / client-auth method remain on the existing OAuth surfaces — no parallel surface.
- `AuthGuard` gates agent-specific menu items on `agentIdentityMode`.
- Applications list shows and filters by agent identity mode.
- CIMD settings hint text updated to reflect that the template ID may reference a blueprint agent.

**Gateway (auth & tokens)**
- Agent client assertions routed via RFC 7523 `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`; `ClientBasicAuthProvider.canHandle` now defers to `ClientAssertionAuthProvider` when `client_assertion_type` is `jwt-bearer`.
- New `JWT_BEARER` entry on `ClientAuthenticationMethod`.
- `ClientAssertionServiceImpl` / `ClientAssertionAuthProvider` extended for the agent jwt-bearer assertion shape; blueprint JWKS used as trust anchor.
- Agent jwt-bearer assertions are validated with audience pinned to the AM token endpoint only (no base-issuer fallback), and with the same `allowedClockSkew=0` semantics as `DefaultJWTParser` — aligning with the standard RFC 7523 path.
- `ClientAuthHandlerImpl` always attempts a direct `findByClientId` lookup first, including for URL-shaped client ids, so pre-registered blueprints with canonical URL client ids resolve without needing CIMD enabled.
- `TokenServiceImpl`: injects `act` claim on access tokens (RFC 8693, `sub`-only per spec) and preserves the `act` chain on token exchange.
- `IDTokenServiceImpl`: emits `act` and `client_profile` on ID tokens for blueprint agents.
- Dynamic Client Registration **rejects** agent identity applications — `DynamicClientRegistrationServiceImpl.validateAgentConstraints` errors on any DCR request whose `software_id` resolves to an agent blueprint. Blueprints must be created through the Management API.

**Management / service**
- `ApplicationServiceImpl` enforces per-persona defaults (grant types, client type inference, redirect-URI rules) and maintains the `agent_applications` side table on create/update/delete.
- `DomainServiceImpl` updates + covering test.
- `AgentAuditBuilder` and updates to `ClientAuthAuditBuilder` / `ClientTokenAuditBuilder` for agent flows.
- `AgentType` / `AgentSettings` / `agentIdentityMode` / `listAgentApplications` surfaced in `openapi.yaml`.

## :warning: Known limitations (v1) — follow-ups

Tracked as follow-up stories rather than in this PR:

- **Per-instance revocation.** The only way to invalidate an instance is to remove its `kid` from the blueprint JWKS, which invalidates every holder of that private key. No per-instance denylist or `jti` tracking in v1.
- **`kid` ↔ `sub` binding strength.** Any process holding any registered key can claim any `sub` under the same blueprint; private-key custody is the only enforcement.

## :memo: Test scenarios

**Unit**
- `ApplicationRepositoryTest#testAgentSettingsRoundTrip` and side-table coverage (Mongo + JDBC parity via the shared repository suite).
- `DomainRepositoryTest` round-trip additions.
- `ClientAssertionAuthProviderTest`, `ClientAssertionServiceTest`, `TokenServiceTest`, `IDTokenServiceTest` — extended for agent flows and `act` claim.
- `DynamicClientRegistrationServiceTest#create_rejectsAgentIdentityReference` — covers the new DCR rejection.
- `AgentAuditBuilderTest`, `DomainServiceTest`.
- `AgentApplicationsResourceTest` — REST layer coverage for the new list endpoint.

**Jest — gateway**
- `specs/gateway/blueprint-agents/flows/agent-auth-errors.jest.spec.ts`
- `specs/gateway/blueprint-agents/flows/autonomous-client-creds-token-exchange.jest.spec.ts`
- `specs/gateway/blueprint-agents/flows/hosted-delegated-workload-jwt.jest.spec.ts`
- `specs/gateway/blueprint-agents/flows/user-embedded-pkce.jest.spec.ts`
- `specs/gateway/blueprint-agents/workload-jwt-assertion.jest.spec.ts` (+ `fixtures/workload-jwt-fixture.ts`)
- `specs/gateway/mfa/fixtures/mfa-type-mismatch-fixture.ts` (fixture-only addition)

**Jest — management**
- `specs/management/blueprint-application/blueprint-application.jest.spec.ts` (+ `fixtures/blueprint-fixture.ts`).
- `specs/management/agent-applications/agent-applications.jest.spec.ts` (+ `fixtures/agent-applications-fixture.ts`) — covers filtering, `q` search, pagination, empty-domain, and unauthenticated-access for `GET /applications/agents`.

Integration suites exercised against `mongo` and `REPOSITORY_TYPE=jdbc`.

_Blueprint agent creation wizard, Agent Identity settings screen — see attached screenshots._
01-new-app-select-agent-type
02-agent-subtype-selector

- `AgentType` values map to the three GMA personas:
  - `USER_EMBEDDED` (GMA-47): public client, PKCE + S256 forced, `act` claim on access tokens.
  - `HOSTED_DELEGATED` (GMA-48): confidential client, `authorization_code` + `client_credentials`, optional `private_key_jwt` / `jwt-bearer` assertion.
  - `AUTONOMOUS` (GMA-49): confidential client, `client_credentials` + RFC 8693 token exchange; redirect URIs forbidden.
- Agent client authentication uses the RFC 7523 `urn:ietf:params:oauth:client-assertion-type:jwt-bearer` assertion type; blueprint JWKS is reused from `ApplicationOAuthSettings.jwks` rather than a parallel agent-keys surface.
- `agentType` is immutable after creation; PATCH attempts are rejected.
- JDBC: new `agent_applications` side table created by Liquibase changelog `4.12.0-agent-applications-table.yml`. Main application fields still live in the existing JSON settings column; the side table is an index-only projection to drive the new list endpoint.
- **Out of scope for this PR** (tracked separately): CIMD fetcher / URI-based `iss` resolution (GMA-50), SAML encrypted assertions, BOM bumps, `client_id` + percent-encoding (AM-4872), SERVICE-app flow-designer restriction. Follow-up refactors: per-instance revocation, `kid`↔`sub` binding strength (see Known limitations above).

@gravitee-io/am-team

[GMA-47]: https://gravitee.atlassian.net/browse/GMA-47
[GMA-48]: https://gravitee.atlassian.net/browse/GMA-48
[GMA-49]: https://gravitee.atlassian.net/browse/GMA-49


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
```diff
@@ -175,6 +177,12 @@ private void resolveClient(HttpServerRequest request, Handler<AsyncResult<Client
                 handler.handle(Future.succeededFuture());
                 return;
             }
+            // For CIMD: when client_id is a URI (metadata document), skip the DB lookup.
+            // The ClientAssertionAuthProvider will handle resolution via CIMD metadata fetch.
+            if (isUri(clientId)) {
```
lgw-gravitee: q: do we need to test if CIMD is enabled? AFAIK it should be possible to register client with url-like id

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/Claims.java
```diff
@@ -145,6 +145,13 @@ public interface Claims {
      */
     String ACT = "act";
 
+    /**
+     * draft-mora-oauth-entity-profiles-00 - Client profile claim used to advertise
+     * the operating profile of the requesting client (e.g. agent profile).
+     * See <a href="https://datatracker.ietf.org/doc/html/draft-mora-oauth-entity-profiles-00#name-client_profile-claim-2">draft-mora § client_profile</a>
```
wajb: suggestion: can we reference latest version `https://datatracker.ietf.org/doc/html/draft-mora-oauth-entity-profiles-01`?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
wajb: Agreed. It is a requirement of our CIMD implementation that you can preregister client ids that are canonical URLs, so this lookup should still happen.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
```diff
@@ -215,7 +227,17 @@ private void parseClientId(HttpServerRequest request, Handler<AsyncResult<String
                 clientId = ClientBasicAuthProvider.urlDecode(clientId);
             } else if(clientAssertion != null && clientAssertionType != null) {
                 JWT jwt = JWTParser.parse(clientAssertion);
-                clientId = jwt.getJWTClaimsSet().getSubject();
+                String iss = jwt.getJWTClaimsSet().getIssuer();
+                String sub = jwt.getJWTClaimsSet().getSubject();
+                // For blueprint-agent jwt-bearer assertions the blueprint client_id is in
+                // "iss" (not "sub") because "sub" carries the agent instance ID which is
+                // not a registered client. Detected by shape: iss is a URI (CIMD) or
+                // iss != sub. Standard RFC 7523 assertions have iss == sub == client_id.
```
wajb, line 238: thought: is using `iss` as the client just conforming to [JWT spec](https://datatracker.ietf.org/doc/html/rfc7523#section-3)? I don't see this specified in CIMD at all, unless the implication is the CIMD source server is the issuer? 🤔

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/provider/ClientBasicAuthProvider.java
```diff
@@ -57,6 +58,9 @@ public ClientBasicAuthProvider(SecretService appSecretService) {
 
     @Override
     public boolean canHandle(Client client, RoutingContext context) {
+        if (ClientAuthenticationMethod.JWT_BEARER.equals(context.request().getParam(Parameters.CLIENT_ASSERTION_TYPE))) {
```
wajb, line 61: aside: the code docs for `Parameters.CLIENT_ASSERTION_TYPE` say it will be an absolute URI. Are these docs just wrong?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
```diff
@@ -247,12 +289,140 @@ private static boolean verifyJws(Client client, SignedJWT signedJWT) throws JOSE
...
+            String sub = claims.getSubject();
+            List<String> aud = claims.getAudience();
+            Date exp = claims.getExpirationTime();
+
+            if (iss == null || iss.isEmpty() || sub == null || sub.isEmpty()
+                    || aud == null || aud.isEmpty() || exp == null) {
+                return Maybe.error(NOT_VALID);
+            }
+
+            if (exp.toInstant().isBefore(Instant.now().minusSeconds(AGENT_ASSERTION_CLOCK_SKEW_SECONDS))) {
```
wajb: question: is this special leniency for agent assertions a requirement? FWIW the `DefaultJWTParser` has `allowedClockSkewMillis = 0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
```diff
@@ -247,12 +289,140 @@ private static boolean verifyJws(Client client, SignedJWT signedJWT) throws JOSE
...
+                return Maybe.error(new InvalidClientException("assertion has expired"));
+            }
+
+            OpenIDProviderMetadata discovery = openIDDiscoveryService.getConfiguration(basePath);
+            if (discovery == null || discovery.getTokenEndpoint() == null) {
```
wajb: minor: the discovery model and the token endpoint can never be null, it's just calculated from the `basePath`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
```diff
@@ -247,12 +289,140 @@ private static boolean verifyJws(Client client, SignedJWT signedJWT) throws JOSE
...
+                return Maybe.error(new ServerErrorException("Unable to retrieve discovery token endpoint."));
+            }
+            if (aud.stream().noneMatch(discovery.getTokenEndpoint()::equals)
+                    && (discovery.getIssuer() == null || aud.stream().noneMatch(discovery.getIssuer()::equals))) {
```
wajb: issue: the proposal states this should always be the AM token endpoint. Please check if allowing the base issuer is correct. (Note that `getIssuer()` here never returns null.)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
```diff
@@ -247,12 +289,140 @@ private static boolean verifyJws(Client client, SignedJWT signedJWT) throws JOSE
...
+                    if (!jwsService.isValidSignature(signedJWT, jwk)) {
+                        return Maybe.error(unableToValidateClientException());
+                    }
+                    return Maybe.just(blueprint);
+                });
+    }
+
+    private static Client buildAgentClient(Client blueprint, String agentInstanceId) {
+        Client agentClient = new Client(blueprint);
+        agentClient.setBlueprintClientId(blueprint.getClientId());
```
wajb: question: could you describe how this is used? From what I can see the blueprint client id will always match the client id 🤔. Equally, the agent instance id comes from the sub claim, which can be the same value.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/clientregistration/impl/DynamicClientRegistrationServiceImpl.java
```diff
@@ -961,6 +962,59 @@ private Single<DynamicClientRegistrationRequest> validateCibaSettings(DynamicCli
...
     }
 
+    /**
+     * When the DCR request includes grant types that don't require redirect_uri
+     * (e.g., client_credentials only for autonomous agents), ensure we don't
+     * reject the request for missing redirect_uri. The redirect_uri validation
+     * in validateRedirectUri already handles this — this method validates
+     * additional agent-specific constraints if agent metadata extensions are present.
+     */
+    private Single<DynamicClientRegistrationRequest> validateAgentConstraints(DynamicClientRegistrationRequest request) {
```
wajb, line 977: thought: do we want to support DCR at all for agent registrations? It's a little at odds with CIMD and might need further consideration. We should probably ignore CIMD settings here altogether if we want this functionality IMO. The pre-registered blueprint app means CIMD is skipped right?

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -916,8 +915,8 @@ private Completable validateCIMDSettings(Domain domain) {
             return Completable.error(new InvalidDomainException("templateId must be provided when Client ID Metadata Document is enabled"));
         }
 
-        return applicationService.findById(templateId).filter(Application::isTemplate)
-                .switchIfEmpty(Maybe.error(new InvalidDomainException("templateId must be a valid application id configured as template"))).ignoreElement();
+        return applicationService.findById(templateId).filter(app -> app.isTemplate() || app.isAgentIdentityMode())
```
wajb, line 918: question: should an agent identity app/client be treated as a template so it's protected in the same way? I'm not sure I follow the distinction in some places.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
sclarkamlab: No, there is no special requirement here, I will look at the DefaultJWTParse, thanks

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/Claims.java
sclarkamlab: Done

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/ClientAssertionServiceImpl.java
sclarkamlab: Good catch `buildAgentClient` does `new Client(blueprint)` which already copies `clientId`, then we overwrite `blueprintClientId` with the same value, so on the cloned agent client the two fields are always equal. The `act.sub` injection in `TokenServiceImpl` / `IDTokenServiceImpl` can just read `client.getClientId()` directly.

I'll drop `blueprintClientId` from `Client` and simplify the token/ID-token paths. `agentInstanceId` stays — it carries the `sub` from the assertion and isn't redundant with anything else.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: The `iss` → client_id mapping isn't CIMD-specific — RFC 7523 §3 says a client-auth assertion MUST carry the client_id in `iss`. For agent assertions we're deliberately diverging from that: `iss` = blueprint client_id, `sub` = agent instance id. The URI-shape OR `iss != sub` check here is just a heuristic to tell agent assertions apart from standard RFC 7523 ones at the pre-lookup stage.

CIMD only enters at the next step (`clientLookupService.findByClientId(iss)`), where a URL-shaped `iss` resolves via the domain's CIMD metadata flow. Non-URL ids fall through to normal lookup.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/provider/ClientBasicAuthProvider.java
sclarkamlab: Docs are correct. `ClientAuthenticationMethod.JWT_BEARER = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer"` — that's an absolute URI (urn: scheme, per RFC 3986 §4.3). The equality check here works because both sides are the same canonical URN string.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/clientregistration/impl/DynamicClientRegistrationServiceImpl.java
sclarkamlab: Agreed that DCR shouldn't mint agent blueprints — that's what `validateAgentConstraints` enforces: any DCR whose `software_id` resolves to an agent-identity app is rejected with `InvalidClientMetadataException`. Blueprints are Management-API-only.

CIMD settings on the domain are orthogonal to this path — they gate client *lookup* (URL-shaped client_id → metadata fetch), not registration. DCR can't produce a blueprint regardless of CIMD state. So nothing to change here unless you're seeing a gap I'm missing?

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: The similarity is intentional at the CIMD config layer: `templateId` must resolve to either a template or a blueprint because both are pre-registered, protected clients the domain uses as a source-of-truth for resolved clients. They're distinct concepts elsewhere though — a template is a *factory* (seeds new apps at creation time), a blueprint is a *live* client that authenticates N agent instances at runtime.

Happy to unify the protection semantics behind a single predicate (e.g. `Application.isProtected()` covering both) if that'd make the distinction clearer across the codebase. WDYT?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: Good question — we don't need a CIMD-enabled gate at this layer. The router logic uses URI-shape OR `iss != sub` to pick the `iss` branch. A URL-like client_id registered on a non-CIMD domain takes that branch and is handed to `clientLookupService.findByClientId(iss)`, which does a direct lookup when CIMD is off and a CIMD-aware lookup when it's on. Both paths work, so the auth handler stays CIMD-agnostic.

sclarkamlab:
Merged as it is just targeting gamma for now, will get overall review before merging to master

PR #7768, opened 2026-04-21
fix: user profile not accessible in the context during token exchange

lgw-gravitee:
fixes [AM-6873]

[AM-6873]: https://gravitee.atlassian.net/browse/AM-6873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/grant/impl/TokenExchangeStrategy.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/grant/impl/TokenExchangeStrategy.java
@@ -15,17 +15,21 @@
  */
 package io.gravitee.am.gateway.handler.oauth2.service.grant.impl;
 
+import io.gravitee.am.common.jwt.Claims;
 import io.gravitee.am.common.oauth2.GrantType;
-import io.gravitee.am.gateway.handler.common.user.UserGatewayService;
 import io.gravitee.am.gateway.handler.oauth2.service.grant.GrantStrategy;
 import io.gravitee.am.gateway.handler.oauth2.service.grant.TokenCreationRequest;
 import io.gravitee.am.gateway.handler.oauth2.service.request.TokenRequest;
+import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.ActorTokenInfo;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.TokenExchangeService;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.oidc.Client;
 import io.reactivex.rxjava3.core.Single;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
 
 /**
  * Strategy for RFC 8693 OAuth 2.0 Token Exchange.
@@ -39,11 +43,9 @@ public class TokenExchangeStrategy implements GrantStrategy {
     private static final Logger LOGGER = LoggerFactory.getLogger(TokenExchangeStrategy.class);
 
     private final TokenExchangeService tokenExchangeService;
-    private final UserGatewayService userGatewayService;
 
-    public TokenExchangeStrategy(TokenExchangeService tokenExchangeService, UserGatewayService userGatewayService) {
+    public TokenExchangeStrategy(TokenExchangeService tokenExchangeService) {
         this.tokenExchangeService = tokenExchangeService;
-        this.userGatewayService = userGatewayService;
     }
 
     @Override
@@ -69,7 +71,7 @@ public class TokenExchangeStrategy implements GrantStrategy {
     public Single<TokenCreationRequest> process(TokenRequest request, Client client, Domain domain) {
         LOGGER.debug("Processing token exchange request for client: {}", client.getClientId());
 
-        return tokenExchangeService.exchange(request, client, domain, userGatewayService)
+        return tokenExchangeService.exchange(request, client, domain)
                 .doOnSuccess(result -> LOGGER.debug("Token exchange successful for subject: {}", result.user().getId()))
                 .map(result -> TokenCreationRequest.forTokenExchange(
                         request,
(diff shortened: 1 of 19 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/grant/impl/TokenExchangeStrategy.java
```diff
@@ -69,19 +71,56 @@ public boolean supports(String grantType, Client client, Domain domain) {
...
+                            result.subjectTokenType(),
+                            result.actorTokenId(),
+                            result.actorTokenType(),
+                            result.actorInfo(),
+                            result.jtisOfParents()
+                    );
+
+                    Map<String, Object> tokenExchangeExecutionContext = buildTokenExchangeExecutionContext(result.actorInfo());
+                    if (!tokenExchangeExecutionContext.isEmpty()) {
+                        tokenCreationRequest = tokenCreationRequest.withExecutionContext(tokenExchangeExecutionContext);
```
wajb: suggestion: `TokenCreationRequest.forTokenExchange` is only ever used here, so passing the execution context directly to that factory method makes more sense than instantiating two objects in a wither pattern IMO. It doesn't matter if the map is empty either.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/SubjectTokenUserResolver.java
```diff
@@ -0,0 +1,98 @@
...
+                user.setSource(subjectManager.extractSourceId(gioInternalSubValue));
+                user.setExternalId(subjectManager.extractUserId(gioInternalSubValue));
+            }
+        }
+        user.setAdditionalInformation(additionalInformation);
+
+        return Maybe.just(user);
+    }
+
+    private Maybe<User> fetchTheUser(ValidatedToken subjectToken) {
```
wajb: nitpick: `findUser` (mirrored by a `synthesizeUser` above) would be more conventional.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/SubjectTokenUserResolver.java
```diff
@@ -0,0 +1,98 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+@RequiredArgsConstructor
+@Slf4j
+public class SubjectTokenUserResolver implements TokenExchangeUserResolver {
```
wajb: thought: this class is more or less token-agnostic and could be named in more generalised terms IMO. The only thing tying it to subject tokens is the appending of the `subject_token_id` attribute. Could this be moved to `TokenExchangeServiceImpl.enrichUser` instead?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/TokenUserResolver.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        JWT subjectJwt = new JWT();
+        subjectJwt.setSub(subjectToken.getSubject());
+        Object internalSub = subjectToken.getClaim(Claims.GIO_INTERNAL_SUB);
+        if (internalSub instanceof String internalSubValue) {
+            subjectJwt.setInternalSub(internalSubValue);
+        }
+
+        return subjectManager.findUserBySub(subjectJwt)
```
wajb, line 85: issue: when the subject token does not have a valid internal sub (for example when it was generated with `client_credentials`) I see an error logged and the custom claims aren't generated.

```
14:49:41 14:49:41.825 [vert.x-eventloop-thread-3] [] ERROR i.g.a.g.h.m.subject.SubjectManagerV2 - malformed internal sub value '{sub=1234567890}'
```

This is probably fine since it falls back to synthesising the user, but perhaps we could check for the `hasValidInternalSub` to at least avoid the noise? Or is that messy because of the differences in SubjectManagerV1/SubjectManagerV2?

lgw-gravitee: I cannot  test "hasValidInternalSub" before, as V1 returns false by default.
I moved verbose message to the exception and removed log.error - the error should not be swallowed and printed at the end.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/SubjectTokenUserResolver.java
lgw-gravitee: nice1

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/grant/impl/TokenExchangeStrategy.java
```diff
@@ -81,7 +83,36 @@ public Single<TokenCreationRequest> process(TokenRequest request, Client client,
...
+        if (actorInfo.hasActorTokenActClaim()) {
+            actorContext.put("actor_token_act", actorInfo.actorTokenActClaim());
+        }
+
+        if (actorInfo.hasClaims()) {
+            actorContext.put("actor_token_claims", actorInfo.claims());
+        }
+
+        return Map.of("token_exchange", Map.of("actor", actorContext));
+    }
```
leleueri: question: could it be a method in the TokenCreationRequest record ? Looks like all the information are coming from the actorInfo, nothing comes from the outside

wajb (approved):

PR #7770, opened 2026-04-23
feat: add Java SDK module for the Management API

sclarkamlab:
Adds `gravitee-am-management-api-sdk-java`, a new Maven module that generates a Java client for the Access Management Management API from `docs/mapi/openapi.yaml` via openapi-generator (`java` / `okhttp-gson`). The module is self-contained: it applies local pre- and post-generation regex patches to work around two spec/generator quirks that strictly-typed Java clients reject (discriminator enum mismatch on the `New/Update ProtectedResourceFeature` subclasses and a redundant `String` assignment in generator-emitted subclass constructors). Upstream Java source and the committed OpenAPI spec are left untouched. Generated sources are not committed — the reactor regenerates them on every build, and consumers depend on the published artifact.

- New module `gravitee-am-management-api-sdk-java` added to the root reactor (`pom.xml`).
- `pom.xml` wires `openapi-generator-maven-plugin` at `generate-sources`: input = `docs/mapi/openapi.yaml`, generator = `java`, library = `okhttp-gson`.
- Pre-generation patch: rewrites discriminator enum values for the `New/Update ProtectedResourceFeature` subclasses so the generated Java discriminator matches the schema.
- Post-generation patch: strips a redundant `String` assignment that openapi-generator emits in subclass constructors (doesn't compile under strict settings).
- License-header check skipped for generated sources.
- `README.md` documents module purpose, generator configuration, and the patch rationale.
- `.gitignore` excludes the generated `target/` output — generated sources are not committed; consumers depend on the published artifact.
- No changes to the OpenAPI spec or any other module.

## :memo: Test scenarios

- `mvn -pl gravitee-am-management-api-sdk-java -am clean install` succeeds from a clean checkout (verifies generation, patches, compilation).
- Snapshot publish path: on merge to `master`, `publish_bundle` (`mvn deploy -P gio-artifactory-snapshot`) and `deploy-snapshot-on-nexus` push the SDK jar to Gravitee Artifactory / Nexus snapshots — no extra CI wiring required.
- Release publish path: release workflow (`gio_action=release`) deploys via `gio-artifactory-release,gio-release`; Maven Central via `publish_maven_central`.

N/A — no UI changes.

- See module `README.md` for generator configuration and the two patches applied.
- The patches are regex-based and narrowly scoped; if the underlying spec/generator issues are fixed upstream, the patches should be removed.

### DIFF
--- a/gravitee-am-management-api-sdk-java/pom.xml
+++ b/gravitee-am-management-api-sdk-java/pom.xml
@@ -1,0 +1,232 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.gravitee.am</groupId>
+        <artifactId>gravitee-am-parent</artifactId>
+        <version>4.12.0-alpha.3-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>gravitee-am-management-api-sdk-java</artifactId>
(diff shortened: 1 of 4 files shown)


podlesrafal (approved):

gravitee-am-management-api-sdk-java/pom.xml
```diff
@@ -0,0 +1,232 @@
...
+        </dependency>
+        <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <version>${jakarta.annotation-api.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
```
wajb, line 77: thought: while fine for an initial feature, it would be worthwhile to look into avoiding these collisions earlier on (in the API or the spec) so the same patterns don't show up in other schemas. Are you able to consume the generated SDK versioned from this branch initially?

sclarkamlab: Sorry I literally merged this as your comment came through.

I did initially update the java code so that we would not need to actually have this but it was changing production code which I wanted to avoid for the purpose of this.

I will raise a technical task to "fix" the underlying issue and we should be able to remove this horrible pom logic

sclarkamlab: Follow on tech item: https://gravitee.atlassian.net/browse/AM-6944

PR #7794, opened 2026-04-29
fix(idp): domain whitelist should be case insensitive for users's ema…

lgw-gravitee:
How to test:
1. create Domain, Application
2..Setup OIDC provider for Google graviteesource.com with whitelist GraviteeSourCE.com
3. Enable Provider on Application
4. Login 
5. Confirm User is created 
6. Confirm User's email to be xxx@graviteesource.com

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/UsersDomainWhitelistValidator.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/UsersDomainWhitelistValidator.java
@@ -1,0 +1,62 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/auth/provider/UsersDomainWhitelistValidator.java
```diff
@@ -0,0 +1,67 @@
...
+        return Single.just(endUser);
+    }
+
+    private static boolean isUsernameWhitelisted(List<String> domainWhitelist, String username) {
+        return nonNull(username) && isDomainWhitelisted(username, domainWhitelist);
+    }
+
+    private static boolean isEmailWhitelisted(List<String> domainWhitelist, String email) {
+        return nonNull(email) && isDomainWhitelisted(email, domainWhitelist);
+    }
```
wajb: nitpick: IMO these methods are redundant. We could simply carry out the `nonNull` check in the `isDomainWhitelisted` method....In the same place, `domainWhitelist` could also be checked, simplifying the condition in the "public" method to two clauses. Existing logic I know :)

wajb (commented):
For consistency, should we also consider the comparisons in `VirtualHostValidatorImpl.validateDomainVhosts`?

lgw-gravitee:
@wajb for me its out of scope. I will create a ticket for that is that ok?

wajb (approved):

PR #7807, opened 2026-04-30
test: AM-6249 add coverage for mutual-TLS authentication

wajb:
[AM-6249](https://gravitee.atlassian.net/browse/AM-6249)

Add test coverage for mutual-TLS OAuth 2.0 authentication methods `tls_client_auth` and `self_signed_tls_client_auth`.

## :memo: Test scenarios 
n/a

n/a

https://datatracker.ietf.org/doc/html/rfc8705

@gravitee-io/am

[AM-6249]: https://gravitee.atlassian.net/browse/AM-6249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/api/config/dev.setup.js
+++ b/gravitee-am-test/api/config/dev.setup.js
@@ -35,5 +35,6 @@ process.env.INTERNAL_FAKE_SMTP_PORT = '5025';
 process.env.SFR_URL = 'http://localhost:8181';
 process.env.INTERNAL_SFR_URL = 'http://wiremock:8080';
 process.env.AM_GATEWAY_SYNC_GRACE_PERIOD = '5000';
+process.env.MTLS_URL = 'https://localhost:1443/am';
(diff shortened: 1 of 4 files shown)


lgw-gravitee:
The header-only approach makes sense for isolating the RFC 8705 auth-method logic (especially the "no cert" / "wrong DN" negatives, which couldn't pass through gateway-mtls's ssl_verify_client on). 

Is there value in a follow-up test that also goes through the gateway-mtls localstack component (port 1443) with the CA-signed certs from gravitee-am-test/certs/client-a/, to cover the full nginx-terminated mTLS -> X-Client-Cert forward -> AM path end-to-end? Would catch regressions in the edge config too, complementary rather than replacement.

lgw-gravitee (approved):

PR #7808, opened 2026-04-30
fix: new sessions should be MFA challenged by ACTIVATED factors (POST fix)

lgw-gravitee:
In 4.9+ MFAChallangeEndpoint was refactored into MFAChallangeEndpoint and MFAChallangePostEndpoint.
Changes made in 4.8 (a previous fix) were applied only on MFAChallangeEndpoint thus on the POST, the DEACTIVATED factor was picked up.
 
The goal of this PR is to make common, abstract class where the same methods can be used in Get and Post ChallangeEndpoint class (as well as getFactor())

How to test:
Perform the same test like last time but try to login one more time and challange the SMS code

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengePostEndpointTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengePostEndpointTest.java
@@ -17,7 +17,6 @@ package io.gravitee.am.gateway.handler.root.resources.endpoint.mfa;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.gravitee.am.common.audit.EventType;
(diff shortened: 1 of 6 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengePostEndpointTest.java
```diff
@@ -308,7 +302,7 @@ public void shouldVerify_withEmail_tidRemovedFromSession() throws Exception {
         int attempt = 20;
         while (attempt > 0) {
             try {
-                assertFalse(spyRoutingContext.session().data().containsKey(MFAChallengeEndpoint.PREVIOUS_TRANSACTION_ID_KEY));
+                assertFalse(spyRoutingContext.session().data().containsKey(MFAChallengeGetEndpoint.PREVIOUS_TRANSACTION_ID_KEY));
```
wajb: nitpick: can revert this to avoid referencing unrelated derived class.

wajb (approved):

PR #7828, opened 2026-05-07
GMA-172 - feat: CIMD Application create toggle option

sclarkamlab:
[GMA-172](https://gravitee.atlassian.net/browse/GMA-172)

Adds a **CIMD bootstrap path** to the application creation wizard. When CIMD is enabled on a domain, users get a Manual / CIMD toggle on step 2; in CIMD mode the user supplies only a Client Identity Metadata Document URL, and the wizard fetches and previews the document before creating the application.

### Backend

- **`CimdMetadataFetcher`** (`gravitee-am-service`) — new component that fetches and validates a CIMD URL using Vert.x `WebClient` (built via the shared `WebClientBuilder` so HTTP traffic flows through the same stack as the rest of the platform). Mirrors the gateway-runtime trust + validation rules: scheme/host check, allowed-domain match, private-IP guard (literal + DNS resolution), bounded response body, configurable timeout, RFC 7591 metadata validation (required `redirect_uris`, forbidden `client_secret` / `client_secret_expires_at`, forbidden secret-based `token_endpoint_auth_method`, `private_key_jwt` requires `jwks` or `jwks_uri`).
- **`CimdMetadataDocumentService.fetchAndValidate(domain, url)`** — wraps the fetcher and returns a `CimdPreview`.
- **`ApplicationService.createFromCimd(...)`** — builds the `Application` from the preview (clientId is forced to the URL), runs through the existing `create0` path, and upserts the `CimdMetadataDocument` to pre-warm the gateway cache.
- **`CimdResource`** — exposes `POST /domains/{id}/cimd/validate` and `POST /domains/{id}/cimd/applications`, wired into `DomainResource`.
- **`NewCimdApplication`** payload (`name`, `type`, `cimdUrl`, optional `clientName`).

### Frontend (`gravitee-am-ui`)

- Manual / CIMD toggle on step 2, gated on `domain.oidc.cimdSettings.enabled`.
- CIMD mode is a single field — the **Document URL** — placed on its own at the top of the step.
- After server-side validation the wizard advances to a new **CIMD confirm** step that renders the parsed metadata read-only. The application name is auto-populated from the document's `client_name`; if the document does not provide one, the confirm step shows an editable Client Name input that mirrors into `application.name`.
- Editing the URL after a successful validation clears the cached preview, so we never create against a stale fetch.
- Toggling Manual ↔ CIMD resets transient state (redirect_uri / client_id / client_secret on one side; cimdUrl / cimdPreview / cimdClientName / name on the other).

### Tests

- `CimdMetadataFetcherTest` — 10 tests covering happy path, missing/forbidden fields, redirect_uris validation, JSON shape failures, 4xx responses, disabled domain, non-http URL, host-not-in-allowed-domains, response-size limit.
- `cimd-application-creation.jest.spec.ts` — integration spec using the existing wiremock CIMD stub: validate happy path, 4xx for unreachable URL, end-to-end app creation with `clientId` equal to the URL.

## :memo: Test scenarios

A companion repo ([am-tester `feat/cimd-harness`](https://github.com/gravitee-io-labs/am-tester/tree/feat/cimd-harness)) hosts a catalogue of **32 static CIMD JSON fixtures** (20 valid, 12 invalid) served at `http://host.docker.internal:1111/cimd/<name>.json`, with a listing page at `/cimd-fixtures.html` for one-click "Copy URL" against AM. Coverage:

- **Validation rules (negative)** — missing/empty/blank/non-string `redirect_uris`, top-level non-object, forbidden `client_secret` and `client_secret_expires_at`, all three forbidden auth methods (`client_secret_basic`/`_post`/`_jwt`), `private_key_jwt` without keys / with non-object `jwks` / with blank `jwks_uri`.
- **Validation rules (positive)** — minimal, RFC 7591 kitchen sink, `private_key_jwt` with inline `jwks` and with `jwks_uri`.
- **Preview surface** — populated logo + branding, multiple redirect URIs, document with no `client_name`, document with empty-string `client_name`, document declaring a `client_id` ≠ URL.
- **Grant / response type variants** — `client_credentials`-only, hybrid flow (`code id_token token`), implicit-only, native app (custom-scheme + RFC 8252 loopback).
- **Advanced features** — DPoP, mTLS (`tls_client_auth`), CIBA poll mode, RFC 7591 §2.2 localized claims (`#fr` / `#de` / `#ja`), pairwise subject + `sector_identifier_uri`, OIDC front- and back-channel logout.

### Manual smoke

1. Enable CIMD on the dev domain (Settings → CIMD → enable, Allowed domains: include `host.docker.internal` or leave empty).
2. Run am-tester (`feat/cimd-harness`): `docker compose up -d` from `am-tester/`.
3. Browse to `http://localhost:1111/cimd-fixtures.html`, copy any URL.
4. Applications → Create → pick any app type → step 2: switch to CIMD, paste URL, Next. Confirm step shows parsed metadata; Create.

### Out of scope / follow-ups

- OpenAPI spec entries for the two new endpoints (additive, no breaking changes — SDK can be regenerated when management API runs locally).
- Angular unit tests for the new components were not added; behaviour is covered end-to-end by the Jest spec.
- The product spec's "user can set client_id when missing" was intentionally not honoured — `client_id` is always the URL to keep the app/doc cache-key alignment with the gateway runtime.

image

image

The CIMD validator is intentionally duplicated between gateway runtime (`CimdMetadataServiceImpl` / `CimdUriTrustValidator`) and management API (`CimdMetadataFetcher`). The latter is reactive on the same `WebClient` stack but lives in `gravitee-am-service` because the management API does not have a per-domain `WebClient`.

[GMA-172]: https://gravitee.atlassian.net/browse/GMA-172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-service/src/main/java/io/gravitee/am/service/cimd/CimdMetadataFetcher.java
```diff
@@ -0,0 +1,356 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeoutException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Fetches and validates a CIMD (Client Identity Metadata Document) URL from the management API
+ * context, returning a {@link CimdPreview} suitable for surfacing in the application creation UI.
+ *
+ * <p>Mirrors the gateway-runtime trust + validation rules in
```
wajb, line 59: discussion: you say the duplication is intentional but it's risky for several reasons. What is the barrier to abstracting the logic so it is reusable? The WebClient and Domain can be parameterised for the central logic from what I can see, then the `CimdMetadataDocumentManager` + `CimdLogoCacheService` could be replaced with a strategy for domain-and-logo-caching vs API-metadata-db-only 🤔

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -398,6 +404,81 @@ public Single<Application> create(Domain domain, Application application) {
...
+            oAuthSettings.setScopeSettings(preview.scopes().stream()
+                    .map(scope -> {
+                        ApplicationScopeSettings s = new ApplicationScopeSettings();
+                        s.setScope(scope);
+                        return s;
+                    })
+                    .toList());
+        }
+        applicationSettings.setOauth(oAuthSettings);
+        application.setSettings(applicationSettings);
```
wajb, line 501: question: do we need to care about default SAML 2.0 settings? (Set for regular applications in `create()`.)

gravitee-am-service/src/main/java/io/gravitee/am/service/CimdMetadataDocumentService.java
```diff
@@ -40,4 +41,11 @@ public interface CimdMetadataDocumentService {
     Single<CimdMetadataDocument> upsert(Domain domain, String clientId, String metadataJson, Duration ttl);
 
     Completable delete(String domainId, String clientId);
+
+    /**
+     * Fetches the CIMD document at {@code url} (server-side, with trust + size + timeout enforcement)
+     * and returns a parsed preview suitable for surfacing in the application creation UI. Does not
+     * persist the document.
+     */
+    Single<CimdPreview> fetchAndValidate(Domain domain, String url);
```
wajb: thought: this does not feel like a relevant concern. Cohesively, the implementation manages records and interacting with the repo layer; ephemeral previews and remote URLs don't sit well here IMO. Can the fetcher service be used directly by consumers?

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DomainResource.java
```diff
@@ -222,6 +222,11 @@ public ApplicationsResource getApplicationsResource() {
         return resourceContext.getResource(ApplicationsResource.class);
     }
 
+    @Path("cimd")
```
wajb, line 225: issue: can you regenerate the OpenAPI spec?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -398,6 +404,81 @@ public Single<Application> create(Domain domain, Application application) {
                 });
     }
 
+    @Override
+    public Single<Application> createFromCimd(Domain domain, NewCimdApplication newApplication, User principal) {
+        LOGGER.debug("Create a new CIMD-bootstrapped application {} for domain {}", newApplication.getName(), domain.getId());
+        return cimdMetadataDocumentService.fetchAndValidate(domain, newApplication.getCimdUrl())
+                .flatMap(preview -> {
```
wajb, line 418: minor: in this context, the CIMD client data isn't really a preview AFAICT, since it's used to actually construct the application. It does beg the question about related parameter and class names in general though.

gravitee-am-service/src/main/java/io/gravitee/am/service/cimd/CimdMetadataFetcher.java
sclarkamlab: Good push. I've pulled the most-likely-to-drift bit out into [`CimdValidationRules`](https://github.com/gravitee-io/gravitee-access-management/blob/GMA-172-cimd-app-create/gravitee-am-service/src/main/java/io/gravitee/am/service/cimd/CimdValidationRules.java) (shared `FORBIDDEN_SECRET_BASED_AUTH_METHODS`) so the gateway and mgmt rule sets can't fork. The full strategy refactor (per-domain WebClient + cache + logo strategy on the gateway side vs single shared WebClient + db-only on mgmt) is a meaningful piece of work that I'd rather scope into its own PR — happy to open a follow-up issue tracked separately so this PR doesn't grow another ~1k lines of churn.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Good catch — done in `f6e9998e`. The CIMD app builder now mirrors the regular `create()` SAML defaults (`entityId` from the redirect URI's root, plus `attributeConsumeServiceUrl`), and silently skips when the redirect URI is a custom-scheme deep link.

gravitee-am-service/src/main/java/io/gravitee/am/service/CimdMetadataDocumentService.java
sclarkamlab: Agreed. Removed `fetchAndValidate` from `CimdMetadataDocumentService` in `f6e9998e`. `ApplicationServiceImpl` and `CimdResource` now inject `CimdMetadataFetcher` directly; the document service only handles persistence and event publishing.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/DomainResource.java
sclarkamlab: Done — regenerated in `f9c1f8bd` via `scripts/regen-oas.sh`. The two new endpoints (`POST /domains/{id}/cimd/validate`, `POST /domains/{id}/cimd/applications`) plus their request/response schemas land in `docs/mapi/openapi.yaml`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Renamed to `CimdClientMetadata` in `f6e9998e`. Agree the parsed metadata is the actual abstraction — it's the input to both the validate/preview and create paths, so 'preview' was misleading. The wire DTO at the REST layer keeps the name `CimdValidationResponse` since that's specifically the validate-endpoint response shape.

gravitee-am-service/src/main/java/io/gravitee/am/service/cimd/CimdMetadataFetcher.java
wajb: Yes you're right, I think this larger refactor will bloat the intent here, plus the gateway CIMD service is still evolving. We should create a ticket to track this though and commit to it as soon as it's feasible.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
wajb: The `CimdValidationResponse` isn't wired up as the response schema for the endpoint. Can you add that?
We have to be careful this is the right shape to prevent compatibility issues in the future...would it make sense to expose a nested or base metadata object, compliant with https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#client-metadata, that defines only the metadata properties we require, plus `additionalProperties: true`? That part at least should not change 🤔.

gravitee-am-service/src/main/java/io/gravitee/am/service/cimd/CimdMetadataFetcher.java
sclarkamlab: https://gravitee.atlassian.net/browse/AM-7007

wajb (approved):

PR #7829, opened 2026-05-07
fix: avoid single eventbus address for reporter

leleueri:
this change create one address per domain/organization so when an event an audit is published

 only relevant reporter instances are used

 this change avoid high CPU usage when the number of domain is important

fixes AM-6989

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/vertx/EventBusReporterWrapper.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/vertx/EventBusReporterWrapper.java
@@ -24,7 +24,6 @@ import io.gravitee.am.model.common.event.Payload;
 import io.gravitee.am.reporter.api.Reportable;
 import io.gravitee.am.reporter.api.provider.ReportableCriteria;
 import io.gravitee.am.reporter.api.provider.Reporter;
-import io.gravitee.am.service.reporter.impl.AuditReporterVerticle;
 import io.gravitee.common.component.Lifecycle;
 import io.reactivex.rxjava3.core.Maybe;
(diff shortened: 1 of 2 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/vertx/EventBusReporterWrapper.java
```diff
@@ -54,7 +55,7 @@ public class EventBusReporterWrapper<R extends Reportable,C extends ReportableCr
     public static final Logger logger = LoggerFactory.getLogger(EventBusReporterWrapper.class);
     private final Vertx vertx;
     private final Reporter<R,C> reporter;
-    private MessageConsumer<Reportable> messageConsumer;
+    private final List<MessageConsumer<Reportable>> messageConsumers = new ArrayList<>();
     private Set<Reference> referenceFilter;
```
lgw-gravitee, line 61: q: shouldnt it be ConcurrentArrayList?

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/vertx/EventBusReporterWrapper.java
```diff
@@ -124,9 +125,12 @@ public Lifecycle.State lifecycleState() {
         return reporter.lifecycleState();
     }
 
+    private static String addressFor(Reference reference) {
```
wajb: suggestion: could move this to `AuditReporterVerticle` and import the method instead of just the prefix. Then reuse it there too for better integrity.

lgw-gravitee (approved):

wajb (approved):

PR #7813, opened 2026-05-01
GMA-144

sclarkamlab:
## :id: Reference related issue.

[GMA-144](https://gravitee.atlassian.net/browse/GMA-144)

## :pencil2: A description of the changes proposed in the pull request

Reworks the agent identity feature to model agent applications as a first-class `ApplicationType.AGENT` instead of a separate side table + dedicated REST resource.

### Backend
- **Unified Application model**: agent applications are regular `Application` rows with `type=AGENT` and an `AgentSettings` profile (`USER_EMBEDDED`, `HOSTED_DELEGATED`, etc.). Drops the `agent_applications` side table, the `SpringApplicationAgentRepository`, and the `agentIdentityMode` advanced toggle.
- **Generalized application listing**: `ApplicationsResource.list` accepts an optional `type` query param. `ApplicationService` / `ApplicationRepository` `findByDomain` and `search` overloads take an `ApplicationType` filter, with default-method shims for the existing untyped callers. Removes the dedicated `AgentApplicationsResource` and `findAgentsByDomain` / `searchAgents` methods.
- **Agent JWT bearer client assertion**: introduces `AGENT_JWT_BEARER` (`urn:ietf:params:oauth:client-assertion-type:agent-jwt-bearer`) as a distinct assertion type. `ClientAssertionServiceImpl` routes agent assertions through the new path; standard `JWT_BEARER` is now strictly RFC 7523 (`iss == sub == client_id`). FAPI signing-algorithm checks apply to agent JWTs.
- **Agent constraints**: `USER_EMBEDDED` and `HOSTED_DELEGATED` agents require `redirect_uris`. Agent defaults set `response_types` so `authorization_code` survives validation. DCR rejects agent apps marked as templates and permits blueprint id as `software_id`.
- **CIMD lookup consolidation**: collapses the lookup paths in `CimdAwareClientLookupServiceImpl` and `DefaultClientLookupServiceImpl`, gates `jwt-bearer` on the configured token-endpoint auth method.
- **Token / claims**: emits `client_profile` and `sub_profile` claims, propagates through `act` delegation chains in token exchange, ID token, and audit builders. Drops `AgentAuditBuilder` in favor of `ClientAuthAuditBuilder` / `ClientTokenAuditBuilder`.
[...]

PR #7781, opened 2026-04-27
GMA-115 - feat: SPIFFE JWT verification support

sclarkamlab:
## :id: Reference related issue

[GMA-115](https://gravitee.atlassian.net/browse/GMA-115)

## :pencil2: Summary

Adds first-class support for SPIFFE workload identity as a client authentication method for AM. Workloads present a JWT-SVID issued by their SPIRE/SPIFFE issuer; AM resolves the trust domain, fetches the bundle, verifies the SVID, and authenticates the client.

### Backend

- **Trust Domain model & persistence** — new `TrustDomain` entity scoped to a domain, with name, JWKS URL, allowed algorithms and refresh interval. MongoDB + JDBC repositories.
- **Trust Domain REST API** — CRUD under the management API; service layer enforces validation and emits audit events.
- **JWKS URL safety** — `PrivateAddressGuard` rejects URLs resolving to private/loopback/link-local addresses (configurable per domain via `allowPrivateIpAddress`). Enforced at config time and at fetch time.
- **Trust Bundle service** — caches fetched bundles per trust domain, honours per-trust-domain refresh interval, and serves the last known good bundle on transient fetch errors.
- **`spiffe_jwt` client authentication** — new `ClientAuthenticationMethod.SPIFFE_JWT` and `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-spiffe`. The gateway dispatches SPIFFE assertions to a dedicated path and returns a clear error when a SPIFFE ID is presented on the standard `jwt-bearer` type.
- **`SpiffeJwtSvidValidator`** — validates the SVID per [SPIFFE JWT-SVID spec](https://github.com/spiffe/spiffe/blob/main/standards/JWT-SVID.md): `typ=JWT`, allowed signature algorithms (no `none`/HMAC), `sub` is a SPIFFE ID inside the configured trust domain, exact-match against the client's configured subject, audience contains the token endpoint, and bounded `iat`/`exp`/`nbf` with clock-skew tolerance.
- **Per-application SPIFFE settings** — clients configured for `spiffe_jwt` carry `trustDomain` + exact `subject` (e.g. `spiffe://prod.example/agent/billing`). Validated at create/update time.

### Frontend

- New **Workload Identity** menu under domain settings, grouping SPIFFE domain settings and Trust Domain management screens.
- Trust Domain CRUD UI (list / create / edit) including JWKS URL, allowed algorithms, refresh interval, allow-private-ip toggle.
[...]

PR #7847, opened 2026-05-11
Gamma

sclarkamlab:
Long-lived gamma branch carrying the Agent Identity work on top of master. Just rebased.

## What gamma adds

Agent Identity (GMA-144, GMA-169):
- New `ApplicationType.AGENT` with `Application.subType` replacing the earlier `AgentSettings`
- Blueprint agent model + persistence, DCR support, `act` / `client_profile` / `sub_profile` claims
- Read-only `/applications/agents` endpoint backed by an `agent_applications` side table
- Agent identity UI screens, audit events, and CIMD client metadata document integration

CIMD:
- Agent CIBA flows and template scopes (`scope-intersection`, `grant-types-intersection`, `agent-hosted-delegated-ciba` wiremock fixtures)
- Loopback redirect URIs allowed for URL-shaped `client_id`s per RFC 8252
- Introspection falls back to protected-resource validation when CIMD lookup errors
- Tightened agent metadata, grants, UI menu visibility

SPIFFE / Workload Identity:
- Trust Domain model, persistence, REST API and management UI screens
- Gateway client auth via JWT-SVID, `jwt-spiffe` assertion type
- JWKS URL safety (private-address guard, stale-bundle fallback, per-domain refresh interval)
- SPIRE compose overlay for local-stack, containers run as non-root with agent healthcheck

Build / CI:
- Gamma version line (`4.12.0-gamma.x-SNAPSHOT`) and `release-gamma-version` workflow
- Gitops deploy + Azure registry image push for gamma
- Java SDK pinned to JDK 21 for consumer compatibility, case-insensitive enums, sparse PATCH semantics

Plus the usual PR-review fixes and test coverage backfills.

## Test plan

- Full Jest suite (management + gateway, Mongo and JDBC)
- Local stack with SPIRE overlay for SPIFFE flows
- CIBA agent blueprint spec covers the delegated flow end-to-end

## Previous PRs:
[GMA-169](https://github.com/gravitee-io/gravitee-access-management/pull/7825)
[GMA-144](https://github.com/gravitee-io/gravitee-access-management/pull/7813)
[GMA-115](https://github.com/gravitee-io/gravitee-access-management/pull/7781)
[Agent Identity](https://github.com/gravitee-io/gravitee-access-management/pull/7745)

[GMA-169]: https://gravitee.atlassian.net/browse/GMA-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker/local-stack/dev/spire/bootstrap/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Tiny shell-capable image that carries the spire-server CLI for one-shot
```
graviteeio: ### :warning: Aqua detected misconfiguration in your code  
**Misconfiguration ID:**    
**Check Name:** No HEALTHCHECK defined   
**Severity:** LOW   
**Message:** Add HEALTHCHECK instruction in your Dockerfile  
**Resolution:** Add HEALTHCHECK instruction in Dockerfile  
[This comment was created by Aqua Pipeline]  
  
Read more at https://avd.aquasec.com/misconfig/ds-0026

docker/local-stack/dev/spire/agent/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Shell-capable wrapper around the distroless spire-agent image so the
```
graviteeio: ### :warning: Aqua detected misconfiguration in your code  
**Misconfiguration ID:**    
**Check Name:** Image user should not be 'root'   
**Severity:** HIGH   
**Message:** Specify at least 1 USER command in Dockerfile with non-root user as argument  
**Resolution:** Add 'USER <non root user name>' line to the Dockerfile  
[This comment was created by Aqua Pipeline]  
  
Read more at https://avd.aquasec.com/misconfig/ds-0026

docker/local-stack/dev/spire/bootstrap/Dockerfile
graviteeio: ### :warning: Aqua detected misconfiguration in your code  
**Misconfiguration ID:**    
**Check Name:** Image user should not be 'root'   
**Severity:** HIGH   
**Message:** Specify at least 1 USER command in Dockerfile with non-root user as argument  
**Resolution:** Add 'USER <non root user name>' line to the Dockerfile  
[This comment was created by Aqua Pipeline]  
  
Read more at https://avd.aquasec.com/misconfig/ds-0026

docs/mapi/openapi.yaml
```diff
@@ -17,7 +17,7 @@
 openapi: 3.0.1
 info:
   title: Gravitee.io - Access Management API
-  version: 4.12.0-alpha.3
+  version: 4.12.0-gamma.2
```
leleueri: todo: this will have to be updated before the merge

docs/mapi/openapi.yaml
```diff
@@ -12217,6 +12432,12 @@ components:
             $ref: "#/components/schemas/ClientSecret"
         settings:
           $ref: "#/components/schemas/ApplicationSettings"
+        subType:
```
leleueri: nickpick: I understand the meaning of this subType, but could it be confusing with the "sub_type" in the token which represent the entity profile type of the entity which is used to initialize the "sub" claim?

leleueri: what about:
* category
* variant
* kind
* profile
* entityProfile
* specifier

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/Claims.java
```diff
@@ -145,6 +145,28 @@ public interface Claims {
...
+     * See <a href="https://datatracker.ietf.org/doc/html/draft-mora-oauth-entity-profiles-01#name-client_profile-claim-2">draft-mora § client_profile</a>
+     */
+    String CLIENT_PROFILE = "client_profile";
+
+    /**
+     * draft-mora-oauth-entity-profiles-01 §3.3 registry token identifying an AI agent
+     * client. Used as the prefix portion of {@link #CLIENT_PROFILE} values
+     * (e.g. {@code "ai_agent autonomous"}).
+     */
+    String CLIENT_PROFILE_AI_AGENT = "ai_agent";
```
leleueri: suggestion: as it is a claim value what about moving it in a dedicated structure either an Enum or an interface with constants ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/impl/TrustBundleServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+                if (privateAddr.isPresent()) {
+                    return "jwksUrl host " + host + " resolves to a private/loopback address ("
+                            + privateAddr.get().getHostAddress() + ")";
+                }
+            } catch (UnknownHostException e) {
+                return "jwksUrl host " + host + " could not be resolved";
+            }
+        }
+        return null;
+    }
```
leleueri, line 165: question: don't we have the same kind of check for CIMD ? is there anything we can mutualize ?

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewTrustDomain.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.List;
+
+public class NewTrustDomain {
+    private String name;
+    private String description;
+    private SpiffeBundleSource bundleSource;
+    private String jwksUrl;
+    private Integer refreshIntervalSeconds;
+    private List<String> allowedAlgorithms;
```
leleueri, line 28: question: do we need to specify allowedAlgorithm at domain level if we can override them at application level. Should we remove it here and make the app setting required?

docs/mapi/openapi.yaml
sclarkamlab: `kind` seems like a good alternative

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/impl/TrustBundleServiceImpl.java
sclarkamlab: Yeah I think we can decrease the duplication here

docker/local-stack/dev/spire/bootstrap/Dockerfile
graviteeio: ### :warning: Aqua detected misconfiguration in your code  
**Misconfiguration ID:**    
**Check Name:** No HEALTHCHECK defined   
**Severity:** LOW   
**Message:** Add HEALTHCHECK instruction in your Dockerfile  
**Resolution:** Add HEALTHCHECK instruction in Dockerfile  
[This comment was created by Aqua Pipeline]  
  
Read more at https://avd.aquasec.com/misconfig/ds-0026

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewTrustDomain.java
sclarkamlab: There is currently no capability to override these settings at the applications level.

docs/mapi/openapi.yaml
```diff
@@ -16706,6 +17023,44 @@ components:
...
         persistent:
           type: boolean
+    SpiffeApplicationSettings:
+      type: object
+      properties:
+        subject:
+          type: string
+        trustDomain:
+          type: string
+    SpiffeDomainSettings:
```
leleueri, line 17033: discussion: do we have to introduce Spiffe as domain and app settings? I would go with something like "workload" or "workloadIdentity". For now only spiffe is managed but if we have to manage another "workload identity" protocol, we will be able to introduce a "type" attribute.

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/ClientProfile.java
leleueri, line 1: comment: no need to change it now but, if we introduce this client_profile claim, it could be nice to maange the entity profile also for native, web, spa, service application so all the token will have this claim with the right entity profile. We can also imagine the ai_agent profile concatenated with service (for autonomous agent, or SPA for user embedded) as proposed in the draft specification. something to keep in mind.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -120,15 +131,80 @@ private void parseRedirectUriParameter(RoutingContext context, TokenPurpose oper
...
+
+    private static boolean loopbackMatches(String requested, String registered) {
+        if (!isLoopbackUri(registered)) {
+            LOGGER.info("[redirect-uri-validation] candidate registered={} skipped (not loopback)", registered);
+            return false;
+        }
+        try {
+            URI req = new URI(requested);
+            URI reg = new URI(registered);
+            boolean schemeMatch = Objects.equals(req.getScheme(), reg.getScheme());
```
leleueri, line 182: question: should we also consider the port ? (if so we have to manage the case of default port 80/443 which lay be present in the requested but not in the registered and vice versa)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/spring/CIBAConfiguration.java
```diff
@@ -35,8 +35,8 @@
 public class CIBAConfiguration implements ProtocolConfiguration {
 
     @Bean
-    public ClientAssertionService clientAssertionService() {
-        return new ClientAssertionServiceImpl();
+    public ClientAssertionService clientAssertionService(java.util.List<io.gravitee.am.gateway.handler.oauth2.service.assertion.ClientAssertionValidator> validators) {
```
leleueri: nickpick: add import to avoid absollute class name for ClientAssertionValidator

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
```diff
@@ -175,18 +177,28 @@ private void resolveClient(HttpServerRequest request, Handler<AsyncResult<Client
...
-            // get client - first try regular client, then fallback to protected resource
+
             clientLookupService
                     .findByClientId(clientId)
                     .subscribe(
                             client -> handler.handle(Future.succeededFuture(client)),
                             error -> handler.handle(Future.failedFuture(error)),
-                            () -> handler.handle(Future.failedFuture(new InvalidClientException(ClientAuthHandler.GENERIC_ERROR_MESSAGE)))
+                            () -> {
+                                if (isUri(clientId)) {
```
leleueri: question: not sure to understand why this specific case has been introduced. Can you explain ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/AgentJwtBearerClientAssertionValidator.java
```diff
@@ -0,0 +1,172 @@
...
+        return ClientAuthenticationMethod.AGENT_JWT_BEARER;
+    }
+
+    @Override
+    public Maybe<Client> validate(String assertion, String basePath, String clientIdHint) {
+        return parseJwt(assertion).flatMap(jwt -> validateAgentAssertion(jwt, basePath));
+    }
+
+    private Maybe<Client> validateAgentAssertion(JWT jwt, String basePath) {
+        return Maybe.defer(() -> {
```
leleueri: nickpick: not sure that we need a defer here as the validateAgentAssertion should be executed only if the flatMap is triggered by a non empty result coming from the parseJwt. Relying on Defer may have impact on performances (useless subscribe call)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtAssertionSupport.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+final class JwtAssertionSupport {
+
+    static final InvalidClientException NOT_VALID = new InvalidClientException("assertion is not valid");
+
+    private JwtAssertionSupport() {
+    }
+
+    static Maybe<JWT> parseJwt(String assertion) {
+        return Maybe.defer(() -> {
```
leleueri: question: is it rely useful to defer this ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
```diff
@@ -0,0 +1,271 @@
...
+                    && !isSignAlgCompliantWithFapi(jwt.getHeader().getAlgorithm().getName())) {
+                return Maybe.error(new InvalidClientException("JWT Assertion must be signed with PS256"));
+            }
+
+            return Maybe.just(jwt);
+        });
+    }
+
+    private Maybe<Client> validateSignatureWithPublicKey(JWT jwt) {
+        return Maybe.defer(() -> {
```
leleueri: question: is it rely useful to defer?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
```diff
@@ -0,0 +1,271 @@
...
+        return ClientAuthenticationMethod.JWT_BEARER;
+    }
+
+    @Override
+    public Maybe<Client> validate(String assertion, String basePath, String clientIdHint) {
+        return parseJwt(assertion).flatMap(jwt -> dispatch(jwt, basePath));
+    }
+
+    private Maybe<Client> dispatch(JWT jwt, String basePath) {
+        return Maybe.defer(() -> {
```
leleueri: question: is it rely useful to defer?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
```diff
@@ -0,0 +1,271 @@
...
+                                        return Maybe.just(client);
+                                    }
+                                    return Maybe.error(unableToValidateClient());
+                                });
+                    });
+        });
+    }
+
+    private Maybe<Client> validateSignatureWithHMAC(JWT jwt) {
+        return Maybe.defer(() -> {
```
leleueri: question: is it rely useful to use defer.?

gravitee-am-common/src/main/java/io/gravitee/am/common/oidc/ClientAuthenticationMethod.java
```diff
@@ -73,13 +73,30 @@ public interface ClientAuthenticationMethod {
...
+    /**
+     * URN to identify a ClientAssertion carrying a SPIFFE JWT-SVID.
+     */
+    String JWT_SPIFFE = "urn:ietf:params:oauth:client-assertion-type:jwt-spiffe";
+
+    /**
+     * URN to identify a ClientAssertion issued by an agent instance under a blueprint application.
+     * The JWT {@code iss} is the blueprint {@code client_id}; {@code sub} is the agent instance id.
+     */
+    String AGENT_JWT_BEARER = "urn:ietf:params:oauth:client-assertion-type:agent-jwt-bearer";
```
leleueri: discussion: wondering if this is a good idea to prefix this assertion URN by ietf stuff as it is a specific one. what about `urn:gravitee:params:oauth:client-assertion-type:agent-jwt-bearer` ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
```diff
@@ -0,0 +1,271 @@
...
+                    });
+        });
+    }
+
+    /**
+     * Ensures that all claims required by RFC 7523 are present and valid.
+     * See <a href="https://tools.ietf.org/html/rfc7523#section-3">RFC 7523 §3</a>.
+     */
+    private Maybe<JWT> validateJWT(JWT jwt, String basePath) {
+        return Maybe.defer(() -> {
```
leleueri: don't think we need defer here

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -476,6 +477,33 @@ private JWT createAccessTokenJWT(OAuth2Request request, Client client, User user
...
+            if (client.getAgentType() != null) {
+                act.put(Claims.SUB_PROFILE, client.getAgentType().name().toLowerCase());
+            }
+            jwt.put(Claims.ACT, act);
+        }
+
+        // Agent application - advertise client_profile per draft-mora-oauth-entity-profiles-01 §3.3
+        // Format: "ai_agent <profile_token>" (registry token + space + lowercase sub-profile).
+        if (client.isAgentApplication() && client.getAgentType() != null && jwt.get(Claims.CLIENT_PROFILE) == null) {
+            jwt.put(Claims.CLIENT_PROFILE, ClientProfile.AI_AGENT + " " + client.getAgentType().name().toLowerCase());
```
leleueri, line 494: comment: as mentioned previously, in a near future we may think to apply the "client_profile" for all kind of clients not only agent

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/ActorTokenInfo.java
```diff
@@ -65,6 +66,14 @@ public boolean hasGis() {
         return gis != null;
     }
 
+    /**
+     * Check if the actor token carries a {@code sub_profile} claim (e.g. an agent profile).
+     * Per RFC entity-profiles draft, only profiled actors propagate sub_profile in the act node.
+     */
+    public boolean hasSubProfile() {
+        return subProfile != null;
```
leleueri: suggestion : should we check empty string also using something like StringUtils.hasLength ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/spring/OAuth2Configuration.java
```diff
@@ -166,8 +166,8 @@ public OAuthRevocationTokenService revocationTokenService() {
     }
 
     @Bean
-    public ClientAssertionService clientAssertionService() {
-        return new ClientAssertionServiceImpl();
+    public ClientAssertionService clientAssertionService(java.util.List<io.gravitee.am.gateway.handler.oauth2.service.assertion.ClientAssertionValidator> validators) {
```
leleueri: can we use import instead of using absolute classname ?

docs/mapi/openapi.yaml
sclarkamlab: That seems fair, would we need to consider a plugin approach for this then rather than embedded within AM? We could look at refactoring this for a more generic top level, "Workload Identity" and have a plugin for the SPIFFE implementation? Then as we implement other workload identity flows we can implement those additional plugins?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
sclarkamlab: Yeah worth adding in just to be safe

gravitee-am-common/src/main/java/io/gravitee/am/common/oidc/ClientAuthenticationMethod.java
sclarkamlab: Yeah makes sense to isolate it from the specific ietf oauth params

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: This is the CIMD (Client Identifier Metadata Document) escape hatch. With CIMD enabled, the `client_id` is allowed to be an http(s):// URL pointing at a metadata document instead of a pre-registered ID.

Two shapes need to work:

1. **Blueprint pre-registered** with the URL as its literal clientId. `findByClientId` succeeds, normal flow.
2. **Not pre-registered**, pure CIMD. `findByClientId` returns empty. We must not fail here; we hand off with `succeededFuture(null)` so `ClientAssertionAuthProvider` resolves the client by fetching the CIMD metadata document.

The `isUri` guard exists so non-URL ids (typos, unknown clients) still fail fast with `InvalidClientException` only URL-shaped misses are treated as "maybe this is CIMD, let the assertion provider try".

wajb: Can this be handled earlier if the client isn't needed (combined with the condition on line 176)? We should also add a comment explaining the condition IMO.

Can we also reuse `ClientIds.isUrlShaped`?

.circleci/workflows.yml
```diff
@@ -1199,6 +1199,10 @@ jobs:
         type: boolean
         default: false
         description: "Create maintenance version branch"
+      qualifier_label:
```
wajb: suggestion: `pre_release_qualifier_label` for clarity.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/oauth2/impl/BaseIntrospectionTokenService.java
```diff
@@ -140,6 +141,17 @@ private Single<String> validateAudienceAndGetCertificateId(JWT jwt, String calle
 
     private Single<AudienceMatch> resolveAudience(String domain, String audience) {
         return clientLookupService.findByDomainAndClientId(domain, audience)
+                .onErrorResumeNext(err -> {
+                    // CIMD lookups reject malformed/non-URL audiences with InvalidClientMetadataException;
```
wajb: issue: this isn't true, since a non-URL client id fetch isn't attempted. I'm guessing this happened with a mismatched configuration (unexpected remote resource or template client) but the logs should clarify.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -120,15 +131,80 @@ private void parseRedirectUriParameter(RoutingContext context, TokenPurpose oper
...
+            final CheckMethod checkMethod = getCheckMethod(client, requestedRedirectUri);
             RedirectUriValidator validator = new RedirectUriValidator(checkMethod);
             validator.validate(registeredRedirectUris, requestedRedirectUri, operation);
         }
     }
 
+    private @NonNull CheckMethod getCheckMethod(Client client, String requestedRedirectUri) {
+        final boolean urlShaped = ClientIds.isUrlShaped(client.getClientId());
+        final boolean loopback = isLoopbackUri(requestedRedirectUri);
+        final boolean localhostAllowed = isLocalhostRedirectAllowed();
```
wajb: suggestion: can we use `domain.isRedirectUriLocalhostAllowed()`?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -53,6 +61,9 @@
  */
 public class RedirectUriValidationHandler implements Handler<RoutingContext> {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(RedirectUriValidationHandler.class);
+    private static final Set<String> LOOPBACK_HOSTS = Set.of("localhost", "127.0.0.1", "::1");
```
wajb: issue: normalising and loopback logic should reuse what's in place for `CimdUriTrustValidator` and `UriBuilder` IMO otherwise we'll have decentralised responsibilities and inconsistent edge cases.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/impl/TrustBundleServiceImpl.java
```diff
@@ -0,0 +1,180 @@
...
+ * <ul>
+ *   <li>Soft refresh interval per trust domain (capped by {@link SpiffeDomainSettings#getCacheTtlSeconds()}).
+ *       When an entry is past the soft interval the bundle is re-fetched on the next access.</li>
+ *   <li>Eager refresh on {@code kid} miss: fetch a new bundle without evicting the existing one,
+ *       so a transient fetch failure can still fall back to the previous bundle.</li>
+ *   <li>Stale-on-error: when a refresh fails, the previously cached bundle is returned.
+ *       Hard expiry (a multiple of the soft interval, ≥ 1h) bounds how long stale data is served.</li>
+ * </ul>
+ */
+public class TrustBundleServiceImpl implements TrustBundleService {
```
wajb, line 49: issue: I think the SPIFFE cache settings conflict with - or at least don't take into account - the underlying JWKS cache (gateway process wide). An expired TTL in the SPIFFE settings (300s default) calls through to `jwkService.getKeys()` that uses its own TTL (`jwt.jwks.cache.ttlAfterWriteSeconds:3600`) so will still be stale. This also breaks assumptions in eager refreshing and eviction logic.

`CachedJWKSetFetcher` is fairly new so maybe this is a merge issue but it will need reconsideration.

docs/mapi/openapi.yaml
leleueri: yes maybe we can think about plugin later, for now inner implementation is probably fine.
We may just have to rename the attrobute in the settings object from spiffe to workload (or something else)

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/idtoken/impl/IDTokenServiceImpl.java
```diff
@@ -254,6 +255,30 @@ private IDToken createIDTokenJWT(OAuth2Request oAuth2Request, Client client, Use
...
+        }
+
+        // Agent application - advertise client_profile per draft-mora-oauth-entity-profiles-01 §3.3
+        // Format: "ai_agent <profile_token>" (registry token + space + lowercase sub-profile).
+        // Top-level sub_profile is intentionally NOT set on ID tokens: ID tokens are issued for
+        // an end-user subject, and sub_profile must describe the subject (the user), not the
+        // agent client. The agent profile travels on client_profile and act.sub_profile.
+        if (client.isAgentApplication() && client.getAgentType() != null && idToken.get(Claims.CLIENT_PROFILE) == null) {
+            idToken.put(Claims.CLIENT_PROFILE, ClientProfile.AI_AGENT + " " + client.getAgentType().name().toLowerCase());
+        }
```
leleueri: question: do we have to provide the act claim in the ID_Token 🤔 I don't think. For now, the id_token is only usable by the application/agent on which the user is signing in... not sure that we should have this act claim in it.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -548,11 +592,19 @@ private Application buildCimdApplication(Domain domain, NewCimdApplication newAp
                         s.setScope(scope);
                         return s;
                     })
-                    .toList());
+                    .collect(Collectors.toCollection(ArrayList::new)));
+        } else if (templateScopeSettings != null && !templateScopeSettings.isEmpty()) {
+            // CIMD doc omitted `scope` — inherit from the CIMD template so callers can request the
+            // domain's standard scopes (e.g. `openid`) without a follow-up PATCH.
```
wajb: question: is this actually required? My understanding is that omitted scopes from the metadata results in wholesale adoption of the template app's scopes at client synthesis time.

https://github.com/gravitee-io/gravitee-access-management/blob/2a3b1b5c2860c837cec8e900852057595801f7c7/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/cimd/impl/CimdMetadataServiceImpl.java#L448-L450

leleueri (commented):
TODO: as we are introducing a new permission related to the trusted domain, I think you have to implement an upgrader on the MAPI side to assign the permission to existing XXX_PRIMARY_OWNER and xxx_OWNER.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -572,9 +624,68 @@ private Application buildCimdApplication(Domain domain, NewCimdApplication newAp
         application.setSettings(applicationSettings);
 
         applicationTemplateManager.apply(application);
+
+        // CIMD forbids secret-based client authentication. When the CIMD doc omitted the auth
+        // method, the template manager has just defaulted it to client_secret_basic (WEB/SERVICE).
+        // Replace that with a CIMD-compatible value: private_key_jwt for confidential clients
+        // (SERVICE / client_credentials), none for public clients.
+        ApplicationOAuthSettings finalOauth = application.getSettings().getOauth();
```
wajb: question: again, omission of `token_endpoint_auth_method` in metadata results in "none" as default in the `CimdMetadataServiceImpl`. I don't understand the necessity of this logic. Is this pathway resolving the metadata by some other means?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/idtoken/impl/IDTokenServiceImpl.java
sclarkamlab: OK this has been removed

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
sclarkamlab: Done

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/Claims.java
sclarkamlab: Done in `8c5d9a21`. Pulled `ai_agent` into `ClientProfile`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/spring/CIBAConfiguration.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/AgentJwtBearerClientAssertionValidator.java
sclarkamlab: Done in `469c91b0`, dropped the defer.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtAssertionSupport.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/JwtBearerClientAssertionValidator.java
sclarkamlab: Done in `469c91b0`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/ActorTokenInfo.java
sclarkamlab: Done in `7f88806f`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/spring/OAuth2Configuration.java
sclarkamlab: Done in `469c91b0`.

.circleci/workflows.yml
sclarkamlab: Done in `48ec72c6`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/oauth2/impl/BaseIntrospectionTokenService.java
sclarkamlab: Good catch. Fixed in `26b27b9b`, log now reflects what actually happened.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
sclarkamlab: Done in `bba402b4`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/impl/TrustBundleServiceImpl.java
sclarkamlab: Good catch. Fixed in `cb988151`, SPIFFE path now bypasses the JWKS cache so the SPIFFE TTL is the only thing driving refresh.

docs/mapi/openapi.yaml
sclarkamlab: Done in `a00341f2` (+ `e456d3d7` for the UI wire keys). Renamed to `workloadIdentitySettings`. Plugin split we can revisit later.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Yeah you're right. Dropped in `eb62d73a`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Same, the fetcher already defaults to `none`. Cleaned up in `04e4d339`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/auth/handler/impl/ClientAuthHandlerImpl.java
sclarkamlab: Reused `ClientIds.isUrlShaped` in `b8a7051067`. I'll leave the early-return refactor for a follow-up, the flow around line 176 has a few other branches I don't want to disturb in this PR.

docs/mapi/openapi.yaml
sclarkamlab: Will sort this out before merge.

gravitee-am-common/src/main/java/io/gravitee/am/common/jwt/ClientProfile.java
sclarkamlab: Yeah agreed, makes sense to generalise once the agent profile lands. I'll raise a follow-up ticket so we don't lose it.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
sclarkamlab: Same as above, tracked for a follow-up.

.circleci/workflows.yml
```diff
@@ -924,6 +924,9 @@ jobs:
             az acr run --cmd "acr purge --filter 'am-gateway:master-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
             az acr run --cmd "acr purge --filter 'am-management-api:master-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
             az acr run --cmd "acr purge --filter 'am-management-ui:master-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
+            az acr run --cmd "acr purge --filter 'am-gateway:gamma-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
+            az acr run --cmd "acr purge --filter 'am-management-api:gamma-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
+            az acr run --cmd "acr purge --filter 'am-management-ui:gamma-.*' --untagged --ago 1d --keep 3" --registry graviteeio /dev/null
```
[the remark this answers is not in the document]
sclarkamlab: Won't fix in this PR. Happy to clean up the purge block separately if we add more branches.

docker/local-stack/dev/spire/bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,41 @@
...
+    -socketPath "${SOCK}" \
+    -parentID spiffe://am.local/agent \
+    -spiffeID spiffe://am.local/agent/billing \
+    -selector unix:uid:0 || true
+
+"${SPIRE}" entry create \
+    -socketPath "${SOCK}" \
+    -parentID spiffe://am.local/agent \
+    -spiffeID spiffe://am.local/agent/test/sample \
+    -selector unix:uid:0 || true
```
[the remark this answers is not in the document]
sclarkamlab: Won't fix. This is local-stack bootstrap, the `|| true` is intentional for idempotent re-runs.

sclarkamlab:
@leleueri on the upgrader TODO for the new `DOMAIN_TRUST_DOMAIN` permission, the existing `DefaultRoleUpgrader` and `OrganizationRolesUpgrader` already rebuild role permissions from `Permission.allPermissionAcls(referenceType)`, which picks up the new permission automatically since it's declared with ORG/ENV/DOMAIN reference types. The `upsert` flow then detects the changed permission set and updates the role.

I bumped both upgrader `VERSION`s in `f142335f` so they re-run on existing installs. That should backfill the permission onto existing `*_PRIMARY_OWNER` and `*_OWNER` roles. Does that match what you had in mind, or were you thinking of a dedicated upgrader?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
```diff
@@ -120,15 +130,85 @@ private void parseRedirectUriParameter(RoutingContext context, TokenPurpose oper
...
+            final CheckMethod checkMethod = getCheckMethod(client, requestedRedirectUri);
             RedirectUriValidator validator = new RedirectUriValidator(checkMethod);
             validator.validate(registeredRedirectUris, requestedRedirectUri, operation);
         }
     }
 
+    private @NonNull CheckMethod getCheckMethod(Client client, String requestedRedirectUri) {
+        final boolean urlShaped = ClientIds.isUrlShaped(client.getClientId());
+        final boolean loopback = isLoopbackUri(requestedRedirectUri);
+        final boolean localhostAllowed = domain.isRedirectUriLocalhostAllowed();
```
wajb: nitpick: can short-circuit lookups and parsing effort by moving below the first condition.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -526,14 +543,16 @@ private Application buildCimdApplication(Domain domain, NewCimdApplication newAp
...
         if (preview.grantTypes() != null && !preview.grantTypes().isEmpty()) {
-            oAuthSettings.setGrantTypes(preview.grantTypes());
+            final List<String> filteredGrantTypes = filterUnsupportedCimdGrantTypes(preview, isAgent);
+            if (!filteredGrantTypes.isEmpty()) {
+                oAuthSettings.setGrantTypes(filteredGrantTypes);
+            }
         }
         if (preview.responseTypes() != null && !preview.responseTypes().isEmpty()) {
-            oAuthSettings.setResponseTypes(preview.responseTypes());
+            oAuthSettings.setResponseTypes(filterUnsupportedCimdResponseTypes(preview, isAgent));
```
wajb, line 580: suggestion: can this filtering be applied in the `CimdMetadataFetcher`? That way, the improvement defined in https://gravitee.atlassian.net/browse/AM-7007 can more easily combine the overall consistent requirements.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/handler/common/RedirectUriValidationHandler.java
sclarkamlab: Done in `65792aea`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Yeah agreed, the filter belongs in the fetcher so AM-7007 can build on it. Leaving it here for this PR to keep the diff focused, but I'll raise a follow-up to move it across.

wajb (approved):

PR #7893, opened 2026-05-18
fix: add repository fallback to ClientSyncServiceImpl for immediate D…

leleueri:
…CR client visibility

When multiple Gateway instances are deployed and a client is created via DCR, the originating gateway adds it to its local cache immediately. However, other gateway instances only discover it when their event-polling sync cycle fires (every ~5 seconds by default). During this window, requests for the new client routed to other gateways fail with "client not found".

This fix adds a repository fallback to findById() and findByDomainAndClientId() in ClientSyncServiceImpl. When the in-memory cache misses, the code queries ApplicationRepository directly, then warms the local cache with the result. Disabled clients are filtered out and not cached. The fix is backward-compatible and has zero performance impact on cache hits.

Resolves the distributed synchronization gap without requiring event-push infrastructure or reducing the existing 5-second sync interval.

fixes AM-7020

fixes gravitee-io/issues#11443

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientSyncServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientSyncServiceImpl.java
@@ -17,10 +17,13 @@ package io.gravitee.am.gateway.handler.common.client.impl;
 
 import io.gravitee.am.gateway.handler.common.client.ClientManager;
 import io.gravitee.am.gateway.handler.common.client.ClientSyncService;
+import io.gravitee.am.model.Application;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.oidc.Client;
+import io.gravitee.am.repository.management.api.ApplicationRepository;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientSyncServiceImpl.java
```diff
@@ -42,9 +48,30 @@ public class ClientSyncServiceImpl implements ClientSyncService {
...
+    private ApplicationRepository applicationRepository;
+
     @Override
     public Maybe<Client> findById(String id) {
-        return ofNullable(clientManager.get(id)).map(Maybe::just).orElseGet(Maybe::empty);
+        return ofNullable(clientManager.get(id))
+                .map(Maybe::just)
+                .orElseGet(Maybe::empty)
+                .switchIfEmpty(Maybe.defer(() -> {
+                    if (applicationRepository == null) {
```
wajb: nitpick: cannot be null.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/client/impl/ClientSyncServiceImpl.java
```diff
@@ -58,7 +85,23 @@ public Maybe<Client> findByDomainAndClientId(String domain, String clientId) {
         return fromIterable(clientManager.entities())
                 .filter(client -> !client.isTemplate())
                 .filter(client -> client.getClientId().equals(clientId) && client.getDomain().equals(domain))
-                .firstElement();
+                .firstElement()
+                .switchIfEmpty(Maybe.defer(() -> {
+                    if (applicationRepository == null) {
```
wajb: nitpick: cannot be null.

wajb (approved):

wajb (changes requested):
Test failures.

wajb (approved):

PR #7897, opened 2026-05-19
feat: AM-6780 Automation API

wajb:
[AM-6780](https://gravitee.atlassian.net/browse/AM-6780)

Introduce Automation API.

- New API for idempotent GET/PUT/DELETE operations for Domains, Identity Providers, Certificates and Reporters
- Schemas leverage human-readable `key`s to map to unique internal UUIDs
- Sub-resources expose `default` flag that is immutable to mould into "system" defaults for domains
  - Implicit creation of default identity provider, reporter and certificate for new domains is bypassed
- Updates are handled with a declarative reconciliation approach, meaning changes are made with "desired state" payloads (e.g. omitted entities are removed)
- Dangling references are not strictly validated at creation/modification of domains (and at gateway runtime), allowing for "eventual consistency" between declared entities
- Uses JWT Bearer token authentication
- OpenAPI specification `docs/automation/openapi.yaml` is defined with:
  - Script to regenerate it
  - Staleness and breaking-changes detection hooked into the PR workflow ([example run](https://app.circleci.com/pipelines/github/gravitee-io/gravitee-access-management/30150/workflows/d9a1e8d4-1725-4836-9bd8-f1b4103b075c/jobs/410788))
  - Automatic version update on release
- Models and database schema have been prepared with `managedBy` fields for isolating entities for future read-only usage (not implemented here)

Based on PoC https://github.com/gravitee-io/gravitee-access-management/pull/7575.

### Notable omissions and limitations
- Domain-level CIMD settings are omitted (dependency on `Application`)
- Identity provider `passwordPolicy` is omitted (dependency on `PasswordPolicy`)
- `SCIMSettings.idpSelectionRule` is used verbatim (stores EL that evaluates to IDP ID - there is no lookup from hrid)

## :memo: Test scenarios 
See `gravitee-am-management-api/gravitee-am-management-api-automation/WALKTHROUGH.md` and generated OpenAPI specification `docs/automation/openapi.yaml`.

n/a

n/a

@gravitee-io/am

[AM-6780]: https://gravitee.atlassian.net/browse/AM-6780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/mapi/openapi.yaml
+++ b/docs/mapi/openapi.yaml
@@ -13512,8 +13512,13 @@ components:
           type: array
           items:
             type: string
+        key:
+          type: string
+          readOnly: true
         loginSettings:
           $ref: "#/components/schemas/LoginSettings"
+        managedBy:
+          $ref: "#/components/schemas/ManagedBy"
         master:
           type: boolean
         name:
@@ -14157,6 +14162,11 @@ components:
               type: string
         id:
           type: string
+        key:
+          type: string
+          readOnly: true
+        managedBy:
+          $ref: "#/components/schemas/ManagedBy"
         mappers:
           type: object
           additionalProperties:
@@ -14295,6 +14305,13 @@ components:
           $ref: "#/components/schemas/StepUpAuthenticationSettings"
         stepUpAuthenticationRule:
           type: string
+    ManagedBy:
+      type: string
+      enum:
+      - NONE
+      - TERRAFORM
+      - GKO
+      - AUTOMATION_API
     McpTool:
       type: object
       allOf:
@@ -16584,6 +16601,11 @@ components:
           type: string
         inherited:
           type: boolean
+        key:
+          type: string
+          readOnly: true
+        managedBy:
+          $ref: "#/components/schemas/ManagedBy"
         name:
           type: string
         reference:
--- a/gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/4.12.0-dataplane-cimd-client-states.yml
+++ b/gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/4.12.0-dataplane-cimd-client-states.yml
@@ -17,12 +17,26 @@ databaseChangeLog:
             columnNames: id
             tableName: cimd_client_states
 
-        - createIndex:
-            columns:
-              - column:
-                  name: domain_id
-              - column:
-                  name: client_id
-            indexName: idx_cimd_client_states_domain_client
-            tableName: cimd_client_states
-            unique: true
+        # PostgreSQL: use functional unique index (client_id is too long for a direct index)
+        - sql:
+            dbms: postgresql
+            splitStatements: false
+            sql: "CREATE UNIQUE INDEX uq_cimd_client_states_domain_client ON cimd_client_states(domain_id, md5(client_id))"
+
+        # SQL Server: add persisted computed column using HASHBYTES('SHA2_256', client_id)
+        - sql:
+            dbms: mssql
+            splitStatements: false
+            sql: "ALTER TABLE cimd_client_states ADD client_id_hash AS LOWER(CONVERT(VARCHAR(64), HASHBYTES('SHA2_256', client_id), 2)) PERSISTED"
+
+        # MySQL/MariaDB: add generated column using SHA2(client_id, 256)
+        - sql:
+            dbms: mysql, mariadb
+            splitStatements: false
+            sql: "ALTER TABLE cimd_client_states ADD COLUMN client_id_hash VARCHAR(64) GENERATED ALWAYS AS (SHA2(client_id, 256)) STORED"
+
+        # Unique constraint for MySQL/MariaDB/MSSQL (PostgreSQL uses functional index above)
+        - sql:
+            dbms: mysql, mariadb, mssql
+            splitStatements: false
+            sql: "ALTER TABLE cimd_client_states ADD CONSTRAINT uq_cimd_client_states_domain_client UNIQUE (domain_id, client_id_hash)"
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
@@ -238,7 +238,7 @@ public class UserServiceImpl implements UserService {
         final var rawPassword = user.getPassword();
         // validate user and then check user uniqueness
         return userValidator.validate(user)
-                .andThen(userService.findByUsernameAndSource(user.getUsername(), source).isEmpty()
+                .andThen(Single.defer(() -> userService.findByUsernameAndSource(user.getUsername(), source).isEmpty()
                         .flatMapMaybe(checkUserPresence(user, source))
                         .switchIfEmpty(Single.error(() -> new UserProviderNotFoundException(source)))
                         .flatMap(userProvider -> userProvider.create(convert(user)))
@@ -258,7 +258,7 @@ public class UserServiceImpl implements UserService {
                                 .client(user.getClient())
                                 .principal(principal)
                                 .type(EventType.USER_REGISTERED)
-                                .throwable(throwable))));
+                                .throwable(throwable)))));
     }
 
     private static RegistrationResponse buildRegistrationResponse(Optional<AccountSettings> accountSettings, User user) {
--- a/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/model/AutomationAccountSettings.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/model/AutomationAccountSettings.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.handlers.automation.model;
+
+import io.gravitee.am.model.account.FormField;
+import io.swagger.v3.oas.annotations.media.Schema;
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.List;
+
+/**
+ * Automation API mirror of {@link io.gravitee.am.model.account.AccountSettings}, wrapped
+ * because {@link #defaultIdentityProviderForRegistration} is a cross-resource reference
+ * expressed in the key-only contract: the {@code key} of an identity provider that exists
+ * under this domain (created via the identity-providers endpoints, resolved against the IdP
+ * rows at apply time). A reference to an IdP that does not exist is rejected with {@code 400}.
+ *
+ * @author GraviteeSource Team
+ */
+@Getter
+@Setter
+public class AutomationAccountSettings {
+
+    private boolean inherited = true;
+
+    private boolean loginAttemptsDetectionEnabled;
+    private Integer maxLoginAttempts;
+    private Integer loginAttemptsResetTime;
+    private Integer accountBlockedDuration;
+
+    private boolean sendRecoverAccountEmail;
+    private boolean sendVerifyRegistrationAccountEmail;
+    private boolean completeRegistrationWhenResetPassword;
+
+    private boolean autoLoginAfterRegistration;
+    private String redirectUriAfterRegistration;
+    private boolean dynamicUserRegistration;
+
+    /**
+     * The {@code key} of an identity provider that exists under this domain, to use as the
+     * default for user registration. Resolved against the domain's IdPs at apply time; a
+     * value that does not match any existing IdP is rejected with {@code 400}. May be null.
+     */
+    @Schema(description = "key of an identity provider that exists under this domain")
+    private String defaultIdentityProviderForRegistration;
+
+    private boolean autoLoginAfterResetPassword;
+    private String redirectUriAfterResetPassword;
+    private boolean deletePasswordlessDevicesAfterResetPassword;
+
+    private boolean rememberMe;
+    private Integer rememberMeDuration;
+
+    private boolean resetPasswordCustomForm;
+    private List<FormField> resetPasswordCustomFormFields;
+    private boolean resetPasswordConfirmIdentity;
+    private boolean resetPasswordInvalidateTokens;
+
+    private boolean mfaChallengeAttemptsDetectionEnabled;
+    private Integer mfaChallengeMaxAttempts;
+    private Integer mfaChallengeAttemptsResetTime;
+    private boolean mfaChallengeSendVerifyAlertEmail;
+}
--- a/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationIds.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationIds.java
@@ -1,0 +1,64 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.handlers.automation.resource;
+
+import io.gravitee.am.management.service.impl.DefaultIdentityProviderServiceImpl;
+
+import java.nio.charset.StandardCharsets;
+import java.util.UUID;
+
+/**
+ * Computes the deterministic internal UUID for an Automation-API-managed resource by folding
+ * the resource's external {@code key} with the id of its scope (environment for domains,
+ * domain for identity providers, certificates and reporters). The same {@code (scope, key)} pair
+ * always resolves to the same primary key, which gives idempotent upserts and uniqueness-within-scope
+ * without any database index or constraint.
+ *
+ * @author GraviteeSource Team
+ */
+public final class AutomationIds {
+
+    private AutomationIds() {
+    }
+
+    public static String domainId(String environmentId, String key) {
+        return deterministicId(environmentId, key);
+    }
+
+    public static String identityProviderId(String domainId, String key) {
+        return deterministicId(domainId, key);
+    }
+
+    public static String certificateId(String domainId, String key) {
+        return deterministicId(domainId, key);
+    }
+
+    public static String reporterId(String domainId, String key) {
+        return deterministicId(domainId, key);
+    }
+
+    /**
+     * The conventional id of a domain's system identity provider — the same id the platform assigns to
+     * the built-in default, and the one the gateway falls back to for user registration.
+     */
+    public static String systemIdentityProviderId(String domainId) {
+        return DefaultIdentityProviderServiceImpl.DEFAULT_IDP_PREFIX + domainId.toLowerCase();
+    }
+
+    private static String deterministicId(String scopeId, String key) {
+        return UUID.nameUUIDFromBytes((scopeId + "/" + key).getBytes(StandardCharsets.UTF_8)).toString();
+    }
+}
--- a/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificateResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificateResource.java
@@ -1,0 +1,122 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.handlers.automation.resource;
+
+import io.gravitee.am.management.handlers.automation.mapper.AutomationCertificateMapper;
+import io.gravitee.am.management.handlers.automation.model.AutomationCertificate;
+import io.gravitee.am.management.service.DomainService;
+import io.gravitee.am.model.Acl;
+import io.gravitee.am.model.Certificate;
+import io.gravitee.am.model.Domain;
+import io.gravitee.am.model.ManagedBy;
+import io.gravitee.am.model.permissions.Permission;
+import io.gravitee.am.service.CertificateService;
+import io.gravitee.am.service.exception.CertificateNotFoundException;
+import io.gravitee.am.service.exception.DomainNotFoundException;
+import io.reactivex.rxjava3.core.Maybe;
+import io.reactivex.rxjava3.core.Single;
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import jakarta.ws.rs.DELETE;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.PathParam;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.container.AsyncResponse;
+import jakarta.ws.rs.container.Suspended;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * A single certificate managed under a domain, addressed by its key.
+ *
+ * @author GraviteeSource Team
+ */
+@Tag(name = "Certificates")
+public class CertificateResource extends AbstractAutomationResource {
+
+    @Autowired
+    private DomainService domainService;
+
+    @Autowired
+    private CertificateService certificateService;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "automationGetCertificate", summary = "Get a certificate",
+            description = "Retrieves a single certificate by its key.")
+    @ApiResponse(responseCode = "200", description = "The certificate",
+            content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = AutomationCertificate.class)))
+    public void get(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("certKey") String certKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        checkAnyPermission(principal, organizationId, environmentId, Permission.DOMAIN_CERTIFICATE, Acl.READ)
+                .andThen(resolveDomain(environmentId, domainKey))
+                .flatMap(domain -> resolveCertificate(domain, certKey)
+                        .map(AutomationCertificateMapper::toAutomationCertificate))
+                .subscribe(response::resume, response::resume);
+    }
+
+    @DELETE
+    @Operation(operationId = "automationDeleteCertificate", summary = "Delete a certificate")
+    @ApiResponse(responseCode = "204", description = "Certificate deleted")
+    public void delete(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("certKey") String certKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        checkAnyPermission(principal, organizationId, environmentId, Permission.DOMAIN_CERTIFICATE, Acl.DELETE)
+                .andThen(resolveDomainMaybe(environmentId, domainKey))
+                .flatMap(domain -> certificateService.findByDomain(domain.getId())
+                        .filter(certificate -> certificate.isManagedBy(ManagedBy.AUTOMATION_API) && certKey.equals(certificate.getAutomationKey()))
+                        .firstElement())
+                .flatMapCompletable(certificate -> certificateService.delete(certificate.getId(), principal))
+                .subscribe(() -> response.resume(Response.noContent().build()), response::resume);
+    }
+
+    private Single<Domain> resolveDomain(String environmentId, String domainKey) {
+        return domainService.findById(AutomationIds.domainId(environmentId, domainKey))
+                .switchIfEmpty(Single.error(() -> new DomainNotFoundException(domainKey)))
+                .flatMap(domain -> domain.isManagedBy(ManagedBy.AUTOMATION_API)
+                        ? Single.just(domain)
+                        : Single.error(new DomainNotFoundException(domainKey)));
+    }
+
+    private Maybe<Domain> resolveDomainMaybe(String environmentId, String domainKey) {
+        return domainService.findById(AutomationIds.domainId(environmentId, domainKey))
+                .filter(domain -> domain.isManagedBy(ManagedBy.AUTOMATION_API));
+    }
+
+    private Single<Certificate> resolveCertificate(Domain domain, String certKey) {
+        return certificateService.findByDomain(domain.getId())
+                .filter(certificate -> certificate.isManagedBy(ManagedBy.AUTOMATION_API) && certKey.equals(certificate.getAutomationKey()))
+                .firstElement()
+                .switchIfEmpty(Maybe.error(() -> new CertificateNotFoundException(certKey)))
+                .toSingle();
+    }
+}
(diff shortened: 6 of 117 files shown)


gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/4.12.0-dataplane-cimd-client-states.yml
```diff
@@ -17,12 +17,26 @@ databaseChangeLog:
...
-              - column:
-                  name: client_id
-            indexName: idx_cimd_client_states_domain_client
-            tableName: cimd_client_states
-            unique: true
+        # PostgreSQL: use functional unique index (client_id is too long for a direct index)
+        - sql:
+            dbms: postgresql
+            splitStatements: false
+            sql: "CREATE UNIQUE INDEX uq_cimd_client_states_domain_client ON cimd_client_states(domain_id, md5(client_id))"
```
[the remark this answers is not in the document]
wajb: This replicates the hash index approach used for equivalent columns in `protected_resource_identifiers`.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_12_0/4.12.0-add-cimd-metadata-documents.yml
```diff
@@ -19,12 +19,26 @@ databaseChangeLog:
...
-              - column:
-                  name: client_id
-            indexName: idx_cimd_metadata_documents_domain_client
-            tableName: cimd_metadata_documents
-            unique: true
+        # PostgreSQL: use functional unique index (client_id is too long for a direct index)
+        - sql:
+            dbms: postgresql
+            splitStatements: false
+            sql: "CREATE UNIQUE INDEX uq_cimd_metadata_documents_domain_client ON cimd_metadata_documents(domain_id, md5(client_id))"
```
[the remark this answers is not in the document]
wajb: This replicates the hash index approach used for equivalent columns in `protected_resource_identifiers`.

leleueri (changes requested):
All entities need to expose HRID (even IDP) as this is the way to reference the entities between them (ex reference idp in an application).

Also, currently the domain hrid is not immutable in the domain, so we need to find a way to make it "immutable" (new attribute like automationApi HRID or making the decision to make it immutable for the domain but this is a BC)

Also we need to be able to manage existing entities for that we should schedule a sync with the devex team in order to get their feedback

wajb:
> All entities need to expose HRID (even IDP) as this is the way to reference the entities between them (ex reference idp in an application).
> 
> Also, currently the domain hrid is not immutable in the domain, so we need to find a way to make it "immutable" (new attribute like automationApi HRID or making the decision to make it immutable for the domain but this is a BC)
> 
> Also we need to be able to manage existing entities for that we should schedule a sync with the devex team in order to get their feedback

You're right, there was a gap with IDP hrids. I've updated the schema and mapped by the new field. This also brought to light an issue with the declarative payload requirement, so I brought IDPs back into Domain as embedded entities so that everything can be declared in one PUT.

The hrids of domains and its IDPs are immutable by the design of the API; any change to an hrid means a different entity. As a starting point, changes made in the Automation API are isolated to automation-managed entities so, from this API's perspective, its resources are fair game (protection from mAPI changes is out of scope and part of the follow up read-only story - which can also use the `managedBy` fields in write guards).

Is this heading in the right direction @leleueri? If there is a need for existing entities to be folded into automation management, my thinking is this would be more safely achieved with a separate uniqueness validation + `managedBy` update script.

Some nested entities (e.g. certs) are currently using `name` instead of `hrid` simply to avoid additional columns - but we may have to consider aligning them to reuse consistent pattern constraints. Wdyt?

leleueri:
> > All entities need to expose HRID (even IDP) as this is the way to reference the entities between them (ex reference idp in an application).
> > Also, currently the domain hrid is not immutable in the domain, so we need to find a way to make it "immutable" (new attribute like automationApi HRID or making the decision to make it immutable for the domain but this is a BC)
> > Also we need to be able to manage existing entities for that we should schedule a sync with the devex team in order to get their feedback
> 
> You're right, there was a gap with IDP hrids. I've updated the schema and mapped by the new field. This also brought to light an issue with the declarative payload requirement, so I brought IDPs back into Domain as embedded entities so that everything can be declared in one PUT.
> 
> The hrids of domains and its IDPs are immutable by the design of the API; any change to an hrid means a different entity. As a starting point, changes made in the Automation API are isolated to automation-managed entities so, from this API's perspective, its resources are fair game (protection from mAPI changes is out of scope and part of the follow up read-only story - which can also use the `managedBy` fields in write guards).
> 
> Is this heading in the right direction @leleueri? If there is a need for existing entities to be folded into automation management, my thinking is this would be more safely achieved with a separate uniqueness validation + `managedBy` update script.
> 
> Some nested entities (e.g. certs) are currently using `name` instead of `hrid` simply to avoid additional columns - but we may have to consider aligning them to reuse consistent pattern constraints. Wdyt?

Not sure to understand why the IDPs have to be managed as embedded entities in the domain model 🤔 
I though that the team wanted to avoid this. Certs are embedded because it can cause cyclic dependency for some settings like the SAML handler.

IMO having a dedicated field for the HRID is probably better so all entities will have one field clearly identified as an ID instead of using the name which may contain spaces or special chars. I also don't know if we have to name if HRID as we already have hrid field in the domain model which is serving another purpose (hrid for the domain is the value used to build the GW urls for this domain and it is updated when the name of the domain is updated, here for automation API we want a unique id) - maybe entityId ? 

The entityid/hrid will be used to generate in a predictable manner the internal id of the resource (need to be used in conjonction with the org and env id so the same hird can be used accross different environment)

For existing resources, we will have to provide a way to manage them using the automation API (most probably by providing the current internalId as hrid/entityid and provide to the API an attribute to say "the id is a current internal id").

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/model/AutomationAccountSettings.java
```diff
@@ -0,0 +1,92 @@
...
+    private String redirectUriAfterRegistration;
+    private boolean dynamicUserRegistration;
+
+    /**
+     * The {@code hrid} of an identity provider declared on this domain to use as the
+     * default for user registration. Resolved against the domain's IdPs at apply time;
+     * a value that does not match any declared IdP is rejected with {@code 400}.
+     */
+    @Schema(description = "hrid of an identity provider declared on this domain")
+    private String defaultIdentityProviderForRegistrationHrid;
```
wajb: note: @leleueri this was the reference from domain -> IDP that necessitated combining the resources.

wajb:
> Not sure to understand why the IDPs have to be managed as embedded entities in the domain model 🤔 I though that the team wanted to avoid this. Certs are embedded because it can cause cyclic dependency for some settings like the SAML handler.
> 
> IMO having a dedicated field for the HRID is probably better so all entities will have one field clearly identified as an ID instead of using the name which may contain spaces or special chars. I also don't know if we have to name if HRID as we already have hrid field in the domain model which is serving another purpose (hrid for the domain is the value used to build the GW urls for this domain and it is updated when the name of the domain is updated, here for automation API we want a unique id) - maybe entityId ?
> 
> The entityid/hrid will be used to generate in a predictable manner the internal id of the resource (need to be used in conjonction with the org and env id so the same hird can be used accross different environment)
> 
> For existing resources, we will have to provide a way to manage them using the automation API (most probably by providing the current internalId as hrid/entityid and provide to the API an attribute to say "the id is a current internal id").

I have restructured the API following suggestions and aligning to APIM precedent:
- removed extraneous nested resources from Domain
  - only certificates remain embedded
  - IDPs are managed via their own endpoints under `domain/{domainKey}`
- new fields have been added to Domain, Certificate and IdentityProvider
  - `key` (`automation_key` in jdbc) for driving deterministic IDs (avoids conflation with existing domain `hrid` and certificate `name`)
  - `managedBy` for flagging and isolating resources to Automation API (readonly aspect not implemented - and reportedly soon to be rolled back in APIM)
- prevented all default dependencies when creating a domain - certificate, reporter and IDP

Existing resources CANNOT be managed via the Automation API as it stands. This is consistent with APIM. If there is an additional requirement in AM to support this, can we look into designing a mapping hook in the key -> deterministic UUID transformation? But this should remain out of scope IMO.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
```diff
@@ -0,0 +1,297 @@
...
+        return domain.getManagedBy() == ManagedBy.AUTOMATION_API;
+    }
+
+    static boolean isAutomationManaged(Certificate certificate) {
+        return certificate.getManagedBy() == ManagedBy.AUTOMATION_API;
+    }
+
+    static boolean isAutomationManaged(IdentityProvider idp) {
+        return idp.getManagedBy() == ManagedBy.AUTOMATION_API;
+    }
```
leleueri: suggestion: should be define in the beans a "manageBy(ManageBy)" method so the test is owned by the bean and will not be duplicate in each resource ?

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/internal/model/DomainMongo.java
```diff
@@ -43,6 +43,8 @@ public class DomainMongo extends Auditable {
 
     private String hrid;
 
+    private String key;
```
leleueri: question: should we name it "automationKey" to be consistent with the SQL schema?

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
```diff
@@ -0,0 +1,297 @@
...
+                                checkAnyPermission(principal, organizationId, environmentId, Permission.DOMAIN, Acl.CREATE)
+                                        .andThen(Single.defer(() -> {
+                                            AutomationNewDomain newDomain = new AutomationNewDomain();
+                                            newDomain.setId(domainId);
+                                            newDomain.setKey(key);
+                                            newDomain.setName(definition.getName());
+                                            newDomain.setPath(definition.getPath());
+                                            newDomain.setDescription(definition.getDescription());
+                                            newDomain.setDataPlaneId(definition.getDataPlaneId());
+                                            return domainService.create(organizationId, environmentId, newDomain, principal);
```
leleueri, line 133: issue: I don't think we can reuse the domainService.create like this as we may face issues due to cyclic dependency even by including the cert in the domain resource (check on fallback certificate)

leleueri: with a step back, I wondering if the embedded cert is a good idea. As you spotted, we also have IDP cyclic dependencies. it make me think that we have to create dedicated method or service to create the resource through the automation API to ignore unkwon resources and make sure we have defensive check in the code base when a domain exist but an IDP or Cert is referenced but doesn't exist... 🤔

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/internal/model/DomainMongo.java
wajb: Yes, and it makes sense to rename the core model fields to match too.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
wajb: I agree certificate is the odd one out and moving it to its own resource/endpoints fits the pattern better...We can also create a lighter-weight service/method for domain creation, that makes sense. But what are you referring to with unknown resources? Is this the eventual consistency problem during creation, which, by tolerating unknown resources, causes dangling references at runtime?

So, at creation time, you would expect a fully declarative Domain payload before the dependencies are created - _rather than PUT domain with null refs -> PUT deps -> PUT updated domain_?

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/spring/security/AutomationBearerTokenFilter.java
```diff
@@ -0,0 +1,180 @@
...
+    private String remoteAddress(HttpServletRequest request) {
+        String xForwardedFor = request.getHeader(HttpHeaders.X_FORWARDED_FOR);
+        if (xForwardedFor != null && !xForwardedFor.isEmpty()) {
+            int idx = xForwardedFor.indexOf(',');
+            String remoteAddress = (idx != -1) ? xForwardedFor.substring(0, idx) : xForwardedFor;
+            idx = remoteAddress.indexOf(':');
+            return (idx != -1) ? remoteAddress.substring(0, idx).trim() : remoteAddress.trim();
+        }
+        return request.getRemoteAddr();
+    }
```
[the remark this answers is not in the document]
wajb: This is consistent across mAPI and much of the rest of AM.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificatesResource.java
```diff
@@ -0,0 +1,174 @@
...
+                        if (definition.isDefaultResource() != existing.isSystem()) {
+                            return Single.error(new InvalidParameterException(
+                                    "The 'default' flag is immutable for an existing certificate '" + key
+                                            + "'; delete and recreate it to change it"));
+                        }
+                        Single<Certificate> op = certificateService.update(domain, existing.getId(),
+                                AutomationCertificateMapper.toUpdateCertificate(definition), principal);
+                        return checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_CERTIFICATE, Acl.UPDATE)
+                                .andThen(op)
+                                .map(AutomationCertificateMapper::toAutomationCertificate);
```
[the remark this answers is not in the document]
wajb: RxJava Singles are lazy so I believe permission checks do occur first.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
wajb: Discussed and implemented.

docs/mapi/openapi.yaml
```diff
@@ -13512,8 +13512,18 @@ components:
...
+          readOnly: true
         loginSettings:
           $ref: "#/components/schemas/LoginSettings"
+        managedBy:
+          type: string
+          enum:
+          - NONE
+          - TERRAFORM
+          - GKO
+          - AUTOMATION_API
```
leleueri: question: it there a way to create an enum in the schema so it can be reuse using $ref in each resource ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/service/user/impl/UserServiceImpl.java
```diff
@@ -258,7 +259,31 @@ public Single<RegistrationResponse> register(Client client, User user, io.gravit
...
+        }
+        return identityProviderManager.getUserProvider(configuredSource)
+                .map(userProvider -> configuredSource)
+                .defaultIfEmpty(defaultSource)
+                .doOnSuccess(source -> {
+                    if (!source.equals(configuredSource)) {
+                        logger.warn("Registration identity provider '{}' could not be found for domain '{}'; " +
+                                "falling back to the domain default provider '{}'", configuredSource, domain.getId(), source);
+                    }
+                });
```
leleueri: issue: I don't think we should fallback on default as it is not what the admin may expect and some profiles maybe stored in the wrong place. I'm fine with the UserProviderNotFound exception in that case.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationIds.java
```diff
@@ -0,0 +1,64 @@
...
+        return deterministicId(domainId, key);
+    }
+
+    public static String certificateId(String domainId, String key) {
+        return deterministicId(domainId, key);
+    }
+
+    public static String reporterId(String domainId, String key) {
+        return deterministicId(domainId, key);
+    }
```
leleueri, line 51: discussion: do we need to have a dedicated method per resource if all of them are relying on the same deterministicId method ?

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificateResource.java
```diff
@@ -0,0 +1,115 @@
...
+    public void get(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("certKey") String certKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        resolveDomain(environmentId, domainKey)
+                .flatMap(domain -> checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_CERTIFICATE, Acl.READ)
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even f the user do not have right to perform the action. I would like to do the opposite, first check permission and then resolve the domain. We should be able to compute the domainId using the automationKey.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificateResource.java
```diff
@@ -0,0 +1,115 @@
...
+    public void delete(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("certKey") String certKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        resolveDomain(environmentId, domainKey)
+                .flatMapCompletable(domain -> checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_CERTIFICATE, Acl.DELETE)
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even f the user do not have right to perform the action. I would like to do the opposite, first check permission and then resolve the domain. We should be able to compute the domainId using the automationKey.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificatesResource.java
```diff
@@ -0,0 +1,174 @@
...
+                    array = @ArraySchema(schema = @Schema(implementation = AutomationCertificate.class))))
+    public void list(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        resolveDomain(environmentId, domainKey)
+                .flatMap(domain -> checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_CERTIFICATE, Acl.LIST)
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even f the user do not have right to perform the action. I would like to do the opposite, first check permission and then resolve the domain. We should be able to compute the domainId using the automationKey.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificatesResource.java
```diff
@@ -0,0 +1,174 @@
...
+        return domainService.findById(AutomationIds.domainId(environmentId, domainKey))
+                .switchIfEmpty(Single.error(() -> new DomainNotFoundException(domainKey)))
+                .flatMap(domain -> domain.isManagedBy(ManagedBy.AUTOMATION_API)
+                        ? Single.just(domain)
+                        : Single.error(new DomainNotFoundException(domainKey)));
+    }
+
+    private static String nullToEmpty(String s) {
+        return s == null ? "" : s;
+    }
```
leleueri: suggestion: should it be managed in the abstract super class as other entities may need such conversion

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainResource.java
```diff
@@ -0,0 +1,124 @@
...
+    public void get(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        final String domainId = AutomationIds.domainId(environmentId, domainKey);
+        domainService.findById(domainId)
+                .switchIfEmpty(Single.error(() -> new DomainNotFoundException(domainKey)))
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even if the user do not have right to perform the action.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainResource.java
```diff
@@ -0,0 +1,124 @@
...
+        final String domainId = AutomationIds.domainId(environmentId, domainKey);
+
+        domainService.findById(domainId)
+                .switchIfEmpty(Single.error(() -> new DomainNotFoundException(domainKey)))
+                .flatMapCompletable(domain -> {
+                    if (!domain.isManagedBy(ManagedBy.AUTOMATION_API)) {
+                        return Completable.error(new DomainNotFoundException(domainKey));
+                    }
+                    return checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN, Acl.DELETE)
+                            .andThen(domainService.delete(new GraviteeContext(organizationId, environmentId, domain.getId()), domain.getId(), principal));
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even if the user do not have right to perform the action. In addition, if the domain is not found should we return a 204 DELETED as at the end the state is the expected one (domain is not present in the system anymore) 🤔

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/CertificateResource.java
leleueri: In addition, I think we should return 204 on delete request if the user has the permission to perform the action and the certificate is not found

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProviderResource.java
```diff
@@ -0,0 +1,122 @@
...
+    public void get(
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("idpKey") String idpKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        resolveDomain(environmentId, domainKey)
+                .flatMap(domain -> checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_IDENTITY_PROVIDER, Acl.READ)
```
leleueri: issue: resolving the domain before checking the permission will raise a domain not found exception even f the user do not have right to perform the action

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProviderResource.java
```diff
@@ -0,0 +1,122 @@
...
+            @PathParam("orgId") String organizationId,
+            @PathParam("envId") String environmentId,
+            @PathParam("domainKey") String domainKey,
+            @PathParam("idpKey") String idpKey,
+            @Suspended final AsyncResponse response) {
+
+        final var principal = getAuthenticatedUser();
+        resolveDomain(environmentId, domainKey)
+                .flatMapCompletable(domain -> checkAnyPermission(principal, organizationId, environmentId, domain.getId(), Permission.DOMAIN_IDENTITY_PROVIDER, Acl.DELETE)
+                        .andThen(resolveIdentityProvider(domain, idpKey))
```
leleueri: I would return 204 if the deletion is performed on NotFound IDP entity

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-container/src/main/java/io/gravitee/am/management/standalone/server/ManagementApiServer.java
```diff
@@ -52,36 +70,89 @@ public ManagementApiServer() {
...
 
-        // X-Forwarded-* support
         context.addFilter(ForwardedHeaderFilter.class, "/*", EnumSet.allOf(DispatcherType.class));
+        context.addFilter(new FilterHolder(new DelegatingFilterProxy("springSecurityFilterChain")), "/*", EnumSet.allOf(DispatcherType.class));
+
+        return context;
+    }
+
+    private ServletContextHandler configureAutomationAPI() {
+        String automationPath = entrypoint + "/automation";
```
leleueri: suggestion: I would create a automationEntrypoint to have flexibility to configure automation endpoint with a path whoch doesn't contains /management

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProvidersResource.java
```diff
@@ -0,0 +1,181 @@
...
+                                        + (definition.isDefaultResource() ? " (the domain already has a default identity provider)" : "")));
+                    }
+                    if (definition.isDefaultResource() && allExisting.stream()
+                            .anyMatch(idp -> idp.isManagedBy(ManagedBy.AUTOMATION_API) && idp.isSystem())) {
+                        return Single.error(new InvalidParameterException(
+                                "The domain already has a default identity provider"));
+                    }
+                    AutomationNewIdentityProvider newIdp = AutomationIdentityProviderMapper.toNewIdentityProvider(definition);
+                    newIdp.setId(idpId);
+                    Single<IdentityProvider> op = identityProviderService.create(domain, newIdp, principal, definition.isDefaultResource());
```
leleueri: The "default" resource is a System resource which is based on the gravitee.yalm config, in that case the newIdp has o be generated (the configuration part), I know this is counter intuitive with automation part but this is to align the behaviour for something which is a "system" IDP (same remark for Certificates and Reporters)

leleueri: in addition, maybe the "default" boolean in the model shoud be named "system" to reflect the attribute used in DB.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationIds.java
wajb: Strictly, no. I'll take a look at this again.

wajb: On balance, I feel the descriptive signatures is useful, even if there is implementation duplication. Perhaps we can review the pattern when further types are handled?

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
```diff
@@ -0,0 +1,164 @@
...
+                                            newDomain.setId(domainId);
+                                            newDomain.setAutomationKey(key);
+                                            newDomain.setName(definition.getName());
+                                            newDomain.setPath(definition.getPath());
+                                            newDomain.setDescription(definition.getDescription());
+                                            newDomain.setDataPlaneId(definition.getDataPlaneId());
+                                            return domainService.create(organizationId, environmentId, newDomain, principal);
+                                        }))))
+                        .flatMap(domain -> applyAndRespond(domain, definition, principal)))
+                .subscribe(response::resume, response::resume);
```
leleueri, line 136: question: if the domain doesn't exists, is it expected to execute the create and then the update (into the applyAndResponse) ?

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -366,22 +368,30 @@ public Flowable<Domain> findByIdIn(Collection<String> ids) {
...
         if (dataPlaneRegistry.getDataPlanes().stream().map(DataPlaneDescription::id).noneMatch(id -> id.equals(newDomain.getDataPlaneId()))) {
             return Single.error(new InvalidDataPlaneException("An error occurred while trying to create a domain. Data Plane with provided Id doesn't exist."));
         }
-        return domainRepository.findByHrid(ReferenceType.ENVIRONMENT, environmentId, hrid)
-                .isEmpty()
+        // Automation domains carry a deterministic id derived from their key, so existence is
+        // checked by id; other domains keep the hrid-uniqueness guard.
+        Single<Boolean> domainIsEmpty = automationDomain != null
+                ? domainRepository.findById(automationDomain.getId()).isEmpty()
+                : domainRepository.findByHrid(ReferenceType.ENVIRONMENT, environmentId, hrid).isEmpty();
```
leleueri: discussion: hrid is derived from the domain [name](https://github.com/gravitee-io/gravitee-access-management/pull/7897/changes#diff-2f4b249c4c3232fcfee253763ba2dfa78bbf2241e7e43bb5306f320033c450acR373).So I would keep only the findByHrid to make sure that the HRID is uniq (the hrid is used to rout the request to the right domain)

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -478,10 +497,15 @@ public Single<Domain> update(String domainId, Domain domain) {
                     domain.setVersion(existingDomain.getVersion());
                     domain.setReferenceId(existingDomain.getReferenceId());
                     domain.setReferenceType(existingDomain.getReferenceType());
-                    domain.setHrid(IdGenerator.generate(domain.getName()));
+                    domain.setHrid(existingDomain.getHrid());
```
leleueri: issue: until now hrid was updated if the domain name change, we have to keep it otherwise this is a BC. as we have a dedicated "automation_key" for each entity (including the domain), there is no need anymore to enforce hrid immutability

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -514,7 +538,8 @@ public Single<Domain> patch(GraviteeContext graviteeContext, String domainId, Pa
                     if (Boolean.FALSE.equals(accountSettingsValidator.validate(accountSettings))) {
                         return Single.error(new InvalidParameterException("Unexpected forgot password field"));
                     }
-                    toPatch.setHrid(IdGenerator.generate(toPatch.getName()));
+                    // Preserve existing HRID — once created, it is immutable
+                    toPatch.setHrid(oldDomain.getHrid());
```
leleueri: issue: until now hrid was updated if the domain name change, we have to keep it otherwise this is a BC. as we have a dedicated "automation_key" for each entity (including the domain), there is no need anymore to enforce hrid immutability

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
wajb: That's right, since any id collision is handled as an update in the `DomainsResource`. I will revert this 👍

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
wajb: Well spotted. Will fix.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/DomainsResource.java
wajb: `DomainServiceImpl.create()` only bootstraps some fields and has some hardcoded defaults (like `enabled=false`) so the `update()` applies the full declarative state. While it could be optimised, this approach at least ensures state consistency for create and update. The lookup of IDPs in the `applyAndRespond` is necessary for mapping in both paths.

leleueri (approved):

PR #7910, opened 2026-05-21
GMA-329 spiffe instance support v3

sclarkamlab:
[GMA-329](https://gravitee.atlassian.net/browse/GMA-329)

Adds SPIFFE instance support for agent applications and the related UI/listing work. Main areas:

**SPIFFE subject matching (backend)**
- Add a `PREFIX` subject match mode for SPIFFE application settings, alongside exact match.
- `SpiffeClientAssertionValidator` / `SpiffeJwtSvidValidator` honour the configured match mode; PREFIX requires a trailing slash on the configured subject so prefixes only match at path boundaries.
- Null-safe handling of `subjectMatchMode` in application validation (defaults to exact match when unset).

**Agent applications**
- Allow agent applications to be marked as templates.
- Support a multi-valued `type` filter on the `/applications` listing so the API can return or exclude agents (`ApplicationCriteria` / `ApplicationFilter` plumbed through Mongo + JDBC repositories and `ApplicationServiceImpl`).

**UI**
- New top-level **Agents** nav section with its own list, resolver and creation flow.
- Applications list now routes through the multi-type filter to exclude agents.
- Expose the SPIFFE subject match mode in the OAuth2 grant-flows settings, including PREFIX trailing-slash hint.
- Reset SPIFFE form state when saving with a non-SPIFFE auth method; surface errors on agent list/resolver/creation subscriptions; send `kind` instead of `subType` on agent creation.

**Contracts**
- Regen `docs/mapi/openapi.yaml` and the management SDK for the new SPIFFE settings and multi-type filter.

## :memo: Test scenarios

- Unit tests for `SpiffeClientAssertionValidator`, `SpiffeJwtSvidValidator`, application type classification, and SPIFFE validation in `ApplicationServiceImpl`.
- Jest coverage: SPIFFE PREFIX subject assertion, agent applications as templates, applications-exclude-agents listing, multi-type filter + pagination.
- SPIRE jest tests enabled in the Mongo gateway CI run.

image

## :man: :woman: Reviewers

[GMA-329]: https://gravitee.atlassian.net/browse/GMA-329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
@@ -100,7 +100,18 @@ public final class SpiffeJwtSvidValidator {
             return "client missing spiffe settings";
         }
         String expected = spiffeApplicationSettings.getSubject();
-        if (expected == null || expected.isBlank() || !expected.equals(sub)) {
+        if (expected == null || expected.isBlank()) {
+            return "sub does not match client subject";
+        }
+        SpiffeApplicationSettings.SubjectMatchMode mode = spiffeApplicationSettings.getSubjectMatchMode();
+        if (mode == null) {
+            mode = SpiffeApplicationSettings.SubjectMatchMode.EXACT;
+        }
+        boolean match = switch (mode) {
+            case EXACT -> expected.equals(sub);
+            case PREFIX -> sub.equals(expected) || sub.startsWith(expected + "/");
+        };
+        if (!match) {
             return "sub does not match client subject";
         }
 
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
@@ -109,7 +109,7 @@ public final class SpiffeJwtSvidValidator {
         }
         boolean match = switch (mode) {
             case EXACT -> expected.equals(sub);
-            case PREFIX -> sub.equals(expected) || sub.startsWith(expected + "/");
+            case PREFIX -> sub.startsWith(expected);
         };
         if (!match) {
             return "sub does not match client subject";
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
@@ -26,6 +26,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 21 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/spiffe/SpiffeJwtSvidValidator.java
```diff
@@ -100,7 +100,18 @@ public String validate(SignedJWT jwt,
...
         }
         String expected = spiffeApplicationSettings.getSubject();
-        if (expected == null || expected.isBlank() || !expected.equals(sub)) {
+        if (expected == null || expected.isBlank()) {
+            return "sub does not match client subject";
+        }
+        SpiffeApplicationSettings.SubjectMatchMode mode = spiffeApplicationSettings.getSubjectMatchMode();
+        if (mode == null) {
+            mode = SpiffeApplicationSettings.SubjectMatchMode.EXACT;
+        }
```
leleueri: nickpick: can be simplified with Optional.ofNullable.orElse

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/SpiffeClientAssertionValidatorTest.java
```diff
@@ -346,6 +348,114 @@ void succeeds_onWellFormedSvid() throws Exception {
...
+        String parent = "spiffe://example.org/hotel-agent/";
+        String instanceSpiffeId = parent + "instance-b";
+        String blueprintClientId = "blueprint-2";
+
+        Client blueprint = new Client();
+        blueprint.setClientId(blueprintClientId);
+        blueprint.setTokenEndpointAuthMethod(ClientAuthenticationMethod.SPIFFE_JWT);
+        blueprint.setAppType(ApplicationType.AGENT);
+        blueprint.setAgentType(AgentType.AUTONOMOUS);
+        SpiffeApplicationSettings appSettings = new SpiffeApplicationSettings();
```
leleueri, line 404: question: I was think you replace spiffe by workload in previous PR 🤔 is that correct? if so, should we change it also at app level?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -1490,6 +1490,21 @@ private Single<Application> validateSpiffeSettings(Application application) {
                     "spiffe.subject must start with " + anchor));
         }
 
+        io.gravitee.am.model.application.SpiffeApplicationSettings.SubjectMatchMode mode = spiffe.getSubjectMatchMode();
```
leleueri: todo: can we import the enum so the initialization can bee simplified ? we can also use 'var' here.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/SpiffeClientAssertionValidatorTest.java
sclarkamlab: Yes we can change that here, good shout

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Done

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/service/assertion/impl/SpiffeClientAssertionValidatorTest.java
sclarkamlab: Actually thinking about it more I'm gonna leave the class as SpiffeApplicationSettings. The Client accessor is already workloadIdentitySettings which covers the app-level naming, but the settings themselves are genuinely spiffe-specific (trust domain, subject match mode etc) so the name still fits. Renaming the class would ripple through the mongo model, repo, service and mgmt-api for not much gain.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -106,14 +101,7 @@
 
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.HashSet;
```
leleueri, line 112: nickpick: the team convension is to avoid wildcard import. you can configure IDEA to not use wildcards (in fact you have to specify a high number of import line before it can use wildcard

Image

leleueri (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Ah ok noted for next time, apologies

PR #7911, opened 2026-05-21
GMA-217 - View Agents as separate list

sclarkamlab:
## :id: Reference

[GMA-217](https://gravitee.atlassian.net/browse/GMA-217)

## :pencil2: Changes

Splits Agents out from Applications in the management UI so they show up as a top-level nav entry with their own list view, while keeping a single `/applications` endpoint on the backend.

**Backend**
- `/applications` listing now accepts a multi-valued `type` filter so the UI can ask for "everything except AGENT" or "AGENT only" in one call. Single-valued `type=X` still works.
- OAS + management SDK regenerated.

**UI**
- New top-level **Agents** nav entry; Applications list now scopes itself via the multi-type filter so agents don't bleed into it.
- Agent creation sends `kind` (was `subType`) to match the backend contract.
- Agent list / resolver / creation now surface subscription errors instead of failing silently.

## :memo: Test scenarios

- Jest: `applications-exclude-agents.jest.spec.ts` covers multi-type filter, pagination across filtered pages, totalCount accuracy, server-side filtering (no post-fetch trimming).
- Jest: `agent-applications.jest.spec.ts` covers `type=AGENT` listing, pagination, name search, projection shape.
- Java: `ApplicationFilter` capture asserts the multi-valued type is forwarded from resource -> service.
- Manual: seeded a domain with 100 apps (50 agents across all three kinds + 50 normal apps weighted across WEB/SERVICE/BROWSER/NATIVE/RESOURCE_SERVER + CIMD enabled) and walked the Applications and Agents lists end-to-end.

## :computer: Screenshots
Agent List View
image

Search
image

Add Agent
image

## :books: Notes

- Agent **kind** filtering (USER_EMBEDDED / HOSTED_DELEGATED / AUTONOMOUS) is intentionally out of scope here - \`ApplicationFilter\` doesn't yet support \`kind\` on the repo query. Easy follow-up once that's in.

[GMA-217]: https://gravitee.atlassian.net/browse/GMA-217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-test/specs/management/applications/applications-exclude-agents.jest.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      expect(first.data).toHaveLength(2);
+      expect(second.data).toHaveLength(2);
+      expect(first.totalCount).toEqual(pagingNonAgentIds.length);
+      expect(second.totalCount).toEqual(pagingNonAgentIds.length);
+    });
+  });
+
+  // Note: ApplicationType enum coverage is enforced by ApplicationTypeClassificationTest
+  // on the Java side. Keeping the tripwire on the JVM means it fires at build time when a
+  // new enum value is added, before code reaches the environment this Jest spec runs in.
```
wajb: nitpick: can remove IMO.

gravitee-am-test/specs/management/applications/applications-exclude-agents.jest.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  (t) => !AGENT_TYPES.includes(t),
+);
+
+let accessToken: string;
+let domain: Domain;
+const createdAppIds: string[] = [];
+const agentIds: string[] = [];
+const nonAgentIds: string[] = [];
+
+const trackAgent = async (name: string): Promise<Application> => {
```
wajb: suggestion: could we adapt the existing `applications-fixture.ts` that handles cleanup etc.?

gravitee-am-ui/src/app/services/application.service.ts
```diff
@@ -22,6 +22,12 @@ import { AppConfig } from '../../config/app.config';
 
 import { AuthService } from './auth.service';
 
+// Mirrors io.gravitee.am.model.application.ApplicationType (server-side enum).
+// Keep in sync — see TripwireSpec in gravitee-am-test for the enforcing test.
```
wajb: nitpick: does TripwireSpec exist?

wajb (approved):

gravitee-am-ui/src/app/services/application.service.ts
sclarkamlab: It was created as part of the dev work but don't think it got committed, will remove this

gravitee-am-test/specs/management/applications/applications-exclude-agents.jest.spec.ts
sclarkamlab: Done in `609a2349`. Whole block went away in the fixture refactor.

gravitee-am-test/specs/management/applications/applications-exclude-agents.jest.spec.ts
sclarkamlab: Done in `609a2349`. Using `setupApplicationsFixture` + `createAndTrackApplication` now, dropped about 75 lines of bespoke setup.

gravitee-am-ui/src/app/services/application.service.ts
sclarkamlab: Good catch, doesn't exist. Stale ref, dropped the comment in `3c88b355`.

PR #7912, opened 2026-05-22
fix(token-exchange): support parameterized scopes in downscoping and delegation

AgRenaud:

[AM-7026](https://gravitee.atlassian.net/browse/AM-7026)

Make OAuth2 Token Exchange (RFC 8693) aware of **parameterized scopes** (e.g. `read:<resource_id>`) when computing granted scopes. Previously the service compared scopes by exact string equality, so any `` `<base>:<suffix>` `` was rejected even when the parameterized `` `<base>` `` was authorized on the client / protected resource.

Three call sites in `TokenExchangeServiceImpl` are updated:

- **Downscoping** (`computeDownscopingAllowedScopes`) — a parameterized base scope `` `<s>` `` in the client/resource pool now admits any `` `<s>:<suffix>` `` coming from the subject/actor token cap.
- **Delegation cap** (new `parameterizedAwareIntersection` used for the `subject ∩ actor` intersection) — keeps the more-specific suffixed form when only one side declares the bare base (`` `<s>` `` ∩ `` `<s>:x` `` → `` `<s>:x` ``). Lateral moves (`` `<s>:a` `` ∩ `` `<s>:b` ``) still collapse to empty.
- **Requested-vs-allowed validation** (`computeGrantedScopes`) — accepts a requested suffixed scope whose base is in the allowed set.

`ScopeManager` is injected via `OAuth2Configuration` so the service can ask whether a scope is declared parameterized in the domain.

## :memo: Test scenarios

Unit tests in `TokenExchangeServiceImplTest` (added in this PR) cover:

- Downscoping: parameterized base in the client/resource pool accepts a suffixed cap scope from the subject/actor token.
- Downscoping: non-parameterized scope still requires exact match.
- Delegation: `subject={<s>}` ∩ `actor={<s>:x}` → `{<s>:x}` (and the symmetric case).
- Delegation: lateral move `subject={<s>:a}` ∩ `actor={<s>:b}` → `∅`.
- Delegation: identical suffixes are preserved.
- Requested-scope validation: requesting `` `<s>:<suffix>` `` succeeds when `` `<s>` `` is allowed and parameterized; fails when `` `<s>` `` is not parameterized.
- Empty / null scope sets on either side are handled.

Manual:

1. Configure a domain with a parameterized scope `read` (declared parameterized in scope management).
2. Issue a subject token with `read:resource-abc`.
3. Call `POST /oauth/token` with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange` on a client whose scopes include `read`.
4. Verify the issued token preserves `read:resource-abc`.
5. Repeat with a non-parameterized scope and confirm exact-match behavior is unchanged.
6. Trigger a lateral move (subject `read:a`, actor `read:b`) and confirm `invalid_scope`.

N/A — backend-only change in `gravitee-am-gateway-handler-oidc`. No UI surface affected.

- Behavior aligns parameterized-scope semantics across grant types: token-exchange now matches what the standard authorization-code / client-credentials flows already do via `ScopeManager.isParameterizedScope`.
- The "more-specific wins" rule for the delegation intersection is worth calling out in the Token Exchange docs: `` `<base>` `` ∩ `` `<base>:<suffix>` `` resolves to `` `<base>:<suffix>` ``, not `` `<base>` ``.
- Lateral moves between two different suffixes of the same base are intentionally rejected (no fall-back to the bare base) — surface this in the troubleshooting section.
- No config flag is introduced; the change is on by default for any scope declared parameterized in scope management.

[AM-7026]: https://gravitee.atlassian.net/browse/AM-7026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/TokenExchangeServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/TokenExchangeServiceImpl.java
@@ -21,12 +21,14 @@ import io.gravitee.am.common.oauth2.Parameters;
 import io.gravitee.am.common.oauth2.TokenType;
 import io.gravitee.am.gateway.handler.oauth2.exception.InvalidScopeException;
 import io.gravitee.am.gateway.handler.oauth2.service.request.TokenRequest;
+import io.gravitee.am.gateway.handler.oauth2.service.scope.ScopeManager;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.ActorTokenInfo;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.TokenExchangeUserResolver;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.TokenValidator;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.TokenExchangeResult;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.TokenExchangeService;
 import io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange.ValidatedToken;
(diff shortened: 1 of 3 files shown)


leleueri (approved):
I have asked another view but looks ok. Waiting for another feedback before merging

wajb (approved):

PR #7914, opened 2026-05-22
feat(GMA-322): set act.sub to agent instance id for user-bound agent …

sclarkamlab:

[GMA-322](https://gravitee.atlassian.net/browse/GMA-322)

For user-bound agent kinds (`USER_EMBEDDED`, `HOSTED_DELEGATED`) the issued token's top-level `sub` is the end user, not the agent. That meant the `act.sub` actor claim was falling back to the blueprint `client_id`, which isn't really the actor doing the work.

This sets `act.sub` to the agent's instance id when we know it, so the actor in the delegation chain points at the actual agent instance rather than the blueprint. The logic lives in a small `resolveAgentActSubject` helper in `TokenServiceImpl`:

- `USER_EMBEDDED` / `HOSTED_DELEGATED` with an instance id → `act.sub` = agent instance id
- same kinds with no instance id → fall back to `client_id`
- `AUTONOMOUS` → unchanged (its top-level `sub` already carries the instance id)

`act.sub_profile` still gets the lower-cased agent type as before.

## :memo: Test scenarios

Unit tests in `TokenServiceTest`:
- USER_EMBEDDED with instance id sets `act.sub` to the instance id
- HOSTED_DELEGATED with instance id sets `act.sub` to the instance id
- HOSTED_DELEGATED without an instance id falls back to `client_id`

Updated the `hosted-delegated-workload-jwt` gateway integration spec to expect `act.sub` to equal the agent instance id.

n/a

`act.sub` now means "the agent instance acting", not "the blueprint client". Worth noting for anyone consuming the delegation chain downstream.

[GMA-322]: https://gravitee.atlassian.net/browse/GMA-322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
@@ -46,6 +46,7 @@ import io.gravitee.am.gateway.handler.oidc.service.discovery.OpenIDDiscoveryServ
 import io.gravitee.am.gateway.handler.oidc.service.idtoken.IDTokenService;
 import io.gravitee.am.model.TokenClaim;
 import io.gravitee.am.model.User;
+import io.gravitee.am.model.application.AgentType;
 import io.gravitee.am.model.oidc.Client;
 import io.gravitee.am.model.safe.ClientProperties;
 import io.gravitee.am.model.safe.UserProperties;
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
wajb, line 446: question: there are conditions in this call that have some overlap with the changes. Could greater clarity be achieved by combining the conditions so that a `sub` value is determined and set in one place?

sclarkamlab: Sounds good will do that now, thanks.

wajb (approved):

PR #7916, opened 2026-05-22
fix: restore certificate alias as token kid

leleueri:
fixes AM-7019

fixes gravitee-io/issues#11442


wajb (changes requested):
The `JWTServiceImpl.extractKid` ignores non-UUID values and causes a fallback to the client's default cert. We can probably simplify that logic but we need to make sure the cert can be resolved by UUID or alias, at least for a transition period.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/jwt/impl/JWTServiceImpl.java
```diff
@@ -149,20 +153,58 @@ public Single<EncodedJWT> encodeAuthorization(JWT jwt, Client client) {
...
+        String kid = signedJWT.getHeader().getKeyID();
+        if (kid == null) {
+            return Optional.empty();
+        }
+
+        String issuerDomain = extractDomain(signedJWT);
+
+        return Stream.concat(certificateManager.providers().stream(), Stream.of(certificateManager.defaultCertificateProvider()))
+                .filter(Objects::nonNull)
+                .filter(provider -> kid.equals(provider.getKeyId()))
```
wajb: issue: this matches by alias only, which means existing tokens (with cert id `kid`s) won't match and will fall through to the clients' defaults. When the default cert does not happen to be the correct signing cert, the token is deemed invalid.

Can we allow matching by cert id too (`provider.getCertificateInfo().certificateId`?), at least for the next 4.10 release?

wajb (approved):
Tested successfully. However Postman introspection tests are failing.

PR #7932, opened 2026-05-26
fix: master domain should introspect token generated in all other dom…

lgw-gravitee:
…ains


wajb (approved):
Tested successfully.

leleueri (approved):

PR #7947, opened 2026-05-27
chore: application endpoint jest test

sagarchauhan-qa:
[AM-5416] - jest test for applications endpoints

[AM-5416]: https://gravitee.atlassian.net/browse/AM-5416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/management/applications/application-endpoints.jest.spec.ts
+++ b/gravitee-am-test/specs/management/applications/application-endpoints.jest.spec.ts
@@ -86,13 +86,12 @@ describe('GET domain entrypoints', () => {
     }
   });
 
-  it('should use the non-default entrypoint for tag-based routing when multiple entrypoints are configured', () => {
-    if (entrypoints.length > 1) {
-      const selected = selectEffectiveEntrypoint(entrypoints);
-      expect(selected.defaultEntrypoint).toBeFalsy();
-    } else {
-      expect(selectEffectiveEntrypoint(entrypoints)).toBe(entrypoints[0]);
-    }
+  it('should return the sole entrypoint in a single-entrypoint environment', () => {
+    // No other test creates or modifies environment-level entrypoints so the count
+    // is always 1 in this environment. The non-default tag-based routing path in
+    // selectEffectiveEntrypoint requires a multi-entrypoint environment to exercise directly.
+    expect(entrypoints.length).toBe(1);
+    expect(selectEffectiveEntrypoint(entrypoints)).toBe(entrypoints[0]);
   });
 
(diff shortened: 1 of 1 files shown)


gravitee-am-test/specs/management/applications/application-endpoints.jest.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  it('should include a url on each entrypoint that serves as the base for all application endpoint URLs', () => {
+    for (const entrypoint of entrypoints) {
+      expect(entrypoint.url).toBeDefined();
+      expect(entrypoint.url).toMatch(/^https?:\/\//);
+    }
+  });
+
+  it('should resolve to standard OAuth2/OIDC endpoint paths from the entrypoint base URL', async () => {
+    const entrypoint = entrypoints.length === 1 ? entrypoints[0] : (entrypoints.find((e) => !e.defaultEntrypoint) ?? entrypoints[0]);
```
wajb: suggestion: this non-default, multiple tag-based routing is subtle and not really called out anywhere. Could it be a discrete test with its own self-documenting description?

This test here could reuse a helper function to get the same effective entrypoint.

gravitee-am-test/specs/management/applications/application-endpoints.jest.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  it('should include a url on each entrypoint that serves as the base for all application endpoint URLs', () => {
+    for (const entrypoint of entrypoints) {
+      expect(entrypoint.url).toBeDefined();
+      expect(entrypoint.url).toMatch(/^https?:\/\//);
+    }
+  });
+
+  it('should use the non-default entrypoint for tag-based routing when multiple entrypoints are configured', () => {
+    if (entrypoints.length > 1) {
```
wajb: question: it's a bit of smell that this isn't deterministic. It means we can't tell what assertions are actually being made. We could either miss coverage of one scenario or provide false confidence by tolerating an inconsistent behaviour.

Is there a genuine cleanup leak in the tests affecting global state or can we find out what always happens here and target that behaviour unconditionally?

sagarchauhan-qa: If I am not mistaken, there is no cleanup leak. The full test suite has been checked and no other test creates or modifies environment-level entrypoints. The count is always 1 in this environment, so we can assert unconditionally.

wajb (approved):

PR #7963, opened 2026-06-03
fix: AM-7079 repair system IDP reference

wajb:
[AM-7079](https://gravitee.atlassian.net/browse/AM-7079)

Context:
- When a domain is PUT using the automation API it can reference an IDP by its automation key in the `account.defaultIdentityProviderForRegistration` field, but, to support eventual consistency, the IDP does not have to exist (yet).
- When an IDP is PUT with the `system` flag (meaning it is created as the default database-source IDP for a domain) it changes the behaviour of the internal id generation of the IDP from deterministic to randomised with a prefix of `default-idp-`.

This change repairs the reference for the scenario where the domain is new (initial PUT) and the referenced IDP does not yet exist, but is then added with `system` flag.

## :memo: Test scenarios 
See Jira.

n/a

n/a

@gravitee-io/am

[AM-7079]: https://gravitee.atlassian.net/browse/AM-7079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProvidersResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProvidersResource.java
@@ -25,11 +25,13 @@ import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.IdentityProvider;
 import io.gravitee.am.model.ManagedBy;
 import io.gravitee.am.model.ReferenceType;
+import io.gravitee.am.model.account.AccountSettings;
 import io.gravitee.am.model.permissions.Permission;
 import io.gravitee.am.service.IdentityProviderService;
 import io.gravitee.am.service.exception.DomainNotFoundException;
 import io.gravitee.am.service.exception.InvalidParameterException;
 import io.gravitee.am.service.model.AutomationNewIdentityProvider;
+import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Single;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.media.ArraySchema;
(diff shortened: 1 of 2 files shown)


gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProvidersResource.java
```diff
@@ -193,6 +197,23 @@ private static Single<AutomationIdentityProvider> rejectIfMissingIdentityProvide
...
+    private Completable reconcileDomainReference(Domain domain, String idpKey, String realIdpId) {
+        AccountSettings account = domain.getAccountSettings();
+        if (account == null
+                || !idpKey.equals(account.getDefaultIdentityProviderForRegistrationKey())
+                || realIdpId.equals(account.getDefaultIdentityProviderForRegistration())) {
+            return Completable.complete();
+        }
+        account.setDefaultIdentityProviderForRegistration(realIdpId);
+        return domainService.update(domain.getId(), domain, false).ignoreElement();
+    }
```
leleueri, line 215: question: what is the behaviour if I apply the domain definition after idp creatio to update some settings ? I would say the realIdpId will be process once again using the key so the issue will remains isn't it?

wajb: For a domain PUT when the IDP already exists, [AutomationDomainMapper::identityProviderKeyToId](https://github.com/gravitee-io/gravitee-access-management/blob/234e45c733871f538b71db3f54718bd8d06dfd35/gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/mapper/AutomationDomainMapper.java#L293) picks up the actual id, so it isn't an issue.

leleueri (approved):

PR #7974, opened 2026-06-03
fix: evaluate the error parameter from context and request params

leleueri:
move the evaluation after the initial query params restore to avoid NPE on client

fixes AM-7078

related-to AM-6304


wajb (approved):
Tested successfully, lgtm.

PR #7977, opened 2026-06-04
chore: introduce oss-bundle and ee-bundle Maven profiles

leleueri:
Split com.graviteesource.* EE artifacts out of the default build into a new ee-bundle profile (activeByDefault=true) in both gateway and management-api distribution poms. An oss-bundle profile is provided to build without any EE plugins by deactivating ee-bundle.

Update CI mvn invocations that use full-bundle to also activate ee-bundle, preserving the existing behaviour.

### DIFF
--- a/.circleci/scripts/ee-schema-compat-check.sh
+++ b/.circleci/scripts/ee-schema-compat-check.sh
@@ -240,7 +240,7 @@ for entry in "${TO_CHECK[@]}"; do
     fi
   else
     echo "  ❌  New ZIP not found in any distribution target directory."
-    echo "      Run 'mvn install -P full-bundle' first to build the distributions."
+    echo "      Run 'mvn install -P ee-bundle,addons-bundle' first to build the distributions."
     echo ""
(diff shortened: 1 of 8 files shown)


wajb (commented):
There are a couple of messages that we could update to include `ee-bundle` in https://github.com/gravitee-io/gravitee-access-management/blob/ab5ba45c825b0697a538744bcfe211235a8417fe/.circleci/scripts/ee-schema-compat-check.sh#L243 and https://github.com/gravitee-io/gravitee-access-management/blob/ab5ba45c825b0697a538744bcfe211235a8417fe/docs/agent-standards/skills/check-schemas/SKILL.md?plain=1#L183 and possibly https://github.com/gravitee-io/gravitee-access-management/blob/ab5ba45c825b0697a538744bcfe211235a8417fe/gravitee-am-test/GUIDELINES.md?plain=1#L1103

Is `full-bundle` perhaps a misnomer? (`extra-bundle`, `util-bundle`...?)

wajb (approved):

PR #7980, opened 2026-06-05
fix: try all providers sharing the same kid before rejecting token va…

leleueri:
…lidation

When two certificates share the same kid (e.g. both named "Default"), the gateway now attempts verification against each matching provider in sequence and accepts the first success, propagating the error only if all providers fail.

Fixes AM-7099

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/jwt/impl/JWTServiceImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/jwt/impl/JWTServiceImpl.java
@@ -42,9 +42,11 @@ import java.security.Key;
 import java.security.KeyPair;
 import java.text.ParseException;
 import java.util.Base64;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/jwt/impl/JWTServiceImpl.java
```diff
@@ -168,7 +174,16 @@ private Optional<CertificateProvider> findProviderByKid(SignedJWT signedJWT) {
...
+                .collect(Collectors.toList());
+    }
+
+    private Single<JWT> verifyWithAnyProvider(String jwt, List<CertificateProvider> providers, TokenType tokenType) {
+        Single<JWT> chain = decodeAndVerify(jwt, providers.get(0), tokenType);
+        for (int i = 1; i < providers.size(); i++) {
+            final CertificateProvider provider = providers.get(i);
+            chain = chain.onErrorResumeNext(err -> decodeAndVerify(jwt, provider, tokenType));
+        }
+        return chain;
```
wajb: suggestion: perhaps more consistent using stream/reduce
```suggestion
    return providers.stream()
            .map(provider -> decodeAndVerify(jwt, provider, tokenType))
            .reduce((chain, next) -> chain.onErrorResumeNext(err -> next))
            .get();
```

wajb (approved):
Tested successfully by duplicating system cert and introspecting signed tokens.

PR #7981, opened 2026-06-08
AM-7103 fix: application cursor search with custom mapi entrypoint

lgw-gravitee:
fixes AM-7103


leleueri (approved):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ApplicationsResourceSearcher.java
wajb: suggestion: could remove this part since it isn't used. We only need to build the query params.

Could also tweak the test setup so that it uses a custom/proxied entrypoint.

PR #8007, opened 2026-06-09
AM-7070 fix: return 403 instead of 3xx in case of smtp error. Added counter t…

lgw-gravitee:


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/resources/endpoint/mfa/MFAChallengeSendEndpoint.java
```diff
@@ -81,7 +82,7 @@ private void resendMfaChallenge(RoutingContext routingContext) {
             sendMfaChallenge(factorProvider, routingContext, factor, endUser, true, true, resChallenge -> {
                 if (resChallenge.failed()) {
                     logger.error("An error has occurred when resending MFA challenge", resChallenge.cause());
-                    routingContext.fail(resChallenge.cause());
+                    respondJsonChallengeFailure(routingContext, SEND_CHALLENGE_FAILED, SEND_CHALLENGE_FAILED_DESCRIPTION, HttpStatusCode.FORBIDDEN_403);
```
leleueri: question: 403 may not be always appropriate, should we evaluate the exception to see if a "getStatus" or "getCode" is available which may represent the status more accurately ?

lgw-gravitee: I added 503 arbitrary. The Cause in simple Throwable so it complicates determining other codes

leleueri (approved):

PR #8008, opened 2026-06-09
fix: manage to move the root extension point per path

leleueri:
this is a BC with a setting ti restore the legacy change

settings in gravitee.yaml legacy.handler.rootFlow.global to set to true to restore the previous behaviour

fixes AM-7113


wajb (approved):
Tested successfully - confirmed `authFlow` context attributes are set during _Enrich Authentication Flow_ by default and behaviour is reverted with legacy flag.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/root/RootProvider.java
```diff
@@ -347,6 +348,9 @@ public class RootProvider extends AbstractProtocolProvider {
     @Value("${handlers.request.logger.accesslog.enabled:false}")
     private boolean accessLogEnabled = false;
 
+    @Value("${legacy.handler.rootFlow.global:false}")
+    private boolean globalRootFlow = false;
```
wajb, line 352: minor: is this consistent with other legacy flags? 🤔 How about `legacy.handler.globalRootFlow.enabled` and the private field `legacyGlobalRootFlowEnabled`?

PR #8018, opened 2026-06-10
AM-7132 - fix: exclude agents from application listings by default

sclarkamlab:
The app listings were returning agents mixed in with normal applications. Agents are just `Application` rows with `type = AGENT`, so any listing that doesn't filter by type picks them up too.

Turns out fixing the `/applications` endpoint alone wasn't enough. The console list actually uses the cursor endpoint `/applications/search`, and that one was silently ignoring the `type` filter entirely. The cursor request carried `types` and the resource forwarded them, but neither the Mongo nor JDBC cursor repo ever applied them to the query. So that gap had to be closed first.

What this does, across all three listing paths (`/applications`, `/applications?q=`, and `/applications/search`):

- adds `ApplicationType.defaultingToNonAgent(...)` so when no `type` is given we default to every type except AGENT (via `EnumSet.complementOf`, so any future type is included automatically). Agents are still reachable with an explicit `type=AGENT`.
- wires `types` into the cursor query in `MongoApplicationCursorRepository` and `JdbcApplicationCursorRepository`.

Left the same `status`/`enabled` cursor gap alone, out of scope here.

Test plan:
- `ApplicationsResourceTest`, `ApplicationsResourceFilterTest`, `ApplicationsResourceSearcherTest` updated, plus new cases for the no-type default and `type=AGENT` passthrough.
- `ApplicationCursorRepositoryTest` has a new type-filter test passing against both Mongo and JDBC (8/8 each), so the exclusion is verified at the DB level.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ApplicationsResource.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ApplicationsResource.java
@@ -55,7 +55,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 
 import java.net.URI;
 import java.util.Collection;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


leleueri (commented):
discussion: I'm not convinced by the implementation, IMO we shouldn't exclude AGENT application by default. They are an application like another and the default call to /applications should return everything. I'm more in favor of the UI asking/filtering for the app type expected for the screen.

wajb:
> discussion: I'm not convinced by the implementation, IMO we shouldn't exclude AGENT application by default. They are an application like another and the default call to /applications should return everything. I'm more in favor of the UI asking/filtering for the app type expected for the screen.

Agreed. The UI already defines both `AGENT_APPLICATION_TYPES` and `NON_AGENT_APPLICATION_TYPES` in the `application.service.ts`, which could be leveraged with the `cursorSearch` method for both pages.

sclarkamlab:
@leleueri @wajb 

Agreed, I will drop this PR and recreate a new one with the much simpler UI changes

PR #8022, opened 2026-06-11
AM-7132 - fix: filter agent applications out of the applications page from the UI

sclarkamlab:
Reworked take on #8018, following the review feedback there: the backend no longer changes default behaviour, the UI now asks for the types it wants.

The applications page passes the non-agent type list into the cursor search, so agents stay off that screen. The agents page already filters with type=AGENT so it needed no change. The only backend change is making the Mongo and JDBC cursor repositories actually honour the type filter, which was accepted by the search endpoint but silently ignored at the repository level. The type params already survive cursor pagination since the next-cursor URL keeps the original query params.

https://gravitee.atlassian.net/browse/AM-7132


gravitee-am-ui/src/app/services/application.service.ts
wajb, line 40: thought: should we also align the `AgentsComponent` to use cursor searching?

sclarkamlab: Yeah might as well while we are here

sclarkamlab: @wajb done in [0869393](https://github.com/gravitee-io/gravitee-access-management/pull/8022/commits/0869393b1b8f9af768ea30b484a119f00afac064)

wajb (approved):

PR #8027, opened 2026-06-12
feat: AM-7140 manage existing resources with automation API

wajb:
[AM-7140](https://gravitee.atlassian.net/browse/AM-7140)

Expose existing, non-automation-managed resources to the Automation API. Such brownfield resources can be referenced using the key format `id:<uuid>`, where the use of such a format bypasses the normal resource filtering of the automation endpoints.

Note that there is no requirement for resources created outside of the Automation API to be "adopted" when referenced directly by UUID. Resources retain the `managedBy` value assigned at creation.

## :memo: Test scenarios 
See Jira.

n/a

n/a

@gravitee-io/am

[AM-7140]: https://gravitee.atlassian.net/browse/AM-7140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationResourceResolver.java
```diff
@@ -0,0 +1,130 @@
...
+    public Single<Reporter> resolveReporter(Domain domain, AutomationRef ref) {
+        return resolveReporterMaybe(domain, ref)
+                .switchIfEmpty(Maybe.error(() -> new ReporterNotFoundException(ref.raw())))
+                .toSingle();
+    }
+
+    public Maybe<Reporter> resolveReporterMaybe(Domain domain, AutomationRef ref) {
+        return switch (ref) {
+            case AutomationRef.IdRef(String id) -> reporterService.findById(id)
+                    .filter(reporter -> Reference.domain(domain.getId()).equals(reporter.getReference()));
```
leleueri, line 123: question: why are we checking the consistency with the ReferenceId only on the case where the id is provided? Should we do it also when we are using the automationKey? (same comment for all the "resolve" method in this class.

leleueri (approved):

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/AutomationResourceResolver.java
wajb: Domain scope is already enforced by folding it into the deterministic UUID. If you're referencing by key, the parent domain of an existing resource has to be the one used when creating it (and it must have been created with the AAPI).

PR #8065, opened 2026-06-16
fix: AM-7144 surface all writable domain fields in AAPI

wajb:
[AM-7144](https://gravitee.atlassian.net/browse/AM-7144)

Define all configurable domain fields for use in the Automation API.

Additionally:
- Amend enum casing to match data on the wire (avoids downstream unmarshalling problems with Terraform)
- Omit calculated `loginSettings.enforcePasswordPolicyEnabled` field from both APIs (it is already not included in JSON)
- Wrap and map boolean fields in `oidc.ClientRegistrationSettings` (e.g. `allowedScopesEnabled` -> `isAllowedScopesEnabled`) to avoid mismatches
- Retype `webAuthnSettings.certificates` to map of `String` for strict correctness (avoids downstream unmarshalling problems with Terraform)
- Resolve identity providers' `external` flag automatically from the plugin provider types at creation time, avoiding the need for surfacing the internal field (AAPI only)
- Remove unused `ManagedBy` enum values (affects MAPI OAS only)
- Document properties of `SpiffeDomainSettings.java`
- Also fix [AM-7156](https://gravitee.atlassian.net/browse/AM-7156) - for domain `PUT` with `id:<UUID>` reference format, check domain existence before permissions in order to return consistent 404 response

## :memo: Test scenarios 
n/a

n/a

n/a

@gravitee-io/am 

[AM-7144]: https://gravitee.atlassian.net/browse/AM-7144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AM-7156]: https://gravitee.atlassian.net/browse/AM-7156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/automation/openapi.yaml
+++ b/docs/automation/openapi.yaml
@@ -1645,6 +1645,62 @@ components:
             \ a certificate created via the domain's certificate endpoints."
           example: default
       title: Certificate settings
+    AutomationClientRegistrationSettings:
+      type: object
+      description: OpenID Connect Dynamic Client Registration configuration for the
+        domain.
+      properties:
+        allowHttpSchemeRedirectUri:
+          type: boolean
+          default: false
+          description: Whether the unsecured http scheme is permitted in redirect
+            URIs.
+        allowLocalhostRedirectUri:
+          type: boolean
+          default: false
+          description: Whether localhost is permitted as a redirect URI host.
+        allowRedirectUriParamsExpressionLanguage:
+          type: boolean
+          default: false
+          description: Whether expression language is permitted in redirect URI parameters.
+        allowWildCardRedirectUri:
+          type: boolean
+          default: false
+          description: Whether wildcards are permitted in redirect URIs.
+        allowedScopes:
+          type: array
+          description: Scopes permitted on client registration requests when the allowed
+            list is enabled.
+          items:
+            type: string
+            description: Scopes permitted on client registration requests when the
+              allowed list is enabled.
+        allowedScopesEnabled:
+          type: boolean
+          default: false
+          description: Whether registered client scopes are restricted to an allowed
+            list.
+        clientTemplateEnabled:
+          type: boolean
+          default: false
+          description: Whether a client may be used as a template for dynamic client
+            registration.
+        defaultScopes:
+          type: array
+          description: Default scopes added to every client registration request.
+          items:
+            type: string
+            description: Default scopes added to every client registration request.
+        dynamicClientRegistrationEnabled:
+          type: boolean
+          default: false
+          description: Whether Dynamic Client Registration is enabled for the domain.
+        openDynamicClientRegistrationEnabled:
+          type: boolean
+          default: false
+          description: Whether open (unauthenticated) Dynamic Client Registration
+            is enabled for the domain.
+      title: Client registration settings
     AutomationDomain:
       type: object
       description: "A security domain managed by the Automation API. The key field\
@@ -1654,6 +1710,9 @@ components:
       properties:
         accountSettings:
           $ref: "#/components/schemas/AutomationAccountSettings"
+        alertEnabled:
+          type: boolean
+          description: Whether alerting is enabled for the domain.
         certificateSettings:
           $ref: "#/components/schemas/AutomationCertificateSettings"
         corsSettings:
@@ -1691,6 +1750,11 @@ components:
           title: Key
         loginSettings:
           $ref: "#/components/schemas/LoginSettings"
+        master:
+          type: boolean
+          default: false
+          description: Whether this is the master domain of its environment. A master
+            domain may perform cross-domain token introspection.
         name:
           type: string
           description: Human-readable name of the domain.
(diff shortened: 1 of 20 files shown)


docs/automation/openapi.yaml
```diff
@@ -1777,6 +1846,12 @@ components:
             description: "Email domains allowed to authenticate through this identity\
               \ provider. When set, users whose email domain is not listed are rejected."
             example: "[\"example.com\"]"
+        external:
+          type: boolean
+          default: false
+          description: Whether this is an external identity provider (one that delegates
+            authentication to a third party) rather than an AM-managed one. Set at
+            creation and immutable afterwards.
```
leleueri: remark: AFAIK this attribute is not something the client should define, this is an internal flag for IdP. Don't know if it has to be exposed on AutomationAPI. In the MAPI the way it is manage is not good as the UI is asking 2 times the list of IDP (first time with the external query param to false, second time with the param set to true). If possible, I think we should not provide this bool in the automation api and rely on the plugin type internally to assign the value based on the plugin type.

docs/automation/openapi.yaml
```diff
@@ -1645,6 +1645,62 @@ components:
             \ a certificate created via the domain's certificate endpoints."
           example: default
       title: Certificate settings
+    AutomationClientRegistrationSettings:
```
leleueri, line 1648: question: can you elaborate why some settings are prefixed by Automation and some other are not?

docs/automation/openapi.yaml
```diff
@@ -2111,6 +2143,7 @@ components:
           description: URL used for certificate-based authentication.
         enforcePasswordPolicyEnabled:
           type: boolean
+          readOnly: true
```
leleueri: question: why readOnly ?

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/main/java/io/gravitee/am/management/handlers/automation/resource/IdentityProvidersResource.java
```diff
@@ -206,6 +206,12 @@ private Single<AutomationIdentityProvider> updateExisting(Domain domain, Identit
                     "The 'type' is immutable for an existing identity provider '" + key
                             + "'; delete and recreate it to change it"));
         }
+        // 'external' is an immutable identity attribute fixed at creation
+        if (definition.isExternal() != existing.isExternal()) {
+            return Single.error(new InvalidParameterException(
+                    "The 'external' flag is immutable for an existing identity provider '" + key
+                            + "'; delete and recreate it to change it"));
+        }
```
leleueri: remark: if we remove this attribute from the Automation model, this error will not be possible anymore, we just have to make sure we are using the value coming from the existing entity.

leleueri (changes requested):

docs/automation/openapi.yaml
wajb: Only those schemas that can't reuse the core models are prefixed `Automation`, the name coming from the java classes. Those with the prefix are therefore divergent in some (minimal) way compared to the MAPI.

docs/automation/openapi.yaml
wajb: It's a calculated field, inherited by both APIs.
https://github.com/gravitee-io/gravitee-access-management/blob/b81eebb7b1a67d7232fd0657b93ce371e857cc98/gravitee-am-model/src/main/java/io/gravitee/am/model/login/LoginSettings.java#L95-L100
The alternative is to create a `AutomationLoginSettings` wrapper and omit the field. Wdyt? In fact the UI doesn't reference it either so we could drop it entirely and document as a breaking change (but the MAPI OAS in prod is wrong).

docs/automation/openapi.yaml
wajb: You are right that deriving the flag from the IdP provider is the only robust approach. I've removed the field.

docs/automation/openapi.yaml
wajb: In fact the getter is used as the source of truth for authentication (`WebAuthnEnforcePasswordHandler.java` and `UserAuthenticationServiceImpl.java`), meaning it does provide definitive state, so I think it is reasonable to expose it in the APIs. But it'll stay readonly.

wajb: Upon review, I chose to hide the property in both MAPI and AAPI OAS. The property is naturally omitted from runtime JSON responses according to the visibility settings in `ObjectMapperResolver` so this aligns the OAS. Furthermore, the previously documented flag is knowable from other values, so my previous justification for keeping it was weak.

leleueri (approved):

PR #8066, opened 2026-06-16
fix: purge audits of the in-memory platform reporter

leleueri:
## Description

Fixes [AM-7133](https://gravitee.atlassian.net/browse/AM-7133): audit logs purge does not respect the retention period for the `reporter_audits_PLATFORM` collection.

### Root cause

The internal **platform** reporter is created in memory at startup (`ManagementAuditReporterManager.doStart()` → `ReporterService.createInternal()` with a `PLATFORM/PLATFORM` reference). Because of that reference, its data is written to `reporter_audits_PLATFORM` (Mongo) / table suffix `PLATFORM` (JDBC).

`ReporterAuditSweeper` only iterates `ReporterService.findAll()`, which returns **DB-persisted** reporters (domain/organization). The platform reporter is never persisted, so it is never returned and its collection/table grows unbounded.

### Fix

- Expose the in-memory internal reporter via `AuditReporterManager.getInternalReporter()` (returns `Optional<Reporter>`).
- `ReporterAuditSweeper` now purges the internal reporter explicitly, inside the cluster lease, with the same graceful per-reporter error handling so a failure there does not abort the sweep.

### Tests

Added new test methods to `ReporterAuditSweeperTest` (existing methods untouched):
- `should_purge_internal_platform_reporter`
- `should_handle_internal_platform_reporter_error_gracefully`
- `should_complete_when_internal_platform_reporter_not_initialized`
- `should_not_purge_internal_platform_reporter_when_lease_not_acquired`

All 11 tests pass.

[AM-7133]: https://gravitee.atlassian.net/browse/AM-7133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8069, opened 2026-06-17
feat: reduce impact of audit retention purge (AM-7041)

leleueri:
## Description

Improves the audit retention purge introduced in 4.11 to reduce its impact on the backend (AM-7041).

Changes:
- **Sequential processing** — reporters are now purged one at a time (`concatMapCompletable`) instead of up to 4 in parallel.
- **Capped, configurable batches** — audits are deleted in batches with a **hard maximum of 2000** rows per batch whatever the backend (bounds memory, lock footprint and bind-parameter count — e.g. SQL Server's 2100-parameter limit). Any larger configured value is clamped with a warning.
- **Wait between batches** — a configurable delay between batches lets the backend sync its replication journal.
- **Shared global timeout** — a single deadline is computed once and shared by every reporter. Once reached, no further reporter is started and any in-flight purge stops gracefully; the remaining work resumes on the next scheduled execution.

### New configuration (under `services.purge.audits`)

| Key | Default | Description |
|-----|---------|-------------|
| `batchSize` | 2000 | audits deleted per batch (also the hard max) |
| `batchDelay` | 1000 | wait in ms between two batches |
| `globalTimeout` | 3600 | max duration in seconds of a purge execution shared across all reporters |

## Notes
- This branch is **stacked on [#8066](https://github.com/gravitee-io/gravitee-access-management/pull/8066)** (AM-7133, branch `am-7133`). Until #8066 is merged into `4.11.x`, this PR's diff also shows the `fix: purge audits of the in-memory platform reporter` commit; GitHub will update it automatically once #8066 lands.
- The Mongo id lookup deliberately reads from the **primary** (not a secondary) to avoid re-reading already-deleted ids during the read→delete loop.

## Tests
- `ReporterAuditSweeperTest` updated and passing (11/11).
- Affected modules compile under JDK 21.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/purge/ReporterAuditSweeperTest.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/purge/ReporterAuditSweeperTest.java
@@ -32,6 +32,7 @@ import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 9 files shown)


gravitee-am-reporter/gravitee-am-reporter-jdbc/src/main/java/io/gravitee/am/reporter/jdbc/audit/JdbcAuditReporter.java
```diff
@@ -204,6 +205,12 @@ public class JdbcAuditReporter extends AbstractService<Reporter> implements Audi
     @Value("${services.purge.audits.retention.days:0}")
     private int retentionDays;
 
+    @Value("${services.purge.audits.retention.batchDelay:500}")
+    private long batchDelayMs;
+
+    @Value("${services.purge.audits.retention.batchSize:" + DELETE_BATCH_SIZE + "}")
+    private int batchSize = DELETE_BATCH_SIZE;
```
wajb: nitpick: IMO just remove the `DELETE_BATCH_SIZE` constant and use `PURGE_MAX_BATCH_SIZE`.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/purge/ReporterAuditSweeperTest.java
```diff
@@ -58,6 +59,8 @@ public class ReporterAuditSweeperTest {
     @Before
     public void setUp() {
         MockitoAnnotations.openMocks(this);
+        // @Value defaults are not applied in unit tests; set a non-zero timeout so reporters are not skipped
+        ReflectionTestUtils.setField(sweeper, "globalTimeout", 3600);
```
wajb, line 64: thought: I wonder if it's worth moving to a `@ConfigurationProperties(prefix = "services.purge.audits")` pattern for less brittle encapsulation. I don't know if `@Value()` annotations are too convenient.

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/main/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporter.java
```diff
@@ -274,58 +276,99 @@ protected void doStop() throws Exception {
...
+
+        return oneBatch
+                .repeat()
+                .takeUntil(deleted -> deleted == 0L)
                 .ignoreElements();
     }
 
+    private Single<List<Object>> findAuditIdsToDelete(Date threshold, int limit) {
+        // the id lookup is read straight from the primary: the deleteMany below removes these ids on the
+        // primary, so reading the next page from a lagging secondary would re-return already-purged ids
```
wajb: minor: the comment seems to have survived a refactor and "deleteMany below" is longer accurate.

Also, would it guarantee accuracy by an explicit `withReadPreference(ReadPreference.primary())` or is it really just aligning with the MongoClient read preference / driver default / gravitee.yml? 🤔

leleueri: The recommandation in AM deployment is to define refedPreference to primary by default. But I can enforce it.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/purge/ReporterAuditSweeperTest.java
leleueri: AFAIK, ConfigurationProperties is a SPringBoot specific annotation and we are using SpringFramwork not spring boot.

gravitee-am-reporter/gravitee-am-reporter-jdbc/src/main/java/io/gravitee/am/reporter/jdbc/audit/JdbcAuditReporter.java
leleueri: yes good point

gravitee-am-management-api/gravitee-am-management-api-standalone/gravitee-am-management-api-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -154,6 +154,9 @@ services:
 #    audits:                                     # audits retention period in days (0 means disabled)
 #      retention:
```
wajb, line 155: thought: should we add `leaseDuration`?

wajb (approved):
Sanity tested with no issues.

PR #5321, opened 2025-01-09
Am 4165 dataplane split

leleueri:
This PR is about to merge the branch which contains all the changes for the DataPlane split.

PR #8075, opened 2026-06-18
chore: AM-6696 repository layer cleanup

wajb:
[AM-6696](https://gravitee.atlassian.net/browse/AM-6696)

Repository-layer cleanup following the 4.7 data-plane refactor.

For 4.7 refactor, see https://github.com/gravitee-io/gravitee-access-management/pull/5321.

#### Superseded by the data plane
Justification: the 4.7 data plane reimplemented these independently (`io.gravitee.am.dataplane.*`).
- **Management scope**: `AccessPolicy`, `Credential`, `Device`, `PasswordHistory`, `PermissionTicket`, `Resource`, `UserActivity`, and the domain `User`.
- **Gateway scope**: `LoginAttempt`, `ScopeApproval`.

####  Orphan interfaces
Justification: declared but never implemented and never referenced.
- `AlertEventRepository`, `OrganizationGroupRepository` (note the name was later reused, see consolidations).

#### Transitive
Justification: became unreachable once the above were deleted.
- Repository-api `LoginAttemptCriteria` (only the data-plane copy is used elsewhere).
- `PermissionRequestConverter` and two dead dozer.xml mappings (`JdbcPermissionTicket`, domain `JdbcUser`).

### Consolidations
Justification: after the domain-user removal these abstractions each had exactly one remaining implementor and no external consumers.
- `CommonUserRepository` → merged into `OrganizationUserRepository` (now extends `CrudRepository` directly and carries `UpdateActions`).
- Mongo `AbstractUserRepository<T>` → merged into `MongoOrganizationUserRepository` (generic dropped).
- `UserMongo` → folded into `OrganizationUserMongo`.

### Renaming
Justification: the management `GroupRepository` only ever serves organization groups.
- `GroupRepository` → `OrganizationGroupRepository` (+ `Mongo*`/`Jdbc*` impls, the repository test, and `OrganizationGroupServiceImpl`/test).

### Marked classes NOT REMOVED
- Two remaining `// TODO [DP] class to remove` comments on `GroupMongo` and `MongoOrganizationGroupRepository` are ignored (comment removed) since only _domain_ groups were migrated to the data-plane and these two classes relate to _organization_ groups in the active control-plane layer.

### Simplified repository scope configuration
- Split `RepositoryScopeProvider` implementations so management server only needs to configure `management` scope repositories (`gateway` and `oauth2` are only defined for gateway). **This is a potential breaking change for environments where Mongo IDPs fall back to use non-management connections** (exotic but possible).
- Isolate gateway-scope-dependent `AuthenticationFlowContextService` to gateway service package to remove risk of wiring it in management server. **This is a breaking change for external plugin consumers** that import `io.gravitee.am.service.AuthenticationFlowContextService`, which must switch to `io.gravitee.am.gateway.handler.common.service.AuthenticationFlowContextService.`

## :memo: Test scenarios 
n/a

n/a

n/a

@gravitee-io/am

[AM-6696]: https://gravitee.atlassian.net/browse/AM-6696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


leleueri (commented):
Question: is it possible to provide 2 implementation of the RepositoryScopeProvider one for MAPI with only the Management Scope and another one for the Gateway with all the scopes ? In that way the gravitee.yaml in the management side only need to define the "management" scope

wajb:
> Question: is it possible to provide 2 implementation of the RepositoryScopeProvider one for MAPI with only the Management Scope and another one for the Gateway with all the scopes ? In that way the gravitee.yaml in the management side only need to define the "management" scope

Yes, I've split the `RepositoryScopeProvider` implementations, removing all but `management` scope from the management `gravitee.yml`. I added a safeguard to filter the sole gateway-scope service (`AuthenticationFlowContextServiceImpl`) from the management's `StandaloneConfiguration.java` (really a precaution since the gateway-scoped `AuthenticationFlowContextRepository` is already is wired with `@Lazy`, preventing it from being instantiated). Unfortunately we can't easily move `AuthenticationFlowContextService` out of the common service package because saml2-idp plugin depends on it.

The one concern is `repositories.system-cluster: gateway` is no longer supported in the management server - a mongo IDP without a dataPlaneId used here would fail to init. Enough to document or too unlikely to be a real concern? I have removed the "All changes are non-functional" statement from the PR description.

leleueri:
> > Question: is it possible to provide 2 implementation of the RepositoryScopeProvider one for MAPI with only the Management Scope and another one for the Gateway with all the scopes ? In that way the gravitee.yaml in the management side only need to define the "management" scope
> 
> Yes, I've split the `RepositoryScopeProvider` implementations, removing all but `management` scope from the management `gravitee.yml`. I added a safeguard to filter the sole gateway-scope service (`AuthenticationFlowContextServiceImpl`) from the management's `StandaloneConfiguration.java` (really a precaution since the gateway-scoped `AuthenticationFlowContextRepository` is already is wired with `@Lazy`, preventing it from being instantiated). Unfortunately we can't easily move `AuthenticationFlowContextService` out of the common service package because saml2-idp plugin depends on it.
> 
> The one concern is `repositories.system-cluster: gateway` is no longer supported in the management server - a mongo IDP without a dataPlaneId used here would fail to init. Enough to document or too unlikely to be a real concern? I have removed the "All changes are non-functional" statement from the PR description.

arf, didn't think about this use-system-cluster 🤔 but maybe we can safely fallback to the "default" dataplane in that case and in the documentation highlight this change a BC and advice before moving to 4.13 to use the datasource instead of the "use-system-cluster". For the AuthenticationFlowContextService, IMO it is OK to move it as soon as we update the saml2-idp plugin with a BC in the commit message(+ reference the change in the documentation & compatibility matrix) . Let me think about it.

lgw-gravitee:
q: should we update Helm scripts as well? Currently there is a _configmap-default-repositories.tpl which creates all scopes

leleueri (approved):

PR #8089, opened 2026-06-19
feat: migration tool - seeding

lgw-gravitee:
To test create local K8s cluster or use remote
`kind create cluster`

Mongo:
`AM_LICENSE_PATH=/home/lgw/Downloads/gravitee-universe-v4.key ./scripts/migration-test.mjs --from-tag 4.10.15 --to-tag 4.11.8 --provider k8s  run`

Postgres:
`AM_LICENSE_PATH=/home/lgw/Downloads/gravitee-universe-v4.key ./scripts/migration-test.mjs --from-tag 4.10.15 --to-tag 4.11.8 --provider k8s --db-type postgres run`

You can apply --namespace param to override default `gravitee-am` 

Sample output
```
🛠️ Running Migration Orchestration
📝 Tags: 4.10.15 -> 4.11.8

▶️ Stage: clean
🧹 Cleaning up K8s environment...
🗑️  Uninstalling Helm releases...
🧹 Cleaning up MongoDB...
🗝️  Cleaning up secrets...
💀 Deleting namespace: gravitee-am...
✅ Cleanup complete.

▶️ Stage: k8s:setup
🏗️  Setting up K8s environment...
🍃 Deploying standalone MongoDB...

▶️ Stage: deploy-from
🚀 Deploying AM version 4.10.15...
🔌 Starting port-forward: svc/am-mapi-management-api 8093:83...
🔌 Starting port-forward: svc/am-mapi-management-ui 8002:8002...
🔌 Starting port-forward: svc/am-gateway-dp1-gateway 8091:82...
🔌 Starting port-forward: svc/am-gateway-dp2-gateway 8092:82...

▶️ Stage: seed-baseline

> gravitee-am-test@4.10.0-SNAPSHOT migration:seed
> cross-env TS_NODE_PROJECT=tsconfig.json TS_NODE_TRANSPILE_ONLY=1 node -r ts-node/register -r tsconfig-paths/register migration-seeding/bootstrap.ts --version 4.10 --label alpha


▶️ Stage: verify-all-alpha
🔍 Verifying alpha domain — full stack...
PASS specs/migration/gateway.jest.spec.ts (33.787 s)
  migration Gateway data [data plane dp1]
    ✓ logs in the user and redirects to an MFA step (2754 ms)
    ✓ logs in with MFA, creates a token and introspects it (3160 ms)
    ✓ grants consent for a requested scope and creates a token carrying it (1616 ms)
    ✓ logs in a user backed by the custom identity provider and creates a token (1731 ms)
    ✓ exchanges a JWT assertion for a token via the seeded jwt-bearer extension grant and introspects it (232 ms)
  migration Gateway data [data plane dp2]
    ✓ logs in the user and redirects to an MFA step (2633 ms)
    ✓ logs in with MFA, creates a token and introspects it (2700 ms)
    ✓ grants consent for a requested scope and creates a token carrying it (1287 ms)
    ✓ logs in a user backed by the custom identity provider and creates a token (1583 ms)
    ✓ exchanges a JWT assertion for a token via the seeded jwt-bearer extension grant and introspects it (214 ms)

PASS specs/migration/mapi-data.jest.spec.ts
  migration MAPI data [data plane dp1]
    ✓ keeps the versioned MAPI data readable through the generated SDK (560 ms)
  migration MAPI data [data plane dp2]
    ✓ keeps the versioned MAPI data readable through the generated SDK (369 ms)

Test Suites: 2 passed, 2 total
Tests:       12 passed, 12 total
Snapshots:   0 total
Time:        35.865 s
Ran all test suites matching /specs\/migration|specs\/migration/i.
⚠️  7 ERROR line(s) in am-mapi:
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:15:23.109 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:15:29.100 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:15:36.695 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:15:37.288 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:15:37.593 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:18:05.507 [RxComputationThreadPool-1] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-mapi-management-api-8595bdccb-xvwdt/am-mapi-management-api] 10:18:10.603 [RxComputationThreadPool-1] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
⚠️  58 ERROR line(s) in am-gateway-dp1:
    [pod/am-gateway-dp1-gateway-797786d994-8cr25/am-gateway-dp1-gateway] 10:16:46.420 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-gateway-dp1-gateway-797786d994-8cr25/am-gateway-dp1-gateway] 10:16:46.429 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-gateway-dp1-gateway-797786d994-8cr25/am-gateway-dp1-gateway] 10:16:46.430 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-gateway-dp1-gateway-797786d994-8cr25/am-gateway-dp1-gateway] 10:16:46.430 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)
    [pod/am-gateway-dp1-gateway-797786d994-8cr25/am-gateway-dp1-gateway] 10:16:46.430 [graviteeio-node] [] ERROR i.g.node.monitoring.metrics.Metrics - Gravitee metrics is disabled. You need to enable it first (services.metrics.enabled=true)

  migration MAPI data [data plane dp2]
    ✓ keeps the versioned MAPI data readable through the generated SDK (501 ms)

### DIFF
--- a/gravitee-am-test/migration-seeding/bootstrap.ts
+++ b/gravitee-am-test/migration-seeding/bootstrap.ts
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export type SeedModule = {
+  seed(label: string): Promise<void>;
+};
+
+export function toMinorVersion(version: string): string {
+  const match = version.match(/^(\d+)\.(\d+)/);
+  if (!match) {
+    throw new Error(`Invalid AM version: ${version}`);
+  }
+  return `${match[1]}.${match[2]}`;
+}
+
+export function applyMissingSeedEnvironmentDefaults(): void {
+  process.env.AM_MANAGEMENT_URL ||= 'http://localhost:8093';
+  process.env.AM_MANAGEMENT_ENDPOINT ||= `${process.env.AM_MANAGEMENT_URL}/management`;
+  process.env.AM_GATEWAY_URL ||= 'http://localhost:8091';
+  process.env.AM_DOMAIN_DATA_PLANE_ID ||= 'dp1';
+  process.env.AM_DEF_ORG_ID ||= 'DEFAULT';
+  process.env.AM_DEF_ENV_ID ||= 'DEFAULT';
+  process.env.AM_ADMIN_USERNAME ||= 'admin';
+  process.env.AM_ADMIN_PASSWORD ||= 'adminadmin';
+}
+
+/**
+ * Seed one version's data set under a channel label.
+ * @param version  Which seed module to run (the data shape), e.g. "4.10". Normalized to major.minor.
+ * @param label    The channel/instance label used to name the seeded entities, e.g. "alpha" (baseline) or "beta" (newline).
+ */
+export async function runSeed(version: string, label: string): Promise<void> {
+  applyMissingSeedEnvironmentDefaults();
+  const minorVersion = toMinorVersion(version);
+  const seedModule = (await import(`./versions/${minorVersion}/seed`)) as SeedModule;
+  await seedModule.seed(label);
+}
+
+function readArg(name: string): string | undefined {
+  const index = process.argv.indexOf(name);
+  return index >= 0 ? process.argv[index + 1] : undefined;
+}
+
+if (require.main === module) {
+  const version = readArg('--version');
+  const label = readArg('--label');
+  if (!version) {
+    throw new Error('--version is required');
+  }
+  if (!label) {
+    throw new Error('--label is required');
+  }
+  runSeed(version, label).catch((error) => {
+    console.error(error);
+    process.exit(1);
+  });
+}
--- a/gravitee-am-test/specs/migration/fixture/gateway-fixture.ts
+++ b/gravitee-am-test/specs/migration/fixture/gateway-fixture.ts
@@ -1,0 +1,294 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  extractXsrfTokenAndActionResponse,
+  getWellKnownOpenIdConfiguration,
+  performFormPost,
+  performGet,
+  performPost,
+} from '@gateway-commands/oauth-oidc-commands';
+import { getBase64BasicAuth } from '@gateway-commands/utils';
+import { initiateLoginFlow, login } from '@gateway-commands/login-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { expect } from '@jest/globals';
+import { TOTP } from 'otpauth';
+import { extractSharedSecret } from '../../gateway/mfa/fixture/mfa-extract-fixture';
+import { getApplicationApi, getDomainApi, getUserApi } from '../../../api/commands/management/service/utils';
+import { Application } from '../../../api/management/models/Application';
+import { Domain } from '../../../api/management/models/Domain';
+import {
+  getApplicationName,
+  getApplicationWithCustomIdpName,
+  getConsentApplicationName,
+  getCustomIdpUserName,
+  getDefaultStoreUserName,
+  getDomainName,
+  getServiceApplicationName,
+  JWT_BEARER_GRANT_TYPE,
+} from '../../../migration-seeding/seed';
+
+export type GatewayFixtureOptions = {
+  applicationName?: string;
+  userName?: string;
+  password?: string;
+};
+
+const DEFAULT_SEED_PASSWORD = 'SomeP@ssw0rd';
+
+export type GatewayFixture = {
+  application: Application;
+  domain: Domain;
+  loginUser: () => Promise<any>;
+  loginUserWithMfa: () => Promise<any>;
+  loginUserAndConsent: () => Promise<any>;
+  requestToken: (finalAuthorizeResponse: any) => Promise<any>;
+  requestJwtBearerToken: (assertion: string) => Promise<any>;
+  introspectToken: (accessToken: string) => Promise<any>;
+};
+
+export function createCustomIdpGatewayFixture(label: string): Promise<GatewayFixture> {
+  return createGatewayFixture(label, {
+    applicationName: getApplicationWithCustomIdpName(label),
+    userName: getCustomIdpUserName(label),
+  });
+}
+
+export function createConsentGatewayFixture(label: string): Promise<GatewayFixture> {
+  return createGatewayFixture(label, {
+    applicationName: getConsentApplicationName(label),
+  });
+}
+
+export function createExtensionGrantGatewayFixture(label: string): Promise<GatewayFixture> {
+  return createGatewayFixture(label, {
+    applicationName: getServiceApplicationName(label),
+  });
+}
+
+export async function createGatewayFixture(label: string, options: GatewayFixtureOptions = {}): Promise<GatewayFixture> {
+  const domainName = getDomainName(label);
+  const applicationName = options.applicationName ?? getApplicationName(label);
+  const userName = options.userName ?? getDefaultStoreUserName(label);
+  const userPassword = options.password ?? DEFAULT_SEED_PASSWORD;
+  const accessToken = await requestAdminAccessToken();
+  const domainApi = getDomainApi(accessToken);
+  const applicationApi = getApplicationApi(accessToken);
+  const userApi = getUserApi(accessToken);
+  const organizationId = process.env.AM_DEF_ORG_ID;
+  const environmentId = process.env.AM_DEF_ENV_ID;
+  const domains = await domainApi.listDomains({ organizationId, environmentId, q: domainName });
+  const domain = domains.data?.find((candidate) => candidate.name === domainName);
+
+  expect(domain).toBeDefined();
+  const applications = await applicationApi.listApplications({
+    organizationId,
+    environmentId,
+    domain: domain.id,
+    q: applicationName,
+  });
+  const matchedApplication = applications.data?.find((candidate) => candidate.name === applicationName);
+
+  expect(matchedApplication).toBeDefined();
+  const application = await applicationApi.findApplication({
+    organizationId,
+    environmentId,
+    domain: domain.id,
+    application: matchedApplication.id,
+  });
+  const openIdConfiguration = (await getWellKnownOpenIdConfiguration(domain.hrid)).body;
+
+  async function loginUser(scope?: string) {
+    const authResponse = await startAuthorize(scope);
+
+    return login(authResponse, userName, application.settings.oauth.clientId, userPassword);
+  }
+
+  async function startAuthorize(scope?: string) {
+    if (!scope) {
+      return initiateLoginFlow(application.settings.oauth.clientId, openIdConfiguration, domain);
+    }
+    const clientId = application.settings.oauth.clientId;
+    const redirectUri = application.settings.oauth.redirectUris[0];
+    const params = `?response_type=code&client_id=${clientId}&redirect_uri=${redirectUri}&scope=${encodeURIComponent(scope)}`;
+
+    return performGet(openIdConfiguration.authorization_endpoint, params).expect(302);
+  }
+
+  async function loginUserWithMfa() {
+    const afterMfaResponse = await loginAndCompleteMfa();
+
+    expect(afterMfaResponse.headers['location']).toContain('code=');
+    return afterMfaResponse;
+  }
+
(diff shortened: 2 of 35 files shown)

### CODE REVIEW

File: scripts/migration-tool/README.md [hunk 1]
```diff
@@ -28,21 +28,38 @@ All commands must be run from the **repository root** (e.g. `./scripts/migration
 
 ## Workflow and design
 
+**Terminology:** seeded data is keyed by a **channel label**, not by version, so the baseline and newline data sets stay distinct even when both sides share the same minor version (e.g. a `4.10.7 → 4.10.8` patch migration). The *baseline* channel (label **`alpha`**) is seeded with the **`--from-tag`** data and lives in domain `migration-seeded-domain-alpha`; the *newline* channel (label **`beta`**) is seeded with the **`--to-tag`** data and lives in domain `migration-seeded-domain-beta`. The `mapi`/`all` prefix on a verify stage only describes the system state at that point (Management API upgraded vs. full stack); the suffix (`baseline`/`newline`) picks which channel is asserted.
```
wajb: discussion: I have never encountered "newline" used in the sense of contrasting to a baseline 🤔. Is this an invention for this particular migration tool or would other readers be confused too? ("Target" was the term generally used previously, and "upgrade" was used in at least one command.)

"Alpha" and "beta" are also used too, hardcoded in places. I'm not sure if these are a different naming convention for the same thing.

scripts/migration-tool/README.md
```diff
@@ -28,21 +28,38 @@ All commands must be run from the **repository root** (e.g. `./scripts/migration
...
+
+> The full pipeline requires a **concrete `--to-tag`** (e.g. `4.11.8`), not `latest`, because it seeds and verifies the new version. The channel asserted is derived automatically from the stage (baseline → `alpha`, newline → `beta`) — no `--test-label` needed.
```
wajb: question: what does "full pipeline" here mean? Also, if `latest` is not supported, is there any reason still to allow this value?

gravitee-am-test/api/commands/gateway/utils.ts
```diff
@@ -22,7 +22,7 @@ export function getBase64BasicAuth(clientId: string, clientSecret: string) {
 }
 
 export function applicationBase64Token(app: Application) {
-  return getBase64BasicAuth(app.settings.oauth.clientId, app.settings.oauth.clientSecret);
+  return getBase64BasicAuth(app.settings.oauth.clientId, app.settings.oauth.clientId);
```
wajb: issue: not sure this makes sense for most usages. If it's needed for some scenarios, could the callers simply replace the secret?

scripts/migration-tool/env/k8s/db/db-mongodb.yaml
```diff
@@ -11,9 +11,22 @@ auth:
   enabled: true
   # Bitnami chart key: K8s secret *name* only (no credentials in repo). Aqua false positive: not a password.
   existingSecret: "mongo-mongodb-auth"
-  # One user 'gravitee' with access to each dataplane DB (parallel arrays: usernames[i], passwords[i], databases[i])
-  usernames: ["gravitee", "gravitee", "gravitee"]
-  databases: ["gravitee", "gravitee-dp1", "gravitee-dp2"]
+  # Single user 'gravitee' (authSource = gravitee db). Multi-database access (gravitee-dp1/dp2)
+  # is granted via initdbScripts below: the Bitnami parallel-array form creates a username only
+  # once, so listing 'gravitee' three times did NOT grant it readWrite on the dataplane DBs
+  # (the dataplane gateways then failed with "not authorized on gravitee-dp1").
```
wajb: nitpick: nice fix, but this explanation can be removed IMO.

scripts/migration-tool/lib/Orchestrator.mjs
```diff
@@ -48,47 +48,70 @@ export class Orchestrator {
             case 'deploy-from':
                 await this.provider.deploy(this.options.fromTag);
                 break;
-            case 'verify-baseline':
-                await this.runTests('🔍 Running baseline tests...', 'ci:management:parallel', 'specs/management');
+            case 'seed-baseline':
+            case 'seed': // alias (backward compatible)
```
wajb: suggestion: we don't need to keep the old command.

scripts/migration-tool/lib/Orchestrator.mjs
```diff
@@ -48,47 +48,70 @@ export class Orchestrator {
...
-            case 'verify-mapi':
-                await this.runTests('🔍 Verifying MAPI upgrade...', 'ci:management:parallel', 'specs/management');
-                break;
             case 'upgrade-gw':
                 await this.provider.upgradeGw(this.options.toTag);
                 break;
-            case 'verify-all':
-                await this.runTests('🔍 Final verification...', 'ci:gateway', 'specs/gateway');
+            case 'seed-newline':
+            case 'seed-upgrade': // alias (backward compatible)
```
wajb: suggestion: we don't need to keep the old command.

scripts/migration-tool/README.md
```diff
@@ -167,33 +186,69 @@ Requires `CIRCLECI_TOKEN`. Sends parameters (from-tag, to-tag, db-type, provider
...
+The migration tool seeds deterministic data through the generated Management API SDK in `gravitee-am-test/api/management`.
+
+Seed files live in `gravitee-am-test/migration-seeding/versions/<major.minor>/seed.ts`. Each seed file must be idempotent and use stable identifiers, names, or metadata so migration specs can find the same objects after upgrade and downgrade. A seed file exposes `seed(label)` and names every entity by that **label** (not the version), so the same version's data can be seeded twice under different labels without colliding.
+
+The seed runner takes `--version <major.minor>` (which seed module / data shape to run) and `--label <name>` (the channel suffix used for naming). The `seed-baseline` stage runs the `--from-tag` module under label `alpha`; the `seed-newline` stage runs the `--to-tag` module under label `beta`. Because each channel targets a single version+label, this works for patch migrations too (e.g. `4.10.7 → 4.10.8` seeds the `4.10` module twice, as `alpha` and `beta`).
+
+Example local seed run against an already deployed Management API:
+
+```bash
+npm --prefix gravitee-am-test run migration:seed -- --version 4.10 --label alpha
```
wajb: question: why not use the script/orchestrator commands?

scripts/migration-tool/README.md
```diff
@@ -167,33 +186,69 @@ Requires `CIRCLECI_TOKEN`. Sends parameters (from-tag, to-tag, db-type, provider
...
+
+Example migration test run against an already seeded environment:
+
+```bash
+AM_MIGRATION_TEST_LABEL=alpha npm --prefix gravitee-am-test run ci:migration
+```
+
+When running through the migration tool, `AM_MIGRATION_TEST_LABEL` is set automatically per stage (`alpha` for baseline stages, `beta` for newline stages; ad-hoc generic verify stages use `--test-label`, default `alpha`).
+
+Retrospective coverage for older releases is limited by the generated SDK and API contract available in the current branch. Prefer adding seeds before or during the release where the data shape is introduced. For older releases, keep seed data to API fields accepted by that older AM version.
```
wajb: discussion: this is a crucial limitation. We cannot reliably seed data for older versions using the current SDK, meaning we cannot target the edge cases the migration tests are really for.

The CI integration part will be key. Do you intend to sequence different AM checkouts in CircleCI for the seeding stages (our initial idea)? Or maybe there could be a way to embed a self-seeding command in (private?) mapi images?

If this is solved then we won't need to discourage covering behaviours across schema changes.

gravitee-am-test/migration-seeding/bootstrap.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export function toMinorVersion(version: string): string {
+  const match = version.match(/^(\d+)\.(\d+)/);
+  if (!match) {
+    throw new Error(`Invalid AM version: ${version}`);
+  }
+  return `${match[1]}.${match[2]}`;
+}
+
+export function applyDefaultSeedEnvironment(): void {
```
leleueri: nickpick: my understanding is this method is using the env vers or the default if the var is not defined. Should id be named "applySeedEnvironment" instead?

scripts/migration-tool/README.md
lgw-gravitee: I wanted to distinct data created before and after migration. The idea was 
- baseline (before upgrade) newline (after upgrade). They refer to created data.  
- alpha/beta are suffixes for domains (in case of upgrading the same major version 4.10.1 -> 4.10.2)

do you think the follwoing would be better?
domain-base / domain-new
and 
seed base/seed new?

gravitee-am-test/api/commands/gateway/utils.ts
lgw-gravitee: thanks, I made it by a mistake

scripts/migration-tool/README.md
lgw-gravitee: yes, I added checking out GIT worktress capability for seeding from specific version

scripts/migration-tool/README.md
lgw-gravitee: I changed that.

scripts/migration-tool/README.md
lgw-gravitee: I added support for 'latest'. it calculates the real version in the beginigng

gravitee-am-test/migration-seeding/bootstrap.ts
lgw-gravitee: I corrected to applyMissingSeedEnvironmentDefaults

.superpowers/sdd/final-fix-report.md
leleueri, line 1: question: is it something we should commit ?

scripts/migration-tool/README.md
wajb: Yes, I think that's clearer, thanks.

scripts/migration-tool/lib/Orchestrator.mjs
```diff
@@ -106,6 +150,51 @@ export class Orchestrator {
...
+    /**
+     * Seed one version under a channel label. When a SeedWorktree is configured (and enabled), the
+     * seed runs from a worktree of `tag` so it uses that version's own SDK + scripts; otherwise (or
+     * when the tag predates the seeding framework) it falls back to the current checkout.
+     */
+    async seedStage(tag, label) {
+        const args = ['--version', toMinorVersion(tag), '--label', label];
+        let seedDir = null;
+        if (this.seedWorktree && this.options.seedFromWorktree) {
+            seedDir = await this.seedWorktree.resolveSeedDir(tag);
```
wajb, line 167: thought: it's still going to be difficult to maintain `gravitee-am-test/migration-seeding/versions/<version>` folders in parallel. API changes may easily break seeding of older versions, so, after fixing, what's left will always be identical for each version 🤔.

I wonder if only the worktree approach is valid and we would be better off with a single seeding folder that is free to differ between branches. Wdyt? Is there still value in keeping the seeding as stacked folders I'm overlooking?

gravitee-am-test/specs/migration/fixture/gateway-fixture.ts
```diff
@@ -0,0 +1,294 @@
...
+    const finalAuthorizeResponse = await performGet(consentResponse.headers['location'], '', {
+      Cookie: consentResponse.headers['set-cookie'],
+    }).expect(302);
+
+    expect(finalAuthorizeResponse.headers['location']).toContain('code=');
+    return finalAuthorizeResponse;
+  }
+
+  async function loginAndCompleteMfa() {
+    await resetUserFactors();
```
leleueri, line 170: discussion: I understand why we are doing this resetUserFactor, this is to make sure that the TOTP shared secret is present with the right value. But by doing this we are not asserting that the existing facor instance is usable by the user, we always use a "fresh" enrolled factor.

This is fine for this PR but IMO in addition of this TOTP case, we should use an EMAIL or SMS factor and in the fixture function, we should adapt the execution based on the Location header value (if the header target enrollment endpoint, we enroll then challenge otherwise we are just challenging the user)

scripts/migration-tool/lib/Orchestrator.mjs
leleueri: I tempt to agree with you @wajb , with the worktree option having dedicated seed folder per version may not be useful... 🤔 however wondering if this per version folder can make sens for the test cases because if we had a test case for agent introduced in 4.12 for example, those test cases make no sens for the seed version 4.11...

leleueri (approved):

scripts/migration-tool/lib/Orchestrator.mjs
leleueri: we will discuss this later and iterate on it.

gravitee-am-test/specs/migration/fixture/gateway-fixture.ts
leleueri: we will discuss this later and iterate

PR #8097, opened 2026-06-22
fix: changed resend CTA for Call factor

lgw-gravitee:
fixes AM-7172

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/templates/i18n/messages_fr.properties
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/templates/i18n/messages_fr.properties
@@ -171,12 +171,17 @@ mfa_challenge.recovery.code=Veuillez saisir l'un de vos codes de secours
 mfa_challenge.error=Code invalide
 mfa_challenge.button.submit=Vérifier
 mfa_challenge.button.resend=Renvoyer le code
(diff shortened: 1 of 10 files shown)


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/templates/i18n/messages_fr.properties
```diff
@@ -171,12 +171,17 @@ mfa_challenge.recovery.code=Veuillez saisir l'un de vos codes de secours
 mfa_challenge.error=Code invalide
 mfa_challenge.button.submit=V�rifier
 mfa_challenge.button.resend=Renvoyer le code
+mfa_challenge.button.call.resend=Rappeler
 mfa_challenge.resend.success=Un nouveau code a �t� envoy�
+mfa_challenge.call.resend.success=Un nouvel appel a \u00E9t\u00E9 lanc\u00E9
```
leleueri: question: it is display properly ? other message are not using unicode

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/templates/i18n/messages_fr.properties
wajb: I think these need to be edited as `ISO-8859-1`. `file -I` returns
```
text/plain; charset=iso-8859-1
```

PR #8103, opened 2026-06-23
fix: removed oauth2-repository probe from MAPI

lgw-gravitee:


wajb (approved):

wajb:
@lgw-gravitee this commit was merged as part of https://github.com/gravitee-io/gravitee-access-management/pull/8101 👍.

PR #8111, opened 2026-06-24
fix: prevent domains from being silently dropped during gateway startup

leleueri:
## Summary

The gateway's `EventManagerImpl` had a thread-safety bug where `DomainEvent.DEPLOY` publishes would race with concurrent listener subscriptions during parallel domain deployment, causing a `ConcurrentModificationException` that silently aborted the event delivery. This meant some domains never reached the reactor and were lost — a non-deterministic subset per instance, since thread interleaving differed each run.

## Changes

- **EventManagerImpl**: Use `computeIfAbsent + CopyOnWriteArrayList` so listeners can be added concurrently while events are published, and fix the get-then-put race that could drop listeners.
- **SyncManager**: Materialize `findAll()` with `.toList()` before deploying, so the management cursor is not held open for the full deployment duration (removing exposure to the MongoDB `cursorMaxTime` 60s timeout).
- **EventManagerTest**: Add concurrency regression tests for the CME and listener-loss races.

## Fix for

Fixes [AM-7179](https://jira.graviteesource.com/browse/AM-7179) — domains silently dropping during startup sync on multi-instance deployments.

## Testing

- All existing `EventManagerTest` cases pass (134 parameterized + 2 new concurrency tests).
- Concurrency tests hammer publish + subscribe for 2000 iterations on 16 threads; assert no `ConcurrentModificationException` and no lost listeners.

## Deployment

Domains will now load reliably during gateway startup, with no silent loss due to event-delivery races.

[AM-7179]: https://gravitee.atlassian.net/browse/AM-7179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
@@ -15,7 +15,6 @@
  */
 package io.gravitee.am.gateway.event;
 
-import com.google.common.collect.Lists;
 import io.gravitee.am.common.event.DomainEvent;
 import io.gravitee.am.common.event.EventManager;
 import io.gravitee.am.common.event.ReporterEvent;
@@ -33,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.stream.Collectors;
 
 import static java.util.Optional.ofNullable;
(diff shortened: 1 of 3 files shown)


gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
```diff
@@ -170,14 +172,10 @@ private <T extends Enum<T>> List<EventListenerWrapper> getEventListeners(Class<T
...
-
-        if (listeners == null) {
-            listeners = Collections.synchronizedList(new ArrayList<>());
-            this.listenersMap.put(new ComparableEventType(eventType, domain), listeners);
-        }
-
-        return listeners;
+        // computeIfAbsent makes get-or-create atomic (the previous get-then-put could race and drop
+        // a listener), and CopyOnWriteArrayList allows lock-free, CME-safe iteration on publish while
+        // listeners are added concurrently during parallel domain deployment.
```
wajb: nitpick: this comment is a bit noisy in that it describes behaviour of code that no longer exists.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/test/java/io/gravitee/am/gateway/event/EventManagerTest.java
```diff
@@ -147,6 +158,110 @@ private static Stream<Arguments> params_that_must_publish_event_or_not_without_d
                         Arguments.of(event, predicateNull));
     }
 
+    /**
+     * Regression test for the {@link java.util.ConcurrentModificationException} raised by
```
wajb: thought: IMO this test is more of a general safety net than a specific regression test, since it catches all exceptions.

wajb (approved):

PR #8137, opened 2026-06-30
fix(AM-7194): preload protected-resource permissions in AuthGuard to prevent blank page on refresh

sagarchauhan-qa:

https://gravitee.atlassian.net/browse/AM-7194

**Root cause:** 
AuthGuard runs before McpServerPermissionsResolver loads protected_resource_* permissions, so the guard returns false and cancels navigation.

**Fix:** 
Added a protected_resource branch to AuthGuard to preload permissions inline, matching the existing pattern for domain and application permissions.

## :memo: Test scenarios 

NA
NA

NA

### DIFF
--- a/gravitee-am-ui/src/app/guards/auth-guard.service.ts
+++ b/gravitee-am-ui/src/app/guards/auth-guard.service.ts
@@ -47,59 +47,40 @@ export class AuthGuard {
     const requiredPerms = route.data.perms.only;
     combineSources.push(!this.authService.isAuthenticated() ? this.authService.userInfo() : of(this.authService.user()));
 
-    if ((requiredPerms[0].startsWith('domain') || requiredPerms[0].startsWith('application')) && requiredPerms[0] !== 'domain_create') {
-      // check if the authenticated user can navigate to the next route (domain settings or application settings)
-      const environmentId = route.paramMap.get('envHrid');
-      const domainId = route.paramMap.get('domainId');
-      const appId = route.paramMap.get('appId');
+    const environmentId = route.paramMap.get('envHrid');
+    const domainId = route.paramMap.get('domainId');
+    const appId = route.paramMap.get('appId');
+    const mcpServerId = route.paramMap.get('mcpServerId');
+    const perm = requiredPerms[0];
 
-      if (environmentId && !this.authService.environmentPermissionsLoaded()) {
-        combineSources.push(this.environmentService.permissions(environmentId));
-      }
+    const needsEnv = !!environmentId && !this.authService.environmentPermissionsLoaded() && perm !== 'domain_create';
+    const needsDomain = !!domainId && !this.authService.domainPermissionsLoaded() && perm !== 'domain_create';
+    const needsApp = !!appId && !this.authService.applicationPermissionsLoaded() && perm.startsWith('application');
+    const needsProtectedResource =
+      !!mcpServerId && !this.authService.protectedResourcePermissionsLoaded() && perm.startsWith('protected_resource');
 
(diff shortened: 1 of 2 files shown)


gravitee-am-ui/src/app/guards/auth-guard.service.ts
```diff
@@ -72,6 +74,13 @@ export class AuthGuard {
           ),
         );
       }
+    } else if (requiredPerms[0].startsWith('protected_resource')) {
```
wajb: issue: I suspect we need to preload domain-level (`PROTECTED_RESOURCE*`) ACLs, mirroring what we do with applications. Otherwise this will only reliably fix the issue for users with resource-specific roles.

sagarchauhan-qa: Updated to also preload domain and environment level permissions for protected_resource routes, mirroring the existing application pattern.

wajb: suggestion: this appears functionally correct but the function itself has become complex and hard to maintain, so we could consider separating the permission requests by resource type:

```ts
if (needsEnv) combineSources.push(envPermissions(...));
if (needsDomain) combineSources.push(domainPermissions(...));
if (needsApp) combineSources.push(appPermissions(...));
if (needsProtectedResource) combineSources.push(protectedResourcePermissions(...));
return combineLatest(combineSources).pipe(map(() => isAuthorized(...)));
```
where, for example, `needsProtectedResource` would be determined by
```ts
const mcpServerId = route.paramMap.get('mcpServerId');
const needsProtectedResource =
  !!mcpServerId &&
  !this.authService.protectedResourcePermissionsLoaded() &&
  requiredPerms[0].startsWith('protected_resource');
```

wajb (approved):

gravitee-am-ui/src/app/guards/auth-guard.service.ts
sagarchauhan-qa: Thanks @wajb 
Done. Refactored canActivate to use flat boolean flags (needsEnv, needsDomain, needsApp, needsProtectedResource) instead of the nested if/else-if. Logic is the same, just cleaner and easier to maintain.

gravitee-am-ui/src/app/guards/auth-guard.service.ts
```diff
@@ -45,33 +47,40 @@ export class AuthGuard {
...
+                }
+                return of([]);
+              }),
             ),
           ),
-        );
-      }
+        ),
+      );
     }
```
wajb, line 78: suggestion: IMO we don't need this chain anymore - in fact, it couples fetching relating to the different resources even though we have individual clauses to make the decisions.

We should be able to decouple the preloading like:

```suggestion
    if (needsDomain) {
      combineSources.push(
        this.domainService.getById(domainId).pipe(
          mergeMap((domain) => this.domainService.permissions(domain.id)),
        ),
      );
    }
    if (needsApp) {
      combineSources.push(
        this.domainService.getById(domainId).pipe(
          mergeMap((domain) => this.applicationService.permissions(domain.id, appId)),
        ),
      );
    }
    if (needsProtectedResource) {
      combineSources.push(
        this.domainService.getById(domainId).pipe(
          mergeMap((domain) => this.protectedResourceService.permissions(domain.id, mcpServerId)),
        ),
      );
    }
```

sagarchauhan-qa: @wajb Done.

wajb (approved):

PR #8154, opened 2026-07-01
AM-6693: BREAKING CHANGE: remove passwordSettings from Application level

lgw-gravitee:
BREAKING CHANGE: remove passwordSettings from Application level


gravitee-am-service/src/main/java/io/gravitee/am/service/model/PatchApplicationSettings.java
```diff
@@ -50,6 +50,15 @@ public class PatchApplicationSettings {
...
+    /**
+     * Application-level password settings have been removed. This field is retained only so that
+     * existing clients still sending {@code passwordSettings} do not get a deserialization error;
+     * the value is ignored. Password policy is now resolved from the identity provider's assigned
+     * policy or the domain default policy.
+     *
+     * @deprecated application-level password policy no longer exists; this field is a no-op.
+     */
+    @Deprecated
     private Optional<PatchPasswordSettings> passwordSettings;
```
leleueri: question: why do we have to keep it ? Is the REST API call failing if this field is missing ?

lgw-gravitee: yes

leleueri: not sure to understand why

lgw-gravitee: if PATCH body is
`{"name": "test", "settings": {"passwordSettings": {"minLength": 1}}}`
it must be accepted due to backward compatibilty. If I remove `passwordSettings` field, previous request would be invalid
`{"message":"Property [passwordSettings] is not recognized as a valid property","http_status":400}`

lgw-gravitee: accroding to converstaion we decided to rmeove it.

leleueri (approved):

PR #8155, opened 2026-07-02
fix: make vhost domain comparison case-insensitive

lgw-gravitee:
## Summary
- Make gateway VHost host matching case-insensitive.
- Make VHost validation detect overlapping paths when the same host uses different casing.
- Add regression coverage for routing and validation.

## How to test
Automated tests:
- `mvn test -pl gravitee-am-service -Dtest=VirtualHostValidatorTest`
- `mvn test -pl gravitee-am-gateway/gravitee-am-gateway-reactor -Dtest=VHostRouterTest`

Manual local test:
1. Start a local AM stack with the Gateway available on `localhost:8092`.
2. In the AM Console, open a security domain and go to `Settings` > `Entrypoints`.
3. Enable VHost mode and add a virtual host, for example:
   - Host: `am-test.localhost`
   - Path: `/my-domain` or `/`
4. Save the domain and wait for the Gateway to redeploy/reload it.
5. Verify both host casings route to the same domain response:
   - `curl -i -H "Host: am-test.localhost" http://localhost:8092/my-domain/.well-known/openid-configuration`
   - `curl -i -H "Host: AM-TEST.LOCALHOST" http://localhost:8092/my-domain/.well-known/openid-configuration`
6. If testing in a browser, add `127.0.0.1 am-test.localhost` to `/etc/hosts` and open `http://am-test.localhost:8092/my-domain/.well-known/openid-configuration`. Prefer curl with an explicit `Host` header to verify uppercase host casing because browsers may normalize displayed hostnames.

fixes AM-6973

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/router/VHostRouter.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/router/VHostRouter.java
@@ -84,7 +84,7 @@ public class VHostRouter extends RouterImpl {
         super(vertx);
         this.domain = domain;
         this.vhost = vhost;
-        this.vhostPattern = Pattern.compile(Pattern.quote(vhost.getHost()));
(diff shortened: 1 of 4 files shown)


wajb (approved):
Tested successfully.

PR #8159, opened 2026-07-02
feat(console): identity-provider picker for device-notifier configuration - AM-7205

otwebti:
Part of #11594.

This is the consolidated **federated CIBA** contribution for Gravitee AM — the SPI, core, notifier-plugin, and console changes that were previously split across the stacked PRs (#8157 / #8158, now closed) brought together on a single branch. It lets a domain complete a CIBA (`backchannel`) authentication by delegating the user-consent leg to an upstream OIDC provider, then establishing the local identity from the upstream `id_token`.

## What this adds

**Federated CIBA (core + notifier plugin)**
- Non-local `login_hint` / `login_hint_token` resolution and RFC 9396 `authorization_details` (RAR) relay on `bc-authorize`, with federated completion: at callback the local user is JIT-provisioned from the upstream `id_token`, JWKS-verified through the registered OIDC identity provider's own pipeline.
- A CIBA-federation **device-notifier plugin** performing the `bc-authorize` → poll → callback exchange against the upstream provider. Federation behaviours (RAR, non-local hints) are **intrinsic to the notifier type via `capabilities()`**, not domain flag checkboxes.
- A **cross-witness** guard: the poller fails closed unless the `authorization_details` echoed by the upstream provider match what was relayed.

**Vendor-neutral consent-relay seam (notifier plugin)**
- The notifier applies an optional, per-notifier `ConsentRelayStrategy` to a request's `authorization_details` before relay. Implementations are discovered on the plugin classpath via `SpringFactoriesLoader` and selected by id. **The core plugin ships no strategy implementations** — when `consentRelayStrategy` is blank the details are relayed unchanged (raw RAR); a non-blank value selects a discovered transform and fails fast if the id is unknown. Any vendor-specific consent rendering lives in a separate drop-in module, out of this contribution.

**Console (identity-provider picker + masking)**
- A device-notifier config schema property that references an identity provider renders as a **picker** of the domain's IdPs, falling back to a text input (and surfacing the failure) if the IdPs can't be loaded.
- Sensitive config fields are **masked** in the form (cosmetic; mirrors the identity-provider form). Generic — any device-notifier schema can use these; not specific to federated CIBA.

## Behaviour change / migration

`consentRelayStrategy`'s default changes from `passthrough` to **blank**. Blank relays `authorization_details` unchanged (raw RAR). `passthrough` is no longer a recognized value — a non-blank value must name an installed consent-relay strategy, and an unknown id fails fast. Any stored notifier config still set to `passthrough` must be cleared (blank) or set to a valid installed strategy id. The feature has not shipped in a release, so this affects only in-flight/dev configurations.

## Testing

- Unit + integration coverage across the SPI, the notifier plugin (raw-vs-transformed relay, registry discovery + unknown-id fail-fast, cross-witness), and the console (picker, load-failure paths, schema handling).
- Validated end-to-end natively against a real upstream OIDC provider with a push-based device authenticator, in both flows: with RFC 9396 `authorization_details` (token carries the RAR; cross-witness matches) and identity-only (no RAR; the consent-relay strategy is correctly not invoked).

## Commits

- `feat(authdevice-notifier)`: notifier capabilities + gateway-provided callback
- `feat(identityprovider)`: retrieveUserFromTokenResponse for back-channel
- `refactor(identityprovider)`: add `profile(Token, AuthenticationContext)` overload
- `fix(authdevice-notifier)`: guard null notifier id; close test Vertx
- `feat(ciba)`: federated CIBA support — non-local hints, RFC 9396 RAR, federated completion
- `feat(authdevice-notifier)`: CIBA-federation notifier plugin (incl. the vendor-neutral consent-relay SPI seam)
- `feat(console)`: device-notifier identity-provider picker + sensitive field masking

related-to AM-7205


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/service/request/CibaAuthenticationRequestResolver.java
```diff
@@ -84,6 +88,23 @@ public Single<CibaAuthenticationRequest> resolve(CibaAuthenticationRequest authR
...
+                // username/email. The user-search filter rejects any
+                // value containing the injection-guard chars " \ ; { } $ ^, so feeding such a
+                // hint to findByCriteria throws IllegalArgumentException and the request 500s.
+                // Guard BEFORE the local lookup: if the hint is not locally resolvable, skip
+                // findByCriteria entirely. When federation is on, relay the hint verbatim with
+                // NO subject — identity is established downstream at completion
+                // (validateUserResponse). Otherwise reject cleanly as a 400.
+                if (!isLocallyResolvableHint(authRequest.getLoginHint())) {
+                    if (isFederatedHintResolutionEnabled()) {
+                        return Single.just(authRequest);
```
leleueri, line 103: discussion: the "iss_sub" format has been introduced in the RFC 9493 and provide the iss value into the structure.  It may worth to check that the iss match the IDP if the federation is enabled.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestAcknowledgeHandler.java
```diff
@@ -182,6 +227,54 @@ public void handle(RoutingContext context) {
...
+    private static String buildCallbackUrl(RoutingContext context) {
+        String absoluteUri = context.request().absoluteURI();
+        if (absoluteUri == null) {
+            return null;
+        }
+        // Strip query string
+        int queryIndex = absoluteUri.indexOf('?');
+        String base = (queryIndex >= 0) ? absoluteUri.substring(0, queryIndex) : absoluteUri;
+        if (base.endsWith(CIBAProvider.AUTHENTICATION_ENDPOINT)) {
+            return base + "/callback";
```
leleueri:
```suggestion
            return CIBAProvider.AUTHENTICATION_CALLBACK_ENDPOINT;
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/authdevice/notifier/impl/AuthenticationDeviceNotifierManagerImpl.java
```diff
@@ -124,6 +124,17 @@ public Collection<AuthenticationDeviceNotifierProvider> getAuthDeviceNotifierPro
         return this.deviceNotifierProviders.values();
     }
 
+    @Override
+    public AuthenticationDeviceNotifier getAuthDeviceNotifier(String notifierId) {
+        // deviceNotifiers is a ConcurrentHashMap, which rejects null keys; guard so a missing
+        // notifier id resolves to "no notifier" rather than throwing.
+        return notifierId == null ? null : this.deviceNotifiers.get(notifierId);
+    }
```
leleueri: discussion: I don't think we need this new method here. From what I can see, it's only used in two places to extract the IdentityProviderId from the notifier configuration. To keep our core product clean, we should avoid embedding plugin-specific processing inside it.

Instead, would it make more sense to add a method to AuthenticationDeviceNotifierProvider that handles extracting this information based on its own configuration structure? For example, it could expose a getFederationConfiguration() method that returns a FederationConfiguration interface (maybe with an Optional wrapper), which in turn provides getIdentityProvider().

Thought ?

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
```diff
@@ -0,0 +1,134 @@
...
+    public static final String ID_TOKEN = "id_token";
+    public static final String ACCESS_TOKEN = "access_token";
+
+    @Autowired(required = false) private CibaFederationAuthenticationDeviceNotifierConfiguration configuration;
+    @Autowired(required = false) @Qualifier("cibaFederationWebClient") private WebClient webClient;
+    @Autowired(required = false) private Vertx vertx;
+    @Autowired(required = false) private OidcDiscoveryResolver discoveryResolver;
+
+    /** Non-null only when injected via forTest seam; null in production (client built per-request). */
+    private CibaClient testAuth0;
```
leleueri: suggestion: Is it possible to rely on standard Provider initialization here instead of introducing a dedicated CibaClient just for testing purposes? It would be great to avoid adding test-specific code to the production logic if we can leverage the existing setup.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
```diff
@@ -0,0 +1,134 @@
...
+        final String tid = Objects.requireNonNull(request.getTransactionId(), "transactionId must not be null");
+        final String callbackUrl = Objects.requireNonNull(request.getCallbackUrl(),
+                "callbackUrl must be supplied by the gateway on ADNotificationRequest");
+        final java.util.List<java.util.Map<String, Object>> relayedRar = request.getAuthorizationDetails() == null ? null
+                : consentStrategy.relay(request.getAuthorizationDetails());
+        final String scope = conn.scope();
+        // resourceAudience is a notifier-config value sent as the downstream `audience` form parameter;
+        // the IdP config cannot carry it. Only resolved when building the per-request client (not used
+        // when testAuth0 is injected).
+        final CibaClient auth0 = testAuth0 != null ? testAuth0
```
leleueri: suggestion: Let's rename auth0 to cibaClient. This implementation looks like it should work with any standard provider using the CIBA protocol, so we should keep the naming generic rather than specific to Auth0.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
```diff
@@ -0,0 +1,134 @@
...
+        p.maxLifetimeSeconds = maxLifetimeSeconds; p.wired = true;
+        return p;
+    }
+
+    private synchronized void ensureWired() {
+        if (wired) return;
+        this.consentStrategy = switch (ConsentRelayStrategy.Type.fromConfig(configuration.getConsentRelayStrategy())) {
+            case PASSTHROUGH -> new PassthroughRelayStrategy();
+            case AUTH0_USER_PROFILE -> new Auth0UserProfileRelayStrategy(configuration.getRecipientDisplayName());
+        };
```
leleueri: Discussion: Could we consider introducing an SPI mechanism here? For example, using a factory to load multiple ConsentRelayStrategy implementations via META-INF and iterating through them to see which one can handle the current configuration (defaulting to a "passthrough" behavior).

This would give users the flexibility to support any other provider simply by dropping additional JARs into the classpath—similar to how JDBC drivers are handled. We could potentially leverage SpringFactoriesLoader.loadFactoryNames for this.

While we might lose some strict configuration validation during plugin setup (since the strategy would likely become a text field), it opens up the system to seamlessly support any provider in the future. In that scenario, the Auth0 strategy itself could then be moved out of the core and into its own external library.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestParametersHandler.java
```diff
@@ -133,4 +162,60 @@ private void validateBindingMessage(CibaAuthenticationRequest request) {
...
+        if (details == null) {
+            return;
+        }
+        for (Map<String, Object> element : details) {
+            Object type = element == null ? null : element.get("type");
+            if (!(type instanceof String s) || s.isBlank()) {
+                throw new InvalidRequestException("Each authorization_details entry must have a non-empty string \"type\" (RFC 9396 §2)");
+            }
+        }
+    }
```
leleueri, line 220: discussion: Since RAR (Rich Authorization Requests) support is on the AM roadmap, how should it behave in this federation mode once implemented? While we don't have an official timeline yet, it's worth considering the validation boundary.

Should we require that all authorization_details be valid from both an AM perspective and for the federated IdP? Or should we skip full AM validation entirely and only verify that the type field is present when the notifier has the authorization details capability ?

leleueri (commented):
Hi @otwebti 

Thanks you for your valuable proposal, I manage to comment this final PR as it gives me global overview. 

For now, I didn't test this feature. I only review the code.

otwebti:
> Hi @otwebti
> 
> Thanks you for your valuable proposal, I manage to comment this final PR as it gives me global overview.
> 
> For now, I didn't test this feature. I only review the code.

Thank you Eric,  

I am happy to close the two underlying PRs (#8157 and #8158) and make all updates here if that is preferable.

#8157 and #8158 where only ever intended to provide a clean picture of the separation between what we believed we need to extend in the core/spi vs what we thought belonged in the plugin.

Please let me know what you'd prefer then I'll address the feedback in the appropriate place.

Also if you would like a demonstration of the system running a live CIBA federation then please let me know what slots you have available and we can jump on for 15 mins to demo.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
wajb: thought: can the client be injected as a stateless bean to avoid branching, and can the client's dependencies be wired into its implementation where possible?

The same pattern could apply to `GatewayCallbackClient`.

gravitee-am-ui/src/app/domain/settings/openid/ciba/device-notifiers/dynamic-sources.ts
```diff
@@ -0,0 +1,83 @@
...
+  } else {
+    property['readonly'] = true;
+  }
+}
+
+/**
+ * Recursively walk a JSON-schema object and set `widget = 'password'` on any
+ * leaf that carries `sensitive: true`.
+ *
+ * Mirrors `providers/provider/form/form.component.ts::applyPasswordInputToSensitiveFields`.
```
wajb: note: this is self-referential, since the form component uses this same method.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/test/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/spring/SchemaFormJsonTest.java
```diff
@@ -0,0 +1,42 @@
...
+    @Test
+    void schema_is_valid_json_with_expected_fields() throws Exception {
+        String raw = Files.readString(Path.of("src/main/resources/schemas/schema-form.json"));
+        JsonObject schema = new JsonObject(raw);
+        JsonObject props = schema.getJsonObject("properties");
+        for (String f : new String[]{"identityProviderId","resourceAudience","callbackClientId","callbackClientSecret",
+                "recipientDisplayName","maxLifetimeSeconds","consentRelayStrategy"}) {
+            assertNotNull(props.getJsonObject(f), "missing schema property: " + f);
+        }
+        // callbackBaseUrl and pollIntervalCapSeconds removed in Tasks 11/12; callbackUrl is per-request.
```
wajb: nitpick: could scan for and remove implementation plan notes.

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/oauth2/MongoCibaAuthRequestRepository.java
```diff
@@ -154,6 +159,11 @@ private CibaAuthRequest convert(CibaAuthRequestMongo authReqMongo) {
             authReq.setExternalInformation(new HashMap<>(authReqMongo.getExternalInformation()));
         }
 
+        if (authReqMongo.getAuthorizationDetails() != null) {
+            authReq.setAuthorizationDetails(authReqMongo.getAuthorizationDetails().stream()
+                    .map(doc -> (Map<String, Object>) new HashMap<String, Object>(doc)).toList());
```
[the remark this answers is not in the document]
otwebti: Done. The read path now materialises `authorizationDetails` with `Collectors.toCollection(ArrayList::new)` so the domain-model collection stays mutable for downstream policy / token-enhancer code. The outbound `convert()`→Mongo path was left on `toList()` since that builds the storage entity, not the domain model.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/service/AuthenticationRequestServiceImpl.java
```diff
@@ -178,37 +196,98 @@ public Single<ADNotificationResponse> notify(ADNotificationRequest adRequest) {
...
+                        // non-federated, rejected, or notifier has no IdP configured -> status-only
+                        return this.authRequestRepository.updateStatus(cibaRequest.getId(), status);
+                    }
+                    final AuthenticationContext authContext = new SimpleAuthenticationContext();
+                    return this.identityProviderManager.get(idpId)
+                            .switchIfEmpty(Single.error(() -> new InvalidRequestException("CIBA federation IdP not available: " + idpId)))
+                            .flatMap(provider -> provider.retrieveUserFromTokenResponse(
+                                            userResponse.getAccessToken(), userResponse.getIdToken(), authContext)
+                                    .switchIfEmpty(Maybe.error(() -> new InvalidRequestException("CIBA federation: empty user from IdP")))
+                                    .toSingle())
```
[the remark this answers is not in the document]
otwebti: Done. Dropped the redundant `.toSingle()` — `Maybe.switchIfEmpty(SingleSource)` already returns `Single`, so `Single.error(...)` is passed to `switchIfEmpty` directly.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/service/AuthenticationRequestServiceImpl.java
```diff
@@ -178,37 +196,98 @@ public Single<ADNotificationResponse> notify(ADNotificationRequest adRequest) {
...
+     * Reads the {@code identityProviderId} field from the active notifier's configuration JSON.
+     * The field was removed from {@code CIBASettingNotifier} (model binding) and now lives
+     * exclusively inside the notifier plugin's configuration JSON — same source as the
+     * acknowledge handler uses via {@code readIdpIdFromNotifierConfig}.
+     */
+    private String resolveIdpId(String deviceNotifierId) {
+        var notifier = this.notifierManager.getAuthDeviceNotifier(deviceNotifierId);
+        if (notifier == null || notifier.getConfiguration() == null) {
+            return null;
+        }
```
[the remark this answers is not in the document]
otwebti: Done. `resolveIdpId` now returns `null` when the notifier configuration is blank (treated as absent → status-only completion), rather than falling through to the fail-closed catch on an empty-string parse.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestAcknowledgeHandler.java
```diff
@@ -182,6 +227,54 @@ public void handle(RoutingContext context) {
...
 
+    /**
+     * Reads the {@code identityProviderId} field from the active notifier's configuration JSON.
+     * Returns null when the notifier is not found, has no configuration, or the field is absent.
+     */
+    private String readIdpIdFromNotifierConfig(String notifierId) {
+        var notifier = deviceNotifierManager.getAuthDeviceNotifier(notifierId);
+        if (notifier == null || notifier.getConfiguration() == null) {
+            return null;
+        }
```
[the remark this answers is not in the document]
otwebti: Done. `readIdpIdFromNotifierConfig` now guards a blank configuration (returns `null`) before attempting to parse it.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CrossWitness.java
```diff
@@ -0,0 +1,98 @@
...
+        return expectedHash != null && value != null && expectedHash.equals(hash(value));
+    }
+
+    public static String hash(Object value) {
+        try {
+            byte[] digest = MessageDigest.getInstance("SHA-256")
+                    .digest(canonical(value).getBytes(StandardCharsets.UTF_8));
+            StringBuilder sb = new StringBuilder(digest.length * 2);
+            for (byte b : digest) sb.append(Character.forDigit((b >> 4) & 0xF, 16)).append(Character.forDigit(b & 0xF, 16));
+            return sb.toString();
```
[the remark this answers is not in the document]
otwebti: Done. Replaced the manual hex loop with `java.util.HexFormat.of().formatHex(digest)`. Output is identical (both lower-case) and covered by the existing hash tests.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestAcknowledgeHandler.java
otwebti: Adopted the constant. One caveat: returning `CIBAProvider.AUTHENTICATION_CALLBACK_ENDPOINT` verbatim would drop the scheme/host/domain and yield only the path, which can't serve as the notifier's callback URL. So the handler now replaces the trailing `AUTHENTICATION_ENDPOINT` segment of the absolute base with `AUTHENTICATION_CALLBACK_ENDPOINT` — keeping the absolute base and removing the literal `/callback`. It also strips a trailing slash first (gemini's related note).

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
otwebti: Renamed `auth0` / `testAuth0` → `cibaClient` / `testCibaClient` in the provider (and the matching local in `AuthorizationPoller`) so the client naming is protocol-generic. The genuinely Auth0-specific names — the `auth0-user-profile` consent strategy and the `urn:auth0:` URN — are intentional and left as-is.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/test/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/spring/SchemaFormJsonTest.java
otwebti: Done. Removed the internal "Tasks 11/12" / "New tests" plan notes and the internal "§2.2" design-doc section references across the plugin. Kept the external `RFC 8628 §3.5` citation in `AuthorizationPoller`.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/model/mapper/ListOfMapsToStringConverter.java
```diff
@@ -0,0 +1,44 @@
...
+    @Override
+    public String convertTo(List source, String target) {
+        return JSONMapper.toJson(source);
+    }
+
+    @Override
+    public List convertFrom(String source, List target) {
+        return JSONMapper.toCollectionOfBean(source, new TypeReference<List<Map<String, Object>>>() { });
+    }
+}
```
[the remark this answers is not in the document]
otwebti: Left this converter unchanged. `JSONMapper.toJson` and `JSONMapper.toCollectionOfBean` already null-guard internally (`if (x != null) … return null`), so the proposed `null` checks are redundant, and the sibling converters (`ListToStringConverter`, `MapToStringConverter`) deliberately don't add them — matching that established pattern here. The only unguarded case is an empty-string in `convertFrom`, but `convertTo` never emits `""` (it returns null or valid JSON), so that branch is unreachable. Happy to add the guards if you'd prefer defence-in-depth despite the redundancy.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/authdevice/notifier/impl/AuthenticationDeviceNotifierManagerImpl.java
otwebti: Fully agree — parsing the notifier's config JSON inside core (and the `getAuthDeviceNotifier` accessor that enables it) is the wrong layering. I'd like to go a bit further than the literal `getFederationConfiguration()` shape and keep **core's vocabulary limited to "identity provider"** — no "federation" concept in core at all. Proposed shape, and I'd value your feedback before actioning:

**1. A role interface, not a method on the base provider.** Rather than adding `getFederationConfiguration()` (which every notifier would carry as an empty default), introduce a small opt-in interface that only IdP-backed notifiers implement:
```java
public interface IdentityProviderDependent {
    Optional<String> getIdentityProviderId();
}
```
Core reads it through the existing `getAuthDeviceNotifierProvider(...)` with an `instanceof` check — mirroring the `provider instanceof OpenIDConnectAuthenticationProvider` idiom already in `AuthenticationRequestAcknowledgeHandler`. `AuthenticationDeviceNotifierProvider` stays federation-agnostic, and non-IdP notifiers (e.g. the HTTP one) implement nothing. Named generically so core never learns the word "federation"; the plugin keeps its own typed config internally. Core needs only the IdP id (it resolves the actual connection itself — see 3), so a single-method interface beats a config wrapper.

**2. Collapse `FEDERATED_HINT_RESOLUTION` into it.** That capability's only consumer (`CibaAuthenticationRequestResolver`) is a boolean proxy for exactly this predicate, so it becomes `provider instanceof IdentityProviderDependent` — same fallback semantics (resolve locally first, forward the hint only if local resolution fails). Will keep `AUTHORIZATION_DETAILS` as a capability: it's a valueless behavioural flag and orthogonal to IdP binding.

**3. Completion reads the notifier's response, not its config.** At completion the resolver stream has already dropped the provider reference (it maps provider → `ADUserResponse`), so instead of re-deriving the IdP id I'd add a nullable `identityProviderId` to `ADUserResponse` — the notifier self-describes which IdP established identity. Core then reacts to a generic nullable; `resolveIdpId` and the notifier-config lookup are deleted.

**4. Single source of truth for the connection stays in core.** The OIDC client creds still come from the registered AM IdP (not duplicated into notifier config), so core keeps a small initiation seam: given the IdP id, resolve and inject the connection. That's generic IdP machinery keyed on a nullable, not a federation branch.

**5. Fail-closed at config-binding.** A notifier that is `IdentityProviderDependent` with no IdP configured is rejected when its config binds, so it can't deploy — core needs no runtime guard for the "dependent but empty" case.

Net: `getAuthDeviceNotifier` deleted, all `readTree`/`identityProviderId` parsing out of core, no `instanceof FederationConfiguration` — core speaks only "identity provider". Lands on #8157 (SPI: `IdentityProviderDependent` + `ADUserResponse.identityProviderId`, minus `FEDERATED_HINT_RESOLUTION`) and #8158 (core call-sites + plugin). Happy to adjust naming or keep the capability if you'd rather — flagging the shape first as you suggested.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestAcknowledgeHandler.java
leleueri: oh yes, you are right in my suggestion we were loosing the absolute url sorry about that.

With a step back, I wondering if we can simplify this method with something like this:

`UriBuilderRequest.resolveProxyRequest(context.request(), context.get(CONTEXT_PATH)) + CIBA_PATH + AUTHENTICATION_CALLBACK_ENDPOINT `

It need to be tested to make sure the result is the rigth one.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/authdevice/notifier/impl/AuthenticationDeviceNotifierManagerImpl.java
leleueri: Yes, I'm fine with an interface at NotifierProvider level.

leleueri:
> > Hi @otwebti
> > Thanks you for your valuable proposal, I manage to comment this final PR as it gives me global overview.
> > For now, I didn't test this feature. I only review the code.
> 
> Thank you Eric,
> 
> I am happy to close the two underlying PRs (#8157 and #8158) and make all updates here if that is preferable.
> 
> #8157 and #8158 where only ever intended to provide a clean picture of the separation between what we believed we need to extend in the core/spi vs what we thought belonged in the plugin.
> 
> Please let me know what you'd prefer then I'll address the feedback in the appropriate place.
> 
> Also if you would like a demonstration of the system running a live CIBA federation then please let me know what slots you have available and we can jump on for 15 mins to demo.

I understand the intension behind those PRs, I thanks you for that. 
As you stacked everything, I guess it will be easiest for you to update this final PR. As soon as the changes are properly fixing up the right commit to have a clean history, I'm ok to work only on this one if you agree.

About the demo, I can be available between 9:00am and 11:30am CET everyday of the next week except the 14th of July if this is possible for you.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/service/request/CibaAuthenticationRequestResolver.java
otwebti: Thanks @leleueri, looking for your steer here in case I missed the point.

Did you mean we should check this ahead of federating to the downstream IdP?

If so, we might need another extension to the core/spi as I can't see a clean way to extract the issuer via OpenIDConnectAuthenticationProvider.  So I'm keen to understand what the goal/benefit is before proposing more changes at that layer.

leleueri: I’m wondering if it’s worth checking that the iss in this structure matches the one linked to the NotifierProvider.

* Pro: It saves an unnecessary network redirection and allows us to fail fast.

* Con: Arguably, it’s the responsibility of the "notified service" to perform validation and ensure it's the correct audience.

Looking at the format attribute in the structure, we might need to extract the value to perform a local search (e.g., if the format is email). If the format is iss_sub, we would likely need to verify that iss matches the AM domain URL so a search by sub becomes possible (though I know we can't search by sub in AM just yet).

Honestly, your current implementation is likely enough for now until we fully support this login_hint JSON structure in AM. Thanks for bearing with my brain dump!

otwebti: I agree, the responsibility lies with the with downstream service to perform validation before notifying the registered user.  Will close the comment as resolved for now.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/authdevice/notifier/impl/AuthenticationDeviceNotifierManagerImpl.java
otwebti: Implemented, please let me know if you're happy.

gravitee-am-ui/src/app/domain/settings/openid/ciba/device-notifiers/dynamic-sources.ts
otwebti: Good catch — the "mirrors" wording was circular. `applyPasswordInputToSensitiveFields` was extracted from the IdP-provider form's private method (`settings/providers/provider/form/form.component.ts`); the CIBA notifier form now delegates to this shared copy rather than keeping its own. Reworded the comment to "extracted from … so the notifier form can share one implementation" so it no longer reads as self-referential.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
otwebti: Done — added a stateless `CibaClientFactory` bean. `notify()` now always calls `cibaClientFactory.create(conn, resourceAudience)`, so the `testCibaClient != null ? … : new CibaClient(…)` branch and the test-only client field are both gone. `DefaultCibaClientFactory` is a `@Bean` holding the shared web client + discovery resolver; it builds the per-request client from the inbound `FederatedConnection`.

The client itself can't be a single stateless bean — its endpoint and credentials come from the per-request connection — so the *factory* is the stateless bean and the client is built through it. Tests inject a stub factory (a lambda) rather than a special client.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
otwebti: Removed the dedicated test-only `CibaClient`. The provider now obtains the client through an injected `CibaClientFactory` bean (`DefaultCibaClientFactory`), so production and tests follow one path — no test-specific field or branch in `notify()`. Tests inject a stub factory (a lambda) instead of a bespoke client.

This keeps to the standard Spring initialization you suggested: the factory bean is wired from the existing `cibaFederationWebClient` + `oidcDiscoveryResolver` beans in `CibaFederationProviderSpringConfiguration`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/ciba/resources/handler/AuthenticationRequestAcknowledgeHandler.java
otwebti: Good call — agreed and adopted. `resolveProxyRequest` is the right tool here: the hand-rolled `absoluteURI()` parsing wasn't proxy-aware, which made this the odd one out among AM's self-referential URLs.

One small tweak to your snippet — I passed the full path *into* `resolveProxyRequest` rather than concatenating onto its result, to match how `/login/callback` and `/logout/callback` are built (`LoginAuthenticationHandler` / `LogoutEndpoint`):

```java
private static String buildCallbackUrl(RoutingContext context) {
    return UriBuilderRequest.resolveProxyRequest(context.request(),
            context.get(CONTEXT_PATH) + CIBAProvider.CIBA_PATH + CIBAProvider.AUTHENTICATION_CALLBACK_ENDPOINT);
}
```

That removes the whole `absoluteURI()` query-strip / trailing-slash / `endsWith("/authenticate")` dance (and its null/warn fallback), and now honours `X-Forwarded-*` and the domain context path. The resolved path is unchanged: `…/oidc/ciba/authenticate/callback`.

On your "needs to be tested": added a handler unit test asserting the resolved URL both with and without `X-Forwarded-*` (`X-Forwarded-Proto: https` + `X-Forwarded-Host` → `https://<host>/oidc/ciba/authenticate/callback`), and validated it end-to-end through the full CIBA flow against a real Auth0 tenant — the notifier's back-channel callback POST round-trips back into the gateway correctly for both the RAR and identity-only scenarios. Folded into the core commit.

otwebti:
> > > Hi @otwebti
> > > Thanks you for your valuable proposal, I manage to comment this final PR as it gives me global overview.
> > > For now, I didn't test this feature. I only review the code.
> > 
> > 
> > Thank you Eric,
> > I am happy to close the two underlying PRs (#8157 and #8158) and make all updates here if that is preferable.
> > #8157 and #8158 where only ever intended to provide a clean picture of the separation between what we believed we need to extend in the core/spi vs what we thought belonged in the plugin.
> > Please let me know what you'd prefer then I'll address the feedback in the appropriate place.
> > Also if you would like a demonstration of the system running a live CIBA federation then please let me know what slots you have available and we can jump on for 15 mins to demo.
> 
> I understand the intension behind those PRs, I thanks you for that. As you stacked everything, I guess it will be easiest for you to update this final PR. As soon as the changes are properly fixing up the right commit to have a clean history, I'm ok to work only on this one if you agree.
> 
> About the demo, I can be available between 9:00am and 11:30am CET everyday of the next week except the 14th of July if this is possible for you.


Thanks @leleueri. I've closed #8157 and #8158 and will continue to fold changes into the appropriate commit so the history stays clean.  I've also scheduled the demo for 11:30 today but can't get the invite over to you.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
otwebti: Added ConsentRelayStrategy SPI discovered via spring.factories, defaulting to a vendor-neutral passthrough so stock deployments need no strategy at all. Additional providers can be added by dropping a module on the classpath, exactly as you suggested.

Following your last point, the Auth0-specific strategy is no longer in core — we've extracted it into a standalone module that we'll maintain on our side, and the core tree is now free of any Auth0 references. It's hooked in purely by configuration (a consentRelayStrategy key) rather than the iterate-until-one-matches approach — selection is explicit as the administrator should know what downstream IdP they are integrating with and what custom behaviour they need if any.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaClient.java
```diff
@@ -0,0 +1,107 @@
...
+        this.wellKnownUri = wellKnownUri;
+        this.clientId = clientId;
+        this.clientSecret = clientSecret;
+        this.audience = audience;
+    }
+
+    public Single<BcAuthorizeResult> bcAuthorize(String loginHint, String loginHintToken,
+                                                 String scope, String bindingMessage, Object relayedAuthorizationDetails) {
+        MultiMap form = MultiMap.caseInsensitiveMultiMap()
+                .set("client_id", clientId).set("client_secret", clientSecret)
```
leleueri: todo: this implementation assume that the client using the `client_secret_post` as authentication method. IMO we should use the IdentityProvider configuration to adapt the method (we should at least support `client_secret_post` and `client_secret_basic`). Sorry I missed this point during my initial review.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaClient.java
```diff
@@ -0,0 +1,107 @@
...
+                    throw new IllegalStateException("CIBA federation: back-channel authorization response omitted auth_req_id");
+                }
+                return new BcAuthorizeResult(authReqId,
+                        body.getInteger("expires_in", 120), body.getInteger("interval", 5));
+            }));
+    }
+
+    public Single<PollResult> pollToken(String authReqId) {
+        MultiMap form = MultiMap.caseInsensitiveMultiMap()
+                .set("client_id", clientId).set("client_secret", clientSecret)
```
leleueri: todo: same comment as for the `bcAuthorize` method. this implementation assume that the client using the client_secret_post as authentication method.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public Set<NotifierCapability> capabilities() {
+        return Set.of(NotifierCapability.AUTHORIZATION_DETAILS);
+    }
+
+    /** A CIBA-federation notifier's identity is always established by a registered AM identity
+     *  provider; core reads this to resolve the federated connection. */
+    @Override
+    public Optional<String> getIdentityProviderId() {
+        return Optional.ofNullable(configuration != null ? configuration.getIdentityProviderId() : null);
```
leleueri:
```suggestion
        return Optional.ofNullable(configuration).map(CibaFederationAuthenticationDeviceNotifierConfiguration::getIdentityProviderId);
```

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/GatewayCallbackClient.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public GatewayCallbackClient(WebClient client, String clientId, String clientSecret) {
+        this.client = client; this.clientId = clientId; this.clientSecret = clientSecret;
+    }
+
+    public Completable postCallback(String callbackUrl, String state, String tid, boolean validated,
+                                    String idToken, String accessToken) {
+        MultiMap form = MultiMap.caseInsensitiveMultiMap()
+                .set("state", state).set("tid", tid).set("validated", Boolean.toString(validated))
+                .set("client_id", clientId).set("client_secret", clientSecret);
```
leleueri: question: do we have to force the usage of client_secret_post ? Should we allow client_secret_basic in the notifier configuration ?

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaClient.java
otwebti: Done — `bcAuthorize` no longer assumes `client_secret_post`. The method now comes from the IdP config's `clientAuthenticationMethod`, carried to the notifier on the `FederatedConnection` the gateway builds (`AuthenticationRequestAcknowledgeHandler.toFederatedConnection`). A small package-private `ClientAuthentication` helper applies it: `client_secret_post` → secret in the body, `client_secret_basic` → `Authorization: Basic` header (URL-encoded per RFC 6749 §2.3.1 so it round-trips through AM's `ClientBasicAuthProvider`), with the secret sent by exactly one mechanism. Both methods are supported; any other value is rejected fail-closed at client build.

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaClient.java
otwebti: Done — `pollToken` routes through the same `ClientAuthentication` helper as `bcAuthorize`, so the token call honours the same configured method (secret in the body for `client_secret_post`, `Authorization: Basic` header for `client_secret_basic`).

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/CibaFederationAuthenticationDeviceNotifierProvider.java
otwebti: Applied verbatim — `getIdentityProviderId` is now `Optional.ofNullable(configuration).map(CibaFederationAuthenticationDeviceNotifierConfiguration::getIdentityProviderId)`. Thanks!

gravitee-am-authdevice-notifier/gravitee-am-authdevice-notifier-ciba-federation/src/main/java/io/gravitee/am/authdevice/notifier/cibafederation/provider/GatewayCallbackClient.java
otwebti: Good point — the callback leg is configurable now too. Added a `callbackClientAuthMethod` notifier config field (default `client_secret_post`) with a schema-form dropdown; `GatewayCallbackClient` applies it through the same `ClientAuthentication` helper (body for post, `Authorization: Basic` header for basic). It's validated fail-closed at bind time and must match the registered callback application's `tokenEndpointAuthMethod`.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/oauth/changelogs/v4_12_0/4.12.0-ciba-add-authorization-details.yml
```diff
@@ -0,0 +1,15 @@
+databaseChangeLog:
+  - changeSet:
+      id: 4.12.0-ciba-add-authorization-details
```
leleueri: TODO: this PR is targeting master which will be the 4.13.0. Can you please rename the id, the file and folder accordingly ?

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/oauth-master.yml
```diff
@@ -18,4 +18,6 @@ databaseChangeLog:
   - include:
       - file: liquibase/oauth/changelogs/v4_11_0/4.11.0-add-tokens-parent-jti.yml
   - include:
-      - file: liquibase/oauth/changelogs/v4_11_0/4.11.0-add-tokens-parent-jtis.yml
\ No newline at end of file
+      - file: liquibase/oauth/changelogs/v4_11_0/4.11.0-add-tokens-parent-jtis.yml
+  - include:
+      - file: liquibase/oauth/changelogs/v4_12_0/4.12.0-ciba-add-authorization-details.yml
```
leleueri: TODO: this PR is targeting master which will be the 4.13.0. Can you please adapt this path once https://github.com/gravitee-io/gravitee-access-management/pull/8159/changes#r3602143736 is managed?

leleueri (changes requested):
When RDBMS is used the flow is failing with a 500 error because the `subject` column in the `ciba_auth_requests` table cannot be null. In the federation case, the subject is not initialized so we have to remove the null constraint. May I ask you to create the appropriate liquibase changeset for it please ?

once those latest comments will be managed, I will approve and merge.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/oauth/changelogs/v4_12_0/4.12.0-ciba-add-authorization-details.yml
otwebti: On the version rename: Done — renamed the changeSet id, file and folder to 4.13.0, folded into the federated-CIBA commit so it reads as 4.13.0 from the start.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/oauth-master.yml
otwebti: On the oauth-master.yml path: Done — repointed the authorization-details include to the v4_13_0 path and appended the subject-nullable include after it.

otwebti:
On the RDBMS 500 / null-constraint (review): Added a forward-only changeset 4.13.0-ciba-subject-nullable dropping the NOT NULL on ciba_auth_requests.subject. Validated on Postgres, MySQL, MariaDB and MSSQL; columnDataType is matched to the v4_7_0 definition to avoid type drift on MODIFY-style engines.

leleueri (approved):

leleueri:
Thanks you for your contribution.

PR #8160, opened 2026-07-02
feat: AM-7200 required scopes for consent

wajb:
[AM-7200](https://gravitee.atlassian.net/browse/AM-7200)

Introduce *required scopes*, configurable for OAuth 2.0 clients of applications. During OAuth2/OIDC authorisation flows when any user consent options are presented, those marked required are disabled and pre-selected.

## :memo: Test scenarios 
See Jira.

### Application scope settings
image

### Management "User consent" template preview
image

### Consent screen with >10 total scopes
image

n/a

@gravitee-io/am

[AM-7200]: https://gravitee.atlassian.net/browse/AM-7200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/consent/UserConsentEndpoint.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/consent/UserConsentEndpoint.java
@@ -20,6 +20,7 @@ import io.gravitee.am.gateway.handler.common.vertx.utils.UriBuilderRequest;
 import io.gravitee.am.gateway.handler.common.vertx.web.auth.user.User;
 import io.gravitee.am.gateway.handler.oauth2.service.consent.UserConsentService;
 import io.gravitee.am.gateway.handler.oauth2.service.request.AuthorizationRequest;
+import io.gravitee.am.gateway.handler.oauth2.service.utils.RequiredScopeUtils;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.Template;
 import io.gravitee.am.model.oauth2.Scope;
@@ -35,6 +36,7 @@ import io.vertx.rxjava3.ext.web.templ.thymeleaf.ThymeleafTemplateEngine;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
@@ -74,7 +76,15 @@ public class UserConsentEndpoint implements Handler<RoutingContext> {
                 return;
             }
             List<Scope> requestedScopes = h.result();
-            routingContext.put(ConstantKeys.SCOPES_CONTEXT_KEY, requestedScopes);
+            Set<String> requiredScopeKeys = RequiredScopeUtils.requiredScopeKeys(client);
+            List<Scope> requiredScopes = requestedScopes.stream().filter(scope -> requiredScopeKeys.contains(scope.getKey())).collect(Collectors.toList());
+            List<Scope> optionalScopes = requestedScopes.stream().filter(scope -> !requiredScopeKeys.contains(scope.getKey())).collect(Collectors.toList());
+            // required scopes are always displayed first
+            List<Scope> orderedScopes = new ArrayList<>(requiredScopes);
+            orderedScopes.addAll(optionalScopes);
+            routingContext.put(ConstantKeys.SCOPES_CONTEXT_KEY, orderedScopes);
+            routingContext.put(ConstantKeys.REQUIRED_SCOPES_CONTEXT_KEY, requiredScopes);
+            routingContext.put(ConstantKeys.OPTIONAL_SCOPES_CONTEXT_KEY, optionalScopes);
             routingContext.put(ConstantKeys.ACTION_KEY, action);
             routingContext.put(ConstantKeys.PRESELECT_ALL_SCOPES, client == null || !client.isOptInScopeSelection());
             CimdConsentPageAttributes.putIfApplicable(routingContext, domain, client);
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/consent/UserConsentProcessHandler.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/consent/UserConsentProcessHandler.java
@@ -18,8 +18,10 @@ package io.gravitee.am.gateway.handler.oauth2.resources.handler.authorization.co
 import io.gravitee.am.common.jwt.Claims;
 import io.gravitee.am.common.utils.ConstantKeys;
 import io.gravitee.am.gateway.handler.common.vertx.web.auth.user.User;
+import io.gravitee.am.gateway.handler.oauth2.exception.AccessDeniedException;
 import io.gravitee.am.gateway.handler.oauth2.service.consent.UserConsentService;
 import io.gravitee.am.gateway.handler.oauth2.service.request.AuthorizationRequest;
+import io.gravitee.am.gateway.handler.oauth2.service.utils.RequiredScopeUtils;
 import io.gravitee.am.identityprovider.api.DefaultUser;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.oauth2.ScopeApproval;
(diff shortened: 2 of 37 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/consent/UserConsentEndpoint.java
```diff
@@ -118,6 +128,18 @@ private void fetchConsentInformation(Set<String> requestedConsents, boolean prom
...
+        if (client == null || client.getScopeSettings() == null) {
+            return Collections.emptySet();
+        }
+        Set<String> requestedKeys = requestedScopes.stream().map(Scope::getKey).collect(Collectors.toSet());
+        return client.getScopeSettings().stream()
+                .filter(ApplicationScopeSettings::isRequiredScope)
+                .map(ApplicationScopeSettings::getScope)
+                .filter(requestedKeys::contains)
+                .collect(Collectors.toSet());
+    }
```
leleueri: question: why are we filtering on the requested scopes? I was expecting to only rely on the client scope definition for that 🤔

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/consent/UserConsentProcessHandler.java
```diff
@@ -143,6 +155,16 @@ private void saveConsent(RoutingContext context,
...
 
+    private static Set<String> requiredScopeKeys(Client client) {
+        if (client == null || client.getScopeSettings() == null) {
+            return Collections.emptySet();
+        }
+        return client.getScopeSettings().stream()
+                .filter(ApplicationScopeSettings::isRequiredScope)
+                .map(ApplicationScopeSettings::getScope)
+                .collect(Collectors.toSet());
+    }
```
leleueri: suggestion: this is the method I was expected in the UserConsentEndpoint and if we agree on relying only on the client settings, then this method can be mutualized.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/consent/UserConsentProcessHandlerTest.java
```diff
@@ -216,6 +217,62 @@ public void shouldPresentAllScopes_promptConsent() throws Exception {
...
+        client.setScopeSettings(List.of(requiredScopeSetting("admin")));
+        when(userConsentService.checkConsent(any(), any())).thenReturn(Single.just(Collections.emptySet()));
+
+        testRequest(HttpMethod.POST, "/oauth/consent",
+                req -> writeForm(req, "scope.read=true&scope.admin=false&user_oauth_approval=true"),
+                HttpStatusCode.OK_200, "OK", null);
+
+        List<ScopeApproval> approvals = captureSavedApprovals();
+        assertEquals(Set.of("read", "admin"), approvedScopes(approvals));
+        assertTrue(deniedScopes(approvals).isEmpty());
```
leleueri: discussion: Should we really "approve" required scopes when the scope entry is missing or false, even if user_oauth_approval is true? I don't think so.

IMO, we should reject the request with an error and redirect the user either to the consent page or the login page. If we keep the current implementation, a template that fails to display a required scope could inadvertently grant permissions without explicit user consent. By enforcing the presence of the parameter with the correct value, we ensure proof of informed consent. I'm not aligned with keeping this permissive behavior as it introduces a security risk. Wdyt?

gravitee-am-model/src/main/java/io/gravitee/am/model/application/ApplicationScopeSettings.java
```diff
@@ -29,6 +29,11 @@ public class ApplicationScopeSettings {
      * Scope approval duration times
      */
     private Integer scopeApproval;
+    /**
+     * True if the scope is required: it cannot be deselected by the user during consent
+     * and is always granted whenever the user allows the authorization request.
```
leleueri: this comment may not be valid if the UserConsentProcessHandler implementation change to enforce the presence of the required scopes

gravitee-am-test/specs/gateway/oauth2/oauth2-grant-authorization-code.jest.spec.ts
```diff
@@ -156,6 +157,79 @@ describe('OAuth2 - RFC 6746 - Authorization Code Grant', () => {
     await logoutUser(fixture.oidc.end_session_endpoint, postConsentRedirect);
   });
 
+  it('must always grant a required scope even when its consent field is not submitted', async () => {
```
leleueri: this test may not be valid according the outcomes of the discussion started on the UserConsentProcessHandler

gravitee-am-ui/src/app/domain/components/oauth2-settings/scopes/scopes.component.html
```diff
@@ -153,7 +164,26 @@ <h5>Scopes</h5>
...
+          <strong>Default</strong> — added to the authorization request automatically when the client starts an authorization flow without
+          requesting any specific scopes.
+          <small>
+            When <strong>Enhance scopes</strong> (above) is enabled, the authenticating user's role scopes are also granted when the request
+            carries no scope, or only <code>openid</code>.
+          </small>
+        </li>
+        <li>
+          <strong>Required</strong> — mandatory scopes that the user cannot deselect on the consent screen when they are requested; they are
+          always consented to whenever the form is submitted.
```
leleueri: may not be true according to the discussion outcomes

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/consent/UserConsentEndpoint.java
wajb: This was an interpretation of required meaning consent is unconditional when requested and presented. To present all required scopes regardless of the request confuses the purpose of default scopes somewhat 🤔.

A scope that is both required and default serves this purpose as long as no scopes are provided in the request. Could a client ever choose to omit certain scopes, even if they are required when presented? Or do you feel required fields should be used as an override?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/test/java/io/gravitee/am/gateway/handler/oauth2/resources/handler/authorization/consent/UserConsentProcessHandlerTest.java
wajb: Redirecting with an error is a reasonable approach - probably the only alternative to making the handler tamper resilient. But you're right that the assumption that a user sees what was provided by the consent endpoint is flawed due to customisable template and the fickleness of browsers. I'll implement as a redirect 👍

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/resources/endpoint/authorization/consent/UserConsentEndpoint.java
leleueri: my point was the following, you are using the `requiredScopeKeys` to filter the requestedScopes in 2 groups (required, optional). So I do not see the point of filtering inside the `requiredScopeKeys` method, returning directly the list of required scopes coming from the client should land in the same result.

```
            Set<String> requiredScopeKeys = requiredScopeKeys(client, requestedScopes);
            List<Scope> requiredScopes = requestedScopes.stream().filter(scope -> requiredScopeKeys.contains(scope.getKey())).collect(Collectors.toList());
            List<Scope> optionalScopes = requestedScopes.stream().filter(scope -> !requiredScopeKeys.contains(scope.getKey())).collect(Collectors.toList());
```

> A scope that is both required and default serves this purpose as long as no scopes are provided in the request. Could a client ever choose to omit certain scopes, even if they are required when presented? Or do you feel required fields should be used as an override?

Good question, IMO the client should be accountable of the scopes requested in that sens if there 2 scopes marked as required in the settings but the client request only one then there is not reason to force the second one. But if the client is not specifying scopes then default scopes are apply and in that case in conjuction with the required scopes the admin can decide which one are mandatory. For example, if openid is part of the default scope IMO it should be required as this scope is more a technical scope to enable the OpenID flows but the user can decide to expose or not the profile or email information (in case of deny, an IDTOken can still be generated only with the sub)

does it make sens?

wajb: I misunderstood your comment. I have now extracted the filtering method for reuse.

gravitee-am-ui/src/app/domain/components/oauth2-settings/scopes/scopes.component.html
wajb: I believe this is still the practical implication of marking a scope as required. Lmk if you feel the wording still needs to change.

gravitee-am-model/src/main/java/io/gravitee/am/model/application/ApplicationScopeSettings.java
leleueri:
```suggestion
     * True if the scope is required: it cannot be deselected by the user during consent and has to be granted by the user allows the authorization request.
```

leleueri (approved):

PR #8198, opened 2026-07-08
fix: preserve value containing '&' in URLParametersUtils.parse

leleueri:
An unencoded '&' inside a query parameter value produces a bare fragment without a '=' separator once the query is split, which made pair.substring(0, idx) throw StringIndexOutOfBoundsException. Such a fragment belongs to the previous value, so re-attach it (with its '&') instead of crashing.

fixes AM-7256

### DIFF
--- a/gravitee-am-common/src/test/java/io/gravitee/am/common/web/URLParametersUtilsTest.java
+++ b/gravitee-am-common/src/test/java/io/gravitee/am/common/web/URLParametersUtilsTest.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


gravitee-am-common/src/test/java/io/gravitee/am/common/web/URLParametersUtilsTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+        assertEquals("abc", result.get("code"));
+        assertEquals("xyz", result.get("state"));
+    }
+
+    @Test
+    public void parse_shouldKeepAmpersandInValue_whenFragmentHasNoSeparator() {
+        // An unencoded '&' inside a value produces a bare fragment with no '=' once the query is split;
+        // that fragment belongs to the previous value and must be re-attached.
+        Map<String, String> result = URLParametersUtils.parse("access_token=abc&raw&state=xyz");
```
wajb, line 41: issue: the fix won't work with multiple `&` in a parameter, e.g. `"access_token=abc&raw1&raw2&state=xyz"`.

leleueri: 🤦 indeed

wajb (approved):

PR #8204, opened 2026-07-10
fix: AM-7223 patch Login settings at application level

lgw-gravitee:
fixes AM-7223


leleueri (approved):

lgw-gravitee:
JUnit ran locally and it wokred.

PR #8210, opened 2026-07-12
feat: DPoP sender-constrained tokens (RFC 9449)

lgw-gravitee:
Implements OAuth 2.0 Demonstrating Proof of Possession across the gateway, management API, model/repositories and console UI.

- Centralized DPoPProofValidator (RFC 9449 proof parsing/validation)
- Issue DPoP-bound tokens for all token-endpoint grants (jkt binding), with client_credentials and refresh-token key continuity
- Bind authorization code to the DPoP key via dpop_jkt (§10)
- Enforce DPoP-bound tokens at the shared bearer gate (OAuth2AuthHandler), including SCIM/users resource enforcement
- Configurable require-DPoP policy: per-app flag + per-domain floor
- Per-domain signing-algorithm allowlist (enforcement + discovery)
- Advertise dpop_signing_alg_values_supported in OIDC discovery
- Persist token jkt (Mongo + JDBC liquibase changelog)
- Domain/Application DPoP settings, patch models, and console UI
- DPoP cluster replay cache
- Auditing on token miting

[dpop-spec.html](https://github.com/user-attachments/files/29946529/dpop-spec.html)

[dpop-rfc9449-v1.md](https://github.com/user-attachments/files/29946554/dpop-rfc9449-v1.md)
[dpop-rfc9449-v2.md](https://github.com/user-attachments/files/29946555/dpop-rfc9449-v2.md)
[dpop-rfc9449-v3.md](https://github.com/user-attachments/files/29946556/dpop-rfc9449-v3.md)

Testing guide
[dpop-testing-guide.html](https://github.com/user-attachments/files/30650991/dpop-testing-guide.html)

image

image

Token AUDIT:
image

image

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
@@ -16,11 +16,15 @@
 package io.gravitee.am.gateway.handler.common.vertx.web.auth.handler.impl;
 
 import io.gravitee.am.common.exception.oauth2.InsufficientScopeException;
+import io.gravitee.am.common.exception.oauth2.InvalidDPoPProofException;
 import io.gravitee.am.common.exception.oauth2.InvalidRequestException;
 import io.gravitee.am.common.exception.oauth2.InvalidTokenException;
 import io.gravitee.am.common.exception.oauth2.OAuth2Exception;
 import io.gravitee.am.common.jwt.JWT;
+import io.gravitee.am.common.jwt.SignatureAlgorithm;
 import io.gravitee.am.common.utils.ConstantKeys;
+import io.gravitee.am.gateway.handler.common.dpop.DPoPProofValidator;
+import io.gravitee.am.gateway.handler.common.vertx.core.http.VertxHttpServerRequest;
 import io.gravitee.am.gateway.handler.common.jwt.SubjectManager;
 import io.gravitee.am.gateway.handler.common.vertx.web.auth.handler.OAuth2AuthHandler;
 import io.gravitee.am.gateway.handler.common.vertx.web.auth.handler.OAuth2AuthResponse;
@@ -32,9 +36,17 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.ext.web.handler.HttpException;
+import io.vertx.rxjava3.core.http.HttpHeaders;
 import io.vertx.rxjava3.core.http.HttpServerRequest;
 import io.vertx.rxjava3.ext.web.RoutingContext;
 import lombok.CustomLog;
+
+import java.util.stream.Collectors;
+
+import static io.gravitee.am.common.utils.ConstantKeys.ACCESS_TOKEN_KEY;
+import static io.gravitee.am.common.utils.ConstantKeys.BEARER_AUTH_SCHEME;
+import static io.gravitee.am.common.utils.ConstantKeys.DPOP_AUTH_SCHEME;
+import static io.gravitee.gateway.api.http.HttpHeaderNames.WWW_AUTHENTICATE;
 
 /**
  * @author Titouan COMPIEGNE (titouan.compiegne at graviteesource.com)
@@ -43,9 +55,12 @@ import lombok.CustomLog;
 @CustomLog
 public class OAuth2AuthHandlerImpl implements OAuth2AuthHandler {
 
-    private static final String BEARER = "Bearer";
     private static final String REALM = "gravitee-io";
     private static final String ACCESS_TOKEN = "access_token";
+    private static final String CONTEXT_DPOP_ENFORCED = "dpopEnforced";
+    private static final String DPOP_SUPPORTED_ALGS = SignatureAlgorithm.DPOP_SUPPORTED_ALGORITHMS.stream()
+            .sorted()
+            .collect(Collectors.joining(" "));
     private final OAuth2AuthProvider oAuth2AuthProvider;
     private String requiredScope;
     private boolean extractRawToken;
@@ -58,7 +73,15 @@ public class OAuth2AuthHandlerImpl implements OAuth2AuthHandler {
     private boolean offlineVerification;
     private String resourceParameter;
     private String resourceRequiredScope;
+    private DPoPProofValidator dpopProofValidator;
     private final SubjectManager subjectManager;
+
+    private record PresentedToken(String value, String scheme){
+
+        boolean isDPoP(){
+            return DPOP_AUTH_SCHEME.equalsIgnoreCase(scheme);
+        }
+    };
 
     public OAuth2AuthHandlerImpl(OAuth2AuthProvider oAuth2AuthProvider, SubjectManager subjectManager) {
         this.oAuth2AuthProvider = oAuth2AuthProvider;
@@ -78,14 +101,14 @@ public class OAuth2AuthHandlerImpl implements OAuth2AuthHandler {
                 return;
             }
 
-            final String jwtToken = parseHandler.result();
+            final PresentedToken presented = parseHandler.result();
 
             // set raw token to the current context
             if (extractRawToken) {
-                context.put(ConstantKeys.RAW_TOKEN_CONTEXT_KEY, jwtToken);
+                context.put(ConstantKeys.RAW_TOKEN_CONTEXT_KEY, presented.value());
             }
 
-            oAuth2AuthProvider.decodeToken(jwtToken, offlineVerification, handler -> {
+            oAuth2AuthProvider.decodeToken(presented.value(), offlineVerification, handler -> {
                 if (handler.failed()) {
                     processException(context, handler.cause());
                     return;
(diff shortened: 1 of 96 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
```diff
@@ -59,8 +66,16 @@ public class OAuth2AuthHandlerImpl implements OAuth2AuthHandler {
     private boolean offlineVerification;
     private String resourceParameter;
     private String resourceRequiredScope;
+    private DPoPProofValidator dpopProofValidator;
     private final SubjectManager subjectManager;
 
+    private record AuthorizationDetails(String authToken, String authScheme){
```
leleueri: suggestion: I would not name it AuthorizationDetails as it will bring confusion when we will introduce RAR which is managing an "authorization_details" query paramter

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
```diff
@@ -106,52 +121,83 @@ public void handle(RoutingContext context) {
...
-                        }
-                    }
-
-                    if (forceEndUserToken && token.getSub().equals(token.getAud())) {
-                        // token for end userId must not contain clientId as subject
-                        processException(context, new InvalidTokenException("The access token was not issued for an End-User"));
+                if (token.getDPoPConfirmationThumbprint() != null) {
+                    context.put(CONTEXT_DPOP_ENFORCED, true);
+                    if (dpopProofValidator == null || !auth.isDPoP()) {
+                        processException(context, new InvalidTokenException("A DPoP-bound access token must be presented using the DPoP scheme with a valid proof"));
```
leleueri: todo: the `authorizeDPoP` method is doing the exact same validation, IMO we can remove this one so everytime the `authorizeDPoP` is used the validation is done

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
```diff
@@ -106,52 +121,83 @@ public void handle(RoutingContext context) {
...
+            return;
+        }
+        dpopProofValidator.validateForResource(new VertxHttpServerRequest(context.request().getDelegate()), auth.authToken, thumbprint)
+                .subscribe(
+                        ignored -> authorize(context, token),
+                        error -> {
+                            if (error instanceof InvalidDPoPProofException) {
+                                processException(context, new InvalidTokenException(error.getMessage()));
+                            } else {
+                                context.fail(error);
```
leleueri, line 156: question: unless the general failure handler is logging the stacktrace, should we log at least a WARN to make sure we know the cause ?

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -153,6 +154,14 @@ public Maybe<Token> getRefreshToken(String refreshToken, Client client) {
                 .flatMapMaybe(jwt -> tokenRepository.findRefreshTokenByJti(jwt.getJti()).map(refreshToken1 -> convertRefreshToken(jwt, null)));
     }
 
+    @Override
+    public Maybe<String> getStoredRefreshTokenJkt(String refreshToken, Client client) {
+        return jwtService.decodeAndVerify(refreshToken, client, REFRESH_TOKEN)
+                .onErrorComplete()
+                .flatMap(jwt -> tokenRepository.findRefreshTokenByJti(jwt.getJti()))
+                .flatMap(rt -> rt.getJkt() == null ? Maybe.empty() : Maybe.just(rt.getJkt()));
```
leleueri: discussion: considering that the refresh token should also be binded to the public key, do you think we can avoid the DB lookup at this stage as at some point the RefreshToken revokation will be validated later in the refresh process ? 


https://datatracker.ietf.org/doc/html/rfc9449#section-3 
> If a refresh token is issued to a public client, it is also bound to the public key of the DPoP proof.

gravitee-am-test/specs/gateway/dpop/dpop-binding.jest.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+    for (let i = 0; i < 2; i++) {
+      const refreshProof = await key.proof({ htu: fixture.oidc.token_endpoint, htm: 'POST' });
+      const refreshed = await refreshAccessToken(fixture, refreshToken, refreshProof);
+
+      expect(refreshed.status).toBe(200);
+      expect(refreshed.body.token_type).toBe('DPoP');
+      expect(cnfJkt(refreshed.body.access_token)).toBe(key.jkt);
+      refreshToken = refreshed.body.refresh_token ?? refreshToken;
+    }
+  });
```
leleueri, line 107: issue: according to the specification the refresh token should also be binded to the key but here we do not validate this requirement

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -874,6 +875,18 @@ private Completable validateDomain(Domain domain) throws URISyntaxException {
             } catch (Exception e) {
                 return Completable.error(new InvalidRedirectUriException(e.getMessage()));
             }
+
+            final List<String> dpopSigningAlgorithms = domain.getOidc().getDpopSettings() != null
+                    ? domain.getOidc().getDpopSettings().getDpopSigningAlgorithms() : null;
```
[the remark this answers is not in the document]
lgw-gravitee: This block is already inside the enclosing `if (domain.getOidc() != null)` guard (a few lines above), so `domain.getOidc()` is guaranteed non-null at this line — the extra null check would be redundant. Dismissing.

gravitee-am-ui/src/app/domain/settings/oauth/dpop/dpop.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    private snackbarService: SnackbarService,
+    private authService: AuthService,
+    private domainStore: DomainStoreService,
+  ) {}
+
+  ngOnInit() {
+    this.domainStore.domain$.subscribe((domain) => (this.domain = deepClone(domain)));
+    this.domainId = this.domain.id;
+    this.requireDpopForAll = this.domain.oidc?.dpopSettings?.requireDpopForAll ?? false;
+    this.dpopSigningAlgorithms = this.domain.oidc?.dpopSettings?.dpopSigningAlgorithms ?? [];
```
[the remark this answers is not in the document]
lgw-gravitee: Fixed (1) reactive init: `domainId`, `requireDpopForAll` and `dpopSigningAlgorithms` are now set inside the `domain$` subscription callback so they track later emissions. Leaving (2): the un-scoped `domain$` subscription matches the existing convention across the sibling settings components (self-service-account, saml2, secrets, uma), so not introducing `takeUntil`/`ngOnDestroy` here for consistency.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
lgw-gravitee: Good call — renamed to `PresentedToken` (components `value`/`scheme`) so it no longer overlaps with RAR (RFC 9396) `authorization_details`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/web/auth/handler/impl/OAuth2AuthHandlerImpl.java
lgw-gravitee: Added logging on the DPoP resource-validation failure path: `WARN` for unexpected errors (mirrors the general `authorize()` handler), and `DEBUG` for `InvalidDPoPProofException` — that one is a routine client-side rejection already surfaced to the caller as a 401 + error code, so keeping it at DEBUG avoids flooding logs on probes/replays.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
lgw-gravitee: Good point — done. The refresh token now carries its own `cnf.jkt` (RFC 9449 §3), set in `createRefreshTokenJWT`, so `getStoredRefreshTokenJkt` reads the binding straight from the signed token with no store lookup; revocation is still enforced later in the refresh grant as before. `jkt` remains persisted on the token record (consistent with access tokens) for audit / future revoke-by-key, just no longer read on this path.

gravitee-am-test/specs/gateway/dpop/dpop-binding.jest.spec.ts
lgw-gravitee: Added two integration cases to the refresh-continuity suite: refreshing a DPoP-bound token with a proof signed by a different key is rejected (`invalid_dpop_proof`), and refreshing it with no proof is rejected (`invalid_request`) — covering the RFC 9449 §3 binding requirement end-to-end.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -153,6 +155,13 @@ public Maybe<Token> getRefreshToken(String refreshToken, Client client) {
                 .flatMapMaybe(jwt -> tokenRepository.findRefreshTokenByJti(jwt.getJti()).map(refreshToken1 -> convertRefreshToken(jwt, null)));
     }
 
+    @Override
+    public Maybe<String> getStoredRefreshTokenJkt(String refreshToken, Client client) {
```
leleueri: suggestion: now that we are not doing a lookup in the database, maybe we can rename it getRefreshTokenJkt 🤔 thought?

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
lgw-gravitee: Agreed — renamed to `getRefreshTokenJkt` (interface, impl, call site in `StrategyGranterAdapter`, tests). Also fixed a compile break in `AuthorizationCodeStrategy` spotted while building: the two `dpop_jkt` validation branches referenced `LOGGER` while the class uses `@CustomLog` (field `log`).

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/dpop/impl/ReplayCacheImpl.java
```diff
@@ -0,0 +1,67 @@
...
+        this.cache = cache;
+        this.replayTtlSeconds = replayTtlSeconds;
+        this.redisFallback = redisFallback;
+    }
+
+    @Override
+    public Single<Boolean> register(String jti) {
+        return Single.defer(() -> {
+            String marker = UUID.randomUUID().toString();
+            return Single.defer(() -> cache.rxComputeIfAbsent(jti, ignored -> marker).toSingle())
```
leleueri: **suggestion**: should we have a specific implementation of this ReplyCache for redis to avoid the exception then a fallback to a valid implementation ? Afraid that the exception will have an impact on the performance.

lgw-gravitee: Agreed, done in `39edcb6e30`. The concern is real: `RedisCache.rxComputeIfAbsent` in gravitee-node 9.5.0 throws unconditionally on its first instruction, so every proof validation was constructing an exception, filling in a stack trace through an RxJava stack, and routing it through `onErrorResumeNext` before doing the real work — on both the token endpoint and every resource call.

Split into `ComputeIfAbsentReplayCache` and `SetNxReplayCache` over a shared `AbstractReplayCache` (which keeps the marker/TTL logic), with `ReplayCacheFactory` selecting from `cacheType` — the value it already had, and which was previously passed in only to permit the catch. The `redisFallback` flag and the `onErrorResumeNext` are gone; no path throws on the nominal flow.

Added `shouldNeverProbeComputeIfAbsentOnSetNxBackend` so a reintroduced probe fails loudly — the fake throws the real plugin message. `ReplayCacheImplTest` -> `ReplayCacheTest`, 10 tests green.

leleueri (approved):

PR #8215, opened 2026-07-14
feat: cap thread executors for smtp resource and auditing

lgw-gravitee:


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AuditServiceImpl.java
```diff
@@ -58,33 +53,28 @@ public class AuditServiceImpl implements AuditService, InitializingBean, Disposa
     private Set<String> auditsToExclude = Collections.synchronizedSet(new HashSet<>());
     private boolean excludeSuccessfulAudits;
 
-    private ExecutorService executorService;
-
     @Override
     public void report(AuditBuilder<?> auditBuilder) {
-        executorService.execute(() -> {
+        if (canReport(auditBuilder)) {
             final var audit = auditBuilder.build(objectMapper);
```
wajb: issue: auditing should never fail a request or any primary work. Should we wrap in try/catch?

leleueri: agree, we have to make sure that the exception are properly catched to avoid interruption of the main execution flow

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AuditServiceImpl.java
```diff
@@ -58,33 +55,32 @@ public class AuditServiceImpl implements AuditService, InitializingBean, Disposa
...
-            final var audit = auditBuilder.build(objectMapper);
-            if (canReport(audit)) {
+        try {
+            if (canReport(auditBuilder)) {
+                final var audit = auditBuilder.build(objectMapper);
                 auditReporterService.report(audit);
             }
-        });
+        } catch (Throwable ex) {
+            log.error("An error occurs while trying to report auditType {}", auditBuilder.getType(), ex);
```
wajb: nitpick: can we use "An error occurred..." for new messages? It reads better IMO, even though we have a mix of phrasing.

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/builder/AuditBuilder.java
```diff
@@ -320,6 +320,14 @@ public Audit build(ObjectMapper mapper) {
         return audit;
     }
 
+    public boolean isFailure(){
```
wajb, line 323: nitpick: can we just have `isSuccess`?

wajb (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/builder/AuditBuilder.java
lgw-gravitee: I find `!isSuccess()` as less readable :P

lgw-gravitee: unless we use as `!!!isSuccess()` :P

PR #8262, opened 2026-07-20
feat: per-user console preferences and redesigned domain picker

sclarkamlab:
Summary

Adds per-user console preferences so you can set a default domain and pin the ones you use most. The navbar domain picker is rebuilt around that: a current section, search, pinning, and a more compact menu. Backend stores the prefs on the org user, exposes them through a new current-user preferences endpoint, and audits updates. Also caps the domains `ids` query param at 50 so the picker can't ask for an unbounded list.

Test plan

Unit tests for the new preferences resource, the org user service and the domains `ids` cap. Jest management spec covering console user preferences and the `ids` filter. Manual pass over pin/unpin, set default and search in the picker.

Still a POC, opening as draft for early eyes.


sclarkamlab:
Superseded by #8269 after renaming the branch poc/domain-picker → feat/domain-picker (the rename closed this one instead of retargeting it).

PR #8269, opened 2026-07-20
AM-7306 - feat: per-user console preferences and redesigned domain picker

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7306

Adds per-user console preferences so you can set a default domain and pin the ones you use most, and rebuilds the navbar domain picker around them.

The picker now opens from the domain label in the top bar instead of the user avatar. It has a search, a Current section, an always-visible Default section and a Pinned section, in a compact menu. The default domain is always surfaced even when it isn't pinned, and the domain chip shows a star when the current domain is your default.

Backend stores the prefs on the org user and exposes them through a new current-user preferences endpoint, with updates audited. The domains `ids` query param is capped at 50 so the picker can't ask for an unbounded list.

## :memo: Test scenarios 

- Open the picker from the domain label, search, pin then unpin, set then unset a default.
- The default domain always shows in its own section, and the chip shows a star when the current domain is the default.
- Set a default domain, reload or log back in, you land on it.
- Sign in as a second user, prefs don't leak across accounts.
- Call the current-user preferences endpoint directly, the update lands on the org user and shows in the audit log.
- Domains `ids` filter with more than 50 ids gets capped at 50.

Covered by unit tests (preferences resource, org user service, `ids` cap), a Jest management spec (`user-preferences.jest.spec.ts`) for the endpoint and `ids` filter, and Playwright specs (`domain-picker.spec.ts`) for the picker UI. Full QA steps are on the ticket: https://gravitee.atlassian.net/browse/AM-7306

The picker, opened from the domain label, with the Current / Default / Pinned sections:

image

The domain chip when the current domain is your default:

image

All Domains Page with Default and Pin options inline:

image

New endpoint is in `openapi.yaml`. JDBC adds a liquibase changelog for the console prefs column, Mongo stores it on the org user document.

@gravitee-io/am


podlesrafal (commented):
## Review

Nicely scoped vertical slice — clean persistence (Mongo + JDBC + liquibase), an audited `/user/preferences` endpoint, per-domain permissions correctly enforced on the new `?ids` path, and strong test coverage (unit + integration + Playwright). A few things worth addressing before merge, ordered by severity.

### 🐛 Correctness

**1. [Medium] The picker can request more than 50 ids → backend replies 400**

In `navbar.component.ts` `loadPinnedDomains()`:

```ts
const ids = [...new Set([currentId, defaultId, ...pinnedIds].filter(Boolean))];
...
this.domainService.findByIds(ids).subscribe((response) => (this.domains = response.data));
```

`pinnedDomainIds` is capped at **50** (`@Size(max = 50)`), but here we also prepend `currentId` and `defaultId`. When both are outside the pinned set, `ids.length` becomes **52**, while the `?ids` query param is bounded by `@Size(max = 50)` on `DomainsResource.list` → **400 Bad Request**. The `subscribe` has no `error` callback, so the picker is left with a stale/empty list and an unhandled error.

Repro: pin 50 domains, then navigate into a 51st (non-pinned) domain and open the picker.

Suggested fix: raise the backend `ids` cap to 52, or clamp client-side while guaranteeing current + default stay in, or fetch current/default separately via `getByIdSilently`.

**2. [Low] No error handling on the mutating subscriptions**

`togglePin` / `toggleDefault` (both `navbar.component.ts` and `domains.component.ts`) and `findByIds` subscribe without an `error` handler. On a failed PUT/GET the UI silently does nothing — no snackbar, no rollback. Local state stays consistent (there is no optimistic update — `tap` in `update()` only runs on success), so this is mainly missing user feedback. Consider an `error` → snackbar.

**3. [Low] Read-modify-write without locking (possible lost update)**

`OrganizationUserServiceImpl.updateConsolePreferences` does `findById` → `setConsolePreferences` → `update(user)` on the whole user, and the frontend also does `{...this.preferences(), ...}`. Two quick actions (e.g. pinning from two tabs) can clobber each other. Low risk for personal prefs — flagging for awareness, not a blocker. Side effect: every toggle rewrites the full org-user record and bumps `updatedAt`.

### 🔒 Security — looks good (verified in code)

- **`?ids` does not bypass permissions.** After the `ids` branch selects `domainSource`, `DomainsResource.list` still applies `checkAnyPermission(DOMAIN, LIST)` + global `DOMAIN:READ` or the per-domain `getResourceIdsWithPermission` filter. A user cannot see domains they lack access to, even by passing their ids. `shouldGetDomains_byIds` confirms the cross-env domain is dropped.
- **Env isolation.** `findByIdIn` is a single unscoped `in(_id)` query; the added `.filter(referenceType == ENVIRONMENT && environmentId.equals(referenceId))` correctly scopes it to the path env.
- **Preferences are self-only** (`getAuthenticatedUser().getId()`), gated by `isAuthenticated()`, and storing dead/invalid ids gives no escalation (reads are still permission-filtered; a dead default falls back to the env page via `SKIP_404_REDIRECT`).

### ⚡ Performance — good

- `findByIdIn` is a single `in(...)`, no N+1; the cap bounds query/URL size.
- Picker search uses `debounceTime(300)` + `distinctUntilChanged` + `switchMap` — correct, cancels stale requests.

### 🎨 Conventions / style

- **The magic `50` lives in ~4 places** (`ConsoleUserPreferences.@Size`, `DomainsResource.MAX_DOMAIN_IDS`, `openapi.yaml` ×2, and implicitly the frontend). The `DomainsResource` comment acknowledges the coupling — but raising the cap later is easy to get partially wrong (see #1).
- **`any` in the navbar** (`domain: any`, `currentDomainRow`, `pinnedDomainRows`) is consistent with the existing component, though `UserPreferencesService` already exposes the `ConsoleUserPreferences` type.
- **`defaultDomainId(): string`** actually returns `string | undefined` (`defaultDomainId?`), and `toggleDefaultDomain` assigns `null` to a `string?` field — works functionally, mismatches under strict TS.

### ✅ Tests — strong

Java unit (resource get/update/>50→400/403, service get/empty/update+audit), repository round-trip (Mongo + JDBC), Jest management spec, and Playwright UI. Gaps: no test for the >50-ids picker scenario (#1) and none for the error-feedback path (#2).

### Verdict

Good to merge once **#1** is addressed (the only real bug); the rest are nice-to-haves.

sclarkamlab:
Thanks for the thorough review. Sorted #1 with the client side clamp: current and default lead the id list and it gets sliced to 50, so worst case a couple of trailing pinned entries drop out of the picker instead of a 400. While I was in there I tied DomainsResource.MAX_DOMAIN_IDS to a new ConsoleUserPreferences.MAX_PINNED_DOMAINS so the java side of the magic 50 is one constant now, and fixed the string | null typing nits in the preferences service.

On #2, the global HttpRequestInterceptor already pops a snackbar for any failed request (401, 403 and everything else) unless a call opts out via SKIP_ERROR_SNACKBAR, and none of these calls do. Component level handlers would double toast, so I've left them as is.

#3 I'll pick up as a follow-up, agree it's low risk for personal prefs.

Jest and playwright suites re-run green after the changes.

podlesrafal (approved):

PR #8276, opened 2026-07-21
test: fix and instrument flaky gateway jest suites

sclarkamlab:
Chipping away at the flaky gateway jest suites. Two straight fixes: forgot-password handed jest a 20s timeout shorter than the 30s+60s domain-start barrier its beforeAll waits on, and the saml metadata-url login wasn't polling long enough for the login button to render under load.

The third commit adds opt-in metrics tracking (ENABLE_METRICS_TRACKING, default off) that samples gateway/management heap, GC, threads, fds and domain-sync state at the start and end of each suite and stores it as a CI artifact. Most of the flaky dataset is one intermittent whole-job contention cascade rather than lots of separate bugs, and I want the CI-side numbers to pin down what's actually starving. Ran the full gateway suite locally with it on and everything's flat (0 full GCs, sync never backs up, same 512m heap cap as CI), so it's looking like CPU contention on the shared executor, not a leak or heap.

https://gravitee.atlassian.net/browse/AM-7300

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -514,7 +514,7 @@ jobs:
       - run:
           name: Run Gateway Jest tests (sequential)
           command: ENABLE_METRICS_TRACKING=1 JEST_RETRIES=2 JEST_JUNIT_OUTPUT_NAME=junit-gateway.xml RUN_SPIRE_TESTS=true npm --prefix gravitee-am-test run ci:gateway
-      # AM-7300: capture per-suite JVM/sync metrics even when the gateway suite fails
+      # capture per-suite JVM/sync metrics even when the gateway suite fails
       - store_artifacts:
           path: gravitee-am-test/jest-reports/metrics
           destination: gateway-metrics
@@ -573,7 +573,7 @@ jobs:
       - run:
           name: Run Gateway Jest tests (sequential)
           command: ENABLE_METRICS_TRACKING=1 JEST_RETRIES=2 JEST_JUNIT_OUTPUT_NAME=junit-gateway.xml REPOSITORY_TYPE=jdbc npm --prefix gravitee-am-test run ci:gateway
-      # AM-7300: capture per-suite JVM/sync metrics even when the gateway suite fails
+      # capture per-suite JVM/sync metrics even when the gateway suite fails
       - store_artifacts:
           path: gravitee-am-test/jest-reports/metrics
           destination: gateway-metrics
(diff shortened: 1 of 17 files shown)


.circleci/workflows.yml
```diff
@@ -567,108 +572,12 @@ jobs:
             curl -fsS --retry 12 --retry-delay 5 --retry-all-errors -o /dev/null -u admin:adminadmin 'http://localhost:18092/_node/health?probes=http-server,security-domain-sync'
       - run:
           name: Run Gateway Jest tests (sequential)
-          command: JEST_RETRIES=2 JEST_JUNIT_OUTPUT_NAME=junit-gateway.xml REPOSITORY_TYPE=jdbc npm --prefix gravitee-am-test run ci:gateway
-      - store_test_results:
-          path: gravitee-am-test/jest-reports/junit
+          command: ENABLE_METRICS_TRACKING=1 JEST_RETRIES=2 JEST_JUNIT_OUTPUT_NAME=junit-gateway.xml REPOSITORY_TYPE=jdbc npm --prefix gravitee-am-test run ci:gateway
+      # AM-7300: capture per-suite JVM/sync metrics even when the gateway suite fails
```
wajb: minor: git history should provide context so we don't normally add task references (except for temporary TODOs).

gravitee-am-test/api/commands/gateway/monitoring-commands.ts
```diff
@@ -17,7 +17,11 @@
 import { DomainState } from '@gateway-apis/MonitoringApi';
 import { retryUntil } from '@utils-commands/retry';
 
-const DEFAULT_TIMEOUT_MS = 30000;
+// AM-7300: raised from 30s. On slow CI executors, domain deploys that normally take ~1.5s stretch
```
wajb: minor: this comment is useful for reviewing but not persisting IMO.

gravitee-am-test/specs/gateway/saml2/setup.ts
```diff
@@ -611,6 +744,30 @@ async function waitForSamlMetadataReady(providerDomain: Domain): Promise<void> {
   await withRetry(() => performGet(metadataUrl, '').expect(200), 60, 500);
 }
 
+/**
+ * Poll until the domain's OIDC endpoint stops serving (undeployed), so a disable→enable cycle produces a
+ * genuine redeploy instead of being coalesced into a no-op by the gateway's periodic sync.
+ */
+async function waitForOidcDown(domainHrid: string, options?: { timeoutMs?: number; intervalMs?: number }): Promise<void> {
```
wajb: suggestion: this could be moved to `domain-management-commands.ts`.

wajb (approved):

gravitee-am-test/specs/gateway/saml2/setup.ts
sclarkamlab: Done

gravitee-am-test/api/commands/gateway/monitoring-commands.ts
sclarkamlab: Done

.circleci/workflows.yml
sclarkamlab: Done

PR #8282, opened 2026-07-22
feat: AM-7238 restrict gateway plugins by org license

wajb:
[AM-7238](https://gravitee.atlassian.net/browse/AM-7238)

In managed cloud mode, gate initialisation of EE features according to the active organization's license.
- Changes in an organization's license trigger a redeploy of dependant domains.
- `PluginLicenseGate` is wrapped for domain-specific license checks and synchronises its results with the `DomainReadinessService`.
  - The internal readiness endpoint serves unlicensed features as successful but caveated with a message about the restriction by licensing - e.g.
```
        ...
        "saml2-idp": {
            "id": "saml2-idp",
            "type": null,
            "success": true,
            "message": "Plugin 'saml2-idp' requires the feature 'am-idp-gateway-handler-saml' which is not included in your organization's license",
            "lastSync": 1784728625081
        },
        ...
```
- Restricted plugins are skipped, meaning behaviours and HTTP routes can become unavailable.
- License checks in the gateway run at midnight and when the server is restarted.

Non-cloud mode behaviour is unaffected.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7238]: https://gravitee.atlassian.net/browse/AM-7238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
@@ -20,6 +20,7 @@ import io.gravitee.am.common.event.IdentityProviderEvent;
 import io.gravitee.am.common.event.Type;
 import io.gravitee.am.gateway.handler.common.auth.idp.IdentityProviderManager;
 import io.gravitee.am.gateway.handler.common.certificate.CertificateManager;
+import io.gravitee.am.gateway.handler.common.license.DomainPluginLicenseGate;
 import io.gravitee.am.identityprovider.api.AuthenticationProvider;
 import io.gravitee.am.identityprovider.api.UserProvider;
 import io.gravitee.am.model.Domain;
@@ -31,6 +32,7 @@ import io.gravitee.am.plugins.idp.core.AuthenticationProviderConfiguration;
 import io.gravitee.am.plugins.idp.core.IdentityProviderPluginManager;
 import io.gravitee.am.repository.management.api.IdentityProviderRepository;
 import io.gravitee.am.monitoring.DomainReadinessService;
+import io.gravitee.am.service.PluginLicenseGate;
 import io.gravitee.common.event.Event;
 import io.gravitee.common.event.EventListener;
 import io.gravitee.common.service.AbstractService;
@@ -79,6 +81,9 @@ public class IdentityProviderManagerImpl extends AbstractService implements Iden
 
     @Autowired
     private DomainReadinessService domainReadinessService;
+
+    @Autowired
+    private DomainPluginLicenseGate domainPluginLicenseGate;
 
     private final ConcurrentMap<String, AuthenticationProvider> providers = new ConcurrentHashMap<>();
     private final ConcurrentMap<String, IdentityProvider> identities = new ConcurrentHashMap<>();
@@ -178,7 +183,7 @@ public class IdentityProviderManagerImpl extends AbstractService implements Iden
 
     private void removeIdentityProvider(String identityProviderId) {
         logger.info("Domain {} has received identity provider event, delete identity provider {}", domain.getName(), identityProviderId);
-        clearProvider(identityProviderId);
+        clearProvider(identityProviderId, true);
     }
 
     private Single<IdentityProvider> updateAuthenticationProvider(IdentityProvider identityProvider) {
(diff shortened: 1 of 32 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/idp/impl/IdentityProviderManagerImpl.java
```diff
@@ -193,6 +198,12 @@ private Single<IdentityProvider> updateAuthenticationProvider(IdentityProvider i
     private Single<IdentityProvider> forceUpdateAuthenticationProvider(IdentityProvider identityProvider) {
         String identityProviderId = identityProvider.getId();
         try {
+            if (!domainPluginLicenseGate.check(PluginLicenseGate.TYPE_IDENTITY_PROVIDER, identityProvider.getType(), identityProviderId)) {
+                if (hasExistingProvider(identityProviderId)) {
+                    clearProvider(identityProviderId);
```
sclarkamlab: `check()` just recorded this idp as unlicensed via `pluginUnlicensed`, then `clearProvider` calls `pluginUnloaded` which wipes that record straight back out of the readiness state. Only bites when an already loaded provider goes unlicensed on an update event, but the ticket wants the 404 handler reading this state so the record needs to survive. Same pattern in `AuthorizationEngineManagerImpl.deployProvider`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/license/DomainPluginLicenseGate.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param pluginType one of the {@link PluginLicenseGate} {@code TYPE_*} constants
+     * @param pluginId the plugin identifier (the {@code type} of the configured instance)
+     * @param instanceId the configured instance identifier, used to record the domain state
+     * @return {@code true} when the plugin may be loaded, {@code false} when it is not licensed
+     */
+    public boolean check(String pluginType, String pluginId, String instanceId) {
+        try {
+            pluginLicenseGate.check(Reference.domain(domain.getId()), pluginType, pluginId).blockingAwait();
+            return true;
+        } catch (LicenseFeatureRequiredException e) {
```
sclarkamlab, line 68: Only `LicenseFeatureRequiredException` is caught here, anything else from the first domain to org resolution (DB hiccup, `DomainNotFoundException`) propagates and every call site handles it differently. The cert initial load kills the whole subscribe and leaves the in flight cert `syncState=false` forever, the authenticator predicate in `VertxSecurityDomainHandler` has no catch at all so domain startup dies, protocols catch per item. Worth picking one behaviour for unexpected gate errors and applying it everywhere.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/VertxSecurityDomainHandler.java
```diff
@@ -187,7 +195,8 @@ private void startSecurityDomainProtocols() {
 
     private void startSecurityDomainAuthenticators() {
         logger.info("Start security domain authenticators");
-        List<AuthenticatorProvider> providers = authenticatorPluginManager.createAll(applicationContext);
+        List<AuthenticatorProvider> providers = authenticatorPluginManager.createAll(applicationContext,
+                pluginId -> domainPluginLicenseGate.check(PluginLicenseGate.TYPE_AUTHENTICATOR, pluginId, pluginId));
```
sclarkamlab, line 199: `createAll` iterates `findAll()`, so this checks every authenticator deployed on the gateway, not just the ones this domain uses. Each denial writes a `pluginUnlicensed` entry into this domain's readiness, so domains pick up unlicensed entries for plugins they never configured. That feeds false positives straight into the 404 handler once it lands on `DomainState`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/license/DomainPluginLicenseGate.java
wajb: Went with the fail-open approach for unhandled exceptions. It logs a warning but does not prevent plugin initialisation for license checks.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/VertxSecurityDomainHandler.java
wajb: Authenticator plugin types are node-wide rather than configured per domain, so it's arguably better to over-record license restrictions than to invent a new mechanism for finding out if a domain exercises each type. Wdyt? The 404 handler does not inspect the domain state so the unlicensed plugins are never publicly attributed.

sclarkamlab: OK that makes sense, thanks for clarifying.

sclarkamlab (approved):

PR #8283, opened 2026-07-22
feat(entrypoint): resolve domain entrypoint URL from environment in cloud mode

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7268

In managed cloud the app Overview and Endpoints pages showed the internal gateway URL instead of the environment's Cockpit access point. Both pages read the same `listEntryPoint` endpoint so I fixed it there once. In cloud mode it resolves the entrypoint by environment from the `EntryPointManager` cache, falls back to the gateway URL if the env entrypoint hasn't synced yet, and picks one deterministically if an environment somehow has more than one access point. Non-cloud behaviour is unchanged.

## :memo: Test scenarios

Unit (`DomainServiceTest`) covers the four cases: single access point, none/fallback, several, and non-cloud.

Cloud jest (`gravitee-am-test/specs/cloud/domain-entrypoint-url.jest.spec.ts`) provisions env access points through the cockpit mock and asserts the endpoint resolves the environment URL: several then one, single, and a non-empty fallback when there are none.

Full manual QA steps for the handover are on the ticket. They drive `GET .../domains/{domain}/entrypoints` on a `--cloud` stack with curl + jq (single access point, several, pre-sync fallback), plus an optional Console check on the Overview/Endpoints pages and a non-cloud regression. Verified by hand on a local cloud stack, both pages render the environment URL.

No UI code changes. The Overview and Endpoints pages are untouched, they just receive the corrected URL from the endpoint. The user-facing UI check is Test D in the ticket QA steps.

Nothing extra. The fix lives in `DomainServiceImpl`, behind the cloud-mode check, so self-managed installs see no change.

@gravitee-io/am

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
@@ -132,7 +132,6 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Objects;
@@ -276,7 +275,6 @@ public class DomainServiceImpl implements DomainService {
     @Autowired
     private EntryPointManager entryPointManager;
 
-    // Spring environment (node config), distinct from the io.gravitee.am.model.Environment used elsewhere in this class.
     @Autowired
     private org.springframework.core.env.Environment springEnvironment;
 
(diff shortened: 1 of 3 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -796,6 +807,30 @@ public Single<List<Entrypoint>> listEntryPoint(Domain domain, String organizatio
...
+            // churn and are not stable to order on. Prefer the one flagged default (the overridden access
+            // point), otherwise take the first in the list received from the command.
+            Entrypoint selected = environmentEntrypoints.stream()
+                    .filter(Entrypoint::isDefaultEntrypoint)
+                    .findFirst()
+                    .orElseGet(() -> environmentEntrypoints.get(0));
+
+            if (environmentEntrypoints.size() > 1 && !selected.isDefaultEntrypoint()) {
+                LOGGER.warn("Environment {} resolves to {} entrypoints and none is flagged default; using the first", domain.getReferenceId(), environmentEntrypoints.size());
+            }
```
wajb: suggestion: this could be coupled to the fallback `orElseGet` above as a side effect rather than a separate condition. You can still include the `if (environmentEntrypoints.size() > 1)` clause there.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -796,6 +807,30 @@ public Single<List<Entrypoint>> listEntryPoint(Domain domain, String organizatio
                 });
     }
 
+    private Single<List<Entrypoint>> listEnvironmentEntryPoint(Domain domain) {
+        return Single.defer(() -> {
+            List<Entrypoint> environmentEntrypoints = entryPointManager.findByEnvironmentId(domain.getReferenceId());
+            if (environmentEntrypoints.isEmpty()) {
+                Entrypoint fallback = new Entrypoint();
+                ofNullable(dataPlaneRegistry.getDescription(domain).gatewayUrl()).ifPresent(fallback::setUrl);
```
wajb: question: the logic in the preceding method indicates a dataplane description doesn't necessarily define a gateway url. If so, do we need to fall back to some other default or throw? Maybe an entrypoint is required for cloud mode, or can we be lenient because this is consumed for cosmetic purposes?

The comment in the test admits "In managed cloud the DataPlane gatewayUrl may be absent, so the fallback entrypoint can have no url". Is this value null? 🤔

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: done, folded the warn into the orElseGet so the size check rides along with the fallback

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: good catch, it was null and it turns out that's not as cosmetic as I thought. The UI's `resolveBaseUrl` does `entrypoint.url + path` so the overview page would render "null/mydomain", and in vhost mode it calls `entrypoint.url.match(...)` which throws outright.

I've changed the empty case to fall back to the organization entrypoint resolution instead of synthesizing a bare entrypoint. The org default entrypoint is always created with a url (`gateway.url` property, or the domain restriction host), so the endpoint now degrades to exactly the non-cloud behaviour rather than inventing a third code path with a nullable url. Tests updated to pin the fallback url.

wajb (approved):

PR #8289, opened 2026-07-23
refactor(logging): route loggers through NodeLoggerFactory and enable archunit rules

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7325

We'd been skipping the archunit rules from gravitee-parent. This turns them on (gravitee.archrules.skip=false) and does the migration they need. The logging rule bans org.slf4j.LoggerFactory, so I moved everything off @Slf4j and manual LoggerFactory.getLogger onto Lombok @CustomLog, which routes through NodeLoggerFactory via a new root lombok.config, same pattern as gravitee-node and gamma. Shared protected/public loggers that subclasses rely on can't be @CustomLog (it generates a private field), so those keep an explicit field with just the factory swapped. I also added gravitee-node-logging as a provided dep in the root pom so every module can see NodeLoggerFactory, and dropped the now unused slf4j Logger imports.

## :memo: Test scenarios

No new tests, it's a mechanical logging refactor with no behaviour change. Verified with a full reactor mvn verify with the rules on, both checks pass across all 138 modules. The execution-context rule needed no code changes, nothing in AM logs through an enforced ExecutionContext (our EvaluableExecutionContext/ReactableExecutionContext have no withLogger and the legacy gateway.api context is excluded).

No UI changes.

Only src/main is scanned, so tests keep @Slf4j/LoggerFactory as they are. If a class genuinely needs slf4j directly in future, add it to the plugin allowList rather than reintroducing LoggerFactory.

@gravitee-io/am


gravitee-am-service/src/main/java/io/gravitee/am/service/tasks/AbstractTask.java
```diff
@@ -16,19 +16,18 @@
...
 import org.springframework.scheduling.TaskScheduler;
 
 import java.time.Instant;
+import lombok.CustomLog;
 
 /**
  * @author Eric LELEU (eric.leleu at graviteesource.com)
  * @author GraviteeSource Team
  */
+@CustomLog
```
podlesrafal: `AbstractTask` is **abstract** and previously used `LoggerFactory.getLogger(getClass())`, so log lines from base-class methods (e.g. `schedule()` just below) were emitted under the **concrete subclass** logger name at runtime. `@CustomLog` generates `getLogger(AbstractTask.class)` — the declaring class, resolved at compile time — so those lines now log under `AbstractTask` instead (confirmed subclass: `AssignSystemCertificate`).

This is the only abstract class with a dynamic `getClass()` logger that was moved to `@CustomLog`; the ~9 siblings (`AbstractErrorHandler`, `AbstractProtocolProvider`, `AbstractJdbcRepository`, `BaseUserEnhancer`, `AbstractOpenIDConnectAuthenticationProvider`, …) correctly kept an explicit field. Suggest doing the same here to preserve the logger name:

```java
protected final Logger LOGGER = NodeLoggerFactory.getLogger(getClass());
```
(and keep the `LOGGER.warn(...)` call as-is).

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
```diff
@@ -45,9 +44,9 @@
  * @author Nicolas GERAUD (nicolas.geraud at graviteesource.com)
  * @author GraviteeSource Team
  */
+@CustomLog
```
podlesrafal, line 47: Minor / behaviour note: this logger was explicitly named after a **different** class — `io.gravitee.common.event.impl.EventManagerImpl` (the gravitee-common base), not this AM class. `@CustomLog` changes the logger name to `io.gravitee.am.gateway.event.EventManagerImpl`.

The new name is arguably more correct, but it's still a change: any log-level config keyed on the old name stops applying. Fine to keep if intentional — just flagging so it's a conscious choice rather than a silent side effect.

.circleci/workflows.yml
```diff
@@ -926,10 +924,14 @@ jobs:
...
+          name: Build (parallel compile, no tests)
+          command: |
+            # cicd-jdbc deliberately excluded so the JDBC integration tests don't run under parallel load
+            mvn install -T 1C -pl \!gravitee-am-ui -P gio-artifactory-snapshot -P << parameters.jdbcType >> -DskipTests -Dmaven.javadoc.skip=true -s /tmp/.gravitee.settings.xml
+      - run:
+          name: Run JDBC testcontainer tests (sequential)
           command: |
-            # -T 1C runs one build thread per core across the reactor; safe here because -DskipTests means no fixed-port test collisions.
-            mvn install -T 1C -pl \!gravitee-am-ui -P gio-artifactory-snapshot -Pcicd-jdbc -P << parameters.jdbcType >> -DskipTests -Dmaven.javadoc.skip=true -s /tmp/.gravitee.settings.xml
+            mvn test -pl :gravitee-am-repository-jdbc,:gravitee-am-reporter-jdbc,:gravitee-am-dataplane-jdbc -P gio-artifactory-snapshot,cicd-jdbc -P << parameters.jdbcType >> -Dmaven.javadoc.skip=true -s /tmp/.gravitee.settings.xml
```
podlesrafal, line 934: Scope note: splitting the JDBC job into a parallel compile + a sequential `mvn test` is unrelated to the logging refactor (it's AM-7323 test-parallelization work).

Also worth confirming — the previous single `Build` step ran with `-DskipTests`, so it looks like these JDBC testcontainer tests weren't actually executing before and now are. Good change if so, but please confirm coverage is equal-or-better and the extra CI time is acceptable. Same scope note applies to the two flaky-test hardening edits (`JdbcAuditReporterTest`, `OAuth2GenericAuthenticationProviderTest`) — reasonable, but not "logging".

pom.xml
```diff
@@ -872,6 +879,26 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Parent pins archrules 1.0.0 (ArchUnit 1.2.1) whose ASM can't read Java 25 (class major version 69);
+                 override with a Java 25-capable ArchUnit so the rules actually analyse our bytecode -->
+            <plugin>
```
podlesrafal, line 884: This pins ArchUnit `1.4.1` inside the archrules plugin (parent ships it against 1.2.1). Sensible workaround for the Java 25 bytecode issue and `verify` passes — but the plugin itself (archrules 1.0.0) was built against 1.2.1, so it's a silent divergence from the parent.

Suggest a short `TODO(AM-7325)` / tracking note to drop this override once `gravitee-parent` bumps `gravitee-archrules-maven-plugin` to a Java 25-capable ArchUnit, so it doesn't get forgotten.

podlesrafal (commented):
Reviewed the full 598-file diff (pulled via the Files API since it's over GitHub's 300-file `pr diff` limit).

The hard part — keeping `NodeLoggerFactory` on the **runtime** classpath despite the parent's `provided` scope — is handled completely and correctly: overridden to `compile` in every executable artifact (both standalone distributions + both jib images), and those are the only `jib-maven-plugin` modules in the repo, so nothing is missed. `lombok.config` is correct (`stopBubbling`, slf4j field type, `addLombokGeneratedAnnotation`). The bulk of the conversion is behaviour-preserving, and the 9 abstract base classes with a dynamic `getClass()` logger correctly kept an explicit field.

A few inline notes. The one on `AbstractTask` is the only one I'd address before merge — it's a genuine (if minor) logger-name change that contradicts the "no behaviour change" claim and is inconsistent with its sibling abstract classes. The rest are minor / scope notes.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
sclarkamlab: I think this is fine because it is actually using the correct naming now

.circleci/workflows.yml
sclarkamlab: Yes this came up as a ridealong issue from something I worked on yesterday as well, just wanted to sneak it in here to get it resolved for the team. Arguably could have been a separate ticket

gravitee-am-service/src/main/java/io/gravitee/am/service/tasks/AbstractTask.java
sclarkamlab: good catch, fixed in 1a38e1fec0. switched it back to an explicit field, `protected final Logger LOGGER = NodeLoggerFactory.getLogger(getClass())`, so it lines up with the sibling abstract classes and keeps the subclass logger name.

pom.xml
sclarkamlab: added a `TODO(AM-7328)` next to the override in 03ca457b95 and raised AM-7328 to track dropping it once gravitee-parent bumps the archrules plugin to a Java 25 capable archunit.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-container/src/main/java/io/gravitee/am/gateway/event/EventManagerImpl.java
sclarkamlab: intentional, keeping it. the old name pointed at `io.gravitee.common.event.impl.EventManagerImpl`, a different class over in gravitee-common, so it looks like a copy paste slip from way back. the new name actually matches this class, so it's the more correct of the two. anything keying log config off the old name was already aimed at the wrong class.

.circleci/workflows.yml
sclarkamlab: yeah that's AM-7323 test-parallelization work that rode along on this branch, not logging. and you're right, the old single Build step ran with `-DskipTests` so those jdbc testcontainer tests weren't running before and now they are, so coverage is strictly better here. happy to pull it out into its own PR if you'd rather keep this one purely logging.

podlesrafal (approved):

PR #8296, opened 2026-07-24
AM-7303 - ci: run gateway jest tests in parallel

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7303

Also carries the postman work from https://gravitee.atlassian.net/browse/AM-6355

Speeds up the AM test pipeline. Ended up being three things once I started pulling the thread:

- Gateway jest now runs in parallel (`--maxWorkers=2`) instead of `--runInBand`. The suite is wait bound (lots of domain start/sync polling) and was running single threaded on a 4 core box.
- Postman turned out to be the actual pipeline long pole, not jest. Retired the 9 collections we've already ported to jest (19 down to 10) and parallelised the newman runner with bounded concurrency.
- Parallelising jest surfaced a few latent flaky tests, so I fixed those at the root rather than papering over them (see the notes section).

## :memo: Test scenarios

All green across both mongo and psql, jest and postman. Measured gains:

- Gateway jest step: ~16.5 min to ~9 min (about 44%)
- Postman step: ~25 min to ~5.5 min
- Pipeline critical path: ~32 min to ~22 min

Also confirmed a failing postman collection still fails the build under the parallel runner (the non-zero exit propagates through xargs), so we don't lose failure signal.

No UI changes.

A couple of notes for reviewers:

maxWorkers for gateway was never actually wired into CI before, the `ci:gateway:parallel` script had been sitting unused since #6832. What failed previously was the cross container fan out, which is a different mechanism. The tailwind that makes this work now is `services.sync.deploy.parallelism` (AM-7143), the gateway does parallel domain deployment by default since June, which removes the serial deploy contention two concurrent workers would otherwise hit.

The three flaky tests were all latent sync/isolation issues only exposed under parallel load:
- token-exchange had one beforeAll building 9 domains and blowing the jest hook timeout (hook timeouts don't get retried), so I split it into smaller per-concern suites.
- mfa-general is order dependent, so failed tests need to retry in place rather than after the rest of the file has changed the shared state.
- the scim fixture was minting its access token before the app had synced to the gateway, which 401d every call on the slower jdbc runs. It now retries until the app is live.

@gravitee-io/am


postman/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Gravitee.io AM - Tests - Postman",
   "license": "MIT License",
   "scripts": {
-    "test:collections": "for f in collections/*.json; do echo \"Running $f\" && newman run \"$f\" -e environment/docker.json --ignore-redirects --insecure --bail || { echo \"Failed: $f\"; exit 1; }; done"
+    "test:collections": "for f in collections/*.json; do echo \"Running $f\" && newman run \"$f\" -e environment/docker.json --ignore-redirects --insecure --bail || { echo \"Failed: $f\"; exit 1; }; done",
+    "test:collections:parallel": "mkdir -p logs && printf '%s\\n' collections/*.json | xargs -P \"${POSTMAN_CONCURRENCY:-2}\" -I {} sh -c 'echo \"START $1\" && newman run \"$1\" -e environment/docker.json --ignore-redirects --insecure --bail > \"logs/$(basename \"$1\").log\" 2>&1 && echo \"PASS $1\" || { echo \"FAIL $1\"; tail -40 \"logs/$(basename \"$1\").log\"; exit 1; }' _ {}"
```
podlesrafal: Solid runner: the `_ {}` form keeps the filename out of the script string, `--bail` per collection is preserved, and a non-zero child makes `xargs` exit 123 so the build still fails correctly.

One thing worth confirming before we lean on `POSTMAN_CONCURRENCY=3`: are the 10 retained collections mutually independent when run concurrently? They share `environment/docker.json` and the same stack, so if any two mutate **domain-level** OIDC/security config that another depends on, we can get intermittent failures that only show up under load. Green on mongo+psql is good evidence — a quick confirmation that each collection provisions its own domain (or pinning any known-conflicting ones to a single lane) would make this bulletproof.

.circleci/workflows.yml
```diff
@@ -830,8 +830,8 @@ jobs:
           name: Spin up stack with Mongo << parameters.mongo_version >>
           command: MONGO_VERSION=<< parameters.mongo_version >> npm --prefix docker/local-stack run stack:ci:setup:mongo
       - run:
-          name: Run Postman tests
-          command: npm --prefix postman run test:collections
+          name: Run Postman tests (parallel)
+          command: POSTMAN_CONCURRENCY=3 npm --prefix postman run test:collections:parallel
```
podlesrafal, line 834: The parallel runner writes full per-collection logs to `postman/logs/*.log`, but this job only `store_artifacts` the jest metrics dir — on a postman failure we get `tail -40` in stdout and nothing else (the container is torn down and `postman/logs/` is gitignored). Consider adding a `store_artifacts: { path: postman/logs }` step here (and in the psql job around L872) so a failing collection's full run is retrievable. Non-blocking.

Makefile
```diff
@@ -200,17 +200,8 @@ postman: ## Run postman non regression test (require newman npm module)
...
-	@newman run postman/collections/graviteeio-am-client-management-collection-app-version.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
 	@newman run postman/collections/graviteeio-am-flows-collection-app-version.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
-	@newman run postman/collections/graviteeio-am-login-collection-app-version.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
-	@newman run postman/collections/graviteeio-am-logout-collection-app-version.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
-	@newman run postman/collections/graviteeio-am-user-management-collection-app-version.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
-	@newman run postman/collections/graviteeio-am-api-management-collection.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
 	@newman run postman/collections/graviteeio-am-uma2-app-version-collection.json -e postman/environment/dev.json --ignore-redirects --insecure --bail
```
podlesrafal: CI discovers collections via the `collections/*.json` glob, but this target enumerates them by hand. They agree now, but the next retirement/addition has to touch both places or they silently drift. Optional: have `make postman` just call `npm --prefix postman run test:collections` (or `:parallel`) so there's a single source of truth. Non-blocking.

podlesrafal (approved):
Approve — well-executed CI performance work, and the flaky tests were root-caused rather than retried away. I verified the parallel-safety invariants against the branch rather than trusting the description.

**Verified**
- **Token-exchange split is parallel-safe**: `setupTokenExchangeFixture` names domains via `uniqueName(prefix, true)` (faker slug), so the files that each rebuild `defaultFixture`/`restrictedFixture` never collide on a domain name. The renamed delegation file has no orphaned references to the 5 fixtures it dropped — worth checking since babel-jest doesn't type-check, so a stray ref would only fail at runtime.
- **Metrics are already parallel-safe**: `metrics.setup.js` writes `metrics-${process.pid}.ndjson`, one file per worker, so `--maxWorkers=2` can't race the append. `ENABLE_METRICS_TRACKING=1` is fine under parallel.
- **Retry layering is correct**: deferred by default (`retry.setup.js` in `setupFilesAfterEnv`), `retryImmediatelyForThisFile()` overrides only mfa-general and runs after the setup file, so precedence holds and only the order-dependent file retries in place.
- **Hook timeout**: `setup(120000)` raises the `beforeAll` timeout (`jest.setTimeout` governs hooks too), and the split caps domains-per-hook at 4 (was 9) — both address the un-retryable hook timeout.
- **scim fix is a strict improvement**: the old path returned `undefined` on a 401 during sync lag (no `.expect(200)`); it's now a bounded `withRetry` (50×100ms) that asserts 200 + token defined, so genuine breakage surfaces the real error instead of a confusing undefined token downstream.
- **Makefile kept in sync**: the retired collections were removed from `make postman`, so no dangling `newman` path to a moved file.
- **Failure signal preserved**: `--bail` per collection + `xargs` exiting 123 on any non-zero child.

**Non-blocking** (see inline): confirm postman collection isolation under `POSTMAN_CONCURRENCY=3`, store `postman/logs` as CI artifacts for triage, and the Makefile-vs-glob drift.

Nice cleanup overall — this turns one 9-domain `beforeAll` into four bounded, parallelizable suites.

.circleci/workflows.yml
sclarkamlab: good shout. added a store_artifacts for postman/logs with when: always to both the mongo and psql jobs, so a failing collection's full newman log survives the teardown now. in dbce9d0.

Makefile
sclarkamlab: done, make postman just calls npm --prefix postman run test:collections now so the glob is the single source of truth. one wrinkle: that target points at a local dev stack (dev.json) not the docker one, so rather than hardcode docker.json i parametrized the npm script with POSTMAN_ENV (defaults to docker.json for CI) and the makefile passes dev.json. in ec2b2d9.

postman/package.json
sclarkamlab: checked this and they're independent. every collection creates its own uniquely-named domain (flows-app-version, oauth2-app-version/cross-app-version, oauth2-par, oidc-core-app-version, request-object, dcr-app-version, fapi, jarm, scim-domain-app-version, uma2_postman), captures that domain's id/hrid into collection variables (per newman process, so concurrent runs cant see each others), scopes every management write under /domains/{ownId}, and routes every gateway call via /{ownHrid}. i grepped all 10 for management writes outside /domains and there are none, so nothing mutates shared org/env/idp/admin config. the only shared touch is the admin password-grant token which is read-only. so no two can step on each others domain config at concurrency=3. only caveat is the domain names are hardcoded so it relies on a clean stack per run, which CI gives us anyway.

PR #8298, opened 2026-07-24
test: (AM-7330) enhance automated tests for domain picker feature

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7330

 Adds automated regression coverage for the domain picker / console preferences feature (AM-7306), closing gaps found during manual QA.

  **Backend/API:** a Jest test confirming PUT /user/preferences fully replaces the preferences object rather than merging fields, a boundary test for the 50-pinned-domains cap
  (exactly 50 succeeds, 51 rejected), an audit-log check confirming preference updates are recorded, and a ids-query-param cap check on the domains list endpoint.

  **Frontend:** a new unit spec for UserPreferencesService covering the pin/default toggle logic and the one-shot default-domain-redirect flag, independent of the browser. Playwright
  E2E coverage extended for the navbar picker — reload-to-default-domain, graceful fallback when the default domain has been deleted, search filtering and empty-state, pin/unpin
  and set/unset-default via the picker's own controls, the avatar menu staying unaffected, and the "All domains" list's pin/star controls staying in sync with the navbar picker.

## :memo: Test scenarios 
NA

NA

NA

NA

### DIFF
--- a/gravitee-am-test/playwright/pages/home.page.ts
+++ b/gravitee-am-test/playwright/pages/home.page.ts
@@ -47,6 +47,44 @@ export class HomePage extends BasePage {
     return this.page.locator('.domain-information a').filter({ hasText: name }).first();
   }
 
+  get datatable(): Locator {
+    return this.page.locator('ngx-datatable');
+  }
+
+  /** Row container for a domain in the "All domains" list, by name. */
+  domainRow(name: string): Locator {
+    return this.datatable.locator('.datatable-body-row').filter({ hasText: name });
+  }
+
+  /** Toggle the default-domain star for a row on the "All domains" list page. */
+  async toggleDefaultInList(name: string): Promise<void> {
+    await this.domainRow(name)
+      .getByRole('button')
+      .filter({ has: this.page.locator('mat-icon', { hasText: /^star/ }) })
+      .click();
+  }
+
+  /** Toggle the pin control for a row on the "All domains" list page. */
+  async togglePinInList(name: string): Promise<void> {
+    await this.domainRow(name)
+      .getByRole('button')
+      .filter({ has: this.page.locator('mat-icon', { hasText: /bookmark/ }) })
+      .click();
+  }
+
+  /** Default-domain star icon for a row on the "All domains" list page — asserts via toHaveText, not a one-shot read, since the icon updates asynchronously after the toggle click. */
+  defaultIconInList(name: string): Locator {
+    return this.domainRow(name).locator('mat-icon').filter({ hasText: /^star/ }).first();
+  }
+
+  /** Pin/bookmark icon for a row on the "All domains" list page — asserts via toHaveText, not a one-shot read, since the icon updates asynchronously after the toggle click. */
+  pinIconInList(name: string): Locator {
+    return this.domainRow(name)
+      .locator('mat-icon')
+      .filter({ hasText: /bookmark/ })
+      .first();
+  }
+
   /** Navigate to a domain's detail page via the domains list search. */
   async navigateToDomain(domainName: string): Promise<void> {
     await this.gotoDomainsList();
--- a/gravitee-am-test/playwright/pages/home.page.ts
+++ b/gravitee-am-test/playwright/pages/home.page.ts
@@ -51,38 +51,36 @@ export class HomePage extends BasePage {
     return this.page.locator('ngx-datatable');
   }
 
-  /** Row container for a domain in the "All domains" list, by name. */
-  domainRow(name: string): Locator {
+  /** Row container for a domain in the "All domains" list, by name — filters via the search box first, same as navigateToDomain, to avoid pagination issues when many domains exist. */
+  async domainRow(name: string): Promise<Locator> {
+    const searchInput = this.page.getByPlaceholder(/search/i).first();
+    await expect(searchInput).toBeVisible();
+    await searchInput.fill(name);
     return this.datatable.locator('.datatable-body-row').filter({ hasText: name });
   }
 
   /** Toggle the default-domain star for a row on the "All domains" list page. */
   async toggleDefaultInList(name: string): Promise<void> {
-    await this.domainRow(name)
-      .getByRole('button')
-      .filter({ has: this.page.locator('mat-icon', { hasText: /^star/ }) })
-      .click();
+    await this.domainRow(name); // filters the datatable so the row is actually rendered first
+    await this.page.getByTestId(`domainDefaultToggle-${name}`).click();
   }
 
   /** Toggle the pin control for a row on the "All domains" list page. */
   async togglePinInList(name: string): Promise<void> {
-    await this.domainRow(name)
-      .getByRole('button')
-      .filter({ has: this.page.locator('mat-icon', { hasText: /bookmark/ }) })
-      .click();
+    await this.domainRow(name); // filters the datatable so the row is actually rendered first
+    await this.page.getByTestId(`domainPinToggle-${name}`).click();
   }
 
   /** Default-domain star icon for a row on the "All domains" list page — asserts via toHaveText, not a one-shot read, since the icon updates asynchronously after the toggle click. */
-  defaultIconInList(name: string): Locator {
-    return this.domainRow(name).locator('mat-icon').filter({ hasText: /^star/ }).first();
+  async defaultIconInList(name: string): Promise<Locator> {
+    await this.domainRow(name);
+    return this.page.getByTestId(`domainDefaultToggle-${name}`).locator('mat-icon');
   }
 
   /** Pin/bookmark icon for a row on the "All domains" list page — asserts via toHaveText, not a one-shot read, since the icon updates asynchronously after the toggle click. */
-  pinIconInList(name: string): Locator {
-    return this.domainRow(name)
-      .locator('mat-icon')
-      .filter({ hasText: /bookmark/ })
-      .first();
+  async pinIconInList(name: string): Promise<Locator> {
+    await this.domainRow(name);
+    return this.page.getByTestId(`domainPinToggle-${name}`).locator('mat-icon');
   }
 
   /** Navigate to a domain's detail page via the domains list search. */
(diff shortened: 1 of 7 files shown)


gravitee-am-test/playwright/pages/home.page.ts
```diff
@@ -47,6 +47,44 @@ export class HomePage extends BasePage {
...
+
+  /** Row container for a domain in the "All domains" list, by name. */
+  domainRow(name: string): Locator {
+    return this.datatable.locator('.datatable-body-row').filter({ hasText: name });
+  }
+
+  /** Toggle the default-domain star for a row on the "All domains" list page. */
+  async toggleDefaultInList(name: string): Promise<void> {
+    await this.domainRow(name)
+      .getByRole('button')
```
wajb: suggestion: many of these selectors are brittle because they reference generic attributes and use filters on cosmetic aspects. Could you enhance the markup by exposing `data-testid` attributes or techniques recommended in https://github.com/gravitee-io/gravitee-access-management/blob/master/gravitee-am-test/playwright/README.md#selector-strategy?

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
```diff
@@ -125,4 +127,173 @@ test.describe('Navbar domain picker', () => {
     await navbar.open();
     await expect(navbar.row(name)).toBeVisible();
   });
+
+  test('reload lands the user directly on their default domain', async ({ homePage, extraDomains, setPreferences }) => {
```
podlesrafal: With `fullyParallel: true` (3 workers on CI) every test in this file mutates the same admin user's preferences, and the auto `cleanPreferences` reset from a concurrently-running test can land between `setPreferences` and the page load here. Consider `test.describe.configure({ mode: 'default' });` at the top of the file to serialize it within one worker.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
```diff
@@ -125,4 +127,173 @@ test.describe('Navbar domain picker', () => {
...
+      defaultDomainId: extraDomains.toDefault.id,
+      defaultEnvironmentId: process.env.AM_DEF_ENV_ID,
+    });
+
+    await safeDeleteDomain(extraDomains.toDefault.id, adminToken);
+
+    await homePage.navigate('/');
+
+    // never tries to load the now-deleted domain, and no error is shown
+    expect(page.url()).not.toContain(extraDomains.toDefault.id);
```
podlesrafal: `page.url()` is a one-shot read right after `goto` — the preference redirect only happens after the async preferences load, so this asserts before the app could possibly redirect, and both assertions being negative means the test also passes on a blank page. Suggest asserting the positive landing state first (`waitForReady()` + `expectUrlMatches(...)` for where the user *should* land), then `await expect(page).not.toHaveURL(new RegExp(extraDomains.toDefault.id))`.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
```diff
@@ -125,4 +127,173 @@ test.describe('Navbar domain picker', () => {
...
+    // reflected in the navbar picker
+    await navbar.open();
+    await expect(navbar.sectionLabel('Default')).toBeVisible();
+    await expect(navbar.row(extraDomains.toDefault.name)).toBeVisible();
+    await expect(navbar.sectionLabel('Pinned')).toBeVisible();
+    await expect(navbar.row(extraDomains.toPin.name)).toBeVisible();
+
+    // and unpinning from the navbar picker is reflected back on the "All domains" list
+    await navbar.pin(extraDomains.toPin.name);
+    await homePage.gotoDomainsList();
```
podlesrafal, line 220: The unpin click fires a PUT in the background — navigating away immediately can cancel it before it persists (flaky). Add `await expect(navbar.row(extraDomains.toPin.name)).toHaveCount(0);` first (the row only disappears once the PUT response lands), then navigate.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
```diff
@@ -125,4 +127,173 @@ test.describe('Navbar domain picker', () => {
...
+  });
+
+  test('a pinned domain survives a full page reload', async ({ page, homePage, navbar, currentDomain, extraDomains }) => {
+    await homePage.navigateToDomain(currentDomain.name);
+    await navbar.open();
+    await navbar.searchFor(extraDomains.toPin.name);
+    await navbar.pin(extraDomains.toPin.name);
+    await navbar.searchFor('');
+
+    await page.reload();
```
podlesrafal: Same write-then-navigate race: the pin PUT can still be in flight when `reload()` fires, so the pin may never persist. After clearing the search, wait for `await expect(navbar.sectionLabel('Pinned')).toBeVisible();` (only rendered once the PUT resolved) before reloading.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
```diff
@@ -125,4 +127,173 @@ test.describe('Navbar domain picker', () => {
...
+    adminToken,
+    setPreferences,
+  }) => {
+    await setPreferences({ pinnedDomainIds: [extraDomains.toPin.id] });
+    await safeDeleteDomain(extraDomains.toPin.id, adminToken);
+
+    await homePage.navigateToDomain(currentDomain.name);
+    await navbar.open();
+
+    await expect(navbar.row(extraDomains.toPin.name)).toHaveCount(0);
```
podlesrafal, line 268: `toHaveCount(0)` passes instantly, before the picker rows render, so this passes even if the deleted domain would reappear a moment later. Assert a positive render signal first, e.g. `await expect(navbar.sectionLabel('Current')).toBeVisible();`, then the negatives (optionally also `sectionLabel('Pinned')` count 0).

gravitee-am-test/playwright/pages/home.page.ts
```diff
@@ -47,6 +47,44 @@ export class HomePage extends BasePage {
     return this.page.locator('.domain-information a').filter({ hasText: name }).first();
   }
 
+  get datatable(): Locator {
+    return this.page.locator('ngx-datatable');
+  }
+
+  /** Row container for a domain in the "All domains" list, by name. */
+  domainRow(name: string): Locator {
```
podlesrafal: Rows are located directly in the datatable, while `navigateToDomain` below deliberately filters via the search box first ("avoids pagination issues with many domains") — on CI with parallel suites creating domains, the row can sit on page 2 and the click times out. Suggest filtering via the search box before locating the row here too.

gravitee-am-test/specs/management/user-preferences.jest.spec.ts
```diff
@@ -80,6 +112,25 @@ describe('Console user preferences', () => {
...
+  it('should record an audit entry when preferences are updated', async () => {
+    const putResponse = await performPut(
+      managementUrl(),
+      '/user/preferences',
+      { defaultDomainId: domainA.id, defaultEnvironmentId: process.env.AM_DEF_ENV_ID },
+      authHeaders(),
+    );
+    expect(putResponse.status).toBe(200);
+
+    const auditResponse = await performGet(getOrganisationManagementUrl(), '/audits?type=USER_PREFERENCES_UPDATED&size=1', authHeaders());
```
podlesrafal: Two robustness notes: audits are written asynchronously by the reporter, so the entry from the PUT above may not be visible yet when this file runs alone; and `data[0]` matches *any* prior `USER_PREFERENCES_UPDATED` entry (earlier tests in this file already created some), so this doesn't verify the audit of *this* update. Suggest a small poll/retry plus correlating the entry (actor/target) with this request.

gravitee-am-ui/src/app/services/user-preferences.service.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+  let service: UserPreferencesService;
+  const preferencesUrl = `${AppConfig.settings.baseURL}/user/preferences`;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      teardown: { destroyAfterEach: false },
+      providers: [UserPreferencesService, provideHttpClient(withInterceptorsFromDi()), provideHttpClientTesting()],
+    });
+
+    httpTestingController = TestBed.get(HttpTestingController);
```
podlesrafal: nit: `TestBed.get` is deprecated since Angular 9 — `TestBed.inject(HttpTestingController)` adds typing for free. `teardown: { destroyAfterEach: false }` also shouldn't be needed for a plain service spec.

podlesrafal (commented):
Solid coverage overall — I checked the selectors against the navbar template and the unit spec against the service implementation, and they line up. Main inline points: two write-then-navigate races, a false-passing fallback test, and the `fullyParallel` × shared-admin-preferences interaction; the rest are nits.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
wajb, line 1: thought: while the additional coverage is thorough, there are now 17 tests in this suite that carry out small scenarios and that feels too granular and therefore time-expensive for a fully e2e stack. https://output.circle-artifacts.com/output/job/e8507234-fa52-4ee6-badc-f4942b867e83/artifacts/0/playwright-report/index.html. IMO we might need to categorise such tests to avoid overhead in general test runs and have them available only on a "full run mode" -- or we could think about larger, happy-path or combined scenarios -- or keep the low-level UI tests to Angular. Wdyt?

gravitee-am-test/playwright/pages/home.page.ts
sagarchauhan-qa: Addressed. Added data-testid attributes to the pin/default buttons in both the navbar picker and the domains list, and updated the tests to use them instead of the old selectors.

gravitee-am-test/playwright/tests/navbar/domain-picker.spec.ts
sagarchauhan-qa: Went with combining the small tests into a few bigger scenarios instead of a separate "full run" mode — down from 18 tests to 11, same coverage, just grouped better.

wajb (approved):

PR #8308, opened 2026-07-28
feat: prefer the overriding access point when resolving an environment entrypoint

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7298

Cockpit tells us which of an environment's gateway access points is the customer's override, via `AccessPoint.isOverriding()`. We were dropping it, so `listEnvironmentEntryPoint` had to guess which entrypoint to build URLs from.

I carry the bit on the existing `defaultEntrypoint` field, inverted. The access point Cockpit generates itself becomes the environment's default entrypoint, the customer's override does not. Counter intuitive, but it avoids a new field, a migration across six DB dialects and an API contract change. Resolution then drops the default whenever an override exists, which mirrors what the organization path has always done.

The rule sits on `EntryPointManager` in `gravitee-am-service` rather than in the management API, so the gateway can reach it too. That's the "apply this logic everywhere in MAPI/GW" part of the ticket, and it means AM-7229 and AM-7230 become call sites instead of two independent reimplementations.

### Worth knowing before you review

Cockpit already does this selection upstream, so the drop-the-default branch will not fire in practice. `CommandFactory.environmentCommand` groups access points by target and sends either only the overriding ones or the whole group:

```java
var hasOverride = group.stream().anyMatch(AccessPoint::isOverriding);
return hasOverride
    ? group.stream().filter(AccessPoint::isOverriding)
    : group.stream();
```

So AM never receives a generated plus custom mix for GATEWAY, it gets all-overriding or all-generated. Either way the rule returns everything it was given. I have kept it as defence in depth rather than relying on a filter in another repo staying put, and because the flag is still worth persisting as information, but it is not load bearing and I would rather you knew that going in than found it in review.

What that leaves genuinely doing work is the never-empty guard and the console fix. An environment with no custom domain and more than one generated gateway access point comes back all flagged default, which is the normal case rather than an edge case, and the console used to resolve that to undefined.

### Two other decisions worth a look

The cloud branch no longer falls through to `listOrganizationEntryPoint`. `findAll(organizationId)` is organization wide and returns environment scoped rows, so once those are default flagged that path would collect every other environment's default, trip its `size > 1` branch and filter the whole list away. Cutting the delegation means `listOrganizationEntryPoint` is untouched and only reachable in non cloud mode, where environment entrypoints cannot exist.

`overriding` is a primitive boolean, so a payload that omits it deserializes to false and flags everything default. Resolution returns all of them rather than none in that case.

## :memo: Test scenarios

Unit, four seams. `AbstractEntryPointManagerTest` covers the rule itself (none, only generated, generated plus override, several overrides, all flagged default). `EntrypointServiceTest` covers the flag plumbing and deleting a default flagged environment entrypoint, which now goes through the last default guard. `EnvironmentCommandHandlerTest` covers `overriding` true, false and absent. `DomainServiceTest` covers the read path, including a case proving the organization fallback ignores other environments' rows.

Two existing tests encoded the assumption this inverts and are rewritten rather than extended, `shouldGetEntrypoint_cloud_multipleEntrypoints_picksDefaultFlagged` and `..._noneDefault_picksFirstInList`. The five non cloud tests pass untouched as the regression guard.

Jest, `specs/cloud/domain-entrypoint-url.jest.spec.ts` provisions a generated plus an overriding access point and asserts the management API returns the override host.

Run against a managed cloud stack: all 3 `specs/cloud` suites pass (9 tests), plus `applications` (26), `application-endpoints` (9) and `automation/domains` (14). `domain/vhost-mode` fails on a cloud stack but that is AM-7228's vhost ban rejecting with a different message, nothing to do with this branch. 2726 and 642 Java tests green across the two modules, 128 UI tests green, lint and prettier clean.

I also drove the cockpit mock directly. Flags land as expected, `overriding:false` gives `defaultEntrypoint=true`, `overriding:true` gives false, omitted gives true. Resolution returns the override only for a mixed pair, both for two non overriding, the single one when there is one, both for two overrides, and the org default carrying the data plane URL when the environment has none.

No visible UI changes. The one frontend change is a one line fallback in `DomainService.getEntrypoint` so it cannot hand callers an undefined entrypoint.

Some pre existing residue I noticed but deliberately left alone, none of it introduced here. `createEntrypoint` hardcodes `https://` and ignores `AccessPoint.secured`, both from AM-7226. And environment entrypoint deletes now take the last default guard branch instead of short circuiting, so there is an extra organization wide query per entrypoint on each ENVIRONMENT command. It always passes because the organization keeps its own default, and I covered it with a test rather than adding a guard for a case that cannot happen.

@gravitee-io/am

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
@@ -812,27 +812,35 @@ public class DomainServiceImpl implements DomainService {
 
     private Single<List<Entrypoint>> listEnvironmentEntryPoint(Domain domain, String organizationId) {
         return Single.defer(() -> {
-            List<Entrypoint> environmentEntrypoints = entryPointManager.findByEnvironmentId(domain.getReferenceId());
+            List<Entrypoint> environmentEntrypoints = entryPointManager.resolveByEnvironmentId(domain.getReferenceId());
             if (environmentEntrypoints.isEmpty()) {
-                // No environment entrypoint synced yet: degrade to the organization resolution, whose
-                // default entrypoint always carries a url (the UI builds links from it, never null).
-                return listOrganizationEntryPoint(domain, organizationId);
+                return organizationDefaultEntryPoint(domain, organizationId);
             }
-
-            // Cockpit recreates every environment entrypoint on each environment command, so createdAt/id
-            // churn and are not stable to order on. Prefer the one flagged default (the overridden access
-            // point), otherwise take the first in the list received from the command.
-            Entrypoint selected = environmentEntrypoints.stream()
-                    .filter(Entrypoint::isDefaultEntrypoint)
-                    .findFirst()
-                    .orElseGet(() -> {
-                        if (environmentEntrypoints.size() > 1) {
-                            log.warn("Environment {} resolves to {} entrypoints and none is flagged default; using the first", domain.getReferenceId(), environmentEntrypoints.size());
-                        }
-                        return environmentEntrypoints.get(0);
-                    });
-            return Single.just(List.of(selected));
+            return Single.just(environmentEntrypoints);
         });
+    }
+
+    /**
+     * The environment has no entrypoint yet — Cockpit has not synced it. Fall back to the organization's
+     * own default entrypoint carrying the data-plane gateway url, as the organization resolution does.
+     * <p>
+     * Deliberately does not call {@link #listOrganizationEntryPoint}: that one reads every entrypoint in
+     * the organization, environment-scoped rows included, and those are flagged default too — so its
+     * "drop the defaults once something else matched" step could filter the whole list away.
+     */
+    private Single<List<Entrypoint>> organizationDefaultEntryPoint(Domain domain, String organizationId) {
+        return entrypointService.findAll(organizationId)
+                .filter(entrypoint -> entrypoint.getEnvironmentId() == null && entrypoint.isDefaultEntrypoint())
+                .firstElement()
+                .map(entrypoint -> {
+                    ofNullable(this.dataPlaneRegistry.getDescription(domain).gatewayUrl())
+                            .ifPresent(entrypoint::setUrl);
+                    return List.of(entrypoint);
+                })
+                .switchIfEmpty(Single.fromCallable(() -> {
+                    log.warn("Organization {} has no default entrypoint; domain {} resolves to no entrypoint at all", organizationId, domain.getId());
+                    return List.<Entrypoint>of();
+                }));
     }
 
     private Single<Domain> createSystemScopes(Domain domain) {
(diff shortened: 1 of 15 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -812,29 +812,37 @@ private Single<List<Entrypoint>> listOrganizationEntryPoint(Domain domain, Strin
...
 
+    /**
+     * The environment has no entrypoint yet — Cockpit has not synced it. Fall back to the organization's
+     * own default entrypoint carrying the data-plane gateway url, as the organization resolution does.
+     * <p>
+     * Deliberately does not call {@link #listOrganizationEntryPoint}: that one reads every entrypoint in
+     * the organization, environment-scoped rows included, and those are flagged default too — so its
+     * "drop the defaults once something else matched" step could filter the whole list away.
+     */
+    private Single<List<Entrypoint>> organizationDefaultEntryPoint(Domain domain, String organizationId) {
```
wajb: suggestion: this returns 0 or 1 `Entrypoint`, so `Maybe<Entrypoint>` would be a more intuitive fit. The method name is missing a verb too. What about `getOrganizationDefaultEntrypoint`?

gravitee-am-service/src/main/java/io/gravitee/am/service/EntryPointManager.java
```diff
@@ -36,4 +36,23 @@ public interface EntryPointManager extends Service<EntryPointManager> {
...
+     * is flagged {@code defaultEntrypoint} and the customer's overriding one is not (see
+     * {@code EnvironmentCommandHandler}). So whenever an environment resolves to more than its default,
+     * the customer has an override and it wins.
+     * <p>
+     * Never narrows a non-empty environment down to nothing: when no entrypoint survives the filter —
+     * every one of them is flagged default, which is what an access point payload carrying no
+     * {@code overriding} field produces — the full list is returned instead. Callers dereference the
+     * URL of whatever comes back, so an empty result would break them.
+     */
+    default List<Entrypoint> resolveByEnvironmentId(String environmentId) {
```
wajb: thought: the method name does not distinguish the way in which it differs from `findByEnvironmentId`. Since `findByEnvironmentId` is only used for diagnostics, as mentioned in the comments, could that instead be renamed `findAllByEnvironmentId`? (Then the comment can go too IMO.)

gravitee-am-service/src/main/java/io/gravitee/am/service/EntrypointService.java
```diff
@@ -36,7 +36,17 @@ public interface EntrypointService {
 
     Flowable<Entrypoint> findByEnvironment(String organizationId, String environmentId);
 
-    Single<Entrypoint> create(String organizationId, NewEntrypoint entrypoint, User principal);
+    default Single<Entrypoint> create(String organizationId, NewEntrypoint entrypoint, User principal) {
+        return create(organizationId, entrypoint, false, principal);
+    }
+
+    /**
+     * The {@code defaultEntrypoint} flag is deliberately kept off {@link NewEntrypoint}: that DTO is the
```
wajb: nitpick: the docs for this method feels like an odd place for this comment. Perhaps not needed anyway since the OAS checks will highlight any public changes.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: Good shout

wajb (approved):

PR #8310, opened 2026-07-28
feat: build email links from the environment entrypoint in managed cloud

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7229

Targets AM-7298 rather than master, so the diff stays to just this change.

In cloud mode, emails were still built from the data plane gateway url, so a customer got links pointing at the wrong hostname for their environment. `DomainReadServiceImpl.buildUrl` now resolves the environment's entrypoint first and falls back to the data plane url as before. Everything after that in the method is untouched, vhosts still win, the domain path composes the same way.

Which entrypoint is `resolveByEnvironmentId` from AM-7298, narrowed to one by `resolvePrimaryByEnvironmentId`. The list can hold more than one and an email link needs exactly one, so it takes the lowest url. That matters more than it sounds, the cache is a ConcurrentHashMap so "the first one" is not the same on the management api and the gateway, and you would get two emails for one environment with different hosts.

Worth calling out, there is only one `buildUrl` and both planes inject it, so this changes gateway emails too. That is deliberate. It absorbs the entrypoint half of AM-7230 and leaves 7230 with the request object logic and SCIM.

Two things I found on the way that are worth a look, both are in the thread with Eric:

Cockpit only sends the overriding access points once a custom domain exists (`CommandFactory.environmentCommand`), so AM never holds a generated and a custom entrypoint at the same time. That means "always use the default entrypoint" cannot be honoured today, whatever rule AM implements.

The local stack only put the management api into managed cloud. That was fine until now because nothing in the gateway used cloud mode. I have added it to the gateway in the cloud overlay, but it opens the question of whether the production cloud gateways carry `installation.type=managed`. If they do not, the gateway half of this does nothing there and 7230 hits the same wall.

## :memo: Test scenarios

Unit, `AbstractEntryPointManagerTest` covers the selection rule, empty, single, generated plus overriding, several overriding, all flagged default. The generated url deliberately sorts first in one case so it fails if the default is not dropped before the tiebreak.

`DomainReadServiceImplTest` covers the url matrix, non cloud never touches the entrypoint manager, cloud with an entrypoint, trailing slash, no entrypoint falling back to the data plane url, null gateway url falling back to `gateway.url`, and vhost mode still winning. The four existing tests are unchanged apart from the constructor.

Jest, three new specs in `specs/cloud/email-entrypoint-url.jest.spec.ts` for the three request-less flows, pre registration, resend confirmation, and SCIM provisioning. SCIM rather than forgot password on purpose, forgot password has a request so 7230 will change where its host comes from and the assertion would need inverting.

All green against a `--cloud` stack, and the full `ci:cloud` suite still passes with the gateway in cloud mode, 4 suites, 13 tests.

No UI changes.

The vhost branch is unreachable in cloud, `EnvironmentCommandHandler` leaves domainRestrictions unset in managed cloud so `setDeployMode` never turns vhost mode on, and AM-7228 blocks enabling it after. I kept the precedence anyway rather than restructure a branch I cannot test. The one gap is the Automation API, `AutomationDomainMapper.applyTo` sets vhostMode straight from the request body with no cloud guard, that probably wants a ticket against 7228.

@gravitee-io/am

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
@@ -113,7 +113,7 @@ public abstract class AbstractEntryPointManager extends AbstractService<EntryPoi
     }
 
     @Override
-    public Optional<Entrypoint> resolvePrimaryByEnvironmentId(String environmentId) {
+    public Optional<Entrypoint> findPrimaryByEnvironmentId(String environmentId) {
         List<Entrypoint> resolved = findByEnvironmentId(environmentId);
         if (resolved.size() > 1) {
             log.warn("Environment {} resolves to {} entrypoints, building URLs from the first in URL order", environmentId, resolved.size());
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
@@ -31,8 +31,10 @@ import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Flowable;
 import io.reactivex.rxjava3.core.Maybe;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -108,6 +110,15 @@ public abstract class AbstractEntryPointManager extends AbstractService<EntryPoi
         return entrypoints.values().stream()
                 .filter(entrypoint -> environmentId != null && environmentId.equals(entrypoint.getEnvironmentId()))
                 .collect(Collectors.toList());
+    }
+
+    @Override
+    public Optional<Entrypoint> resolvePrimaryByEnvironmentId(String environmentId) {
+        List<Entrypoint> resolved = findByEnvironmentId(environmentId);
+        if (resolved.size() > 1) {
+            log.warn("Environment {} resolves to {} entrypoints, building URLs from the first in URL order", environmentId, resolved.size());
+        }
+        return resolved.stream().min(Comparator.comparing(Entrypoint::getUrl));
     }
 
     @Override
--- a/docker/local-stack/dev/docker-compose.cloud.yml
+++ b/docker/local-stack/dev/docker-compose.cloud.yml
@@ -13,8 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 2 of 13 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
```diff
@@ -110,6 +112,15 @@ public List<Entrypoint> findAllByEnvironmentId(String environmentId) {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public Optional<Entrypoint> resolvePrimaryByEnvironmentId(String environmentId) {
```
wajb: nitpick: `find~` naming would be more consistent.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
```diff
@@ -110,6 +112,15 @@ public List<Entrypoint> findAllByEnvironmentId(String environmentId) {
...
+        List<Entrypoint> resolved = findByEnvironmentId(environmentId);
+        if (resolved.size() > 1) {
+            log.warn("Environment {} resolves to {} entrypoints, building URLs from the first in URL order", environmentId, resolved.size());
+        }
+        return resolved.stream().min(Comparator.comparing(Entrypoint::getUrl));
```
wajb, line 121: thought: we need to define this behaviour carefully. The a/c requires the selection to be the "first in the list" rather than a sorted result. We don't preserve any intrinsic order, but if the _intent_ is that the first entrypoint in the ENVIRONMENT command payload is special, then we may have to amend how we process and store the list 🤔.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
sclarkamlab: Done. Renamed to `findPrimaryByEnvironmentId`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
sclarkamlab: Went digging in the cockpit code to check. There is no ordering intent on their side, `AccessPoint` has no primary flag, the ENVIRONMENT payload builder groups by target via a HashMap and pre-filters the overrides, and gateway domains pass through HashSets before storage so the stored order is incidental. Cockpit's own `EndpointService` resolves a single URL with a min comparator too. In practice we normally get exactly one GATEWAY access point anyway (the hello template only allows one host per target), multiples only appear when a custom AM gateway has several domains and then cockpit only sends the overriding ones. So I am keeping the min-by-URL pick, it is deterministic and identical across the mgmt-api and gateway caches, "first in the list" does not map to anything real upstream.

wajb (approved):

PR #8313, opened 2026-07-29
fix: AM-7343 prevent permanent unlicensed plugin states

wajb:
[AM-7343](https://gravitee.atlassian.net/browse/AM-7343)

Fully initialise plugins for the domain readiness endpoint and set terminal states to avoid sticky unlicensed messages.

- Unlicensed warning entries for **Protocol**, **Authenticator**, **Resource** and **Policy** types are reset when a license change grants them.
- Plugin states for **Protocol**/**Resource**/**Authenticator** are initialised so they are included in domain readiness responses and entries in domain readiness responses always have `type`.
- **Identity providers** are also initialised during redeployment so entries in domain readiness responses always have `type`.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7343]: https://gravitee.atlassian.net/browse/AM-7343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/resource/impl/ResourceManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/resource/impl/ResourceManagerImpl.java
@@ -17,6 +17,7 @@ package io.gravitee.am.gateway.handler.manager.resource.impl;
 
 import io.gravitee.am.common.event.EventManager;
 import io.gravitee.am.common.event.ResourceEvent;
+import io.gravitee.am.common.event.Type;
 import io.gravitee.am.gateway.handler.common.license.DomainPluginLicenseGate;
 import io.gravitee.am.gateway.handler.manager.resource.ResourceManager;
 import io.gravitee.am.model.Domain;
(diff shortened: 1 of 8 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/manager/resource/impl/ResourceManagerImpl.java
```diff
@@ -137,13 +145,20 @@ private void reloadResource(String resourceId) {
...
-                        this.resources.put(res.getId(), res);
-                        this.resourceProviders.put(resourceId, provider);
+                        try {
+                            var provider = resourcePluginManager.create(new ProviderConfiguration(res.getType(), res.getConfiguration()));
+                            provider.start();
+                            this.resources.put(res.getId(), res);
+                            this.resourceProviders.put(resourceId, provider);
+                            domainReadinessService.pluginLoaded(domain.getId(), res.getId());
+                        } catch (Exception e) {
+                            domainReadinessService.pluginUnloaded(domain.getId(), res.getId());
```
sclarkamlab: Did you mean `pluginFailed` here? `pluginUnloaded` deletes the entry without restoring the domain status, so a resource that blows up on reload leaves the domain stuck in INITIALIZING and the readiness probe goes not-ready with no record of why. Then if any unrelated plugin syncs afterwards the status flips back to DEPLOYED and the failure vanishes entirely. `pluginFailed(domain.getId(), res.getId(), e.getMessage())` matches the other managers (`CertificateManagerImpl` etc) and keeps it visible, same for the catch in `afterPropertiesSet`.

wajb: Great point, I overlooked the domain status here. A future improvement might be to recompute the domain state after `pluginUnloaded` calls too, but I'll leave that out of this bug fix. `pluginFailed` is the right call here (and the other new tracked feature managers).

sclarkamlab (approved):

PR #8319, opened 2026-07-29
feat: build email links from the request host in managed cloud

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7230

In managed cloud, reset password, registration and blocked account emails now build their links from the host the user actually arrived on, as long as that host is one of the environment's entrypoints. If it isn't one, or there's no request to take a host from (SCIM, management API), we fall back to the AM-7229 resolution.

Blocked account is in there because that email carries a reset password link too. It fires on a failed login, so the host comes off the same request the login did, and it goes through the same entrypoint check as everything else. Worst case is a link to another host the environment already trusts.

Only a stored entrypoint is ever used to build the url, never the string off the request. So a forged Host header can't point a reset link at someone else's domain, which was the main thing we sorted out in Slack with Eric and William.

Limited to cloud mode as agreed, self hosted behaviour doesn't move.

## :memo: Test scenarios

Unit tests in `DomainReadServiceImplTest` cover the new precedence: request host wins when it matches an entrypoint, matching is case and trailing slash tolerant, an unknown origin is rejected and falls back, the same host on a different scheme or a different port is rejected too, a null origin falls back, and non cloud ignores it completely.

`UriBuilderRequestTest` covers the new `resolveOrigin` helper, including the X-Forwarded handling it inherits and that it carries no path or forwarded prefix.

`UserAuthenticationServiceTest` covers the blocked account side, that the origin survives the hop onto the email thread when the recovery email is on, and that nothing is sent when it's off.

`specs/cloud/email-entrypoint-url.jest.spec.ts` drives the forgot password form end to end against an environment with two entrypoints, once on a host that belongs to the environment and once on a forged one, and asserts the emailed link follows the first and ignores the second.

No UI changes.

Worth knowing the request only ever picks between hosts we already trust, it can't introduce a new one. So behaviour only changes when an environment has more than one host, otherwise it resolves to the same place it did before.

The self hosted `setDeployMode` pick of `vhosts.get(0)` is still arbitrary and has the same underlying problem, we deliberately left it out of scope here. Probably worth its own ticket.

@gravitee-io/am

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
@@ -81,7 +81,7 @@ public interface UserAuthenticationService {
      * @param user End-User to lock
      * @param requestOrigin the origin the failed login came in on, or null when there is no request to
      *                      take one from. The blocked account email carries a reset password link, so it
-     *                      resolves its url the same way the reset password flow does. See AM-7230.
+     *                      resolves its url the same way the reset password flow does.
      * @return
      */
     Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user, @Nullable String requestOrigin);
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
@@ -21,6 +21,7 @@ import io.gravitee.am.model.User;
 import io.gravitee.am.model.account.AccountSettings;
 import io.gravitee.am.model.oidc.Client;
 import io.gravitee.gateway.api.Request;
+import jakarta.annotation.Nullable;
 import io.reactivex.rxjava3.core.Completable;
 import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Single;
@@ -83,7 +84,7 @@ public interface UserAuthenticationService {
      *                      resolves its url the same way the reset password flow does. See AM-7230.
      * @return
      */
-    Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user, String requestOrigin);
+    Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user, @Nullable String requestOrigin);
 
     default Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user) {
         return lockAccount(criteria, accountSettings, client, user, null);
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
@@ -78,9 +78,16 @@ public interface UserAuthenticationService {
      * @param accountSettings account settings
      * @param client oauth2 client
      * @param user End-User to lock
+     * @param requestOrigin the origin the failed login came in on, or null when there is no request to
+     *                      take one from. The blocked account email carries a reset password link, so it
+     *                      resolves its url the same way the reset password flow does. See AM-7230.
      * @return
      */
-    Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user);
+    Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user, String requestOrigin);
+
+    default Completable lockAccount(LoginAttemptCriteria criteria, AccountSettings accountSettings, Client client, User user) {
+        return lockAccount(criteria, accountSettings, client, user, null);
+    }
 
     default Single<User> connect(io.gravitee.am.identityprovider.api.User principal, boolean afterAuthentication) {
         return connect(principal, null, null, afterAuthentication);
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/EmailService.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/EmailService.java
@@ -22,6 +22,7 @@ import io.gravitee.am.model.User;
 import io.gravitee.am.model.oidc.Client;
 
 import io.vertx.core.MultiMap;
+import jakarta.annotation.Nullable;
 import java.io.IOException;
 import java.util.List;
 import java.util.Locale;
@@ -41,7 +42,7 @@ public interface EmailService {
         this.send(template, user, client, queryParams, null);
     }
 
-    void send(Template template, User user, Client client, MultiMap queryParams, String requestOrigin);
+    void send(Template template, User user, Client client, MultiMap queryParams, @Nullable String requestOrigin);
 
     void asyncSend(Template template, User user, Client client, MultiMap queryParams);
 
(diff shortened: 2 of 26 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/EmailService.java
```diff
@@ -37,7 +38,11 @@ default void send(Template template, User user, Client client) {
         this.send(template, user, client, MultiMap.caseInsensitiveMultiMap());
     }
 
-    void send(Template template, User user, Client client, MultiMap queryParams);
+    default void send(Template template, User user, Client client, MultiMap queryParams) {
+        this.send(template, user, client, queryParams, null);
+    }
+
+    void send(Template template, User user, Client client, MultiMap queryParams, @Nullable String requestOrigin);
```
wajb, line 45: thought: the overloads add strain to consumers so they have to do things like instantiate empty MultiMaps. The parameter object pattern might be a better way to go at this point. Just a thought.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
```diff
@@ -78,9 +79,16 @@ public interface UserAuthenticationService {
      * @param accountSettings account settings
      * @param client oauth2 client
      * @param user End-User to lock
+     * @param requestOrigin the origin the failed login came in on, or null when there is no request to
+     *                      take one from. The blocked account email carries a reset password link, so it
+     *                      resolves its url the same way the reset password flow does. See AM-7230.
```
wajb: suggestion: remove jira reference (we can use git to track heritage).

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/utils/UriBuilderRequest.java
```diff
@@ -218,6 +219,48 @@ private static boolean isDefaultPort(String port, String scheme) {
                ("https".equals(scheme) && "443".equals(port));
     }
 
+    /**
+     * This request's public origin as {@code scheme://host[:port]}, using the same {@code X-Forwarded-*}
+     * resolution as {@link #resolveProxyRequest(HttpServerRequest, String, MultiMap, boolean)}.
+     * <p>
+     * Carries no path: callers compare it against configured entrypoints and vhosts, which are hosts
+     * rather than URLs, so neither the request path nor {@code X-Forwarded-Prefix} belongs in it.
```
wajb: nitpick: I'm not a fan of describing usage on utility methods because they become stale. Active references and git history should provide this context IMO, and callers can add comments as needed for clarifications.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/utils/UriBuilderRequest.java
```diff
@@ -218,6 +219,48 @@ private static boolean isDefaultPort(String port, String scheme) {
...
+     */
+    public static @Nullable String resolveOrigin(HttpServerRequest request) {
+        final URI baseUri;
+        try {
+            baseUri = URI.create(resolveProxyRequest(request, "/", (MultiMap) null, false));
+        } catch (IllegalArgumentException e) {
+            // The forwarding headers are caller-controlled and go into the url unencoded, so they can
+            // fail to parse. Callers read null as "no origin" and resolve the url the usual way.
```
wajb: nitpick: buried assumption here about consumers.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/application/DomainReadServiceImpl.java
```diff
@@ -136,4 +140,69 @@ private String resolveEntryPoint(Domain domain) {
...
+        if (matched.isEmpty()) {
+            // Either a forged host or an entrypoint the environment never synced, and the two look the
+            // same from here. Worth saying out loud, because the fallback link silently differs from
+            // the host the user is on.
+            log.warn("Environment {} has no entrypoint matching the request origin, building URLs from the configured entrypoint instead", domain.getReferenceId());
+        }
+        return matched;
+    }
+
+    private static boolean sameOrigin(String entrypointUrl, String requestOrigin) {
```
wajb, line 169: suggestion: could we expose and reuse `UriBuilderRequest::sameOriginAuthority`?

gravitee-am-test/specs/cloud/email-entrypoint-url.jest.spec.ts
```diff
@@ -85,4 +87,84 @@ describe('AM - Cloud - entrypoint url in email links', () => {
     expect(new URL(link).origin).toEqual(expectedOrigin);
     await clearEmails(email);
   });
+
+  it('builds the reset password link from the host the user reached the gateway on', async () => {
+    // The generated access point is a real host for this environment but not the one the request-less
+    // flows above resolve to, so seeing it here is only possible if the request decided it.
+    await clearEmails(fixture.resetPasswordUserEmail);
```
wajb: thought: I don't understand the clearing approach. `clearEmails` accepts calls without an address, where it should remove all emails. Could we not wire that into an `afterEach`?

gravitee-am-test/api/config/dev.setup.js
```diff
@@ -19,6 +19,7 @@ process.env.AM_GATEWAY_URL = 'http://localhost:8092';
 process.env.AM_DOMAIN_DATA_PLANE_ID = 'default';
 process.env.AM_INTERNAL_GATEWAY_URL = 'http://localhost:8092';
 process.env.AM_MONGODB_URI = 'mongodb://localhost:27017';
+process.env.AM_INTERNAL_MONGODB_URI = 'mongodb://mongodb:27017';
```
wajb: issue: this is a dockerised path. Afaik this config relates to the traditional local jvm approach and we can use the CI setup when running via local-stack.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/email/EmailService.java
sclarkamlab: Yeah it is heading that way. Left it for now since the interface already stacks overloads like this, if another param shows up I'll switch it to a params object.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/auth/user/UserAuthenticationService.java
sclarkamlab: Done in 73d1c7a260.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/utils/UriBuilderRequest.java
sclarkamlab: Done in 73d1c7a260, trimmed it to behaviour only.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/vertx/utils/UriBuilderRequest.java
sclarkamlab: Done in 73d1c7a260.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/application/DomainReadServiceImpl.java
sclarkamlab: Couldn't use it directly, handler-common depends on am-service so that'd be a cycle. Moved the comparison into `UriBuilder` in am-common and both sides use it now, 65a4f5688f.

gravitee-am-test/specs/cloud/email-entrypoint-url.jest.spec.ts
sclarkamlab: Done in 7f49bbb986. Kept it per address rather than a blanket delete, the fake smtp inbox is shared with suites running in parallel.

gravitee-am-test/api/config/dev.setup.js
sclarkamlab: Good catch, removed in 8cfb1d99df.

wajb (approved):

PR #8323, opened 2026-07-29
feat: AM-7237 add traceability for organization license changes

wajb:
[AM-7237](https://gravitee.atlassian.net/browse/AM-7237)

Trace organization license changes in management and gateway nodes with logging and audit events (managed cloud mode only).

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7237]: https://gravitee.atlassian.net/browse/AM-7237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
```diff
@@ -65,40 +77,65 @@ public Maybe<License> findByReference(ReferenceType referenceType, String refere
...
+                .flatMap(existing -> {
+                    if (Objects.equals(existing.getLicense(), license)) {
+                        log.debug("License unchanged for referenceType={}, referenceId={}, skipping", referenceType, referenceId);
+                        return Maybe.just(existing);
+                    }
+                    final GraviteeLicense previous = describe(referenceType, referenceId, existing.getLicense());
+                    existing.setLicense(license);
+                    existing.setUpdatedAt(new Date());
+                    return licenseRepository.update(existing)
+                            .flatMap(updated -> emitEvent(referenceType, referenceId, Action.UPDATE).andThen(Single.just(updated)))
```
sclarkamlab, line 94: The audit hooks sit downstream of `emitEvent`, so if the repo write succeeds but `eventService.create` fails, `reportChange` never runs and we record a FAILURE audit for a license that is already persisted. Same shape on the create and delete paths. Verified with a quick test, update succeeds then the event errors and the captured audit is `ORGANIZATION_LICENSE_UPDATED` / FAILURE. Could fix by auditing between the repository write and `emitEvent`, or by auditing the success and the notification failure separately.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
```diff
@@ -65,40 +77,65 @@ public Maybe<License> findByReference(ReferenceType referenceType, String refere
...
-                        toCreate.setUpdatedAt(now);
-                        return licenseRepository.create(toCreate)
-                                .flatMap(created -> emitEvent(referenceType, referenceId, Action.CREATE).andThen(Single.just(created)));
-                    }));
-        });
+        return Maybe.<License>defer(() -> {
+                    validate(license);
+                    return licenseRepository.findById(referenceId, referenceType);
+                })
+                .doOnError(error -> reportFailure(EventType.ORGANIZATION_LICENSE_UPDATED, referenceType, referenceId, error))
```
sclarkamlab, line 84: This fires before the create/update fork, so a failed validation or lookup on a first-time provisioning gets audited as `ORGANIZATION_LICENSE_UPDATED` for a license that never existed. The intent is unknowable up here, so maybe a neutral event type for pre-write failures rather than picking UPDATED.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
```diff
@@ -117,4 +154,60 @@ private Completable emitEvent(ReferenceType referenceType, String referenceId, A
...
+                    .tier(license.getTier())
+                    .packs(license.getPacks())
+                    .features(license.getFeatures())
+                    .expiresAt(license.getExpirationDate())
+                    .isExpired(license.isExpired())
+                    .scope(license.getReferenceType())
+                    .build();
+        } catch (Exception e) {
+            log.warn("Cannot read license details for referenceType={}, referenceId={}", referenceType, referenceId, e);
+            return null;
```
sclarkamlab, line 177: Returning null here means an undecodable license still audits as SUCCESS with no message at all, both sides of the diff are null so the builder skips the detail entirely (`LicenseServiceTest` pins this with `assertNull` on the message). The only trace of why is this warn log outside the trail. Worth putting a marker in the audit payload so an auditor can see the entitlements were unreadable rather than just missing.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
wajb: Agreed it is unknowable. UPDATED was chosen as a reasonable default - with a comment in the test `createOrUpdateReportsFailureAuditWhenLookupFails`. I don't think we should add an in between type for the unlikely edge case. Wdyt?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
wajb: This is fair, though it is the established pattern (see FactorServiceImpl, EmailTemplateServiceImpl, FormServiceImpl, ApplicationServiceImpl, AlertTriggerServiceImpl, etc.). This demonstrates failures during event emission is uncommon. Ok to follow convention here and follow up across all services if the need arises?

sclarkamlab: Yeah following the convention here is fine.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
sclarkamlab: Yeah that is fair, UPDATED makes sense in that case then.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/LicenseServiceImpl.java
wajb: I created the sentinel license to use in this scenario but then determined it is unreachable in production. Invalid org licenses degrade to a static OSS license and the same `LicenseFactory` is used for parsing here. It would be a defensive fix but also dead code so I'm inclined to leave it out. Wdyt?

sclarkamlab (approved):

PR #8337, opened 2026-07-30
chore: add postman collection for the cockpit mock control api

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7353

Adds a Postman collection for the cockpit mock's control API, so poking at the cloud command path locally doesn't mean hand rolling curl every time.

Four folders: status/queue inspection, one request per cockpit command type AM actually handles (ORGANIZATION, ENVIRONMENT, USER, MEMBERSHIP, INSTALLATION, GOODBYE) with valid example payloads, REPLY templates for the commands AM initiates, and a few management API calls to check what the commands produced. Command ids chain between requests automatically via collection variables, so you send a command and the next queue pop already knows what to match on.

The example payloads come from each handler's required field validation, and the descriptions carry the gotchas I kept relearning: the mock has no auth, /_control/send returns 409 until AM's websocket is up, entrypoints are cockpit owned so you can't delete them, and both planes warm their entrypoint caches async so verification may need a retry. Also added a short pointer to it in the mock's README.

Second commit is a small addition to the mock itself. Writing the collection turned up the fact that you cannot see AM's HELLO at all: it only flows AM to cockpit, the mock auto answers the handshake and deliberately keeps it off the queue, and the log line only prints what the mock sent back. That made AM-7326 (installation type plus the access point hostname template in HELLO) untestable outside unit tests. So the mock now retains the last HELLO and serves it on GET /_control/hello, 200 with the payload or 204 before AM connects, and the collection gets a matching request. I kept it off the FIFO queue on purpose, that's what waitForCockpitReply pops through and a HELLO landing there on every connect and reconnect would just be noise the specs have to discard.

## :memo: Test scenarios

No automated tests, it's a JSON collection, a README section and a read only endpoint on a dev only tool. Verified by hand against a local --cloud stack.

Collection: sent the INSTALLATION command through the mock and got the SUCCEEDED reply back on the queue with the matching commandId, and confirmed the management API token request shape works. Didn't run the ENVIRONMENT example since the entrypoints it creates can't be deleted afterwards.

New endpoint: in managed mode GET /_control/hello returns installationType "managed" and the accessPointsTemplate matching the configured template (ENVIRONMENT -> GATEWAY, host {environment}.{organization}.{account}.example.com, secured true), plus node, default org/env ids and additionalInformation. Flipped GRAVITEE_INSTALLATION_TYPE to standalone and restarted the management API, it comes back as "standalone" with the template unpopulated, so the managed gating holds in both directions. That's AM-7326 verified end to end, details are on that ticket.

No UI changes.

The collection descriptions are the documentation, they're written to be read in Postman rather than skimmed here. Two things worth knowing.

While testing AM-7225: there's no cockpit command that deletes an environment (AM registers no DELETE_ENVIRONMENT handler and EnvironmentService has no delete), so the way to test entrypoint removal is to resend the ENVIRONMENT command with an empty accessPoints list, which makes the handler wipe and recreate. AM-7354 is the follow up for rejecting that case in cloud mode.

On the HELLO payload: in standalone the accessPointsTemplate field is still sent, just empty, it's not absent. And there are two installationType fields facing opposite ways, the one on the command AM sends and the one on the reply cockpit sends, which AM ignores. Both are called out in the request description so nobody checks the wrong one.

@gravitee-io/am

### DIFF
--- a/gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json
+++ b/gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json
@@ -1,7 +1,7 @@
 {
   "info": {
     "name": "Gravitee AM Cockpit Mock",
-    "description": "Common requests against the cockpit mock (gravitee-am-cockpit-mock) control API, plus a few management API calls to verify the results in AM.\n\nPrerequisites: stack up in managed-cloud mode (`./local-stack.sh up --cloud` or `--full`), or the mock running standalone (`npm start` in gravitee-am-cockpit-mock). The mock itself has NO auth. `POST /_control/send` returns 409 while no AM instance is connected — check `Connection status` first.\n\nFlow: send a command, note the returned `id` (auto-saved to `{{commandId}}`), then pop the queue until AM's REPLY with that `commandId` appears.",
+    "description": "Common requests against the cockpit mock (gravitee-am-cockpit-mock) control API, plus a few management API calls to verify the results in AM.\n\nPrerequisites: stack up in managed-cloud mode (`./local-stack.sh up --cloud`, or `--with cloud`; note `--full` does NOT include the cloud overlay), or the mock running standalone (`npm start` in gravitee-am-cockpit-mock). The mock itself has NO auth. `POST /_control/send` returns 409 while no AM instance is connected — check `Connection status` first.\n\nFlow: send a command, note the returned `id` (auto-saved to `{{commandId}}`), then pop the queue until AM's REPLY with that `commandId` appears.",
     "schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
   },
   "variable": [
@@ -12,7 +12,10 @@
     { "key": "envId", "value": "postman-cloud-env" },
     { "key": "commandId", "value": "" },
     { "key": "amCommandId", "value": "" },
-    { "key": "token", "value": "" }
+    { "key": "token", "value": "" },
+    { "key": "orgId", "value": "DEFAULT" },
+    { "key": "licenseB64", "value": "" },
+    { "key": "licenseB64Alt", "value": "bm90LWEtcmVhbC1saWNlbnNl" }
   ],
   "item": [
     {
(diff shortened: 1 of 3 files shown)


gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json
```diff
@@ -0,0 +1,357 @@
+{
+  "info": {
+    "name": "Gravitee AM Cockpit Mock",
+    "description": "Common requests against the cockpit mock (gravitee-am-cockpit-mock) control API, plus a few management API calls to verify the results in AM.\n\nPrerequisites: stack up in managed-cloud mode (`./local-stack.sh up --cloud` or `--full`), or the mock running standalone (`npm start` in gravitee-am-cockpit-mock). The mock itself has NO auth. `POST /_control/send` returns 409 while no AM instance is connected — check `Connection status` first.\n\nFlow: send a command, note the returned `id` (auto-saved to `{{commandId}}`), then pop the queue until AM's REPLY with that `commandId` appears.",
```
wajb: minor: local-stack.sh `--full` mode does not include the cloud overlay.

gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json
```diff
@@ -0,0 +1,357 @@
...
+              "        console.log('sent command, saved commandId=' + json.id);",
+              "    }",
+              "}"
+            ]
+          }
+        }
+      ],
+      "item": [
+        {
+          "name": "ORGANIZATION - create/update org",
```
wajb, line 113: suggestion: could we wire up an optional `license` env variable? This would be useful for testing!

leleueri (approved):

gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json
sclarkamlab: Yeah I was just looking at that for testing https://gravitee.atlassian.net/browse/AM-7237

PR #8339, opened 2026-07-30
feat: scope webauthn to the environment entrypoint in managed cloud

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7231

In managed cloud the browser reaches the gateway on the environment's Cockpit access point, so that's the host a WebAuthn ceremony has to be scoped to. We were using `gateway.url` (or whatever the domain had configured), which is the wrong host, and the relying party id was falling back to `localhost` so the browser rejected the ceremony outright.

Origin and relying party id now both come from the environment entrypoint. Two things worth flagging for review:

It resolves per request rather than once when the domain deploys. A Cockpit ENVIRONMENT resync swaps an environment's entrypoints without redeploying its domains, so anything captured at startup goes stale and stays stale. Same reason the three handlers stopped caching the origin in a constructor field.

An environment with several entrypoints matches the host the request actually came in on, falling back to the primary. WebAuthn compares the origin by exact string against `window.location.origin`, so a fixed pick breaks anyone sitting on the other host. That matching already existed for the AM-7230 email links, so I moved it onto `EntryPointManager` rather than writing a second copy of the "never trust the Host header" bit.

A relying party id the domain set for itself is never overridden. The entrypoint only supplies one when `relyingPartyId` is unset. Credentials are bound to the id they were registered under, so replacing a deliberate one (commonly a parent domain, so a single credential covers several subdomains) would stop every authenticator a customer already has from offering it. See the release note below.

The origin is serialized the way a browser serializes `window.location.origin`, meaning lowercased host and no port when it's the scheme default. Vert.x compares it to `clientDataJSON.origin` with `.equals()`, so an entrypoint stored as `https://Auth.Acme.com` or `https://auth.acme.com:443` would match during entrypoint resolution and then fail verification against the very host it matched.

## :memo: Test scenarios

Unit: `DomainDataPlaneTest` for the resolution order and the browser-style origin serialization, `AbstractEntryPointManagerTest` for the request matching, `UriBuilderTest` for `toOrigin`, `WebAuthnHandlerRelyingPartyTest` for the rpId rewrite including the configured-id guard. `DomainReadServiceImplTest` passes unchanged, which is the net for AM-7230 after the extraction. Email links still carry the entrypoint url verbatim, port and casing included, so nothing there moves.

Jest: new `specs/cloud/webauthn-entrypoint-origin.jest.spec.ts` covers all five behaviours against a real managed cloud gateway, including a resync being picked up with no domain restart and a domain whose configured `relyingPartyId` has to survive. Whole `specs/cloud` suite is green.

Manual: walked it end to end on a `--cloud` stack, which is how I found the rpId bug. Playwright webauthn suite is 18 passed, the 2 failures in `webauthn-mfa-remember-device.spec.ts` are `mock-am-factor not deployed` and fail the same way on master unless you build with `-P ee-bundle,addons-bundle`.

One thing I could not cover automatically: the origin itself is only compared server side during verification so it needs a real authenticator. The rpId travels with it and is observable over HTTP, so that's what the assertions read.

No UI changes.

Not a breaking change for existing credentials, but it is worth a release note because an earlier revision of this PR was.

Domains that set `relyingPartyId` themselves keep it, so their credentials keep working. That guard is the point of the fourth commit and it's what makes this safe to ship.

Domains that never set one were advertising `localhost` in cloud. A browser refuses a ceremony whose rpId isn't a registrable suffix of the page origin, so no credential could ever have been registered there. There's nothing to orphan and those users go from broken to working.

The one case that does change: a domain that left `relyingPartyId` unset but set a custom `origin` whose host isn't the environment entrypoint. That combination was working, and the rpId now moves to the entrypoint host, so those credentials would need registering again. Anyone in that position can pin the old behaviour by setting `relyingPartyId` explicitly to the host they were using before upgrading.

@gravitee-io/am

### DIFF
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
@@ -375,6 +375,26 @@ public class UriBuilder {
         return "https".equalsIgnoreCase(uri.getScheme()) ? 443 : 80;
     }
 
+    public static String toOrigin(String url) {
+        if (url == null || url.isBlank()) {
+            return null;
+        }
+        final URI uri;
+        try {
+            uri = URI.create(url.trim());
+        } catch (IllegalArgumentException e) {
+            return null;
+        }
+        if (uri.getScheme() == null || uri.getHost() == null) {
+            return null;
+        }
+        StringBuilder origin = new StringBuilder(uri.getScheme().toLowerCase(Locale.ROOT)).append("://").append(uri.getHost());
+        if (uri.getPort() >= 0) {
+            origin.append(':').append(uri.getPort());
+        }
+        return origin.toString();
+    }
+
     public static String buildErrorRedirect(String baseRedirectUri, ErrorInfo error, boolean fragment, Map<String, String> extraParams) throws URISyntaxException {
         final URI redirectUri = UriBuilder.fromURIString(baseRedirectUri).build();
 
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
@@ -18,6 +18,7 @@ package io.gravitee.am.common.web;
 import io.gravitee.am.common.oauth2.Parameters;
 import io.gravitee.am.common.utils.ConstantKeys;
 
+import java.net.IDN;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -378,21 +379,17 @@ public class UriBuilder {
 
     /**
      * The url's origin, serialized the way a browser serializes {@code window.location.origin}: scheme and
-     * host lowercased, and the port present only when it is not the scheme's default. WebAuthn compares the
-     * ceremony origin against {@code clientDataJSON.origin} by exact string equality, so an origin that
-     * keeps {@code :443} or the stored host's casing fails verification against the very host it matched.
+     * host lowercased, a unicode host in its punycode form, and the port present only when it is not the
+     * scheme's default. WebAuthn compares the ceremony origin against {@code clientDataJSON.origin} by
+     * exact string equality, so an origin that keeps {@code :443} or the stored host's casing fails
+     * verification against the very host it matched.
      */
     public static String toOrigin(String url) {
         if (url == null || url.isBlank()) {
             return null;
         }
-        final URI uri;
-        try {
-            uri = URI.create(url.trim());
-        } catch (IllegalArgumentException e) {
-            return null;
-        }
-        if (uri.getScheme() == null || uri.getHost() == null) {
+        final URI uri = toAsciiUri(url.trim());
+        if (uri == null || uri.getScheme() == null || uri.getHost() == null) {
             return null;
         }
         StringBuilder origin = new StringBuilder(uri.getScheme().toLowerCase(Locale.ROOT))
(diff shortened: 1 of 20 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/DomainDataPlane.java
```diff
@@ -15,27 +15,57 @@
...
     private final DataPlaneDescription dataPlaneDescription;
+    @NonNull
+    private final EntryPointManager entryPointManager;
+    private final boolean managedCloud;
 
-    public String getWebAuthnOrigin() {
+    public String getWebAuthnOrigin(@Nullable String requestOrigin) {
+        Optional<String> entrypointOrigin = getWebAuthnEntrypointOrigin(requestOrigin);
+        if (entrypointOrigin.isPresent()) {
+            return entrypointOrigin.get();
```
leleueri, line 44: discussion: do we have to do it here at the beginning of the method or should we do it if the origin is not defined in the webAuthnSettings?

I tempt to give priority to the defined value in the WebAuthSetting.getOrigin, and resolver the origin from the request if nothing is specified (with a orElse using the dataPlaneDescriptio.gateway.url)

gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
```diff
@@ -369,12 +369,41 @@ private static boolean equalsIgnoreCase(String left, String right) {
...
+            uri = URI.create(url.trim());
+        } catch (IllegalArgumentException e) {
+            return null;
+        }
+        if (uri.getScheme() == null || uri.getHost() == null) {
+            return null;
+        }
+        StringBuilder origin = new StringBuilder(uri.getScheme().toLowerCase(Locale.ROOT))
+                .append("://")
+                .append(uri.getHost().toLowerCase(Locale.ROOT));
```
wajb, line 397: suggestion: edge case unicode handling (e.g. `bücher.example` -> `xn--bcher-kva.example`)
```suggestion
                .append(IDN.toASCII(uri.getHost()).toLowerCase(Locale.ROOT));
```

gravitee-am-service/src/main/java/io/gravitee/am/service/DomainDataPlane.java
sclarkamlab: Yeah I went back and forth on this one. The snag is that a set origin doesn't tell us anyone actually picked it: the console prefills the field from the entrypoint before the form even renders (`webauthn.component.ts:66`) and the mAPI won't save webauthn settings with a blank one (`DomainServiceImpl:931`), so nearly every domain is carrying one. Giving it priority in cloud would pin those back to a single host, which is the thing this PR is trying to fix.

Your point did catch a real bug though. `applyRelyingPartyId` was assuming exactly that about `relyingPartyId`, and the console prefills that one too. It now only leaves an rpId alone when it differs from the configured origin's host, so a deliberate parent domain survives and a prefilled copy follows the entrypoint. Fixed in `ceb49cb9a0`.

gravitee-am-common/src/main/java/io/gravitee/am/common/web/UriBuilder.java
sclarkamlab: Good catch, though I don't think that line ever sees a unicode host. `java.net.URI` won't parse one as server-based, so `getHost()` comes back null and `toOrigin` bails on the guard above it. `https://bücher.example` gives host=null, and `IDN.toASCII` ends up identity for anything `getHost()` will hand back. Handling it properly means converting before the parse rather than after.

Impact looks small either way, a unicode entrypoint url just fails to match and falls back to the primary entrypoint. I'll raise a follow-up for it unless you'd rather it went in here.

sclarkamlab: Done in `43706b8fe2`, converting before the parse rather than after:

```java
if (uri.getHost() != null || uri.getAuthority() == null) {
    return uri;
}
return new URI(uri.getScheme(), toAsciiAuthority(uri.getAuthority()), uri.getPath(), uri.getQuery(), uri.getFragment());
```

`https://bücher.example` now gives `https://xn--bcher-kva.example` instead of null, and it drops out of `sameOrigin` for free since that goes through `toOrigin` on both sides.

wajb: Ah yes, it would have to be converted before parsing. It's not a big deal but probably worth including IMO.

wajb (approved):

leleueri (approved):

PR #8345, opened 2026-07-31
feat: provision data plane definitions via the node internal api

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7259

The things worth knowing before reading the diff:

- New module `gravitee-am-management-api-internal` holding the endpoints, registered as a node component
- Create, get, list and delete for data plane definitions
- New `dataplanes` collection/table in the management repository, MongoDB and JDBC, with a Liquibase changelog
- Served on the node technical API (services.core, 18093, basic auth), not the MAPI, so nothing new is exposed on 8093
- The stored configuration can hold credentials and never leaves the service layer, the reads only ever return a credential free summary
- An environment is bound to at most one data plane, enforced in the service and by a unique constraint in both stores
- New `DATA_PLANE_CREATED` and `DATA_PLANE_DELETED` audit events against the owning organization
- Deliberately inert, nothing loads or connects to a provisioned definition yet

The stored row holds the same information as a gravitee.yml `dataPlanes[i]` entry plus the org and env it's bound to. POST validates the configuration at write time: an explicitly named database and host in both the uri and the discrete key forms, a deployed plugin for the type, and `default` refused as an id since it belongs to the data plane declared in the gravitee.yml. DELETE frees the environment binding, refused with a 409 while any domain still references it (`existsByDataPlaneId` guard on the domains store).

Loading and connecting is AM-7260/AM-7261, and there's no update endpoint yet.

## :memo: Test scenarios

Unit tests for the service and the four endpoints, repository tests against MongoDB and JDBC, and an e2e jest spec covering provisioning, redaction on every read shape, both conflicts, validation, auth and deletion. 21 cases, green on both stores.

Worth a look in review: the uri form has to name a host as well as a database, otherwise the definition silently lands on the driver default. That check reads the uri authority rather than the parsed host, because a replica set uri parses to a null host while still naming its servers, so keying off the host would reject a valid one.

Each spec case provisions what it needs and releases the environment binding around itself, so there is no ordering between them. Verified under `--randomize` on three seeds.

The 409 delete guard is unit tested only. No domain can reference a provisioned data plane until AM-7260 loads them into the registry.

Testing notes with the curl commands are in the ticket description.

No UI changes.

The internal API rides the node technical API port (18093, basic auth), so nothing new is exposed on 8093.

@gravitee-io/am


gravitee-am-management-api/gravitee-am-management-api-internal/src/main/java/io/gravitee/am/management/handlers/internalapi/InternalApiService.java
```diff
@@ -0,0 +1,59 @@
...
+import io.gravitee.am.management.handlers.internalapi.endpoints.DeleteDataPlaneEndpoint;
+import io.gravitee.am.management.handlers.internalapi.endpoints.GetDataPlaneEndpoint;
+import io.gravitee.am.management.handlers.internalapi.endpoints.ListDataPlanesEndpoint;
+import io.gravitee.common.service.AbstractService;
+import io.gravitee.node.management.http.endpoint.ManagementEndpointManager;
+import lombok.CustomLog;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * Registers the Management API endpoints served by the node technical API (services.core, :18093).
```
wajb: thought: `18093` is conventional in gravitee.yml but isn't the default for the port in core services. Will we always override this?

gravitee-am-model/src/main/java/io/gravitee/am/model/DataPlaneDefinition.java
```diff
@@ -0,0 +1,59 @@
...
+    private String name;
+
+    /**
+     * Matches the {@code dataplane-am-<type>} plugin id suffix (mongodb, jdbc, ...).
+     */
+    private String type;
+
+    private String gatewayUrl;
+
+    private String organizationId;
```
wajb, line 41: question: do we denormalise the reference for org id for convenience of auditing? Uniqueness appears bound to environment and the grandparent org is informational. Worth confirming intent here, thanks.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/DataPlaneConfigHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Extracts the handful of fields that are safe to expose on a read.
+     *
+     * This is an allowlist on purpose. The stored blob can carry credentials in a lot of places —
+     * {@code username}/{@code password}, {@code keystore.password}, {@code keystorePassword},
+     * {@code keyPassword}, {@code truststore.password}, and the userinfo of a connection {@code uri}
+     * — and the key surface keeps growing, so nothing is emitted unless it is named here.
+     */
+    DataPlaneConnectionSummary summarise(JsonNode configuration);
```
wajb: minor: should use US `summarize`.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/DataPlaneConfigHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Knows the shape of one data plane type's connection settings: what makes them valid on write, and
+ * what is safe to hand back on read.
+ *
+ * A new data plane type ships its own handler alongside its plugin.
+ *
+ * @author GraviteeSource Team
+ */
+public interface DataPlaneConfigHandler {
```
wajb, line 26: thought: it makes sense to use imperative handlers for sanitising summaries and rejecting defaults, but would it be possible to leverage schema form validation for a first pass, both for consistency and maintainability?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static io.gravitee.am.plugins.dataplane.core.DataPlanePluginManager.PLUGIN_ID_PREFIX;
+import static org.springframework.util.StringUtils.hasText;
+
+/**
+ * Persists a data plane definition. Deliberately inert: no registry entry, no event (AM-7260/7261).
```
wajb: nitpick: are the references TODOs? Otherwise we can remove them IMO.

gravitee-am-model/src/main/java/io/gravitee/am/model/DataPlaneDefinition.java
sclarkamlab: Yeah, denormalised. It's the audit reference, `DataPlaneDefinitionAuditBuilder` writes against `Reference(ORGANIZATION, organizationId)`, and it's the parent in the validation, we check the org exists then that the env belongs to it.

One thing on the env though: domains point at a data plane directly via `Domain.dataPlaneId` and nothing checks the domain's env matches the data plane's, so the unique constraint only really guards provisioning at the moment. I'd assumed that was staging for AM-7261. Can tighten it at domain create or leave it, no strong feeling.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/DataPlaneConfigHandler.java
sclarkamlab: Done

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/DataPlaneConfigHandler.java
sclarkamlab: I'll look into doing this now

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
sclarkamlab: Done

gravitee-am-model/src/main/java/io/gravitee/am/model/DataPlaneDefinition.java
wajb: Thanks. I think that integrity check is worthwhile at some point - not sure it fits well into AM-7261 though.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/DataPlaneConfigHandler.java
sclarkamlab: Yeah, done in `b7bcdc3836`. Both data plane plugins now ship a `schema-form.json` and `DataPlanePluginHandler` registers it. That bit was half built already, it was autowiring the validators registry and never using it.

Schema runs first and owns the shape, types and the driver enum. The handlers keep what a schema can't say, mainly that a uri has to name both a host and a database, and that a mongodb block with no dbname would quietly land on `gravitee-am`. Left `additionalProperties` open on purpose, the providers read a few more keys than are worth describing and some of them are deprecated.

gravitee-am-model/src/main/java/io/gravitee/am/model/DataPlaneDefinition.java
sclarkamlab: You're right, wrong ticket on my part. It's AM-7262, "Update domain creation process", which already says a domain's DP id has to match the one linked to its environment, and that in Cloud Mode a domain can only be created without one if the env has a DP loaded. So the integrity check is scoped, just not in 7261.

Epic is AM-7167 if you want the framing, "extend the data model so environments are associated with a data plane (not only domains)". This row is that association, held on the DP side rather than a column on Environment so that deleting a DP frees the binding in one write and the 1:1 stays a real constraint.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/JdbcDataPlaneConfigHandler.java
```diff
@@ -0,0 +1,99 @@
...
+            requireHostAndDatabaseInUri(block.get("uri").asText());
+            return;
+        }
+
+        List<String> missing = REQUIRED_WITHOUT_URI.stream()
+                .filter(key -> !DataPlaneConfigs.hasText(block, key))
+                .toList();
+        if (!missing.isEmpty()) {
+            throw new InvalidParameterException("configuration.jdbc requires either 'uri' or all of 'driver', 'host' and 'database', missing: " + String.join(", ", missing));
+        }
```
wajb: suggestion: can this bit be pushed to the form schema?
```json
"anyOf":[
  {
    "required": [
      "uri"
    ]
  },
  {
    "required": [
      "driver",
      "host",
      "database"
    ]
  },
  ...
]
```
We do almost exactly this for idp configs already https://github.com/gravitee-io/gravitee-access-management/blob/master/gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/resources/schemas/schema-form.json#L197-L210.

wajb: Hardcoding the list from `REQUIRED_WITHOUT_URI` in the message is a latent consistency issue either way IMO.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/JdbcDataPlaneConfigHandler.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    private void requireHostAndDatabaseInUri(String uri) {
```
wajb, line 57: nitpick: missing newline.

wajb (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/config/JdbcDataPlaneConfigHandler.java
sclarkamlab: Done in `0a184d35cd`, `anyOf` exactly as you wrote it and the hardcoded `REQUIRED_WITHOUT_URI` message is gone with it. Handler now only keeps the bit a schema can't say, that a uri has to name a host and a database.

One thing I added on top. `anyOf` plus `required` only checks the key is present, but the old handler used `hasText`, so `{"uri": ""}` would have satisfied `required: [uri]` and then fallen straight past the uri check and been accepted, which is the silent-default case we're trying to stop. Put `minLength: 1` on uri, host and database so `required` means what `hasText` meant. driver didn't need it, the enum already rejects empty.

Behaviour change worth flagging: `{"uri": "", "driver": ..., "host": ..., "database": ...}` used to be accepted (blank uri ignored, discrete form used) and now gets rejected on the uri. Rejecting an explicitly empty uri felt better than silently ignoring it, but say if you'd rather I loosen it.

Checked it end to end on a local stack as well as the unit tests, since the whole point is that enforcement now lives outside Java. Rejections come back from the schema now:

```
driver+host, no database  ->  400  configuration.jdbc is not valid: #: required key [database] not found
{"uri": ""}               ->  400  configuration.jdbc is not valid: #/uri: expected minLength: 1, actual: 0
```

so the plugin registry really is delivering the schema. Both valid forms still 201 and the uri-names-a-database rule still fires from the handler. Newline nitpick fixed too.

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewDataPlaneDefinition.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private String name;
+
+    private String type;
+
+    private String gatewayUrl;
+
+    private String organizationId;
+
+    private String environmentId;
```
leleueri, line 45: suggestion: I would provide also the hrid for org and env for commodity. If environmentId is present, it take precedence, otherwise we are trying the envHrid and if missing we are using DEFAULT env (same for org)

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-dataplanes-table.yml
```diff
@@ -0,0 +1,32 @@
...
+
+        - addPrimaryKey:
+            constraintName: pk_dataplanes
+            columnNames: id
+            tableName: dataplanes
+
+        - addUniqueConstraint:
+            constraintName: uq_dataplanes_environment_id
+            columnNames: environment_id
+            tableName: dataplanes
```
leleueri: Issue: this constraint will block some customer. Some of them has to create multiple DP per env

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/MongoDataPlaneDefinitionRepository.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String FIELD_ENVIRONMENT_ID = "environmentId";
+
+    private MongoCollection<DataPlaneDefinitionMongo> collection;
+
+    @PostConstruct
+    public void init() {
+        collection = mongoOperations.getCollection(COLLECTION_NAME, DataPlaneDefinitionMongo.class);
+        super.init(collection);
+        // an environment is bound to at most one data plane
+        super.createIndex(collection, Map.of(new Document(FIELD_ENVIRONMENT_ID, 1), new IndexOptions().name("e1").unique(true)));
```
leleueri: issue: this constraint will block some customer. Some of them has to create multiple DP per env

leleueri (changes requested):

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewDataPlaneDefinition.java
sclarkamlab: Done in `ba25422a2d`. `organizationHrid` and `environmentHrid` on the payload, resolving exactly as you described, id wins, then hrid, then DEFAULT. The env hrid resolves inside whichever org came out of the org step, so `{"organizationId": "org-1", "environmentHrid": "prod"}` works too.

There was no env lookup by hrid at all so I added `findByHrid` to `EnvironmentRepository`, mongo plus a join on `environment_hrids` for jdbc. Org reuses the existing `findByHrids`.

One thing worth knowing, nothing enforces hrid uniqueness in either table, so it takes the first match. Say if you'd rather an ambiguous hrid was an error.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-dataplanes-table.yml
sclarkamlab: Done in `ba25422a2d`. The unique constraint is gone, it's a plain `idx_dataplanes_environment_id` index now, and the mongo index lost its `unique(true)` too.

Went further than just the DB. `checkEnvironmentIsFree` and `EnvironmentAlreadyBoundToDataPlaneException` are deleted, so there is no 409 on environment left anywhere, and `findByEnvironmentId` returns a `Flowable` since more than one row per env is legal now. Verified on a psql stack, eight data planes into DEFAULT all come back 201 and the table only has `pk_dataplanes` plus the non unique index.

Two things I can't call on my own though. The AC on AM-7259 still says one DP per env backed by a unique constraint in both stores, so that needs editing. And AM-7262 says a domain's DP id has to match the one linked to its environment, which doesn't really mean anything now an env can have several. Do you want a rule for picking one, or a pointer on the environment side?

Heads up if you pull this branch, I edited the existing changeset in place so you'll want a fresh DB or liquibase will moan about the checksum.

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/MongoDataPlaneDefinitionRepository.java
sclarkamlab: Same commit, `unique(true)` is off the index. Kept the index itself since we still look up by environment. More detail in my reply on the liquibase one.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-dataplanes-table.yml
sclarkamlab: Thanks, that settles it. So it keys off the count for the env either way, 0 is an error, exactly 1 means the domain can leave `dataPlaneId` out and gets that one, more than 1 means it has to say which.

Ignore what I said above about `findByEnvironmentId` possibly going then, that rule is exactly what it is for and the `Flowable` is what makes the count answerable. It stays.

One thing you'll hit in 7262, the check at `DomainServiceImpl:380` has to be replaced rather than extended. It matches `newDomain.getDataPlaneId()` against the whole registry with no environment scoping, so today it would happily take a DP belonging to another env. And `NewDomain.dataPlaneId` has no default, so a null ends up in `id.equals(null)`, false for every id, and the domain just gets rejected. There's no "leave it out and we pick" path there yet.

leleueri (approved):

PR #8354, opened 2026-08-03
test: automate AM cloud license enforcement coverage (AM-7235/7236/7238/7240/7241)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7235
https://gravitee.atlassian.net/browse/AM-7236
https://gravitee.atlassian.net/browse/AM-7238
https://gravitee.atlassian.net/browse/AM-7240
https://gravitee.atlassian.net/browse/AM-7241

`local-stack.sh` never built the `saml2-idp` EE plugin. The `ee-bundle`/`addons-bundle` Maven profiles were already correctly defined in the existing `pom.xml` files; the script's own`mvn clean install` invocation just never activated `addons-bundle`, so the plugin was silently missing from every local build. This made `/saml2/idp/metadata` always return 404 locally regardless of license state, and was blocking the AM-7238 test below from passing at all.
  
  **Test automation**
  - Jest coverage for cockpit ORGANIZATION license commands (AM-7235)
  - Jest coverage for the organization/platform license endpoints (AM-7236)
  - Jest coverage for gateway-side EE plugin enforcement via SAML (AM-7238)
  - Jest coverage for management-API plugin license gating (AM-7241)
  - Playwright spec for the console UI lock-icon / upsell-dialog behaviour (AM-7240 / AM-7241)

## :memo: Test scenarios 
NA

NA

NA

NA


sclarkamlab (commented):
Nice coverage on the cockpit error paths, the blank / non-base64 / missing id / unknown org cases are the good stuff.

Three things I think are blocking though, comments inline. Rest of what I spotted is minor and can wait.

gravitee-am-test/playwright/tests/cloud/plugin-license-ui.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * AM-7240 / AM-7241 Group 6 — Console UI license gating for EE plugins.
+ * Verifies the lock icon + upsell dialog genuinely pre-empt the backend 403 when
+ * unlicensed, and that a licensed org can create the EE plugin normally end-to-end.
+ *
+ * Requires: --cloud stack (local-stack.sh up --cloud).
```
sclarkamlab, line 22: I don't think this can pass in CI as wired up.

`pw:ci` is just `playwright test` with `testDir: './playwright/tests'` and no `testIgnore`, so `tests/cloud/` gets collected. But the playwright jobs run `stack:ci:pw:setup:mongo|psql`, and those compose files don't include `docker-compose.cloud.yml`. No cockpit-mock service, so nothing on :8085 for `sendOrgCommand` to hit.

Second problem even if you fix that: `docker-compose-ci.yml` sets a full `GRAVITEE_LICENSE_KEY`, where the cloud overlay deliberately sets it empty so the platform is unlicensed. Under the CI stack the console would see platform EE entitlements and the lock icon would never render.

Either we need a cloud playwright job, or `tests/cloud/` should be excluded from `pw:ci` and run only against a `--cloud` stack. As it stands the AM-7240/7241 UI coverage doesn't actually execute.

gravitee-am-test/specs/cloud/gateway-org-license.jest.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+    expect(samlEntry).not.toBeNull();
+    expect(samlEntry.success).toBe(true);
+    expect(samlEntry.message).toEqual(expect.stringContaining('requires the feature'));
+  });
+
+  it(jira`should show OSS plugins with a null license message in the domain readiness state ${'AM-7238'}`, async () => {
+    const state = await getDomainState(domainId);
+    const plugins = Object.values(state.creationState ?? {}) as PluginEntry[];
+    const ossEntry = plugins.find((p) => p.message === null && p.success === true);
+    expect(ossEntry).not.toBeNull();
```
sclarkamlab: `find()` returns undefined, not null, so `expect(undefined).not.toBeNull()` passes. That's the only assertion in this test, so it can never fail.

`expect(ossEntry).toBeDefined()` instead.

Same pattern on 149, less bad there because `.success` on the next line throws anyway, but it turns a clear assertion failure into a confusing TypeError. Worth fixing both. The existing `organization-license-enforcement.jest.spec.ts` sidesteps this by indexing `creationState[GATED_PLUGIN]` directly if you'd rather go that way.

gravitee-am-test/specs/cloud/gateway-org-license.jest.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+    const res = await fetch(`${process.env.AM_GATEWAY_URL}/${domainHrid}/saml2/idp/metadata`);
+    expect(res.status).toBe(200);
+    const body = await res.text();
+    expect(body).toContain('<?xml');
+  });
+});
+
+describe('SAML metadata — expired org license', () => {
+  it(jira`should return 404 when an expired org license is sent (treated same as no license) ${'AM-7238'}`, async () => {
+    if (!fixture.hasExpiredLicense) {
```
sclarkamlab: Early return means jest reports this as passed, not skipped. `AM_ORG_LICENSE_EXPIRED_PATH` isn't set anywhere in the repo or in either cloud job, so this is permanently green having never run a single assertion.

`it.skip` (or a conditional describe) would at least make the gap visible in the report.

sclarkamlab (approved):

PR #8355, opened 2026-08-04
fix: permissions key cache

lgw-gravitee:
fixes AM-7410

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
@@ -15,11 +15,14 @@
  */
 package io.gravitee.am.management.service;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import io.gravitee.am.identityprovider.api.User;
 import io.gravitee.am.management.service.permissions.PermissionAcls;
 import io.gravitee.am.model.Acl;
 import io.gravitee.am.model.Group;
 import io.gravitee.am.model.Membership;
+import io.gravitee.am.model.Reference;
 import io.gravitee.am.model.ReferenceType;
 import io.gravitee.am.model.Role;
 import io.gravitee.am.model.permissions.Permission;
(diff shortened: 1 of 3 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
```diff
@@ -120,78 +134,120 @@ protected Single<Boolean> haveConsistentReferenceIds(PermissionAcls permissionAc
...
+    private Maybe<Reference> parentOf(Reference reference) {
+
+        return Maybe.defer(() -> {
+            Reference cached = parentCache.getIfPresent(reference);
+
+            if (cached != null) {
+                return Maybe.just(cached);
+            }
+
+            return resolveParent(reference).doOnSuccess(parent -> parentCache.put(reference, parent));
```
sclarkamlab, line 220: AM-7410 AC #1 asks to cache found and notFound results, but `doOnSuccess` skips the empty case so a missing entity gets re-resolved on every check. Was deferring the notFound caching intentional? Fine either way, just worth closing out on the ticket.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
```diff
@@ -120,78 +134,120 @@ protected Single<Boolean> haveConsistentReferenceIds(PermissionAcls permissionAc
...
             }
 
-            if (environmentId != null) {
-                obs.add(isEnvironmentIdConsistent(environmentId, organizationId));
+            if (applicationId != null) {
+                checks.add(isApplicationConsistent(applicationId, domainId, environmentId, organizationId));
             }
 
-            return Single.merge(obs)
+            return Single.concat(checks)
```
sclarkamlab, line 151: Right call letting technical errors propagate instead of swallowing them into a denial, but it does change observable behaviour: a DB failure here used to come back as a 403 and is now a 500 via `ManagementExceptionMapper`. Worth a line in the PR description / release notes.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
lgw-gravitee: I created a PR and during the impl I realized that storing notFound is useless. It just can make cache bloated. The querying with non-existing resource identifiers cannot happen and will eventually fail

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/PermissionService.java
lgw-gravitee: discussion: which is right I suppose. If AM cannot connect othe database its not 4xx but rather 5xx.

sclarkamlab: Yeah true, ok let's roll with it then!

sclarkamlab (approved):

PR #8357, opened 2026-08-04
feat: load and register provisioned data planes in the management api

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7260

Loads the data plane definitions AM-7259 stores, so domains can be served from a provisioned data plane.

At startup the loader reads the definitions out of the management repository and hands them to the registry alongside the gravitee.yml ones. When one is provisioned over the internal API it registers straight away, so you don't have to restart the node to use it.

The provider never sees its configuration. The stored JSON gets flattened onto dotted keys under dataPlanes.provisioned.<id> and published as a property source, and the plugin resolves its own settings out of the environment exactly like it already does for the yaml ones. Nothing downstream needs to know where a definition came from.

Registering is best effort. The definition is already saved by the time it runs, so a failure gets logged and the plane waits for the next restart rather than failing a provisioning call that actually succeeded.

Also in here:

- DataPlaneRegistryImpl.register throws when no provider could be built, instead of storing a description that a domain can bind to but nothing can serve.
- An id whose registration failed stays free, so a corrected definition can reuse it without a restart.
- Provisioning an id the gravitee.yml already declares is rejected, not just default.

The register call sits in the endpoint rather than the service. ServiceConfiguration component scans io.gravitee.am.service and the gateway imports it, so DataPlaneDefinitionServiceImpl is a bean in the gateway too, and injecting the management API only loader into it would break gateway startup.

## :memo: Test scenarios

Unit and e2e coverage in the branch, all green. Manual test steps with curl commands are in the Jira ticket.

No UI changes.

The runtime registration is idempotent on purpose. AM-7261 adds the sync event that tells the other management API instances about a new data plane, and when that lands it can just call register(id) without worrying about double registering on the node that created it.

@gravitee-io/am

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
@@ -23,17 +23,20 @@ import io.gravitee.am.dataplane.api.DataPlaneDescription;
 import io.gravitee.am.model.DataPlaneDefinition;
 import io.gravitee.am.plugins.dataplane.core.DataPlaneLoader;
 import io.gravitee.am.repository.management.api.DataPlaneDefinitionRepository;
+import io.reactivex.rxjava3.core.Completable;
 import lombok.CustomLog;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.MapPropertySource;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Consumer;
 
 /**
  * Loads the data planes declared in the gravitee.yml, then the ones provisioned through the internal
- * API (AM-7259) and stored in the management repository.
+ * API (AM-7259) and stored in the management repository. {@link #register(String)} adds a definition
+ * provisioned since startup, so a new data plane can be served without restarting the node.
  *
  * A provider never receives its configuration: {@link DataPlaneDescription#propertiesBase()} hands it a
  * property prefix and the plugin resolves the settings out of the Spring environment itself. A stored
@@ -65,6 +68,8 @@ public class ProvisionedDataPlaneLoader implements DataPlaneLoader {
             .disable(StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION)
             .build();
     private final Map<String, Object> properties = new ConcurrentHashMap<>();
+    private final Set<String> registered = ConcurrentHashMap.newKeySet();
+    private volatile Consumer<DataPlaneDescription> storage;
 
     public ProvisionedDataPlaneLoader(DataPlaneLoader configurationLoader,
                                       DataPlaneDefinitionRepository dataPlaneDefinitionRepository,
@@ -77,9 +82,32 @@ public class ProvisionedDataPlaneLoader implements DataPlaneLoader {
 
     @Override
     public void load(Consumer<DataPlaneDescription> storage) {
+        // kept so a definition provisioned later reaches the same registry this one is filling
+        this.storage = storage;
         configurationLoader.load(storage);
         dataPlaneDefinitionRepository.findAll()
                 .blockingForEach(definition -> load(definition, storage));
+    }
+
+    /**
+     * Registers a definition provisioned since startup so domains bound to it can be served straight away,
+     * rather than from the next restart.
+     *
+     * Best effort on purpose: the definition is persisted before this runs, so failing here only costs the
+     * node the live registration — the startup load will pick it up next time — and it must not turn a
+     * successful provisioning call into an error.
+     */
+    public Completable register(String dataPlaneId) {
+        var storage = this.storage;
+        // nothing registered yet means no registry to add to; its own load() will read this definition
+        if (storage == null || !registered.add(dataPlaneId)) {
+            return Completable.complete();
+        }
+        return dataPlaneDefinitionRepository.findById(dataPlaneId)
+                .doOnSuccess(definition -> load(definition, storage))
+                .ignoreElement()
+                .doOnError(e -> log.error("Data plane [{}] could not be read back after being provisioned and will be unavailable until restart", dataPlaneId, e))
+                .onErrorComplete();
     }
 
     /**
@@ -91,6 +119,7 @@ public class ProvisionedDataPlaneLoader implements DataPlaneLoader {
     private void load(DataPlaneDefinition definition, Consumer<DataPlaneDescription> storage) {
         try {
             storage.accept(publish(definition));
+            registered.add(definition.getId());
             log.info("Data plane [{}] of type [{}] loaded from the management repository", definition.getId(), definition.getType());
         } catch (Exception e) {
             log.error("Data plane [{}] of type [{}] could not be loaded and will be unavailable; domains bound to it cannot be served",
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
@@ -51,6 +51,7 @@ public class ProvisionedDataPlaneLoader implements DataPlaneLoader {
 
     private final DataPlaneLoader configurationLoader;
     private final DataPlaneDefinitionRepository dataPlaneDefinitionRepository;
+    private final ConfigurableEnvironment environment;
 
     private final ObjectMapper objectMapper = JsonMapper.builder()
             .disable(StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION)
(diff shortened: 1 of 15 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
```diff
@@ -0,0 +1,130 @@
...
+    private final Map<String, Object> properties = new ConcurrentHashMap<>();
+    private final Set<String> registered = ConcurrentHashMap.newKeySet();
+    private final AtomicReference<Consumer<DataPlaneDescription>> storage = new AtomicReference<>();
+
+    public ProvisionedDataPlaneLoader(DataPlaneLoader configurationLoader,
+                                      DataPlaneDefinitionRepository dataPlaneDefinitionRepository,
+                                      ConfigurableEnvironment environment) {
+        this.configurationLoader = configurationLoader;
+        this.dataPlaneDefinitionRepository = dataPlaneDefinitionRepository;
+        environment.getPropertySources().addLast(new MapPropertySource(PROPERTY_SOURCE_NAME, properties));
```
wajb: suggestion: move this registration to `load` to avoid the smell of side effects in the ctor.

Though, if it needs to be wired at the bean level, then it could be moved to the `@Bean` factory in `StandaloneConfiguration`.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public void load(Consumer<DataPlaneDescription> storage) {
+        this.storage.set(storage);
+        configurationLoader.load(storage);
+        dataPlaneDefinitionRepository.findAll()
+                .blockingForEach(definition -> load(definition, storage));
+    }
+
+    public Completable register(String dataPlaneId) {
+        var storage = this.storage.get();
```
wajb: thought: I agree with sonarcloud that this could be renamed. However, should registering when it's null really be a silent no-op? If `load` has not been called, should we throw an InvalidOperationException?

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
```diff
@@ -0,0 +1,130 @@
...
+        return dataPlaneDefinitionRepository.findById(dataPlaneId)
+                .doOnSuccess(definition -> load(definition, storage))
+                // fires only when findById completes empty
+                .doOnComplete(() -> log.warn("Data plane [{}] was not found when read back after being provisioned and will be unavailable until restart", dataPlaneId))
+                .ignoreElement()
+                .doOnError(e -> log.error("Data plane [{}] could not be read back after being provisioned and will be unavailable until restart", dataPlaneId, e))
+                .onErrorComplete();
+    }
+
+    private void load(DataPlaneDefinition definition, Consumer<DataPlaneDescription> storage) {
```
wajb: nitpick: method name collides with another and they're not really related overloads IMO.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
```diff
@@ -0,0 +1,130 @@
...
+        flatten(propertiesBase, objectMapper.readTree(definition.getConfiguration()), properties);
+        return new DataPlaneDescription(
+                definition.getId(),
+                definition.getName(),
+                definition.getType(),
+                propertiesBase,
+                definition.getGatewayUrl());
+    }
+
+    private static void flatten(String prefix, JsonNode node, Map<String, Object> properties) {
```
wajb, line 123: suggestion: can we leverage the `YamlPropertiesFactoryBean` here?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
```diff
@@ -242,8 +246,8 @@ private DataPlaneDefinition validate(NewDataPlaneDefinition payload) {
         if (!hasText(payload.getId())) {
             throw new InvalidParameterException("'id' is required");
         }
-        if (DataPlaneDescription.DEFAULT_DATA_PLANE_ID.equals(payload.getId())) {
-            throw new InvalidParameterException("'" + DataPlaneDescription.DEFAULT_DATA_PLANE_ID + "' is reserved for the data plane declared in the gravitee.yml");
+        if (DataPlaneDescription.DEFAULT_DATA_PLANE_ID.equals(payload.getId()) || isDeclaredInConfiguration(payload.getId())) {
```
wajb: question: is it safe to remove the "default" id clause if we're checking gravitee.yml? (If not, moving it to the boolean method would at least encapsulate it better IMO.)

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
```diff
@@ -268,6 +272,17 @@ private DataPlaneDefinition validate(NewDataPlaneDefinition payload) {
         return definition;
     }
 
+    // an id colliding with a gravitee.yml plane would persist but never register: the registry keeps
+    // serving the yml plane, and domains binding to the id would silently land on the wrong store
+    private boolean isDeclaredInConfiguration(String id) {
```
wajb: thought: would this fit better in `MultiDataPlaneLoader`?

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
sclarkamlab: Done in `1c20cf33`, it happens in `load` now. `addLast` replaces by name so a second load is still fine, and nothing resolves those keys before the registry starts. Added a test that constructing the loader leaves the environment alone.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
sclarkamlab: Yeah fair, it is `activate` now.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
sclarkamlab: Renamed the field to `storageRef` so the local can stay `storage`.

On throwing, I don't think we should. `load` sets the consumer before it reads the repository, so a definition persisted while it is still null gets picked up by that read anyway, nothing is lost. Throwing would fail a request whose row is already committed and the retry would then hit the already exists error. Added a comment because you can only see that by going and reading `load`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
sclarkamlab: Safe to drop as far as I can tell, `MultiDataPlaneLoader` throws if the yml has no default so we can't boot without one. I kept it but moved it into the boolean like you suggested. `default` is also the id the registry falls back to for a domain with no `dataPlaneId`, so I'd rather reserve it on its own terms than by accident.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/DataPlaneDefinitionServiceImpl.java
sclarkamlab: Done in `3f1e4975`. It's `MultiDataPlaneLoader.isDeclared` now and the service just asks it, so the `dataPlanes[i].id` shape only lives in one place. Small test for it in that module too.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
sclarkamlab: Tried it, but leaving it as is.

You're right about the convention, gravitee-node flattens the gravitee.yml with exactly that class (`PropertiesConfiguration.graviteeProperties`) and I checked the keys come out identical to ours for objects and lists. Snag is it doesn't fail on input that isn't an object, `not json` gives you `{document=not json}` rather than an exception, so a broken definition would register with junk properties instead of being skipped. I'd have to keep the jackson parse for that anyway and then parse a second time with snakeyaml. Shout if you think the parity is worth the double parse.

wajb (approved):

PR #8363, opened 2026-08-04
feat: AM-7393 skip org and env provisioning in managed cloud mode

wajb:
[AM-7393](https://gravitee.atlassian.net/browse/AM-7393)

This spike demonstrates a clean boot of AM without provisioning of any organizations, environments, IdPs, reporters or users.

- AM can skip provisioning of `DEFAULT` organization and environment resources when managed cloud mode is detected without deleterious effect.
   - Handling of edge cases where `DEFAULT` org authentication is assumed and treating as benign with logging.
   - Skipping of registration of transient providers (configurable `gravitee.yml` IdPs and users).
- Initially there is nothing to log into and no users available to do it. Provisioning of these resources is managed by a sequence of cockpit commands - `ORGANIZATION`, `ENVIRONMENT`, `USER`, `MEMBERSHIP`.
- SSO token handling in local stack, mimicking the minting conducted by Cockpit.
   - `POST /_control/sso-token` command in `gravitee-am-cockpit-mock`.
   - In-memory IDP creation using `security.providers[]` (`gravitee.yml` approach for on-prem installations) is replaced with `cloud.connector.ws.ssl` configuration.
   - Jest and playwright tests follow the production sequence and are org-agnostic.
   - Postman collection includes Cockpit SSO requests for manual testing.
- UI gating to hide "Discover cockpit" menu item.

Recommended follow-ups:
- **Console landing page**. What should an unauthenticated console user get in managed cloud when there is no org/env? Either reject or redirect to Cockpit.
- **Deletion command handlers**. Unlike APIM, AM does not yet implement `DELETE_ORGANIZATION`, `DELETE_ENVIRONMENT` or `DELETE_MEMBERSHIP` (nor `DISABLE_ORGANIZATION` / `DISABLE_ENVIRONMENT`). When implemented, jest test fixtures can leverage these immediately for cleanup operations, then they can be updated to use randomised resource IDs (the recommended pattern).
- **Close/narrow user-creation back door**. Config-declared, in-memory user bootstrapping should be restricted to `DEFAULT` org for backward compatibility and not allow users to be created for whichever org is being logged into (applicable outside of managed cloud mode).

## :memo: Test scenarios 
Run `local-stack.sh up --cloud --ui` to inspect AM in an unprovisioned managed cloud state.
Use `gravitee-am-cockpit-mock/postman/gravitee-am-cockpit-mock.postman_collection.json` postman collection to manually provision resources.

n/a

n/a

@gravitee-io/am

[AM-7393]: https://gravitee.atlassian.net/browse/AM-7393?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-ui/src/app/components/submenu/submenu.component.ts
```diff
@@ -43,17 +45,20 @@ export class SubmenuComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
-    this.subscription = this.navigationService.level2MenuItemsObs$.subscribe(
-      (items) =>
-        (this.subMenuItems = items.reduce((r, a) => {
-          r[a.section] = r[a.section] || [];
-          r[a.section].push(a);
-          return r;
-        }, {})),
-    );
+    this.subscription = combineLatest([this.navigationService.level2MenuItemsObs$, this.cloudModeService.isCloudModeEnabled()])
```
sclarkamlab, line 52: This makes the whole submenu depend on the installation call succeeding. If it errors the combineLatest dies, the subscription is gone and the menu stays empty until a full page reload. It fires on self-hosted too, where nothing gets filtered anyway.

Think we just want:

```ts
this.cloudModeService.isCloudModeEnabled().pipe(catchError(() => of(false)))
```

wajb: Thanks. I applied the `catchError` in the shared service, so this covers the same fix for submenu, guard and entrypoints.

sclarkamlab (approved):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DefaultOrganizationUpgrader.java
```diff
@@ -99,11 +102,17 @@ public DefaultOrganizationUpgrader(OrganizationService organizationService,
         this.identityProviderManager = identityProviderManager;
         this.auditService = auditService;
         this.useDefaultAdmin = environment.getProperty("security.defaultAdmin", boolean.class, true);
+        this.managedCloudEnabled = CloudProperties.isManagedCloudEnabled(environment);
     }
 
     @Override
     public boolean upgrade() {
 
+        if (managedCloudEnabled) {
```
leleueri, line 111: question: do we have to disable it as soon as we are in cloud mode or should we add an additional settings to explicitly ask to not create the default org.

`        if (managedCloudEnabled && defaultOrgDisabled) {`

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
```diff
@@ -157,6 +163,17 @@ private Flowable<IdentityProvider> loadProvidersFromConfig() {
...
+     * The 'gravitee' provider is registered for every installation, including managed cloud: it is not a
+     * login path but the write path for organization users, since OrganizationUserServiceImpl forces
+     * {@code source = gravitee} on every user created through the console or the API.
+     * <p>
+     * The DEFAULT reference below is nominal and currently unread — its only consumer is
+     * {@code loadUserProvider} -> {@code PluginLicenseGate.checkPersisted}, which resolves ORGANIZATION
+     * references without a lookup and short-circuits because gravitee-am-idp declares no feature in its
+     * plugin.properties. It becomes meaningful, and wrong under managed cloud where DEFAULT does not
+     * exist, if that plugin ever gains a feature.
+     */
```
leleueri, line 176: discussion: good catch... if this IDP is only linked to the default org, then we have to confirm with the product team that it shouldn't be possible to create org user in a managed mode and all the users have to comes from external IDP or from the GraviteeCloud service

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DefaultOrganizationUpgrader.java
leleueri: answering my own question, let's go this this current change.
We will amend it if it doesn't feat the Gravitee Cloud need but as Cloud is accountable for Org creation, it make more sens to not create the DEFAULT org which is created only for OnPrem deployment.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/IdentityProviderManagerImpl.java
leleueri: I will create a new issue to evolve it. We should be able to add org user from the AM UI

PR #8368, opened 2026-08-05
feat: notify other management api instances when a data plane changes

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7261

Tells the other management API instances about a data plane that was provisioned or deleted on one of them. The node serving the request registers it itself, as it already did, so this is only for the instances that never saw the call.

Provisioning and deleting now write a DATA_PLANE event. Every instance picks it up on its sync tick and registers or drops the definition, so a domain can be created against a provisioned data plane whichever node you hit.

It follows the manager pattern the certificate, entrypoint and license listeners already use, so there is no new mechanism here:

- A deploy rebuilds unconditionally rather than skipping an id already held. The sync poll keeps only the latest event per type and id, so a delete and a re-create of one id inside the window arrive as a single deploy, and skipping would leave the node serving the definition that was deleted.
- A deploy naming a definition that has since gone drops the id instead of failing, which is the same collapse in the other direction.
- Undeploy stops the provider and frees the id, and releases the properties it published, so a later definition reusing that id cannot resolve the settings of the one that has gone.
- Deleting over the internal API also unregisters on the node that served it, matching what create already does.

DataPlaneRegistry gains register and unregister. register still throws on a duplicate id, because that is what stops two gravitee.yml definitions claiming one id at startup, so the deploy path unregisters first.

## :memo: Test scenarios

Unit and e2e coverage in the branch, all green.

I also ran two management API instances against the same mongo, one in the local stack and a second on the host, and drove it by hand. Provision on the first and the second logs `registered from a sync event` a few seconds later, no restart. A domain bound to that data plane is then accepted against the second instance, deleting the data plane while the domain still exists is refused with a 409, and once the domain goes the second instance logs `unregistered`.

Manual steps are written up on AM-7261 if QA want to repeat it. One gotcha if you do: the mongo address in the definition has to resolve from inside docker and from the host, so use the machine's LAN IP, not `mongodb` or `localhost`. The driver connects lazily, so a bad address still logs `registered from a sync event` and only falls over later when something actually uses it.

No UI changes.

DataPlaneEvent carries UPDATE for consistency with the other event enums but nothing emits it, as there is no update endpoint. Updating a data plane was dropped along with the cockpit command approach in the design note, though the epic description still mentions it.

Two things this deliberately does not do. Nothing absorbs the gap between the event being written and a remote node registering, so provisioning and immediately creating a domain against another instance can still be refused. AM-7262 owns that, since its rule turns on whether a data plane is loaded. And an id that exists both as a provisioned row and in the gravitee.yml would be unregistered by an undeploy, which takes a contradictory configuration to reach and is already logged as an error when the two collide at startup.

@gravitee-io/am

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
@@ -82,9 +82,15 @@ public class ProvisionedDataPlaneManager extends AbstractService<ProvisionedData
     }
 
     private void register(DataPlaneDefinition definition) {
+        // the node that served the provisioning request reads its own event back
+        if (provisionedDataPlaneLoader.isServing(definition)) {
+            log.debug("Data plane [{}] is already served at this version, ignoring the event", definition.getId());
+            return;
+        }
         try {
             dataPlaneRegistry.unregister(definition.getId());
             dataPlaneRegistry.register(provisionedDataPlaneLoader.publish(definition));
+            provisionedDataPlaneLoader.markServing(definition);
             log.info("Data plane [{}] of type [{}] registered from a sync event", definition.getId(), definition.getType());
         } catch (Exception e) {
             log.error("Data plane [{}] could not be registered and will be unavailable; domains bound to it cannot be served",
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
@@ -1,0 +1,110 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.management.service.impl;
+
+import io.gravitee.am.common.event.DataPlaneEvent;
+import io.gravitee.am.model.DataPlaneDefinition;
+import io.gravitee.am.model.common.event.Payload;
+import io.gravitee.am.plugins.dataplane.core.DataPlaneRegistry;
+import io.gravitee.am.repository.management.api.DataPlaneDefinitionRepository;
+import io.gravitee.am.service.dataplane.ProvisionedDataPlaneLoader;
+import io.gravitee.common.event.Event;
+import io.gravitee.common.event.EventListener;
+import io.gravitee.common.event.EventManager;
+import io.gravitee.common.service.AbstractService;
+import lombok.CustomLog;
+import org.springframework.context.annotation.Lazy;
+import org.springframework.stereotype.Component;
+
+/**
+ * Applies data plane definitions provisioned on another management API instance. The node serving
+ * the provisioning request registers the definition itself; every other node only ever learns about
+ * it here.
+ * <p>
+ * There is no initial load to do: {@code DataPlaneRegistryImpl} reads the stored definitions when it
+ * starts, and it is listed before this component in the node's components.
+ *
+ * @author GraviteeSource Team
+ */
+@Component
+@CustomLog
+public class ProvisionedDataPlaneManager extends AbstractService<ProvisionedDataPlaneManager> implements EventListener<DataPlaneEvent, Payload> {
+
+    private final DataPlaneRegistry dataPlaneRegistry;
+    private final ProvisionedDataPlaneLoader provisionedDataPlaneLoader;
+    private final DataPlaneDefinitionRepository dataPlaneDefinitionRepository;
+    private final EventManager eventManager;
+
+    public ProvisionedDataPlaneManager(DataPlaneRegistry dataPlaneRegistry,
+                                       ProvisionedDataPlaneLoader provisionedDataPlaneLoader,
+                                       @Lazy DataPlaneDefinitionRepository dataPlaneDefinitionRepository,
+                                       EventManager eventManager) {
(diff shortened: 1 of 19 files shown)


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
```diff
@@ -0,0 +1,99 @@
...
+        dataPlaneDefinitionRepository.findById(dataPlaneId)
+                .subscribe(
+                        this::register,
+                        error -> log.error("Data plane [{}] could not be read and will be unavailable until restart", dataPlaneId, error),
+                        () -> undeploy(dataPlaneId));
+    }
+
+    private void register(DataPlaneDefinition definition) {
+        try {
+            dataPlaneRegistry.unregister(definition.getId());
```
podlesrafal, line 91: The description says the serving node "registers it itself, as it already did, so this is only for the instances that never saw the call" - but management sync reads events with `findByTimeFrame` and isn't filtered by origin, so that node gets its own DEPLOY back. This then runs `unregister()` on the provider it built seconds earlier, stopping the pool and rebuilding it.

That differs from the existing listeners: `licenseManager.registerOrganizationLicense` and the client-secret notifier calls are only ever made from their managers, so a self-event there is genuinely the only path. Here `CreateDataPlaneEndpoint` already did the work, so the self-event is a stop/rebuild rather than a no-op - and if a domain is already using the data plane (which is the point of AM-7260), `stop()` closes the pool under live operations.

I take the point that a collapsed delete+create arrives as a lone DEPLOY, so unconditional register is the safe default. Could the rebuild be gated on something that distinguishes the version - `updatedAt`, or a hash of the configuration held next to the description? The collapsed case still rebuilds, the idempotent self-event doesn't.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
```diff
@@ -43,10 +43,7 @@
 @CustomLog
 public class ProvisionedDataPlaneLoader implements DataPlaneLoader {
 
-    // must not be "graviteeYamlConfiguration": gravitee-node's /_node/configuration endpoint dumps that
-    // source key by key, and these definitions carry credentials
     static final String PROPERTY_SOURCE_NAME = "provisionedDataPlanes";
```
podlesrafal, line 47: Any reason to drop the comment here? It landed yesterday in #8357 and it checks out - gravitee-node's `ConfigurationEndpoint` enumerates key names from the `graviteeYamlConfiguration` source only, so this naming is what keeps the provisioned credentials out of `/_node/configuration`. Without the note, renaming it for consistency looks harmless.

gravitee-am-test/specs/management/internal-api/fixtures/dataplane-provisioning-fixture.ts
```diff
@@ -175,6 +163,18 @@ export async function releaseBoundDomain(bound: BoundDomain | undefined): Promis
   }
 }
 
+export async function canBindADomainTo(dataPlaneId: string): Promise<boolean> {
```
podlesrafal, line 166: nit: the bare catch returns false for an expired token or a 500 as well, so "stops accepting domains on a data plane once it is deleted" could pass for the wrong reason. Worth asserting the rejection status instead of any failure.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManagerTest.java
```diff
@@ -0,0 +1,185 @@
...
+        verify(dataPlaneRegistry).register(DESCRIPTION);
+    }
+
+    /**
+     * The sync poll keeps only the latest event per type and id, so a delete and a re-create of one
+     * id inside the poll window arrive as a lone deploy. Skipping an id already held would leave
+     * this node serving the definition that was deleted.
+     */
+    @Test
+    void shouldReplaceAnIdItAlreadyHoldsOnDeploy() throws Exception {
```
podlesrafal: nit: nothing is registered before the event here, so this verifies the `unregister` -> `register` ordering rather than the replacement the name promises.

podlesrafal (commented):
Read through the whole change plus both SyncManagers, EventKey and the existing management-api listeners. The manager follows the OrganizationLicenseManager shape closely and the error paths are well covered. One behavioural point on the self-event, one question about a dropped comment, and two nits.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManager.java
sclarkamlab: Good catch, and it's not just theory. I still had two instances up from testing the sync and the node that served the POST does exactly this:

```
11:03:45.854  DataPlane provider loaded with id qa-dp
11:03:50.023  Data plane [qa-dp] unregistered
11:03:50.072  DataPlane provider loaded with id qa-dp
```

Fixed in `250726b8b4` with the version gate you suggested. The loader already tracked which ids it had activated, so that set is now id to `updatedAt`, and the manager checks `isServing` before it goes near the registry. A collapsed delete and re-create still rebuilds, because that pair moves the timestamp.

Verified it end to end: provisioning against the patched node now logs the provider once and nothing after it across four sync polls, and the other instance still registers from the event and still drops it on undeploy.

gravitee-am-service/src/main/java/io/gravitee/am/service/dataplane/ProvisionedDataPlaneLoader.java
sclarkamlab: No reason, that was me trimming too hard. Restored in `250726b8b4`.

gravitee-am-test/specs/management/internal-api/fixtures/dataplane-provisioning-fixture.ts
sclarkamlab: Yeah fair. It now rethrows anything that isn't the 400 you get for an unknown data plane, so only a real rejection reads as false. `9d06e3c810`.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/ProvisionedDataPlaneManagerTest.java
sclarkamlab: True, the name promises more than it checks. Renamed to `shouldUnregisterBeforeRegisteringOnDeploy` since the ordering is the thing worth pinning there, and the version cases have proper coverage in `ProvisionedDataPlaneLoaderTest` now. `250726b8b4`.

podlesrafal (approved):

PR #8374, opened 2026-08-05
fix: avoid StackOverflowError when mounting many vhost-mode domains

leleueri:
DefaultReactor previously mounted every vhost-mode domain as its own Vert.x route ("/*") on the shared root router. Since all vhost routes share the same wildcard path, none of them can be skipped by Vert.x's own path matching, so a request has to walk every previously-mounted vhost, in registration order, via RoutingContext.next(). That traversal is recursive - one stack frame per candidate tested - so request latency and stack depth both grew linearly with the number of mounted domains, guaranteed to StackOverflowError past ~3000 domains.

Add VHostDomainIndex: a single Vert.x route backed by a hashmap keyed on the Host header. Dispatch becomes an O(1) lookup plus a linear scan only across the (typically one) candidates registered for that host - no recursion, no dependency on total domain count. Non-vhost (path-based) domains are unaffected; Vert.x's own path matching already skips them via a plain loop, not recursion.

fix AM-7433

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
@@ -19,10 +19,10 @@ import io.gravitee.am.common.event.DomainEvent;
 import io.gravitee.am.gateway.handler.vertx.VertxSecurityDomainHandler;
 import io.gravitee.am.gateway.reactor.Reactor;
 import io.gravitee.am.gateway.reactor.SecurityDomainHandlerRegistry;
+import io.gravitee.am.gateway.reactor.impl.router.VHostDomainIndex;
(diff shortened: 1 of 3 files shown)


lgw-gravitee:
Problem A:
VHost config:
`am.local:8092 + /abc`
Url 
`am.local:8092/abc/oauth... `**DOESNT WORK**

VHost config:
`am.local:8092 + /`
Url 
`am.local:8092/oauth... works`

Problem B:
VHost config:
`am.local:8092 + /`
Url 
`am.local:8092/oauth... client_id=nonexisting `**DOESNT WORK**

getting error:
```
10:44:53.152 [vert.x-eventloop-thread-21] [] ERROR io.vertx.ext.web.RoutingContext - Unhandled exception in router
{"error":"invalid_request"}
	at io.gravitee.am.gateway.handler.root.resources.handler.client.ClientRequestParseHandler.lambda$authenticate$2(ClientRequestParseHandler.java:98)
```
but should be handled by failureHandler

lgw-gravitee (changes requested):
in the comment

leleueri:
another PR has been opened https://github.com/gravitee-io/gravitee-access-management/pull/8379

PR #8378, opened 2026-08-06
fix(dataplane): purge every store when a domain is deleted

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7440
https://gravitee.atlassian.net/browse/AM-7441
https://gravitee.atlassian.net/browse/AM-7442

A domain pins its data plane, so a domain that could not be deleted took its data plane with it: the definition cannot be deleted while a domain still binds to it, and a store that did not answer failed the whole delete. Point a domain at a data plane that does not answer and the pair was stuck for good.

The purge is now best effort. It runs while the domain's reporters are still around to audit it, and a failure is only logged, because the domain has to go regardless of whether its store is reachable.

While in there, the purge was also incomplete. Deleting a domain left its devices, login attempts, webauthn credentials and permission tickets behind. The team called it a long standing bug, so it is fixed here rather than left for later.

Also in here:

- Stores that hold domain data declare themselves through DomainDataPlaneCleanup and the deletion collects them, so it never names a store and the next one added is purged without anyone remembering.
- Devices, login attempts, permission tickets and scope approvals get bulk deletes on the data plane repositories, for mongodb and jdbc.
- Scopes go through a bulk deleteByDomain instead of the per scope delete the deletion used to loop over. That skips unpicking each scope from the roles and applications that reference it, which go with the domain anyway.
- Audit targets with no page in the console render as text instead of a link, so environments and data planes stop leading to a 404 and entrypoints stop leading to the standalone editor.

The purge runs with delayed errors on purpose. One unreachable repository must not skip the stores queued behind it, or a single broken connection leaves most of the domain's data in place.

AM-7442 needed no change. #8368 landed after it was raised and already unregisters a deleted data plane, so what is here is the regression test through the endpoint the new domain form actually reads.

## :memo: Test scenarios

Unit, repository and e2e coverage in the branch, all green. The repository tests run against mongodb and every supported jdbc database. Manual test steps are in the Jira tickets.

No screenshots. The console change removes links from three audit target types, the rows render as plain text instead.

Deleting a domain now removes data that AM used to leave behind, so anyone relying on that residue will notice. Worth calling out in release notes.

The retention rules this falls under are still undefined. AM-7167 lists "data plane decommission and environment deletion need clear retention and migration rules" as an open risk, and nothing has settled it since, so the behaviour here is the code's own intent rather than a spec.

@gravitee-io/am

### DIFF
--- a/gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/spring/JdbcDataPlaneSpringConfiguration.java
+++ b/gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/spring/JdbcDataPlaneSpringConfiguration.java
@@ -90,7 +90,14 @@ public class JdbcDataPlaneSpringConfiguration extends AbstractRepositoryConfigur
     }
 
     @Override
-    public void afterPropertiesSet() throws Exception {
-        initializeDatabaseSchema(getPoolWrapper(), environment, description.propertiesBase() + ".jdbc.");
+    public void afterPropertiesSet() {
+        // resolved outside the catch, is required for API to run
+        var poolWrapper = getPoolWrapper();
+        try {
+            initializeDatabaseSchema(poolWrapper, environment, description.propertiesBase() + ".jdbc.");
+        } catch (Exception e) {
+            LOGGER.error("Schema initialisation failed for data plane [{}], which cannot serve a domain until its database is reachable and the schema is up to date",
+                    description.id(), e);
+        }
     }
 }
--- a/gravitee-am-dataplane/gravitee-am-dataplane-junit/src/main/java/io/gravitee/am/dataplane/junit/gateway/MemoryScopeApprovalRepository.java
+++ b/gravitee-am-dataplane/gravitee-am-dataplane-junit/src/main/java/io/gravitee/am/dataplane/junit/gateway/MemoryScopeApprovalRepository.java
@@ -75,6 +75,11 @@ public class MemoryScopeApprovalRepository extends MemoryRepository<ScopeApprova
     }
 
     @Override
+    public Completable deleteByDomain(String domain) {
+        throw new UnsupportedOperationException("not implemented yet");
+    }
+
+    @Override
     public Completable purgeExpiredData() {
         return ScopeApprovalRepository.super.purgeExpiredData();
     }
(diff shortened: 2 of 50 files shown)


gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/spring/JdbcDataPlaneSpringConfiguration.java
```diff
@@ -90,7 +90,14 @@ public DatabaseDialectHelper databaseDialectHelper(R2dbcDialect dialect) {
     }
 
     @Override
-    public void afterPropertiesSet() throws Exception {
-        initializeDatabaseSchema(getPoolWrapper(), environment, description.propertiesBase() + ".jdbc.");
+    public void afterPropertiesSet() {
+        // resolved outside the catch, is required for API to run
+        var poolWrapper = getPoolWrapper();
```
lgw-gravitee, line 95: q: I dont get this change - why is it requried?

gravitee-am-dataplane/gravitee-am-dataplane-junit/src/main/java/io/gravitee/am/dataplane/junit/gateway/MemoryScopeApprovalRepository.java
```diff
@@ -74,6 +74,11 @@ public Completable deleteByDomainAndClient(String domain, String clientId) {
         throw new UnsupportedOperationException("not implemented yet");
     }
 
+    @Override
+    public Completable deleteByDomain(String domain) {
+        throw new UnsupportedOperationException("not implemented yet");
+    }
```
lgw-gravitee, line 80: wont be implemented ever:P

sclarkamlab: It's only the in memory test implementation so it shouldn't need to get implemented at all.

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/spring/JdbcDataPlaneSpringConfiguration.java
sclarkamlab: `getPoolWrapper()` only builds the pool, it doesn't open a connection, so an unreachable db never fails there. The only thing that actually throws is the liquibase call, and that's the bit I want to swallow.

If the assignment sat inside the try we'd log the warning and then NPE later in `connectionFactory()` on `poolWrapper.getClient()`. Keeping it outside also means a real misconfig (bad driver, missing props) still fails the context instead of being hidden behind a "schema init failed" log.

lgw-gravitee (approved):

PR #8379, opened 2026-08-06
AM-7433 - Refactor VHost routing

leleueri:
avoid StackOverflowError when many vhosts share the same path

Mounting one Route+subRouter per domain/vhost forced Vert.x to walk
and recursively call context.next() across every entry sharing a
mount path. With thousands of domains on the same vhost path
(differentiated only by Host header), this blew the stack.

Group domains/vhosts sharing the same path behind a single
VHostGroupRouter that resolves the match with a map instead of recursive routing, 
keeping stack depth constant regardless of domain count. Also makes domain undeploy actually
remove its routing entries instead of only clearing the inner
    router.

fix AM-7433

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
@@ -19,7 +19,7 @@ import io.gravitee.am.common.event.DomainEvent;
 import io.gravitee.am.gateway.handler.vertx.VertxSecurityDomainHandler;
 import io.gravitee.am.gateway.reactor.Reactor;
 import io.gravitee.am.gateway.reactor.SecurityDomainHandlerRegistry;
-import io.gravitee.am.gateway.reactor.impl.router.VHostRouter;
+import io.gravitee.am.gateway.reactor.impl.router.VHostGroupRouter;
 import io.gravitee.am.gateway.reactor.impl.transaction.TransactionHandlerFactory;
 import io.gravitee.am.model.Domain;
 import io.gravitee.am.model.VirtualHost;
@@ -39,8 +39,11 @@ import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.env.Environment;
 
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.ReentrantLock;
 
 /**
@@ -67,6 +70,15 @@ public class DefaultReactor extends AbstractService implements Reactor, EventLis
     // Root router is mutated only under this lock; heavy per-domain work
     // (Spring context refresh, component start) happens before, in parallel.
     private final ReentrantLock routerLock = new ReentrantLock();
+
+    // One VHostGroupRouter per distinct mount path, shared by every domain/vhost mounted on
+    // that path, so that the number of routes registered on the root router (and therefore the
+    // request-handling recursion depth) no longer grows with the number of domains.
+    private final Map<String, VHostGroupRouter> pathGroups = new ConcurrentHashMap<>();
+
+    // Handles removing a domain's members from the VHostGroupRouter(s) it was mounted on, keyed
+    // by domain id, so undeploy actually forgets the domain instead of only clearing its router.
+    private final Map<String, List<Runnable>> domainUnmountActions = new ConcurrentHashMap<>();
 
     @Autowired
     private TransactionHandlerFactory transactionHandlerFactory;
@@ -120,6 +132,7 @@ public class DefaultReactor extends AbstractService implements Reactor, EventLis
         routerLock.lock();
         try {
             Domain domain = domainHandler.getDomain();
+            List<Runnable> unmountActions = new ArrayList<>();
 
             if (domain.isVhostMode()) {
                 // Mount the same router for each virtual host / path.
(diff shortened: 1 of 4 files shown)


sclarkamlab (commented):
Looks right overall, root router entries now scale with distinct paths instead of domain count, which is what actually kills the StackOverflowError. No blockers, comments inline.

gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
```diff
@@ -131,15 +144,37 @@ public void mountDomain(VertxSecurityDomainHandler domainHandler) {
...
+                    unmountActions.add(() -> group.removeMember(member));
+                });
             } else {
-                this.router.route(sanitizePath(domain.getPath())).subRouter(VHostRouter.router(vertx, domain, domainHandler.router()));
+                VHostGroupRouter group = groupFor(sanitizePath(domain.getPath()));
+                Object member = group.addMember(vertx, domain, domainHandler.router());
+                unmountActions.add(() -> group.removeMember(member));
             }
+
+            domainUnmountActions.put(domain.getId(), unmountActions);
```
sclarkamlab: If `mountDomain` runs twice for the same domain without an unmount in between (a duplicate DEPLOY event would do it, `create()` in the registry mounts unconditionally), this put drops the old handles. The old member keeps the `byHost` slot, so the stored handle's removal is a no-op on the identity check and the domain can never be unmounted after that. I verified this at group level with a throwaway test.

We could clean this up by draining any previous entry at the top of the method, before the new members register:

```java
List<Runnable> previous = domainUnmountActions.remove(domain.getId());
if (previous != null) {
    previous.forEach(Runnable::run);
}
```

gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/router/VHostGroupRouter.java
```diff
@@ -0,0 +1,149 @@
...
+        return byHost.isEmpty() && hostAgnostic.isEmpty();
+    }
+
+    @Override
+    public void handleContext(RoutingContext context) {
+        dispatchTo(resolve(context), context, false);
+    }
+
+    @Override
+    public void handleFailure(RoutingContext context) {
```
sclarkamlab, line 115: The failure path never gets exercised in the tests, all three call `handleContext`. Same for host-agnostic members, the two-arg `addMember` is never used. Both are new routing paths so worth a test each.

gravitee-am-gateway/gravitee-am-gateway-reactor/src/test/java/io/gravitee/am/gateway/reactor/impl/router/VHostGroupRouterTest.java
```diff
@@ -0,0 +1,124 @@
...
+            io.vertx.rxjava3.ext.web.Router delegate = i == targetIndex
+                    ? io.vertx.rxjava3.ext.web.Router.newInstance(targetDelegate)
+                    : io.vertx.rxjava3.ext.web.Router.newInstance(mock(Router.class));
+            group.addMember(vertx, new Domain(), virtualHost("host-" + i + ".gravitee.io", "/"), delegate);
+        }
+
+        RoutingContext routingContext = routingContext("host-" + targetIndex + ".gravitee.io", "/login");
+
+        // Dispatching is a direct map lookup on the Host header, not a scan through every
+        // registered member: a StackOverflowError here would indicate a regression back to
```
sclarkamlab: This comment oversells the test a bit. `RoutingContext` is a mock and `handleContext` is called directly, so there's no vert.x `next()` recursion to overflow, a StackOverflowError can't happen here under any implementation. The map dispatch check is still worth keeping, but either reword this or add a test that drives a real router chain.

gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/router/VHostGroupRouter.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public io.vertx.rxjava3.ext.web.Router asRxRouter() {
+        return io.vertx.rxjava3.ext.web.Router.newInstance(this);
+    }
+
+    /**
+     * Registers a new domain/vhost candidate on this group and returns an opaque handle that
+     * must be kept to later {@link #removeMember(Object)} it (e.g. on domain undeploy).
+     */
+    public Object addMember(io.vertx.rxjava3.core.Vertx vertx, Domain domain, VirtualHost vhost, io.vertx.rxjava3.ext.web.Router delegate) {
```
leleueri: I do prefer an addMember method which return the effective member we just added. In that way the intension is clearer

gravitee-am-gateway/gravitee-am-gateway-reactor/src/test/java/io/gravitee/am/gateway/reactor/impl/router/VHostGroupRouterTest.java
leleueri: rephrase the comment and add an assertion on the number request object access to confirm the single member usage

gravitee-am-gateway/gravitee-am-gateway-reactor/src/main/java/io/gravitee/am/gateway/reactor/impl/DefaultReactor.java
leleueri: I manage to do it at the end by using the result of the put method which provide the previous value if any.
Also added a WARN to highlight the case which shouldn't happen

lgw-gravitee (approved):

PR #8383, opened 2026-08-10
fix: reporter index manipulation should be ran only by MAPI

lgw-gravitee:
introduced reports-only ensureIndex setting property

fixes AM-7454
fixes github-io/issues#11712

### DIFF
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/node/AmNode.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/node/AmNode.java
@@ -1,0 +1,29 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.common.node;
+
+import io.gravitee.node.api.Node;
+
+public interface AmNode extends Node {
+
+    default boolean isGateway() {
+        return false;
+    }
+
+    default boolean isManagementNode() {
+        return false;
+    }
+}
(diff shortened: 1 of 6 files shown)


gravitee-am-reporter/gravitee-am-reporter-mongodb/src/main/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporter.java
```diff
@@ -260,37 +272,34 @@ protected void doStop() throws Exception {
...
-                    .doOnComplete(() -> logger.debug("{} Reporter indexes created", indexes.size()))
-                    .doOnError(throwable -> logger.error("An error has occurred during creation of indexes", throwable));
-
-            // process indexes
-            deleteOldIndexes
-                    .andThen(createNewIndexes)
-                    .subscribe();
+        // drop old indexes
+        // see : https://github.com/gravitee-io/issues/issues/7136
+        Completable deleteOldIndexes = Observable.fromPublisher(reportableCollection.listIndexes())
```
sclarkamlab: Why not use the better implementation for deleting indexs in MongoUtils?

gravitee-am-common/src/main/java/io/gravitee/am/common/node/AmNode.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.common.node;
+
+import io.gravitee.node.api.Node;
+
+public interface AmNode extends Node {
```
sclarkamlab, line 20: Did you add this because it is cleaner than relying on  `node.application().contains("gateway")` to infer gateway node?

lgw-gravitee: yes - do you dislike it?:)

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/main/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporter.java
lgw-gravitee: fixed that.

gravitee-am-common/src/main/java/io/gravitee/am/common/node/AmNode.java
sclarkamlab: no it is cleaner than relying on a text field that may change in the future.

sclarkamlab (approved):

PR #8384, opened 2026-08-10
fix: removed DefaultReporterUpgrader

lgw-gravitee:
Why the upgrader existed
The configuration of a system reporter cannot be edited through the API
(ReporterServiceImpl:213 ignores configuration for system reporters unless
isUpgrader=true), and the gateway builds its reporter provider from the configuration
stored in the database. The upgrader was therefore the only channel propagating
gravitee.yml changes (Mongo/JDBC host, credentials, URI) to system reporters.

Why it is no longer needed
The configuration is fully derived from repositories.management.*
(ReporterServiceImpl.createReporterConfig) plus the reporter reference, and every node
already has those properties in its own gravitee.yml — the gateway distribution has had a
management repository section since at least 3.21, and since 4.6.0 (AM-3680) both containers
register a RepositoriesEnvironment bean resolving repositories.management.* with a
fallback to management.*. The same pattern is already in use for the internal platform
reporter, which is built from a locally generated configuration and never persisted
(ManagementAuditReporterManager:116-119).

In other words, persisting a derived value in the database is what created the need for the
upgrader in the first place.


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/SystemReporterConfigResolver.java
```diff
@@ -0,0 +1,265 @@
...
+    private Optional<String> getMongoServers(RepositoriesEnvironment env) {
+        LOGGER.debug("Looking for MongoDB server configuration...");
+        boolean found = true;
+        int idx = 0;
+        List<String> endpoints = new ArrayList<>();
+
+        while (found) {
+            String serverHost = env.getProperty(Scope.MANAGEMENT.getRepositoryPropertyKey() + ".mongodb.servers[" + (idx) + "].host");
+            int serverPort = env.getProperty(Scope.MANAGEMENT.getRepositoryPropertyKey() + ".mongodb.servers[" + (idx) + "].port", int.class, 27017);
+            idx++;
```
sclarkamlab, line 249: Nice, the old loop was doing `idx++` twice so it paired `servers[0].host` with `servers[1].port` and then jumped to `servers[2]`. Real fix.

Worth a line in the PR description though, it changes the generated uri for anyone running multi node mongo.

Nothing pins it either. `shouldHaveNullPortWhenMongoServersAreDefined` only sets `servers[0]`, so it passes against the old loop too. A two server case would lock it in.

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/SystemReporterConfigResolver.java
```diff
@@ -0,0 +1,265 @@
...
+                          "enableCredentials": false,
+                          "database": "%s",
+                          "reportableCollection": "reporter_audits%s",
+                          "bulkActions": 1000,
+                          "flushInterval": 5
+                        }
+                        """.formatted(
+                    mongoUri,
+                    (mongoHost != null) ?  mongoHost : "",
+                    (mongoPort != null) ?Integer.parseInt(mongoPort) : null,
```
sclarkamlab: Lost a space here.

```suggestion
                    (mongoPort != null) ? Integer.parseInt(mongoPort) : null,
```

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-common/src/main/java/io/gravitee/am/gateway/handler/common/audit/impl/GatewayAuditReporterManager.java
```diff
@@ -235,7 +239,7 @@ private boolean startReporterProvider(Reporter reporter, GraviteeContext context
             return false;
         }
         logger.info("\tInitializing reporter: {} [{}]", reporter.getName(), reporter.getType());
-        var providerConfiguration = new ReporterProviderConfiguration(reporter, context);
+        var providerConfiguration = new ReporterProviderConfiguration(systemReporterConfigResolver.resolve(reporter), context);
```
sclarkamlab: Gateway didn't run any of this config building code before, and a throw here is silent. `resolve()` sits outside the try below, and we're in a `forEach` inside an rx `subscribe` consumer, so it lands in `RxJavaPlugins.onError` rather than the handler on 111, and the rest of the domain's reporters get skipped.

Not a blocker. `Integer.parseInt(jdbcPort)` behind it is pre-existing, and mgmt api already hits the same line at startup via `createInternal()`. Cheap insurance though:

```java
Reporter resolved;
try {
    resolved = systemReporterConfigResolver.resolve(reporter);
} catch (Exception ex) {
    logger.error("Unable to resolve config for reporter {}", reporter.getId(), ex);
    return false;
}
var providerConfiguration = new ReporterProviderConfiguration(resolved, context);
```

gravitee-am-service/src/test/java/io/gravitee/am/service/reporter/SystemReporterConfigResolverTest.java
```diff
@@ -0,0 +1,132 @@
...
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+
+/**
+ * @author GraviteeSource Team
+ */
+class SystemReporterConfigResolverTest {
+
+    private final MockEnvironment environment = new MockEnvironment();
+    private final SystemReporterConfigResolver resolver = new SystemReporterConfigResolver(new RepositoriesEnvironment(environment));
```
sclarkamlab, line 39: Nothing in here covers the jdbc branch, it's all mongo. The sha-256 table suffix path for domain ids over 30 chars isn't hit either.

sclarkamlab (commented):
Had a look through this. Swapping the boot time upgrader for resolving at reporter init makes sense to me, and `resolve()` copying rather than mutating the reporter is nice.

Few comments, nothing blocking.

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/SystemReporterConfigResolver.java
lgw-gravitee: I added test and Im skipping description part

sclarkamlab (approved):

PR #8393, opened 2026-08-12
test: cover LDAP authentication, role mapping and attributes (AM-7447)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7447

Adds LDAP regression coverage for AM-7447. There was previously **no automated LDAP
coverage of any kind** — no specs, and no LDAP server in the local stack.

### Local stack

Adds an OpenLDAP service, seeded and health-gated so tests cannot race the import:

```
./local-stack.sh up --ldap        # or --with ldap, or --full
npm --prefix docker/local-stack run stack:dev:setup:ldap
```

The directory is built so one instance covers every LDAP configuration: `memberOf` as
full DNs (memberof overlay), `memberUid` for posix group search, a nested group, an
alternate group attribute, plaintext / `{SSHA}` / compare-mode credentials, and a second
subtree so two providers can resolve one username.

`openldap-init` installs the memberof overlay and **fails the stack** if the expected
`memberOf` values do not materialise — otherwise a misconfiguration would surface much
later as an inexplicable role-mapper failure.


docker/local-stack/dev/docker-compose-ldap.yml
```diff
@@ -0,0 +1,155 @@
...
+        if ! ldapsearch -x -LLL -H ldap://openldap:1389 -D "cn=admin,dc=gravitee,dc=io" -w admin \
+               -b "uid=ldap-admin,ou=people,dc=gravitee,dc=io" -s base memberOf \
+             | grep -q "memberOf:"; then
+          echo "groups predate the overlay — rewriting them"
+          ldapdelete -H ldap://openldap:1389 -D "cn=admin,dc=gravitee,dc=io" -w admin \
+            "cn=GRAVITEE_ADMINS,ou=groups,dc=gravitee,dc=io" \
+            "cn=GRAVITEE_USERS,ou=groups,dc=gravitee,dc=io"
+          apply "groups (rewrite)" \
+            ldapadd -H ldap://openldap:1389 -D "cn=admin,dc=gravitee,dc=io" -w admin \
+              -f /ldifs/groups.ldif
```
sclarkamlab, line 143: This rewrite branch only deletes `GRAVITEE_ADMINS` and `GRAVITEE_USERS`, but `ldapadd` bails on the first error, so once it hits `workers` (rc 68) it stops and `engineers` + `GRAVITEE_SUPER_ADMINS` never get rewritten either. `apply` reads 68 as already-applied so the whole thing reports ok.

The other half of it is `olcMemberOfRefInt: TRUE`. Deleting the `GRAVITEE_ADMINS` entry strips it out of `GRAVITEE_SUPER_ADMINS`'s member list, and re-adding it doesn't put it back, so the nested group tests go too.

Net effect is john ends up with no `memberOf` and AM-5385 fails with nothing to go on, and the gate below only asserts on ldap-admin so it passes regardless. Could we delete all five here (or `ldapadd -c`) and assert john's `memberOf` in the gate as well?

docker/local-stack/dev/docker-compose-ldap.yml
```diff
@@ -0,0 +1,155 @@
...
+# OpenLDAP directory backing the LDAP identity-provider tests (AM-7447).
+#
+# Start with:  ./local-stack.sh up --with ldap
+#
+# The LDAP IdP plugin already ships inside the AM distribution zips
+# (gravitee-am-identityprovider-ldap), so — unlike the Kerberos overlay — no
+# extra plugin zip needs mounting into the gateway or management containers.
+#
+# IdP configuration that matches this service:
+#   URL                    ldap://openldap:389
```
sclarkamlab: Should be `1389`. Everything else in the file uses it, and the comment on line 43 explains why 389 can't work, so this block sends anyone setting the IdP up by hand straight into connection refused.

gravitee-am-test/specs/gateway/ldap/ldap-user-attributes.jest.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+ */
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from '@jest/globals';
+import { login, getHeaderLocation } from '@gateway-commands/login-commands';
+import { performGet, logoutUser } from '@gateway-commands/oauth-oidc-commands';
+import { deleteUser, listUsers } from '@management-commands/user-management-commands';
+import { setup } from '../../test-fixture';
+import { jira } from '@specs-utils/jira';
+import { LdapFixture, setupLdapFixture, LDAP_ADMIN, REDIRECT_URI } from './fixtures/ldap-fixture';
+
+setup(200000);
```
sclarkamlab, line 24: Did you mean to leave `retryImmediatelyForThisFile()` off this one? Every other spec in the folder has it.

Not a flake risk as far as I can tell, each test re-applies its own mappers first and `beforeEach` drops the user, so a deferred retry still works. Just the odd one out.

sagarchauhan-qa: Deliberate — deferred retry works there for the reason you spotted. But not worth being the odd one out, so I've added it with a comment explaining why.

docker/local-stack/dev/docker-compose-ldap.yml
sagarchauhan-qa: Header now says 1389. Checked the rest of the LDAP files for stale 389 references too; that was the only one.

docker/local-stack/dev/docker-compose-ldap.yml
sagarchauhan-qa: All done — deletes every group listed in groups.ldif rather than a hardcoded pair, re-add uses -c, and the gate now asserts john as well as ldap-admin. Also added a check on the nested group link so that it can't be silently lost either.

Verified by deliberately breaking each case and confirming the stack refuses to start. Suite still green.

Minor note: I couldn't reproduce the olcMemberOfRefInt half — deleting GRAVITEE_ADMINS leaves GRAVITEE_SUPER_ADMINS's member list intact, since that pruning comes from the refint overlay and only memberof is loaded. The outcome you described is real; it just comes from the ldapadd bail. Covered either way.

sclarkamlab (approved):

PR #8394, opened 2026-08-12
test: permissions and roles regression automation (AM-7448)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7448

Automates the **Regression > Permissions and Roles** area — who can do what at organization, domain, application and protected-resource level, and what the Console shows them.

  ## Jest — 13 specs, 223 tests
  
  `gravitee-am-test/specs/management/permissions/`

  - **Endpoint sweeps** — every endpoint refuses a caller without its permission, and opens to one granted exactly that permission. Table generated from`docs/mapi/openapi.yaml`.
  - **Write sweep** — 17 write operations refused, collections re-read to confirm nothing was created.
  - **Role scope confinement** — a role only works at the tier it is assignable to.
  - **Delegated administration** — who may grant roles, plus escalation guards.
  - **Membership cascade** — assignment provisions parent-tier memberships.
  - **Cross-reference isolation** — a resource can't be reached via the wrong parent.
  - **List scoping** — lists narrowed to what the caller may see.
  - **Custom role lifecycle** — create, grant, built-in roles resist modification.
  - **Assignment journey** — onboard, grant, verify, revoke, verify.
  - **Organization baseline** — default role, Console access, group permissions.
  - **Protected-resource tier** — the fifth tier end to end.

  ## Playwright — 5 specs, 23 tests
  
  `gravitee-am-test/playwright/tests/permissions/`

Console visibility per role: which domains and applications appear, which menus are shown or withheld, role creation through the Console, system-role permissions greyed out, and that what the Console offers matches what the API allows for the same user.

  ## Xray automated

  | Spec | Xray |
  |---|---|
  | `org-role-visibility` | AM-6031 |
  | `domain-role-visibility` | AM-6032 |
  | `application-role-visibility` | AM-6036 |
  | `ui-api-parity` | AM-7472 |
  | `application-permission-isolation` | AM-7473 |
  
**A few routes are excluded from the sweeps with a Jira reference where current behaviour is defective.**

## :memo: Test scenarios 
NA

NA

NA

### DIFF
--- a/gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
+++ b/gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
@@ -18,15 +18,18 @@ import { test, expect } from '../../fixtures/permissions.fixture';
 import { ApplicationGeneralSettingsPage } from '../../pages/application-general-settings.page';
 import { Oauth2SettingsPage } from '../../pages/oauth2-settings.page';
 import { linkJira } from '../../utils/jira';
-import { MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
+import { submenuItem } from '../../utils/permissions-helpers';
+import { APPLICATION_OVERVIEW_URL, MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
 
 // Each test signs in as its own persona, so no shared authenticated state.
 test.use({ storageState: { cookies: [], origins: [] } });
 
-type Page = import('@playwright/test').Page;
+// Every test here builds its own permissions world (domains, applications, personas, memberships)
+// and that setup counts against the test timeout, so the extended budget applies to all of them
+// rather than to the few that happened to ask for it.
+test.beforeEach(({}, testInfo) => testInfo.setTimeout(MULTI_PHASE_TEST_TIMEOUT));
 
-/** A submenu entry in the application sidebar, rendered as <a title="{label}">. */
-const submenuItem = (page: Page, label: string) => page.locator(`.gv-submenu a[title="${label}"]`);
+type Page = import('@playwright/test').Page;
 
 /** An application row in the list, rendered as <a [routerLink]="[row.id]">{{ row.name }}</a>. */
 const applicationLink = (page: Page, name: string) => page.getByRole('link', { name, exact: true });
@@ -40,7 +43,6 @@ const oauthSaveButton = (page: Page) => page.getByRole('button', { name: /^SAVE$
 test.describe('Application-level roles - Console visibility', () => {
   test('AM-6036: an application owner can rename the application they own', async ({ page, permissionsWorld, signInAs }, testInfo) => {
     linkJira(testInfo, 'AM-6036');
-    test.setTimeout(MULTI_PHASE_TEST_TIMEOUT);
 
     await signInAs(permissionsWorld.appOwner);
 
@@ -61,7 +63,6 @@ test.describe('Application-level roles - Console visibility', () => {
     signInAs,
   }, testInfo) => {
     linkJira(testInfo, 'AM-6036');
-    test.setTimeout(MULTI_PHASE_TEST_TIMEOUT);
 
     await signInAs(permissionsWorld.appOwner);
 
@@ -100,7 +101,7 @@ test.describe('Application-level roles - Console visibility', () => {
     await signInAs(permissionsWorld.appViewer);
 
     await applicationLink(page, permissionsWorld.ownedApplication.name).click();
-    await page.waitForURL(/.*\/applications\/.*\/overview.*/i);
+    await page.waitForURL(APPLICATION_OVERVIEW_URL);
 
     // Anchor on the menus a reader is entitled to before asserting the rest are absent.
     await expect(submenuItem(page, 'Overview')).toBeVisible();
@@ -122,7 +123,7 @@ test.describe('Application-level roles - Console visibility', () => {
     await signInAs(permissionsWorld.appOwner);
 
     await applicationLink(page, permissionsWorld.ownedApplication.name).click();
-    await page.waitForURL(/.*\/applications\/.*\/overview.*/i);
+    await page.waitForURL(APPLICATION_OVERVIEW_URL);
 
     await expect(submenuItem(page, 'Overview')).toBeVisible();
     await expect(submenuItem(page, 'Settings')).toBeVisible();
--- a/gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
+++ b/gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
@@ -1,0 +1,130 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { test, expect } from '../../fixtures/permissions.fixture';
+import { ApplicationGeneralSettingsPage } from '../../pages/application-general-settings.page';
+import { Oauth2SettingsPage } from '../../pages/oauth2-settings.page';
+import { linkJira } from '../../utils/jira';
+import { MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
+
+// Each test signs in as its own persona, so no shared authenticated state.
+test.use({ storageState: { cookies: [], origins: [] } });
+
+type Page = import('@playwright/test').Page;
+
+/** A submenu entry in the application sidebar, rendered as <a title="{label}">. */
+const submenuItem = (page: Page, label: string) => page.locator(`.gv-submenu a[title="${label}"]`);
+
+/** An application row in the list, rendered as <a [routerLink]="[row.id]">{{ row.name }}</a>. */
+const applicationLink = (page: Page, name: string) => page.getByRole('link', { name, exact: true });
+
+/**
+ * The OAuth2 settings form saves through a plain mat-raised-button, not a submit button, so
+ * BasePage.clickSave() cannot reach it. It stays disabled until the form is dirty.
+ */
+const oauthSaveButton = (page: Page) => page.getByRole('button', { name: /^SAVE$/ });
+
+test.describe('Application-level roles - Console visibility', () => {
+  test('AM-6036: an application owner can rename the application they own', async ({ page, permissionsWorld, signInAs }, testInfo) => {
+    linkJira(testInfo, 'AM-6036');
+    test.setTimeout(MULTI_PHASE_TEST_TIMEOUT);
+
+    await signInAs(permissionsWorld.appOwner);
+
+    const settings = new ApplicationGeneralSettingsPage(page);
+    await settings.navigateTo(permissionsWorld.domain.id, permissionsWorld.ownedApplication.id);
+
+    const renamed = `${permissionsWorld.ownedApplication.name}-renamed`;
+    await settings.fillName(renamed);
+    await settings.clickSave();
+
+    await settings.expectSnackbar(/updated|saved/i);
+    await expect(settings.nameInput).toHaveValue(renamed);
+  });
+
+  test('AM-6036: an application owner can change the client authentication method', async ({
+    page,
+    permissionsWorld,
+    signInAs,
+  }, testInfo) => {
+    linkJira(testInfo, 'AM-6036');
+    test.setTimeout(MULTI_PHASE_TEST_TIMEOUT);
+
+    await signInAs(permissionsWorld.appOwner);
+
+    const oauth = new Oauth2SettingsPage(page);
+    await oauth.navigateTo(permissionsWorld.domain.id, permissionsWorld.ownedApplication.id);
+
+    await expect(oauth.tokenAuthMethodSelect).toBeVisible();
+    await oauth.tokenAuthMethodSelect.click();
+    await oauth.tokenAuthMethodOption(/client_secret_post/).click();
+
+    await expect(oauthSaveButton(page)).toBeEnabled();
+    await oauthSaveButton(page).click();
+
+    await oauth.expectSnackbar(/updated|saved/i);
+    expect(await oauth.getTokenAuthMethodText()).toContain('client_secret_post');
+  });
+
+  test('AM-6036: an application owner sees only the application they are a member of', async ({
+    page,
+    permissionsWorld,
+    signInAs,
+  }, testInfo) => {
+    linkJira(testInfo, 'AM-6036');
+
+    await signInAs(permissionsWorld.appOwner);
+
+    // Positive anchor before the absence assertion — on an unrendered page every row is missing,
+    // so the owned application must be shown first for the exclusion below to mean anything.
+    await expect(applicationLink(page, permissionsWorld.ownedApplication.name)).toBeVisible();
+    await expect(applicationLink(page, permissionsWorld.otherApplication.name)).toHaveCount(0);
+  });
+
+  test('AM-6036: a read-only application role hides the editing menus', async ({ page, permissionsWorld, signInAs }, testInfo) => {
+    linkJira(testInfo, 'AM-6036');
+
+    await signInAs(permissionsWorld.appViewer);
+
+    await applicationLink(page, permissionsWorld.ownedApplication.name).click();
+    await page.waitForURL(/.*\/applications\/.*\/overview.*/i);
+
+    // Anchor on the menus a reader is entitled to before asserting the rest are absent.
+    await expect(submenuItem(page, 'Overview')).toBeVisible();
+    await expect(submenuItem(page, 'Endpoints')).toBeVisible();
+
+    await expect(submenuItem(page, 'Settings')).toHaveCount(0);
+    await expect(submenuItem(page, 'Identity Providers')).toHaveCount(0);
+    await expect(submenuItem(page, 'Design')).toHaveCount(0);
+    await expect(submenuItem(page, 'Analytics')).toHaveCount(0);
+  });
+
+  test('AM-6036: an application owner keeps the editing menus the reader is denied', async ({
+    page,
+    permissionsWorld,
+    signInAs,
+  }, testInfo) => {
+    linkJira(testInfo, 'AM-6036');
+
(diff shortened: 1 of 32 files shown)


gravitee-am-test/specs/management/permissions/endpoint-permission-sweep.jest.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+ * limitations under the License.
+ */
+
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { setup } from '../../test-fixture';
+import { RbacFixture, setupRbacFixture } from './fixtures/rbac-fixture';
+import { PERMISSION_ENDPOINTS, PermissionEndpoint } from './fixtures/permission-endpoints.generated';
+import { performDelete, performGet } from '@gateway-commands/oauth-oidc-commands';
+
+setup();
```
sclarkamlab: All 13 specs in this folder call bare `setup()`, which is the 20s default. `setupRbacFixture` is 2 domains, 2 apps, 4 personas and 5 membership calls, and it runs in 13 files at once under `--maxWorkers=2` with `--bail`, so one slow beforeAll kills the whole management run. `setup(200000)` costs nothing and removes the flake mode.

gravitee-am-test/specs/management/permissions/membership-cascade.jest.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+import {
+  addApplicationMembership,
+  addDomainMembership,
+  getEnvironmentMemberPermissions,
+  listDomainMemberships,
+  userMembership,
+} from '@management-commands/membership-management-commands';
+import { deleteOrganisationUser } from '@management-commands/organisation-user-commands';
+
+setup();
```
sclarkamlab: This file is a narrative (baseline, assign, assert) but doesn't call `retryImmediatelyForThisFile()`, so with `JEST_RETRIES=2` in CI a failed step gets retried after the later steps already mutated state. Same in `protected-resource-tier` and `organization-role-baseline`. `role-assignment-journey` already does it right.

gravitee-am-test/playwright/tests/permissions/domain-role-visibility.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+import { test, expect } from '../../fixtures/domain-permissions.fixture';
+import { ApplicationCreationPage } from '../../pages/application-creation.page';
+import { linkJira } from '../../utils/jira';
+import { uniqueTestName } from '../../utils/fixture-helpers';
+import { MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
+
+// Every test signs in as its own persona, so no shared authenticated state.
+test.use({ storageState: { cookies: [], origins: [] } });
+
+const DOMAIN_LIST = '/environments/default/domains';
```
sclarkamlab: GUIDELINES §3 puts cross-file values in `utils/test-constants.ts`, and this one also bypasses `envHrid` from `base.page.ts` (only hardcoded `environments/default` in playwright/). Same story for the overview regex (x5 across 3 files), the `a[mat-fab], button[mat-fab]` selector (x4) and the no-environment copy (x2).

gravitee-am-test/specs/management/permissions/permission-sufficiency.jest.spec.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const fill = (route: string) =>
+  route
+    .replace('{organizationId}', process.env.AM_DEF_ORG_ID)
+    .replace('{environmentId}', process.env.AM_DEF_ENV_ID)
+    .replace('{domain}', fixture.domain.id)
+    .replace('{application}', fixture.application.id);
+
+/** See endpoint-permission-sweep for why each of these cannot be driven generically. */
+const EXCLUDED: Record<string, string> = {
```
sclarkamlab: This table and the one in `endpoint-permission-sweep` are already diverging, this side has 2 extra exclusions and 2 extra overrides. Worth moving `EXCLUDED` and `DOCUMENTED_PERMISSION_OVERRIDES` into a shared module under `fixtures/` before they drift further apart.

gravitee-am-test/specs/management/permissions/write-endpoint-sweep.jest.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      const response = await performPost(managementUrl(), writeCase.path(), writeCase.body(), headers(fixture.orgUser.token));
+
+      expect(response.status).toBe(403);
+      expect(response.body.message).toEqual('Permission denied');
+    });
+  });
+});
+
+describe('Write endpoint sweep - the refusals left nothing behind', () => {
+  WRITE_CASES.filter((writeCase) => writeCase.collection).forEach((writeCase) => {
```
sclarkamlab, line 225: Only 5 of the 17 cases declare a `collection`, so the other 12 (all six org tier writes, both member adds, both idp creates) never get the residue check this describe says is the point. Most of them have list endpoints, so the same trick works.

sclarkamlab (commented):
Ran a super review over this (code-review + cursor, then checked against both GUIDELINES docs). Inline comments below, the first 5 are worth fixing before merge, the rest take or leave. The test design itself is genuinely nice, the positive anchors and the two-direction sweep especially.

gravitee-am-test/specs/management/permissions/fixtures/rbac-fixture.ts
```diff
@@ -0,0 +1,242 @@
...
+      findOrganizationRoleByName(adminToken, 'APPLICATION_OWNER', ListRolesTypeEnum.Application),
+    ]);
+
+    const [orgUser, domainOwnerPersona, appOwner, assignee] = await Promise.all([
+      createPersona(adminToken, 'orguser'),
+      createPersona(adminToken, 'domainowner'),
+      createPersona(adminToken, 'appowner'),
+      createPersona(adminToken, 'assignee'),
+    ]);
+    personas.push(orgUser, domainOwnerPersona, appOwner, assignee);
```
sclarkamlab: If one `createPersona` rejects, `Promise.all` rejects before this push runs, so the catch block loops an empty `personas` array and the three users that did get created leak. Pushing each persona as it resolves fixes it.

gravitee-am-test/specs/management/permissions/role-scope-confinement.jest.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+});
+
+/**
+ * A role declares the tier it may be assigned to via `assignableType`. Assignment is refused
+ * outright when that does not match the tier of the membership, which is the first of two
+ * independent gates confining a role to its scope.
+ */
+describe('Role scope confinement - a role can only be assigned at its own tier', () => {
+  it('should accept an ORGANIZATION-assignable role at organization level', async () => {
```
sclarkamlab, line 93: These three accept cases have no assertions, they pass on any non-throwing call. Worth asserting the membership actually landed like the delegated-admin cases do.

gravitee-am-test/specs/management/permissions/endpoint-permission-sweep.jest.spec.ts
```diff
@@ -0,0 +1,189 @@
...
+    });
+  });
+});
+
+describe(`Endpoint permission sweep - ${allowed.length} endpoints within the default allowance`, () => {
+  allowed.forEach((endpoint) => {
+    it(`should permit ${endpoint.method} ${endpoint.route} (${requiredPermission(endpoint)})`, async () => {
+      const response = await performGet(managementUrl(), fill(endpoint.route), headers(fixture.orgUser.token));
+
+      expect(response.status).toBe(200);
```
sclarkamlab: `permission-sufficiency` notes some of these answer 204 when the collection is empty, which would make this flake on a fresh instance. `toBeLessThan(400)` like that file uses would be safer. Same in `deny-by-default:173`.

gravitee-am-test/specs/management/permissions/protected-resource-tier.jest.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+  it('should show the resource to the member assigned to it', async () => {
+    const response = await listResourcesAs(member.token);
+
+    expect(response.status).toBe(200);
+    expect(response.body.data.map((resource) => resource.id)).toContain(protectedResourceId);
+  });
+
+  it('should remove them from the resource when the membership is deleted', async () => {
+    const { memberships } = await listProtectedResourceMemberships(fixture.domain.id, protectedResourceId, fixture.adminToken);
+    const theirs = memberships.find((membership) => membership.memberId === member.userId);
```
sclarkamlab: If the membership is missing this dies with a TypeError on `theirs.id` instead of a readable failure. A `toHaveLength(1)` first keeps it useful, like `organization-role-baseline:91` does. Same shape at :198, :208, `role-assignment-journey:121` and `organization-role-baseline:154`.

gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+import { linkJira } from '../../utils/jira';
+import { MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
+
+// Each test signs in as its own persona, so no shared authenticated state.
+test.use({ storageState: { cookies: [], origins: [] } });
+
+type Page = import('@playwright/test').Page;
+
+/** A submenu entry in the application sidebar, rendered as <a title="{label}">. */
+const submenuItem = (page: Page, label: string) => page.locator(`.gv-submenu a[title="${label}"]`);
```
sclarkamlab: This already exists in `utils/permissions-helpers.ts`, the other two specs import it from there.

gravitee-am-test/specs/management/permissions/custom-role-lifecycle.jest.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+import { performPost } from '@gateway-commands/oauth-oidc-commands';
+import { getOrganisationManagementUrl } from '@management-commands/service/utils';
+import { uniqueName } from '@utils-commands/misc';
+import { ListRolesTypeEnum } from '@management-apis/RoleApi';
+import { JWT_FORMAT } from '@specs-utils/jwt-format';
+
+setup();
+
+let accessToken: string;
+const createdRoleIds: string[] = [];
```
sclarkamlab: Only file in the folder without the fixture pattern. Mild since there's no domain to build here, but a tiny fixture would keep the folder consistent.

gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    await oauth.tokenAuthMethodOption(/client_secret_post/).click();
+
+    await expect(oauthSaveButton(page)).toBeEnabled();
+    await oauthSaveButton(page).click();
+
+    await oauth.expectSnackbar(/updated|saved/i);
+    expect(await oauth.getTokenAuthMethodText()).toContain('client_secret_post');
+  });
+
+  test('AM-6036: an application owner sees only the application they are a member of', async ({
```
sclarkamlab, line 83: `permissionsWorld` rebuilds the whole world per test and that setup counts against the test timeout, but only 2 of the 5 tests here set `MULTI_PHASE_TEST_TIMEOUT`. Either they all need it or none do. Same split in `domain-role-visibility`.

gravitee-am-test/scripts/generate-permission-endpoints.js
```diff
@@ -0,0 +1,136 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Regenerates the endpoint/permission table driven by endpoint-permission-sweep.jest.spec.ts.
+ *
+ *   node scripts/generate-permission-endpoints.js
```
sclarkamlab, line 20: Nothing fails CI if the generated table drifts from `openapi.yaml`, so a new guarded endpoint only joins the sweep when someone remembers to rerun this. A staleness check like the check-oas one would close that gap.

gravitee-am-test/specs/management/permissions/organization-role-baseline.jest.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    await addDomainMembership(fixture.domain.id, fixture.adminToken, groupMembership(group.id, fixture.roles.domainOwner.id));
+
+    const response = await performGet(getOrganisationManagementUrl(), domainPath(fixture.domain.id), headers(groupMember.token));
+
+    expect(response.status).toBe(200);
+    expect(response.body.id).toEqual(fixture.domain.id);
+  });
+
+  it('should still confine a group-derived role to the domain it was assigned to', async () => {
+    const response = await performGet(getOrganisationManagementUrl(), domainPath(fixture.otherDomain.id), headers(groupMember.token));
```
sclarkamlab, line 146: A caller with no access at all also gets this 403, so on its own this doesn't distinguish confinement from the role never applying. The previous test anchors the positive so it holds while the file runs in order, just fragile if anyone reorders.

gravitee-am-test/scripts/generate-permission-endpoints.js
sagarchauhan-qa: Closed this with permission-table-freshness.jest.spec.ts rather than a CircleCI step - it asserts the committed table equals a fresh generation from openapi.yaml, needs no stack and so fails the run_jest_tests_* jobs that already run on every PR. Verified by deleting a row and watching it go red. Happy to move it into workflows.yml alongside the oas check if you'd prefer it there.

gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
sagarchauhan-qa: Removed. Also collapsed the permissions-helpers copy into utils/selectors.ts, which already had one — single definition now. 24 permissions tests pass against it.

gravitee-am-test/specs/management/permissions/custom-role-lifecycle.jest.spec.ts
sagarchauhan-qa: Done - added fixtures/role-fixture.ts (admin token, createRole, cleanup, nothing else) and custom-role-lifecycle now opens the same way as the rest of the folder. Kept it separate from setupRbacFixture since that provisions two domains, two applications and four personas this spec never touches. 237 tests passing.

gravitee-am-test/specs/management/permissions/endpoint-permission-sweep.jest.spec.ts
sagarchauhan-qa: Done — all 14 specs in the folder now use setup(200000), matching the convention across specs/management.

gravitee-am-test/specs/management/permissions/membership-cascade.jest.spec.ts
sagarchauhan-qa: Added to all three — membership-cascade, protected-resource-tier and organization-role-baseline — each with a comment noting the steps are one ordered narrative

gravitee-am-test/playwright/tests/permissions/domain-role-visibility.spec.ts
sagarchauhan-qa: Moved to utils/test-constants.ts — added envHrid() and domainListPath() (so the hardcoded environments/default is gone), plus APPLICATION_OVERVIEW_URL and CREATE_FAB_SELECTOR. No inlined values left in the folder.

gravitee-am-test/specs/management/permissions/permission-sufficiency.jest.spec.ts
sagarchauhan-qa: Extracted to fixtures/permission-sweep-tables.ts. Kept the two sufficiency-only exclusions separate (SUFFICIENCY_ONLY_EXCLUDED) — they only apply when a test grants the permission and expects success, so merging them would have dropped 2 endpoints from the negative sweep. Overrides are shared, and the sweep picked up the 2 it was missing.

gravitee-am-test/specs/management/permissions/write-endpoint-sweep.jest.spec.ts
sagarchauhan-qa: Now 16 of 17. Added a residue hook so membership writes fingerprint the would-be member id rather than the marker name. The exception is the org member case — ORGANIZATION_USER is granted at user creation, so the assignee already holds the exact membership being attempted and asserting on it couldn't fail; commented in place.

gravitee-am-test/specs/management/permissions/fixtures/rbac-fixture.ts
sagarchauhan-qa: Fixed — each persona registers for cleanup as it resolves, via a trackPersona wrapper. Also switched to allSettled so none is still in flight when the catch block starts deleting.

gravitee-am-test/specs/management/permissions/role-scope-confinement.jest.spec.ts
sagarchauhan-qa: All three now read the membership back and assert on memberId and roleId, like the delegated-admin cases. Added a note that a create which silently stopped persisting would otherwise leave this half of the matrix green.

gravitee-am-test/specs/management/permissions/endpoint-permission-sweep.jest.spec.ts
sagarchauhan-qa: Changed at both sites. Left the stillThere check in the deletions block at toBe(200) — that's a single-resource GET where 204 doesn't arise.

gravitee-am-test/specs/management/permissions/protected-resource-tier.jest.spec.ts
sagarchauhan-qa: Done at all five — switched each to filter + toHaveLength(1) + index, matching the shape already used at organization-role-baseline:91.

gravitee-am-test/playwright/tests/permissions/application-role-visibility.spec.ts
sagarchauhan-qa: Made it all-or-nothing — replaced the per-test calls with a file-level test.beforeEach(({}, testInfo) => testInfo.setTimeout(MULTI_PHASE_TEST_TIMEOUT)), applied across all five permissions specs so the folder is consistent.

gravitee-am-test/specs/management/permissions/organization-role-baseline.jest.spec.ts
sagarchauhan-qa: Anchored within the test — it now asserts the group-derived role does grant the domain it was assigned to before asserting the 403 on the other, so it no longer depends on file order.

sclarkamlab (approved):

PR #8395, opened 2026-08-13
test: token claim regression coverage (AM-7488)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7488

Adds 73 Jest integration tests for token claims, in 10 new specs plus a shared fixture under specs/gateway/token-claims/. No production code changed. The existing token-claims.jest.spec.ts and its fixture are untouched.

## What's covered

- Subject identity — sub stays the same across the access token, ID token, userinfo, refresh and re-login, and a sub supplied by an identity provider cannot override it.
- Relying party validation — tokens verified against the published JWKS with issuer and audience enforced, the way a real OIDC client does.
- Custom claims — static and expression language, on both token types and at /oidc/userinfo, released only when the right scope was granted, consistent across all three
  surfaces, and recomputed on refresh.
- Resilience — a mistyped expression or a missing user attribute drops just that claim instead of breaking sign-in.
- Claim disclosure — a client receives only what its scopes asked for (OIDC Core §5.4) or the claims it named (§5.5), and internal values are not forwarded.

## Notes
  
- Green on both MongoDB and PostgreSQL: 76 passing, 3 skipped.
- Three tests in claims-request-parameter are skipped pending AM-7525, a product bug these tests found: the claims request parameter is dropped when the process-global
  JSON mapper has been reconfigured by a /_node call. Assertions are unchanged — removing .skip should be part of that fix, and AM-7525 has a note saying so.

NA

NA


gravitee-am-test/specs/gateway/token-claims/subject-identity.jest.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+describe('Subject identity - a profile attribute cannot impersonate the subject', () => {
+  it('ignores a sub supplied by the identity provider when issuing a token', async () => {
+    const tokens = await fixture.passwordGrantFor(fixture.spoofedUser, 'openid email profile');
+    const accessToken = decodeToken(tokens.access_token).payload;
+    const idToken = decodeToken(tokens.id_token).payload;
+
+    expect(accessToken.sub).not.toEqual(SPOOFED_PROFILE_CLAIMS.sub);
+    expect(idToken.sub).not.toEqual(SPOOFED_PROFILE_CLAIMS.sub);
+    // AM's own derivation still holds for this user.
+    expect(accessToken.sub).toEqual(nameUuidFromBytes(accessToken.gis));
```
sclarkamlab, line 140: The spoofed user plants a `gis` too but nothing checks it survived. This check is self referential, if the planted `gis` had displaced ours then `sub` would still match since it is derived from whatever `gis` ended up in the token. Something like

```ts
expect(accessToken.gis).toEqual(`${fixture.spoofedUser.source}:${fixture.spoofedUser.externalId}`);
```

would close the loop on 7118.

gravitee-am-test/specs/gateway/token-claims/internal-claim-exposure.jest.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+ * a stored value dictate a token's lifetime.
+ *
+ * `ConstantKeys.ID_TOKEN_EXCLUDED_CLAIMS` is the filter under test. Every value in
+ * LEAKY_PROFILE_CLAIMS is planted on `leakyUser` before the domain starts.
+ */
+const UPSTREAM_PROVIDER_TOKENS = ['op_id_token', 'op_access_token'] as const;
+/** Timing claims AM sets itself, so the token carries AM's value rather than the planted one. */
+const AM_GENERATED_TIMING_CLAIMS = ['iat', 'exp', 'auth_time'] as const;
+/** Timing claims AM puts in no ID token at all, planted or otherwise. */
+const TIMING_CLAIMS_NEVER_IN_ID_TOKEN = ['nbf', 'updated_at'] as const;
```
sclarkamlab: Did you mean to call these never in ID token? `updated_at` does show up in the ID token under the `profile` scope, the test at line 137 in this file proves it. They are only stripped on the `full_profile` path, worth renaming so the next reader is not misled.

sclarkamlab (approved):

gravitee-am-test/specs/gateway/token-claims/claims-request-parameter.jest.spec.ts
```diff
@@ -54,8 +54,17 @@ afterAll(async () => {
 
 describe('Claims request parameter - naming claims instead of a scope', () => {
   it('puts the named claims in the ID token without granting the profile scope', async () => {
+    // TEMPORARY DIAGNOSTIC - these three tests pass locally and fail in CI. Compare
+    // what the two flows see so the difference is visible in the CI log, then remove.
+    const viaPasswordGrant = await fixture.passwordGrant('openid full_profile');
+    console.log('[diag] password grant, full_profile ID token:', JSON.stringify(decodeToken(viaPasswordGrant.id_token).payload));
+    console.log('[diag] password grant userinfo:', JSON.stringify(await fixture.userinfo(viaPasswordGrant.access_token)));
+
```
lgw-gravitee: question: will you remove it before merging?

sagarchauhan-qa: Yes, that's temporary logging to work out why these tests fail in CI but pass locally — it'll be removed before merge along with whatever fix comes out of it. I've raised the underlying question in the team channel on Slack, as the logs show the claims parameter returning no user claims in the authorization code flow.

PR #8396, opened 2026-08-13
test: cover token revocation triggers and RFC 7009 endpoint (AM-7450)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7450

Adds 22 Jest tests across 6 new spec files plus 5 unit tests, covering token revocation between the management API and the gateway. Revocation was well covered at the unit layer, but those tests stop at "the revoke callback was
  invoked" against a mock — consent revocation was the only trigger verified end to end.

  - `revoke-token-endpoint` (6) — RFC 7009 `/oauth/revoke`: rejection at userinfo and introspection, `token_type_hint`, cross-client refusal, unknown tokens
  - `revoke-on-application-disable` (6) — disable revokes issued tokens, blocks new ones, leaves other applications untouched
  - `revoke-on-user-status-change` (4) — `BY_USER` reaches every application and is scoped to one user; enabling a user does not revoke
  - `revoke-on-user-lifecycle` (2) — admin password reset and user deletion
  - `revoke-on-scim-disable` (2) — SCIM `active: false`, which revokes via `deleteByUser` rather than the `REVOKE_TOKEN` event bus
  - `revoke-cross-domain-isolation` (2) — no leakage across domains, including when both hold an application with the same `clientId`
  
**Unit tests cover** `token_type_hint` fallback in both directions and cross-client refusal on the refresh-token path (`RevocationServiceTest`), plus hint propagation and unrecognised-hint handling (`RevocationTokenEndpointTest`).
  
  `revocation-fixture.ts` is generalised to support multiple clients and users, optional SCIM, and a shared offline-verification helper. `revoke-user-consents.jest.spec.ts` is untouched and still passes.

## :memo: Test scenarios 
NA

NA

NA

### DIFF
--- a/gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
+++ b/gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
@@ -36,13 +36,17 @@ import {
   performPatch,
   performPost,
 } from '@gateway-commands/oauth-oidc-commands';
-import { withRetry } from '@utils-commands/retry';
+import { retryUntil, withRetry } from '@utils-commands/retry';
+import { waitForSyncAfter } from '@gateway-commands/monitoring-commands';
 import { applicationBase64Token } from '@gateway-commands/utils';
 import { uniqueName } from '@utils-commands/misc';
 import { decodeJwt } from '@utils-commands/jwt';
 import { JWT_FORMAT } from '@specs-utils/jwt-format';
 import { Domain } from '@management-models/Domain';
 import { Application } from '@management-models/Application';
+// Type-only: jest's moduleNameMapper has no @management-models entry, so a value import from here
+// resolves under tsc but blows up at runtime. Keep every model import in a type position.
+import type { NewApplication } from '@management-models/NewApplication';
 import { IdentityProvider } from '@management-models/IdentityProvider';
 import { User } from '@management-models/User';
 import request from 'supertest';
@@ -122,6 +126,14 @@ export interface RevocationFixture {
   createAdditionalUser: (index: number) => Promise<User>;
   /** Run the authorization_code flow as an arbitrary user, defaulting to the fixture's primary client. */
   obtainAuthorizationCodeTokensAs: (asUser: User, client?: RevocationClient) => Promise<SubjectTokens>;
+  /**
+   * Poll until the gateway rejects this token at the userinfo endpoint. Use instead of a fixed
+   * wait when asserting a token has been revoked — it tolerates slow propagation and still fails
+   * honestly, because a token that was never revoked keeps returning 200 until the timeout.
+   */
+  waitUntilTokenRejected: (token: string) => Promise<void>;
+  /** As `waitUntilTokenRejected`, but polls the introspection endpoint for `active: false`. */
+  waitUntilTokenInactive: (token: string) => Promise<void>;
   /** Present only when the fixture was built with `withScim: true`. */
   scim?: {
     endpoint: string;
@@ -131,20 +143,35 @@ export interface RevocationFixture {
   };
 }
 
-/** Default of `handlers.oauth2.introspect.offlineVerificationTimerSeconds`, plus a second of slack. */
-const OFFLINE_VERIFICATION_WINDOW_MS = 11000;
+/**
+ * `handlers.oauth2.introspect.offlineVerificationTimerSeconds`. Within this many seconds of a
+ * token being issued the gateway trusts the JWT and never looks it up, so a revoked token still
+ * works. The stack these tests run against disables it
+ * (`GRAVITEE_HANDLERS_OAUTH2_INTROSPECT_OFFLINEVERIFICATIONTIMERSECONDS=0` in
+ * docker/local-stack/dev/docker-compose.yml, which the CI compose layers), hence the default of 0
+ * here. Override to match if you point the suite at a stack that leaves the product default of 10.
+ */
+const OFFLINE_VERIFICATION_SECONDS = Number(process.env.AM_OFFLINE_VERIFICATION_SECONDS ?? 0);
+
+/** Headroom for the revoke event to reach the gateway and the delete to land. */
+const REVOKE_PROPAGATION_SETTLE_MS = 2000;
+
+/** How long to keep polling before deciding a token was never revoked. */
+const REVOCATION_POLL_TIMEOUT_MS = 20000;
 
 /**
- * Block until the gateway will consult the token store for this token rather than trusting the
- * JWT alone. Deliberately a wait and not a poll: the point is to reach a state where a single
- * check is meaningful, and polling for a status would mask a token that is still valid.
+ * Wait until the gateway will consult the token store for this token rather than trusting the JWT.
+ *
+ * Only needed before asserting a token is STILL VALID. Inside the offline-verification window such
+ * an assertion is meaningless — the JWT is trusted, so it passes even if the token has been revoked.
+ * To assert a token HAS been revoked, poll with `waitUntilTokenRejected` / `waitUntilTokenInactive`
+ * instead: polling for the revoked state cannot produce a false pass, because a token that was
+ * never revoked simply stays valid until the timeout.
  */
 export const waitPastOfflineVerification = async (token: string): Promise<void> => {
   const { iat } = decodeJwt(token) as { iat: number };
-  const remainingMs = iat * 1000 + OFFLINE_VERIFICATION_WINDOW_MS - Date.now();
-  if (remainingMs > 0) {
-    await new Promise((resolve) => setTimeout(resolve, remainingMs));
-  }
+  const remainingWindowMs = Math.max(0, (iat + OFFLINE_VERIFICATION_SECONDS) * 1000 - Date.now());
+  await new Promise((resolve) => setTimeout(resolve, remainingWindowMs + REVOKE_PROPAGATION_SETTLE_MS));
 };
 
 const REVOCATION_TEST = {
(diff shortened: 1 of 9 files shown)


gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
```diff
@@ -238,11 +437,64 @@ export const setupRevocationFixture = async (): Promise<RevocationFixture> => {
...
-      return exchangedToken;
+      expect(response.body.access_token).toMatch(JWT_FORMAT);
+      return response.body.access_token;
+    };
+
+    const createAdditionalClient = async (
+      namePrefix: string = REVOCATION_TEST.SECONDARY_CLIENT_NAME,
+      clientOptions: { exactName?: boolean } = {},
+    ): Promise<RevocationClient> => {
+      const additionalApplication = await createRevocationApp(
```
sclarkamlab: No `waitForSyncAfter` here and both callers run the auth code flow immediately after. `createTestApp` doesn't wait either, so this can race the gateway sync.

gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
```diff
@@ -54,40 +65,110 @@ export interface SubjectTokens {
...
+  /** Present only when the fixture was built with `withScim: true`. */
+  scim?: {
+    endpoint: string;
+    accessToken: string;
+    /** PATCH a user's `active` flag. Returns the raw response so tests can assert on it. */
+    setUserActive: (userId: string, active: boolean) => Promise<any>;
+  };
 }
 
+/** Default of `handlers.oauth2.introspect.offlineVerificationTimerSeconds`, plus a second of slack. */
```
sclarkamlab: Once this is configurable, `revoke-on-application-disable.jest.spec.ts:160` needs a wait before it asserts 200. It's fine at 0 because the store is always checked, but on a default stack it'd pass even if the token had been revoked.

gravitee-am-test/specs/gateway/revocation/revoke-cross-domain-isolation.jest.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    // The premise of the test: one clientId, two domains.
+    expect(clientB.application.settings.oauth.clientId).toEqual(clientA.application.settings.oauth.clientId);
+
+    const tokensA = await clientA.obtainAuthorizationCodeTokens();
+    const tokensB = await clientB.obtainAuthorizationCodeTokens();
+
+    expect((await domainA.getUserInfo(tokensA.accessToken)).status).toBe(200);
+    expect((await domainB.getUserInfo(tokensB.accessToken)).status).toBe(200);
+
+    await patchApplication(domainA.domain.id, domainA.accessToken, { enabled: false }, clientA.application.id);
```
sclarkamlab, line 99: Any reason this isn't wrapped in `waitForSyncAfter` like `revoke-on-application-disable.jest.spec.ts:60`? Becomes a real race once the offline wait below goes, there's no margin left at that point.

gravitee-am-test/specs/gateway/revocation/revoke-cross-domain-isolation.jest.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (domainA) {
+    await domainA.cleanup();
+  }
+  if (domainB) {
+    await domainB.cleanup();
+  }
+});
+
+describe('Revocation isolation - across domains', () => {
+  it('should not revoke tokens in another domain when a user is disabled', async () => {
```
sclarkamlab, line 63: Worth a note that this one can't really fail, user ids are globally unique so even `deleteByUserId` with no domain predicate would pass it. The BY_CLIENT one below is the one with teeth.

gravitee-am-test/specs/gateway/revocation/revoke-on-user-status-change.jest.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    await updateUserStatus(fixture.domain.id, fixture.accessToken, fixture.user.id, true);
+
+    await waitPastOfflineVerification(primaryTokens.accessToken);
+    await waitPastOfflineVerification(secondaryTokens.accessToken);
+
+    expect((await fixture.getUserInfo(primaryTokens.accessToken)).status).toBe(401);
+    expect((await fixture.getUserInfo(secondaryTokens.accessToken)).status).toBe(401);
+  });
+
+  it('should let the re-enabled user obtain a working token again', async () => {
```
sclarkamlab: This doesn't disable or re-enable anyone, it just mints a fresh token, so the name is a bit of a stretch. Passes standalone either way.

gravitee-am-test/specs/gateway/revocation/revoke-on-user-lifecycle.jest.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { deleteUser, resetUserPassword } from '@management-commands/user-management-commands';
+import { setup } from '../../test-fixture';
+import { setupRevocationFixture, waitPastOfflineVerification, RevocationFixture } from './fixtures/revocation-fixture';
+
+setup(200000);
+
+/**
+ * Admin lifecycle actions that revoke tokens, separate from the enable/disable status changes in
+ * revoke-on-user-status-change. Both emit BY_USER (ManagementUserServiceImpl lines 486 and 838).
```
sclarkamlab: `lines 486 and 838` will rot, worth naming the methods instead.

sclarkamlab (commented):
Really like the shape of these, re-enabling the user/app before asserting is the bit most revocation tests get wrong and you've done it everywhere and written down why.

One thing I want to sort before this goes in though. The `waitPastOfflineVerification` waits are no-ops against the stack we actually test on. `docker/local-stack/dev/docker-compose.yml:53` already sets `GRAVITEE_HANDLERS_OAUTH2_INTROSPECT_OFFLINEVERIFICATIONTIMERSECONDS=0`, and the CI stack layers that base compose, so the gateway hits the token store on every call. `OFFLINE_VERIFICATION_WINDOW_MS` is pinned to the prod default of 10s instead. That's ~21 sleeps at ~10s, so about 3.5 mins of CI sitting there for nothing, and we're trying to pull pipeline time down not push it up.

Two changes should do it. Read the value rather than hard coding it:

```ts
const OFFLINE_VERIFICATION_SECONDS = Number(process.env.AM_OFFLINE_VERIFICATION_SECONDS ?? 0);
```

and for the "should be revoked" assertions poll rather than sleep, same as you already do in `revoke-on-application-disable.jest.spec.ts:81`:

```ts
await retryUntil(() => fixture.getUserInfo(t.accessToken).then((r) => r.status), (s) => s === 401,
  { timeoutMillis: 20000, intervalMillis: 250 });
```

The negative controls still need a short settle, but that only has to cover the revoke event propagating, so a couple of seconds rather than eleven. Should land around 20s total and it still passes if anyone points the suite at a stack running the default window.

Worth updating the file header comments that cite the 10s default at the same time, otherwise someone puts the sleeps back in six months.

Separately and not for this PR, `introspection-resource-indicators.jest.spec.ts:70` has a `delay(10000)` for the same already-disabled window, and the test at :114 says it exercises the offline path which it hasn't been doing. Another 10s for free if someone picks it up.

For transparency I haven't run the suite, the ~3.5 mins is arithmetic off the config rather than a measurement.

sagarchauhan-qa:
Thanks, you're right. Checked it and the gateway does have it set to 0 - it's in the compose file and I confirmed it on the running container - so those waits weren't doing anything.

Changed:
- reads the value from env now rather than being hardcoded, defaults to 0
- the "should be revoked" checks poll instead of sleeping
- the "still valid" checks keep a short wait, so they'd still hold up against a stack running the default window

Folder is down from 199s to 69s, all 27 still passing.

Also picked up your fixture.ts:134 one while I was in there - the blast radius test was asserting 200 with no wait at all, so it would have passed even if the token had been revoked. Same root cause so I did it in this change.

Raised [AM-7549](https://gravitee.atlassian.net/browse/AM-7549) for the introspection-resource-indicators one.

[AM-7549]: https://gravitee.atlassian.net/browse/AM-7549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
sagarchauhan-qa: Fixed. Wrapped the create in waitForSyncAfter so the client is on the gateway before it gets handed back.
  
Might also explain a one-off flake I saw early on in "should still issue tokens to another application in the domain" - never reproduced it, but an unsynced client fits.

This one got more important after the other change - the old 11s sleeps were hiding any sync lag, and they're down to ~2s now.

Still 27/27, folder now 73s.

gravitee-am-test/specs/gateway/revocation/fixtures/revocation-fixture.ts
sagarchauhan-qa: Done - added the wait before the 200 assertion.
  
Went through the other "still valid" checks as well since they had the same hole: the re-enabled token in this spec, the untouched user in revoke-on-user-status-change, both surviving tokens in the cross-domain one, and the cross-client no-op in revoke-token-endpoint. Those all clear the window now, and only the "should be revoked" ones poll.

sclarkamlab (approved):

gravitee-am-test/specs/gateway/revocation/revoke-cross-domain-isolation.jest.spec.ts
sagarchauhan-qa: The reason is that disabling an app doesn't change the domain's lastSync value, which is the thing waitForSyncAfter is waiting on. So it waits for something that never happens. Checked and the value didn't move for 24 seconds after the patch, even though the domain was showing as synced.

The one you pointed at in revoke-on-application-disable was only working by accident - creating the client was moving that value for it. Once the create started waiting properly (your other comment), that one started timing out too. So I've taken it out of both files.

What it does instead: disable the app, then keep checking the token until it actually stops working. That's the wait now. It can't finish early, and if the token never gets revoked the test fails rather than passing quietly.

One thing worth knowing for the re-enable test - there needs to be a gap between disabling and re-enabling. With them back to back, the revoke arrived late and killed the new token as well as the old one.

27/27 passing, 72s.

gravitee-am-test/specs/gateway/revocation/revoke-cross-domain-isolation.jest.spec.ts
sagarchauhan-qa: Added a note above it saying it can't really fail and why - the two users have different ids, so even an unscoped delete would only ever match the one in domain A.

Kept it for the pairing with the BY_CLIENT one rather than deleting it, but it's labelled now so nobody reads it as real protection.

gravitee-am-test/specs/gateway/revocation/revoke-on-user-status-change.jest.spec.ts
sagarchauhan-qa: Thanks for flagging. It wasn't doing what the name said. Rather than just renaming it, I've made it disable and re-enable itself, so it now proves that re-enabling a user actually gets them working again.

It uses its own user too, so it doesn't lean on the test above it any more.

gravitee-am-test/specs/gateway/revocation/revoke-on-user-lifecycle.jest.spec.ts
sagarchauhan-qa: Changed it to name the methods - ManagementUserServiceImpl#resetPassword and ManagementUserServiceImpl#delete.

PR #8397, opened 2026-08-14
fix: verify a provisioned data plane answers before binding a domain to it

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7489

A provisioned data plane now has to answer before a domain can be bound to it. Building a provider proved nothing on its own, both implementations hand back a client that only connects on its first operation, so a wrong password or an unreachable host registered like any other definition and domain creation accepted it.

DataPlaneProvider gains a healthCheck that does a real round trip, a ping for mongo and a select for jdbc. Both drivers authenticate while they open the connection, so bad credentials surface there.

The check itself lives in DataPlaneVerifier. It runs asynchronously when a data plane is provisioned, and domain creation waits on the same run rather than starting its own, so the common path is a map lookup and the first caller after startup joins a check already in flight.

- Only provisioned data planes are checked. The ones the gravitee.yml declares are the node's own configuration and are served without one.
- A refusal is remembered for a short window, so a store that is down costs one connection attempt per window instead of one per caller.
- Once that window passes the id is checked again, so correcting a definition does not need a restart.
- dataPlaneVerification in the gravitee.yml turns the whole thing off, or changes the timeout and that window.

The verifier is told which ids to check by the loader, not by the registry, because the registry registers the yml definitions through the same call. The sync event path does the same, otherwise only the node that handled the provisioning request would check anything.

## :memo: Test scenarios

Unit and e2e coverage in the branch, all green.

No UI changes.

A check that passes is not repeated, so a data plane that breaks after it was verified still accepts domains. Worth a follow up if we want the check to mean anything ongoing.

That is also why dataplane-unreachable.jest.spec.ts is deleted here. It bound a domain to a broken data plane and asserted the domain could still be deleted, which is a real production case, a plane that was healthy at creation and broke later. There is no longer a way to reach that state through the API, binding is refused and the delete guard stops you breaking a plane after the fact.

@gravitee-io/am


gravitee-am-dataplane/gravitee-am-dataplane-mongodb/src/main/java/io/gravitee/am/dataplane/mongodb/provider/MongoDataPlaneProvider.java
```diff
@@ -96,6 +99,15 @@ public class MongoDataPlaneProvider implements DataPlaneProvider, InitializingBe
     @Qualifier("dataplaneUpgraderRepository")
     private UpgraderRepository upgraderRepository;
 
+    /**
+     * The database the repositories were given, which already holds the wrapper's single reference.
+     * Taking another one here would leave {@link #stop()} one release short of the count the wrapper
+     * closes its client on, so an unregistered data plane would keep its connection pool open.
+     */
```
wajb: nitpick: wiring the instance seems conventional. I'm not sure the docs here provide anything meaningful tbh.

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/provider/JdbcDataPlaneProvider.java
```diff
@@ -107,6 +110,15 @@ public void stop() {
 
     }
 
+    @Override
+    public Completable healthCheck() {
+        return Single.fromPublisher(clientWrapper.create())
+                .flatMapCompletable(connection -> Flowable.fromPublisher(connection.createStatement("SELECT 1").execute())
+                        .flatMap(result -> Flowable.fromPublisher(result.map((row, metadata) -> 1)))
+                        .ignoreElements()
+                        .doFinally(() -> Completable.fromPublisher(connection.close()).onErrorComplete().subscribe()));
```
wajb: suggestion: this query does not necessarily work for all dialects (like the rdbc `validationQuery`) and close errors are dropped. What about leveraging `Connection.validate()`?

```suggestion
    return Completable.fromPublisher(
            Mono.usingWhen(
                    clientWrapper.create(),
                    connection -> Mono.from(connection.validate(ValidationDepth.REMOTE))
                            .filter(Boolean.TRUE::equals)
                            .switchIfEmpty(Mono.error(() -> new IllegalStateException("R2DBC connection validation failed"))),
                    Connection::close));
```

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlaneRegistry.java
```diff
@@ -49,6 +50,12 @@ public interface DataPlaneRegistry {
 
     void unregister(String dataPlaneId);
 
+    void reserveVerification(String dataPlaneId);
+
+    void requireVerification(String dataPlaneId);
```
wajb: thought: `DataPlaneVerifierImpl` makes sense as a collaborator but its complexity is leaking out here. Could reserve/require states be handled by the registry in its `register` impl?

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlaneVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+import io.reactivex.rxjava3.core.Completable;
+
+/**
+ * Decides whether a data plane's store answers with the settings it was provisioned with. Building a
+ * provider proves nothing on its own: both implementations hand back a client that connects on its
+ * first operation, so a wrong password or an unreachable host is only found by asking the store.
+ *
+ * @author GraviteeSource Team
+ */
+public interface DataPlaneVerifier {
```
wajb, line 28: suggestion: this is never mocked and only has one impl, so it could probably be dropped to reduce surface area.

gravitee-am-test/specs/management/internal-api/dataplane-invalid-binding.jest.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+  provisionUnreachableDataPlane,
+  provisionWrongCredentialsDataPlane,
+} from './fixtures/dataplane-provisioning-fixture';
+import { setup } from '../../test-fixture';
+import { uniqueName } from '@utils-commands/misc';
+
+// each rejected bind waits out the verification timeout before it answers
+setup(60000);
+
+/** AM-7489: a data plane whose store does not answer must not be bindable. */
```
wajb: nitpick: avoid internal references when not test case IDs or TODOs.

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlaneRegistry.java
sclarkamlab: Yeah, done in dc3d0b7298. Couldn't fold it into `register` itself because the gravitee.yml planes go through that same method and have to stay exempt, so there's a `registerProvisioned` next to it that does reserve, register, require, and cleans up its own claim if the provider can't be built. Both callers lose the handshake and the duplicated comments that went with it.

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/DataPlaneVerifier.java
sclarkamlab: Went the other way on this one, f24a8115ee. The gateway was already a second impl wearing a boolean, `new DataPlaneVerifierImpl(false, 0, 0)`, so I added a `NoOpDataPlaneVerifier` and deleted the `enabled` flag and its two guards instead. Management picks the type off the property now rather than passing it in.

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/java/io/gravitee/am/dataplane/jdbc/provider/JdbcDataPlaneProvider.java
sclarkamlab: Good catch, applied as is in e0d521f5fc. The close was the worse half of it, that `doFinally` was detaching the close from the returned Completable so the check completed before the connection went back to the pool. Dialect side isn't live today, the driver enum is postgres/mysql/mariadb/sqlserver, but no reason to bank on that staying true.

gravitee-am-dataplane/gravitee-am-dataplane-mongodb/src/main/java/io/gravitee/am/dataplane/mongodb/provider/MongoDataPlaneProvider.java
sclarkamlab: Done in c144312a60.

gravitee-am-test/specs/management/internal-api/dataplane-invalid-binding.jest.spec.ts
sclarkamlab: Done in ba38518ea1.

wajb (approved):

PR #8403, opened 2026-08-14
feat: reject cloud environments without a gateway access point and drop the org default entrypoint

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7354

Follow up from AM-7225, under the AM-7165 epic.

All changes apply to cloud mode only.

- Reject an ENVIRONMENT command that has no usable gateway access point, and reply ERROR to Cockpit. This covers three cases: accessPoints is null or empty, there is no GATEWAY access point, and every GATEWAY access point is overriding. Each case has its own message. The environment is not created or updated when the command is rejected.
- Do not create the org level "Default" entrypoint. Everything in cloud resolves from the environment access points.
- Skip the last default entrypoint guard in `syncEntrypoints`. The org default was satisfying that guard, so without it a single environment org failed on every ENVIRONMENT command after the first.
- `listEnvironmentEntryPoint` builds an entrypoint from the data plane gateway url, falling back to `gateway.url`, when the org lookup returns nothing. The org lookup still runs first, so existing installations are unchanged. The warn for a missing org default is now a debug line.
- Guard a `get(0)` on a possibly empty list in the non cloud entrypoint path.

There is no upgrader. An existing org default entrypoint stays in place and is never resolved.

For the Cockpit integration: Cockpit must always send the generated non overriding access point with the environment.

## :memo: Test scenarios

- `EnvironmentCommandHandlerTest` — null accessPoints, empty list, CONSOLE only and all overriding each reject with no call to the environment service. Two tests pin that none of this applies outside cloud mode.
- `EntrypointServiceTest` — `createDefaults` creates nothing in cloud, with and without domain restrictions, plus the resync sequence that needs the guard change.
- `DomainServiceTest` — replaces the empty list test with one for the data plane gateway url and one for the `gateway.url` fallback.
- Jest `specs/cloud` — new rejection cases in command-validation, and a new organization-default-entrypoint spec asserting a fresh cloud org has no org level entrypoint.
- `setupCloudOrganizationFixture` now sends a gateway access point with its home environment so the command is valid.

Not yet run against a cloud stack.

n/a

The org settings entrypoints screen lists everything in the org unfiltered, so in cloud it now shows only environment rows.

@gravitee-io/am


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -277,6 +277,9 @@ public class DomainServiceImpl implements DomainService {
     @Value("${domains.identities.default.enabled:true}")
     private boolean createDefaultIdentityProvider = true;
 
+    @Value("${gateway.url:http://localhost:8092}")
+    private String gatewayUrl = "http://localhost:8092";
```
wajb: nitpick: the single source of truth for the default can be in the spring annotation so we can drop the duplicate.
```suggestion
    private String gatewayUrl;
```

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -824,7 +830,22 @@ private Maybe<Entrypoint> getOrganizationDefaultEntrypoint(Domain domain, String
...
                 })
-                .doOnComplete(() -> log.warn("Organization {} has no default entrypoint; domain {} resolves to no entrypoint at all", organizationId, domain.getId()));
+                .doOnComplete(() -> log.debug("Organization {} has no default entrypoint, domain {} falls back to the data plane gateway url", organizationId, domain.getId()));
+    }
+
+    /**
+     * Last resort when neither the environment nor the organization has an entrypoint, which is the normal
+     * state for a cloud organization: {@code createDefaults} no longer gives those an org-level default.
+     * Callers dereference a url, so this must never be empty.
+     */
```
wajb: thought: this is a little misleading since the new normal state is to ensure there is an environment entrypoint when in cloud mode. Is this really only reachable in cloud mode in edge cases during sync windows?

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/commands/EnvironmentCommandHandlerTest.java
```diff
@@ -404,17 +405,85 @@ void handleCloudMode_nullAccessPointEntry_repliesWithoutThrowing() {
...
         // The guard runs before the reactive chain is built, so a null entry would escape handle() as a
         // synchronous throw rather than reaching onErrorReturn.
-        EnvironmentCommand command = commandWithAccessPoints(new AccessPoint[]{null});
+        assertRejectedInCloudMode(commandWithAccessPoints(new AccessPoint[]{null}));
+    }
+
+    private void assertRejectedInCloudMode(EnvironmentCommand command) {
+        TestObserver<EnvironmentReply> obs = cut.handle(command).test();
+
+        obs.awaitDone(10, TimeUnit.SECONDS);
```
wajb: minor: could this value be a constant at least for better descriptiveness? The magic number is used without explanation (existing issue).

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/DomainServiceTest.java
```diff
@@ -2081,18 +2081,40 @@ public void shouldGetEntrypoint_cloud_noEnvironmentEntrypoint_nullGatewayUrl_fal
...
+        final Domain mockDomain = cloudDomain();
+        when(entryPointManager.findByEnvironmentId(ENVIRONMENT_ID)).thenReturn(List.of());
+        doReturn(Flowable.empty()).when(entrypointService).findAll(ORGANIZATION_ID);
+        when(dataPlaneRegistry.getDescription(mockDomain))
+                .thenReturn(new DataPlaneDescription("dp1", "legacy", "mongodb", "baseProp", null));
+
+        final var subscriber = domainService.listEntryPoint(mockDomain, ORGANIZATION_ID).test();
+
+        // Never an empty list, whatever the data plane says.
+        subscriber.assertValue(entrypoints -> entrypoints.size() == 1
```
wajb, line 2122: thought: conventionally we would define `gatewayUrl` as a ctor parameter so we can assert the given value. This works to assert the default though so NBD.

gravitee-am-test/specs/cloud/organization-default-entrypoint.jest.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+import { retryUntil } from '@utils-commands/retry';
+import { CloudOrganizationFixture, setupCloudOrganizationFixture } from './fixtures/cloud-organization-fixture';
+import { setup } from '../test-fixture';
+
+setup(120000);
+
+const POLL = { timeoutMillis: 30000, intervalMillis: 1000 };
+
+// Fixed id, as the organization fixture explains. It must stay unused by other specs: createDefaults runs
+// once at creation, so an organization predating AM-7354 still has the default entrypoint and would fail here.
```
wajb: minor: nothing is persisted between tests so this isn't an issue?

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: Done in `ea3fc61551`. The initializer was only there so the unit test saw a value, the test sets the field itself now.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: Good catch, that wording is stale. It is the sync window only: cloud environments each carry an entrypoint now, and the command is rejected if the access point is missing, so this is only reachable before Cockpit has synced. Reworded in `ea3fc61551`.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/commands/EnvironmentCommandHandlerTest.java
sclarkamlab: Done in `ea3fc61551`, it is `HANDLE_TIMEOUT_SECONDS` across all 15 call sites in the file.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/DomainServiceTest.java
sclarkamlab: Went the other way in `ea3fc61551`. Your other comment killed the field default, so the test now sets `gatewayUrl` explicitly and asserts that value.

gravitee-am-test/specs/cloud/organization-default-entrypoint.jest.spec.ts
sclarkamlab: It is, unfortunately. Organizations survive the run, AM has no way to delete one, which is why the fixture uses a fixed id and upserts. So if that id was ever created by a build predating this change it still has the org level default entrypoint and this spec fails. Comment reworded in `ea5e1c2fb6` to actually say that.

wajb: Thanks, this is fine for now. The org deletion inability is a gap I've raised as https://gravitee.atlassian.net/browse/AM-7519.

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/DomainServiceTest.java
```diff
@@ -2081,18 +2081,38 @@ public void shouldGetEntrypoint_cloud_noEnvironmentEntrypoint_nullGatewayUrl_fal
...
-        subscriber.assertValue(List::isEmpty);
+        subscriber.assertValue(entrypoints -> entrypoints.size() == 1
+                && "http://gateway:8092".equals(entrypoints.get(0).getUrl())
+                && entrypoints.get(0).isDefaultEntrypoint());
+    }
+
+    @Test
+    public void shouldGetEntrypoint_cloud_noEnvironmentEntrypoint_noOrganizationDefault_noDataPlaneUrl_fallsBackToConfiguredGatewayUrl() {
+        enableCloudMode();
+        ReflectionTestUtils.setField(domainService, "gatewayUrl", "http://configured-gateway:8092");
```
wajb: minor: could the field be package-private instead?

wajb (approved):

gravitee-am-management-api/gravitee-am-management-api-service/src/test/java/io/gravitee/am/management/service/impl/DomainServiceTest.java
sclarkamlab: Yeah, done in `2002f49cba`. Field is package-private now and the test just assigns it, so the ReflectionTestUtils call is gone.

PR #8406, opened 2026-08-14
fix: avoid NPE on list endpoints when a scoped role carries no matching permission

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7475

The domain, application and protected resource lists return the entries the caller may see when the caller's role carries no permission in the family being filtered on. They returned 500 before.

`PermissionService.getReferenceIdsWithPermission` calls `containsAll` on the permission set it reads out of the membership map. The map holds only the permissions the role carries, so a scoped role that grants nothing in that family gives back null. An absent permission now grants nothing, and the method excludes that membership.

A domain scoped role that grants `application_list` and no domain permission is enough to hit this. The domain list is the Console landing screen, so an affected user cannot use the product at all.

Also in here: the protected resource endpoints declare `type` as the enum it is and let JAX-RS convert it. `RolesResource` and the form and email resources already do this. `list`, `get` and `delete` each parsed the parameter by hand, before the permission check ran, so a missing `type` returned 500 for every caller.

- A missing `type` returns 400, from the `@NotNull` on the parameter.
- An unknown `type` returns 404, because JAX-RS returns 404 for a failed query parameter conversion.
- Neither case returns 500.

The spec is regenerated. `type` is now required, with an `MCP_SERVER` enum on all three operations. oasdiff reports this as breaking, and the version boundary exempts it. A call that omits the parameter already failed, so no working client depends on the parameter being optional.

The conversion is now exact, where `Type.fromString` upper cased the value first. The Console and the test SDK both send `MCP_SERVER`, so nothing depends on the lenient casing.

## :memo: Test scenarios

Unit coverage in the branch, all green.

No UI changes.

No jest regression spec here. The coverage for three tier list scoping belongs with the permissions automation in AM-7448.

@gravitee-io/am


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ProtectedResourceResource.java
```diff
@@ -91,12 +90,10 @@ public void get(
             @PathParam("environmentId") String environmentId,
             @PathParam("domain") String domainId,
             @PathParam("protected-resource") String protectedResourceId,
-            @QueryParam("type") String type,
+            @NotNull @QueryParam("type") ProtectedResource.Type type,
```
wajb, line 93: thought: IMO it's weird to have this type filtering on an explicit GET. We already know the type from the list request and guarding for MCP view is really a UI concern.

Making it required is correct for the implementation, but it exposes this weirdness.

wajb (approved):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ProtectedResourceResource.java
sclarkamlab: I guess it is weird because at this time there is only one type of Protected Resource `MCP_SERVER` but I believe there was a vision of it extending to other Types.

wajb: The weirdness is more that the endpoint requires the type parameter at all. Protected resources are unique by id, the value of which is already provided. It's an existing design decision we could revisit but the spec fix is correct as it stands.

PR #8407, opened 2026-08-14
test: cover database index lifecycle and purge scheduling (AM-7456)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7456

Adds 40 tests covering MongoDB index creation and the scheduled purge, the storage-lifecycle
    layer behind the AM-7457 / AM-7454 P1.

  - OAuth2TokenStoreIndexTest (5) — TTL index on all 7 expiring token stores, plus the 17
  lookup indexes behind token read and revocation
  - ManagementStoreIndexTest (4) — TTL and unique indexes on cimd_metadata_documents and
  cp_action_lease, events pinned to its configured retention rather than just "a TTL exists"
  - EnsureIndexOnStartResolutionTest (12) — which property switches index creation on and off,
  for the management, oauth2 and gateway scopes
  - DataPlaneEnsureIndexOnStartResolutionTest (6) — the same for the data plane scope
  - MongoIndexLifecycleTest (6) — restart idempotency, and MongoUtils.dropIndexes selective
  drop, concurrent removal, missing collection
  - ScheduledPurgeServiceTest (5) — purge target selection and cron scheduling
  - JdbcAuditReporterIndexTest (2) — the idx_audits_timestamp index that shipped untested in
  572bc8b0d1, asserted on PostgreSQL, MySQL, MariaDB and SQL Server

Across ~67 createIndex call sites and 16 TTL indexes, exactly one test asserted a declared index actually existed. That gap is expensive because on MongoDB the TTL index is the only thing deleting expired tokens — the scheduled purge is JDBC-only — and MongoUtils.createIndex only logs failures, so a node with no indexes starts up healthy. Not a security issue: every read path filters on expiry, so the impact is storage and query performance.

Three configuration results are pinned as tests because they are surprising: repositories.oauth2.mongodb.ensureIndexOnStart is ignored by the token stores; the legacy oauth2. key silently disables three scopes at once; and the data plane falls back to the oauth2 keys where the gateway falls back to the management ones, so repositories.management.mongodb.ensureIndexOnStart disables gateway indexes but not data plane ones.

The JDBC reporter test asserts the index's leading column rather than its name: reporter_audits already carries composite indexes containing timestamp as a trailing column, and the name is not portable — PostgreSQL, MariaDB and SQL Server declare it as idx_audits_timestamp___ while MySQL declares it inline and unnamed, so MySQL auto-generates the name timestamp.

## :memo: Test scenarios 
NA

NA

postgresql.schema declares idx_audits_timestamp___ twice — harmless under IF NOT EXISTS, looks like an accidental double-add. Left alone.


gravitee-am-reporter/gravitee-am-reporter-jdbc/src/test/java/io/gravitee/am/reporter/jdbc/audit/PostgresAuditReporterIndexTest.java
```diff
@@ -0,0 +1,139 @@
...
+ * {@code src/main/resources/database/}, one file per database, each written differently. Nothing
+ * checks that a given index made it into every vendor's file, so an index added to three of the
+ * four goes unnoticed.
+ * <p>
+ * The leading column is what is asserted, not merely that a column is indexed somewhere:
+ * {@code reporter_audits} already carries composite indexes that include {@code timestamp} as a
+ * trailing column, so "timestamp appears in some index" would hold even with the dedicated
+ * timestamp index missing - which is the index the audit purge relies on to find expired rows.
+ * <p>
+ * PostgreSQL only. Index introspection has no portable form - MySQL and MariaDB need
```
wajb: issue: IMO this is only worthwhile if it runs against all vendors (like `JdbcAuditReporterTest`). If you switch the query on the db type and drop the column name in the assertion (only leading column matters for all vendors) then it should be doable.

gravitee-am-repository/gravitee-am-repository-mongodb/src/test/java/io/gravitee/am/repository/mongodb/common/EnsureIndexOnStartResolutionTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Test
+    public void tokenStoreIndexesAreDisabledByTheLegacyOauth2Key() {
+        assertThat(oauth2Resolves(env(LEGACY_OAUTH2_KEY, "false"))).isFalse();
+    }
+
+    /**
+     * The token stores read the legacy key only. Setting the modern, scope-prefixed key that every
+     * other repository honours leaves index creation switched on - including the TTL indexes that
+     * are the only thing deleting expired tokens on MongoDB. An operator who standardises on the
+     * {@code repositories.*} naming will believe they have disabled index creation when they have not.
```
wajb, line 102: nitpick: we must use US English - "honors", "standardizes" etc.

gravitee-am-repository/gravitee-am-repository-mongodb/src/test/java/io/gravitee/am/repository/mongodb/common/EnsureIndexOnStartResolutionTest.java
```diff
@@ -0,0 +1,176 @@
...
+ * Pins which configuration keys switch MongoDB index creation on and off, per repository scope.
+ * <p>
+ * Index creation never fails loudly - {@code MongoUtils.createIndex} subscribes without blocking and
+ * only logs - so an operator who disables it by accident gets a healthy-looking node whose token
+ * collections silently stop self-cleaning. The keys are worth pinning because they are not
+ * consistent between scopes: each scope resolves a different chain, and two of the combinations
+ * asserted below are surprising enough that they have caused production incidents.
+ * <p>
+ * The {@code @Value} expressions are read off the production fields by reflection rather than
+ * copied, so this test cannot drift away from the annotations it describes.
```
wajb: thought: what about adapting the oauth2 and management repos to use the gateway repo approach of a test subclass that exposes `getEnsureIndexOnStart`? This would enable us to drop annotation scraping and reflection.

Are we missing coverage of `AbstractDataPlaneMongoRepository` too?

gravitee-am-service/src/test/java/io/gravitee/am/service/purge/ScheduledPurgeServiceTest.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Covers how the scheduled purge decides what to clean and whether to run at all.
+ * <p>
+ * On JDBC deployments this service is the only mechanism removing expired records - the TTL-index
+ * mechanism is MongoDB-only - so its target selection and scheduling are worth pinning.
+ *
+ * @author GraviteeSource Team
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class ScheduledPurgeServiceTest {
```
wajb, line 51: minor: some test names are a little misleading since they describe purging, but it's really the config/scheduling that's covered.

gravitee-am-repository/gravitee-am-repository-mongodb/src/test/java/io/gravitee/am/repository/mongodb/common/EnsureIndexOnStartResolutionTest.java
sagarchauhan-qa: Done - annotation scraping and reflection are both gone.

  Couldn't expose getEnsureIndexOnStart on oauth2/management as they have no such method, just a @Value field (private on oauth2). So the probes drive the real createIndex(...) against a
  mock collection and assert whether createIndexes was called, with a small Spring context populating the @Value. Gateway converted to match.

  Dataplane was indeed missing — added, 6 tests. It's a fifth distinct chain: gateway falls back to the management keys, dataplane to the oauth2 ones, so repositories.management.… disables
  gateway indexes but not dataplane. Pinned now.

wajb (approved):

PR #8410, opened 2026-08-17
feat: resolve data plane against the environment during domain creation

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7262

`dataPlaneId` is now optional on domain creation. Dropped the `@NotNull`, and the openapi required list and the TS SDK are regenerated to match.

Cloud resolves the plane from the ones provisioned against the environment. It picks the single linked one when the caller omits the id, rejects zero or many, and rejects a supplied id that isn't linked. Standalone is looser on purpose, because the planes `gravitee.yml` declares serve every environment. A supplied id just has to exist in the registry, so `default` stays valid for an environment that also has one provisioned against it. An omitted id only falls back to `default` when `gravitee.yml` declares just `default` and nothing is provisioned for the environment.

Any id that comes from the caller or from an environment link is checked against the registry on this node before a domain gets bound to it. A linked row only means the id is claimed. Activation can still fail, and `verified()` returns complete for an id it was never handed, so without the check you get a domain pointing at a store nothing can read or write. The standalone fallback to `default` is the one path that skips it, because it only fires when `default` is the single declared plane.

Entrypoint events publish one event per target plane now instead of taking whichever row the repository hands back first. An environment can hold several linked planes and a gateway pinned to any of the others would never see the change. Standalone adds `default` to the targets, since a domain there can sit on a `gravitee.yml` plane while the environment has another provisioned against it.

The cloud jest specs needed reshaping before any of this could be tested. Each fixture provisioned its own org and environment and no data plane at all, so every cloud domain fell back to `default`. There's now one shared org, environment and data plane in `cloud-shared-fixture.ts`, the compose overlay pins the cloud gateway to `cloud-test` to match, and every cloud spec and fixture moved onto it. Per suite state is still torn down per suite, and `ci:cloud` runs `--runInBand`, so sharing the environment is safe. That's most of the file count in this PR.

Still open, worth knowing: a standalone domain on a `gravitee.yml` plane that isn't `default` won't get entrypoint events. `EntrypointServiceImpl` targets the environment's linked planes plus `default` and nothing else. Reading the declared ids there is doable, the module dependency is already in place, it just felt like more than this PR should carry.

## :memo: Test scenarios 

`DomainServiceTest` and `EntrypointServiceTest` cover every resolution branch, including the not-linked and not-loaded rejections and the entrypoint fan out.

New spec at `specs/cloud/domain-dataplane-resolution.jest.spec.ts` covers omitted, supplied and unlinked against a real gateway. `specs/management/domain-create.jest.spec.ts` covers the standalone side: omitting the id lands the domain on `default`, an unknown id gets a 4xx.

Ran the cloud jest suite and the playwright cloud project on both mongo and postgres, plus the full management suite on `--maxWorkers=2`, all green.

No UI change.

One trade off worth a look. Management nodes register provisioned planes from sync events, so during that propagation window a node that hasn't seen a newly provisioned plane will 400 on domain creation instead of binding the domain to a plane it can't reach. I'd rather have the clear rejection than the silent bad write, but shout if you disagree.

The SDK regen is a two line change to `NewDomain.ts`, kept in its own `chore(sdk)` commit.

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
@@ -390,12 +390,18 @@ public class DomainServiceImpl implements DomainService {
 
     /**
      * Picks the data plane a new domain will bind to. Both modes hold a supplied id to the planes
-     * provisioned against the environment; cloud requires one, while standalone falls back to
-     * {@code default} when the environment has none and only {@code default} is declared in
-     * {@code gravitee.yml}.
+     * provisioned against the environment; cloud requires the domain to land on one of them, while
+     * standalone also serves the planes {@code gravitee.yml} declares and falls back to
+     * {@code default} when the environment has none and only {@code default} is declared.
      */
     private Single<String> resolveDataPlaneId(String environmentId, String requestedId) {
         boolean cloudEnabled = CloudProperties.isManagedCloudEnabled(springEnvironment);
+        // A standalone caller naming a plane is not held to the environment's links: the planes
+        // gravitee.yml declares serve every environment, so `default` stays a valid choice for an
+        // environment that also has one provisioned against it. Saves the repository round-trip too.
+        if (!cloudEnabled && StringUtils.hasText(requestedId)) {
+            return resolveDataPlaneIdForStandalone(requestedId, List.of());
+        }
         return dataPlaneDefinitionService.findByEnvironmentId(environmentId)
                 .map(DataPlaneDefinitionSummary::id)
                 .toList()
@@ -425,12 +431,6 @@ public class DomainServiceImpl implements DomainService {
 
     private Single<String> resolveDataPlaneIdForStandalone(String requestedId, List<String> linkedForEnv) {
         if (StringUtils.hasText(requestedId)) {
-            // The environment's links win whenever it has any, so a domain and the entrypoint events
-            // for its environment (which EntrypointServiceImpl routes by the same links) agree on the plane.
-            if (!linkedForEnv.isEmpty() && !linkedForEnv.contains(requestedId)) {
-                return Single.error(new InvalidDataPlaneException(
-                        "Data Plane [" + requestedId + "] is not linked to this environment."));
-            }
             return requireLoadedOnThisNode(requestedId);
         }
         boolean onlyDefaultDeclared = Set.of(DataPlaneDescription.DEFAULT_DATA_PLANE_ID)
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
@@ -378,31 +378,24 @@ public class DomainServiceImpl implements DomainService {
         String hrid = IdGenerator.generate(newDomain.getName());
 
         return resolveDataPlaneId(environmentId, newDomain.getDataPlaneId())
-                .flatMap(resolvedId -> {
-                    newDomain.setDataPlaneId(resolvedId);
-                    return dataPlaneRegistry.verified(resolvedId)
-                            .onErrorResumeNext(error -> {
-                                log.error("Failed to verify Data Plane configuration", error);
-                                return Completable.error(new InvalidDataPlaneException(
-                                        "An error occurred while trying to create a domain. Data Plane [" + resolvedId
-                                                + "] did not answer with the settings it was provisioned with."));
-                            })
-                            .andThen(createDomain(organizationId, environmentId, newDomain, principal, automationDomain, hrid));
-                });
+                .flatMap(resolvedId -> dataPlaneRegistry.verified(resolvedId)
+                        .onErrorResumeNext(error -> {
+                            log.error("Failed to verify Data Plane configuration", error);
+                            return Completable.error(new InvalidDataPlaneException(
+                                    "An error occurred while trying to create a domain. Data Plane [" + resolvedId
+                                            + "] did not answer with the settings it was provisioned with."));
+                        })
+                        .andThen(createDomain(organizationId, environmentId, newDomain, resolvedId, principal, automationDomain, hrid)));
     }
 
     /**
-     * Picks the data plane a new domain will bind to. Cloud mode drives everything off the DPs
-     * provisioned against the environment; standalone falls back to {@code default} when the caller
-     * hasn't asked for anything and only {@code default} is declared in {@code gravitee.yml}.
+     * Picks the data plane a new domain will bind to. Both modes hold a supplied id to the planes
+     * provisioned against the environment; cloud requires one, while standalone falls back to
+     * {@code default} when the environment has none and only {@code default} is declared in
+     * {@code gravitee.yml}.
      */
     private Single<String> resolveDataPlaneId(String environmentId, String requestedId) {
         boolean cloudEnabled = CloudProperties.isManagedCloudEnabled(springEnvironment);
-        // Standalone with an explicit id doesn't need the environment's linked planes: the id has
-        // to exist in the registry, and that's the whole check. Save the repository round-trip.
-        if (!cloudEnabled && StringUtils.hasText(requestedId)) {
-            return resolveDataPlaneIdForStandalone(requestedId, List.of());
-        }
         return dataPlaneDefinitionService.findByEnvironmentId(environmentId)
                 .map(DataPlaneDefinitionSummary::id)
                 .toList()
@@ -421,25 +414,24 @@ public class DomainServiceImpl implements DomainService {
                 return Single.error(new InvalidDataPlaneException(
                         "Multiple data planes are linked to this environment; dataPlaneId must be provided."));
             }
-            return Single.just(linkedForEnv.get(0));
+            return requireLoadedOnThisNode(linkedForEnv.get(0));
         }
         if (!linkedForEnv.contains(requestedId)) {
             return Single.error(new InvalidDataPlaneException(
                     "Data Plane [" + requestedId + "] is not linked to this environment."));
         }
-        return Single.just(requestedId);
+        return requireLoadedOnThisNode(requestedId);
     }
 
     private Single<String> resolveDataPlaneIdForStandalone(String requestedId, List<String> linkedForEnv) {
         if (StringUtils.hasText(requestedId)) {
-            boolean known = dataPlaneRegistry.getDataPlanes().stream()
-                    .map(DataPlaneDescription::id)
-                    .anyMatch(requestedId::equals);
-            if (!known) {
+            // The environment's links win whenever it has any, so a domain and the entrypoint events
+            // for its environment (which EntrypointServiceImpl routes by the same links) agree on the plane.
+            if (!linkedForEnv.isEmpty() && !linkedForEnv.contains(requestedId)) {
                 return Single.error(new InvalidDataPlaneException(
-                        "An error occurred while trying to create a domain. Data Plane with provided Id doesn't exist."));
+                        "Data Plane [" + requestedId + "] is not linked to this environment."));
             }
-            return Single.just(requestedId);
+            return requireLoadedOnThisNode(requestedId);
         }
         boolean onlyDefaultDeclared = Set.of(DataPlaneDescription.DEFAULT_DATA_PLANE_ID)
                 .equals(dataPlaneConfigurationLoader.declaredIds());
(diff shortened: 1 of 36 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewDomain.java
```diff
@@ -32,7 +32,6 @@ public class NewDomain {
 
     private String description;
 
-    @NotNull
```
wajb, line 35: issue: the automation API relies on the same `DomainServiceImpl` as the management API so it requires the same leniency in `AutomationDomain`.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -367,21 +377,88 @@ public Single<Domain> create(String organizationId, String environmentId, NewDom
...
+
+    private Single<String> resolveDataPlaneIdForStandalone(String requestedId, List<String> linkedForEnv) {
+        if (StringUtils.hasText(requestedId)) {
+            return requireLoadedOnThisNode(requestedId);
+        }
+        boolean onlyDefaultDeclared = Set.of(DataPlaneDescription.DEFAULT_DATA_PLANE_ID)
+                .equals(dataPlaneConfigurationLoader.declaredIds());
+        if (!linkedForEnv.isEmpty() || !onlyDefaultDeclared) {
+            return Single.error(new InvalidDataPlaneException(
+                    "dataPlaneId is required when the environment has linked data planes or gravitee.yml declares non-default planes."));
```
wajb: nitpick: it's potentially inaccurate to assume gravitee.yml is the configuration source even in standalone since the setting could be defined via env variables.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
```diff
@@ -367,21 +377,88 @@ public Single<Domain> create(String organizationId, String environmentId, NewDom
...
+        return requireLoadedOnThisNode(requestedId);
+    }
+
+    private Single<String> resolveDataPlaneIdForStandalone(String requestedId, List<String> linkedForEnv) {
+        if (StringUtils.hasText(requestedId)) {
+            return requireLoadedOnThisNode(requestedId);
+        }
+        boolean onlyDefaultDeclared = Set.of(DataPlaneDescription.DEFAULT_DATA_PLANE_ID)
+                .equals(dataPlaneConfigurationLoader.declaredIds());
+        if (!linkedForEnv.isEmpty() || !onlyDefaultDeclared) {
```
wajb: question: the story requires that an id is required when any *loaded* are linked to the environment. If it's defined but not loaded on the node, can we rely on the system settings, provided it's default?

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/MultiDataPlaneLoader.java
```diff
@@ -59,6 +61,14 @@ public boolean isDeclared(String id) {
         return false;
     }
 
+    public Set<String> declaredIds() {
```
wajb, line 60: minor: I wonder if it's worth lazily evaluating this list and reusing it in the above method.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EntrypointServiceImpl.java
```diff
@@ -240,10 +245,45 @@ private Completable publishEntrypointEvent(Entrypoint entrypoint, Action action)
...
+    private List<String> targetPlanes(List<String> linkedForEnv) {
+        // Cloud binds every domain to one of the environment's linked planes, so those are the only
+        // gateways that can be serving it.
+        if (managedCloudEnabled) {
+            return linkedForEnv.isEmpty() ? List.of(DataPlaneDescription.DEFAULT_DATA_PLANE_ID) : linkedForEnv;
+        }
+        // Standalone also serves the planes gravitee.yml declares, and a domain may sit on one of
+        // those while the environment has another provisioned against it. Dropping `default` here
+        // strands the entrypoints of every domain bound to it.
+        if (linkedForEnv.contains(DataPlaneDescription.DEFAULT_DATA_PLANE_ID)) {
```
wajb: suggestion: could we use a hash set and simply always add the default data plane id when outside managed cloud mode?

gravitee-am-service/src/test/java/io/gravitee/am/service/EntrypointServiceTest.java
```diff
@@ -778,6 +784,55 @@ public void shouldPublishEntrypointEventOnDelete() {
...
+
+        cut.delete(ENTRYPOINT_ID, ORGANIZATION_ID, null).test()
+                .awaitDone(10, TimeUnit.SECONDS)
+                .assertComplete();
+
+        verify(eventService, times(1)).create(argThat(event ->
+                event.getType() == Type.ENTRYPOINT && "dp-a".equals(event.getDataPlaneId())));
+        verify(eventService, times(1)).create(argThat(event ->
+                event.getType() == Type.ENTRYPOINT && "dp-b".equals(event.getDataPlaneId())));
+        // standalone domains can sit on a gravitee.yml plane, so default is a target as well
```
wajb: nitpick: "standalone domains" sounds like something it isn't IMO and this could mislead.

gravitee-am-test/specs/cloud/fixtures/cloud-shared-fixture.ts
```diff
@@ -0,0 +1,177 @@
...
+ */
+const ORGANIZATION_ID = 'cloud-shared';
+const ENVIRONMENT_ID = 'cloud-shared-env';
+const USER_ID = 'cloud-shared-owner';
+const DATA_PLANE_ID = 'cloud-test';
+
+// The definition has to name the same store the gateway is configured with, otherwise the management
+// API writes users the gateway then cannot read. Both stacks run one database, so the branch follows
+// REPOSITORY_TYPE the same way the internal-api data plane fixtures do.
+const dataPlaneStore = () =>
```
wajb, line 49: thought: it seems brittle to hardcode these environment settings. Can we reuse anything from the *.setup.js file or the process env if not the local stack docker config?

gravitee-am-test/specs/cloud/domain-dataplane-resolution.jest.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  beforeAll(async () => {
+    shared = await setupCloudSharedFixture();
+  });
+
+  afterEach(async () => {
+    await safeDeleteCloudDomains(shared, createdDomainIds.splice(0));
+  });
+
+  it(jira`should resolve the environment's data plane when dataPlaneId is omitted ${'AM-7262'}`, async () => {
```
wajb: question: should we be linking tests to stories rather than test cases? 🤔

gravitee-am-test/playwright.config.ts
wajb, line 1: question: is this change needed?

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewDomain.java
sclarkamlab: Good catch. Done in `6443ae4410`, `AutomationDomain.dataPlaneId` is optional now and the omitted case reaches `DomainServiceImpl` the same way. Regenerated `docs/automation/openapi.yaml` too. oasdiff flags it as a response-side break because the model is shared request/response, but we're on the 4.13 boundary so the check passes.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: Yeah fair. Done in `6d8ee055b7`, message and the javadocs say the node's configuration now.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DomainServiceImpl.java
sclarkamlab: Yeah we can. Done in `6d8ee055b7`, standalone now only counts a linked plane if it's actually loaded on this node, so a linked but unloaded one falls back to `default` when that's all that's declared. Covered in `DomainServiceTest`.

gravitee-am-plugins-handlers/gravitee-am-plugins-handlers-dataplane/src/main/java/io/gravitee/am/plugins/dataplane/core/MultiDataPlaneLoader.java
sclarkamlab: Done in `522005428e`, `isDeclared` is just `declaredIds().contains(id)` now. Skipped caching it, it's one walk over the config per call.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/EntrypointServiceImpl.java
sclarkamlab: Done in `522005428e`, `LinkedHashSet` and `default` gets added when not in managed cloud or nothing is linked.

gravitee-am-service/src/test/java/io/gravitee/am/service/EntrypointServiceTest.java
sclarkamlab: Done in `522005428e`, reworded.

gravitee-am-test/specs/cloud/fixtures/cloud-shared-fixture.ts
sclarkamlab: Partly. The hosts now take the same `AM_DATAPLANE_MONGO_HOST` / `AM_DATAPLANE_JDBC_HOST` overrides the internal-api fixtures use (`1e05e0ac03`). The rest is the local stack's own db config, and CI spins up the same compose files, so I don't have anything else to read them from. The setup.js values are the runner's view (localhost) rather than what the management api container needs.

gravitee-am-test/specs/cloud/domain-dataplane-resolution.jest.spec.ts
sclarkamlab: Fair, the tag was written for Xray test keys and I've pointed it at the story because there isn't a test case for this yet, same as the AM-7236 specs did. I'll raise the test cases and swap the keys in a follow up rather than leave these untagged.

gravitee-am-test/playwright.config.ts
sclarkamlab: Yes. The playwright cloud project imports `cloud-shared-fixture`, which calls `createDataPlane` on the management technical api, and that reads `AM_MANAGEMENT_NODE_URL`. Playwright doesn't load the jest `*.setup.js` files so it was undefined there. The indent fix on the cloud project is incidental.

gravitee-am-test/specs/cloud/fixtures/cloud-shared-fixture.ts
wajb: Could we use `AM_MONGODB_URI` as `uri`, for example, rather than separate hosts and ports at least? Anything missing, like database names, would probably be better off defined as new setup variables IMO, unless I'm missing something?

sclarkamlab: Done in `567c1eab68`. Went with `AM_INTERNAL_MONGODB_URI ?? AM_MONGODB_URI` and `AM_INTERNAL_POSTGRES_HOST ?? AM_POSTGRES_HOST`, same fallback `idps-commands.ts` uses, and mongo takes the uri form now instead of host/port. Had to be the internal ones rather than `AM_MONGODB_URI` on its own, the management api opens this connection from inside the container and dev.setup.js points that at localhost. `playwright.config.ts` gets the two defaults since it mirrors dev.setup.js and the cloud project only ever runs against the docker stack.

Left the db name and the postgres creds inline. They're the stack's own config and don't move with the runner, so making them setup vars means the same five constants copied into ci.setup.js, dev.setup.js and playwright.config.ts. Happy to do it if you'd still rather, though.

gravitee-am-test/specs/cloud/domain-dataplane-resolution.jest.spec.ts
sclarkamlab: Fair point, tags gone in `e06f372c1d`. There's no test case to point at and I'm not raising one, so the tests just have plain titles now.

gravitee-am-test/specs/cloud/fixtures/cloud-shared-fixture.ts
wajb: Thanks. Relatedly, I notice `AM_DATAPLANE_MONGO_HOST` and `AM_DATAPLANE_JDBC_HOST` are made up variables in the data plane fixture (outside cloud suite) with hard coded fallbacks. Are these really the same values in both stacks?

sclarkamlab: Yeah, same values, and the `AM_DATAPLANE_*` pair was never set anywhere so it always took the fallback. Folded it into the same vars in `7fa2285888`: the fixture reads `AM_INTERNAL_MONGODB_URI ?? AM_MONGODB_URI` and `AM_INTERNAL_POSTGRES_HOST ?? AM_POSTGRES_HOST` now, like the cloud fixture and idps-commands. It needs a host and port for the payloads rather than a uri, hence the parse. Side effect is a native dev run gets localhost instead of the docker service name, which is what it should have had all along. Also pointed the unreachable payload at the same hosts so it only differs by the port.

wajb (approved):

PR #8417, opened 2026-08-18
feat(token-exchange): propagate subject token claims to the issued token

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7331

Token exchange throws away the subject token claims after it validates them, so a dynamic claim on the subject token cannot reach the exchanged token. Actor token claims already reach EL. This adds the subject half, and a declarative mapper for admins who do not want to write EL.

**Expression language.** A custom token claim can now read `#context.attributes['token_exchange']['subject']['subject_token_claims']['<name>']`. `SubjectTokenInfo` carries the sub, the gis and the full claims map of the validated subject token. `TokenExchangeResult.buildExecutionContext` assembles the map: the `subject` entry is always there, the `actor` entry only for delegation. `ActorTokenInfo.buildTokenExchangeExecutionContext` becomes `buildContext` and returns the bare actor map, because the `token_exchange` wrapper now has two contributors.

**Claims mapper.** `TokenExchangeOAuthSettings` gains a `claimsMapper` list. Each entry names its source token, the claim to read, and the name to write. The gateway resolves the list at exchange time and copies the values onto the issued token before the custom claim enhancement, so a `tokenCustomClaims` entry of the same name wins. It works on the access token and on an exchanged id_token. Admins configure it in the Console, on the domain defaults and per application.

Both paths omit a claim that is absent from its source token, and issuance still succeeds. Only the validators produce the claims map, so the claims are always cryptographically verified. Only the token exchange strategy fills this part of the execution context, so no other grant sees it.

Two things to know at review:

- `token_exchange` is now present during impersonation, holding `subject` alone. An expression that tests `['token_exchange'] != null` to detect delegation flips. Test `['token_exchange']['actor'] != null` instead. Nothing in the repo does this.
- The mapper cannot write a protocol-critical claim. `TokenExchangeClaimsMapperValidator` rejects `sub`, `iss`, `aud`, `exp`, `iat`, `nbf`, `jti`, `act`, `client_id`, `scope` and `gis`. A mapping can read from an external trusted-issuer subject token, so writing `sub` or `iss` would let a remote issuer redefine the identity of the exchanged token.

Inheritance stays wholesale, as it already is for `scopeHandling`: an application that turns off `Inherit configuration` supplies its own settings and does not fall back to the domain mapper.

## :memo: Test scenarios

Java unit, repository round-trip and Jest coverage in the branch, all green. The repository tests cover both Mongo and JDBC. The Jest cases prove the value comes from the subject token rather than the request or the user profile. I also drove the Console against a local stack: added mappings on the domain defaults, saved, reloaded, and confirmed the mapper persisted and the reserved-claim validation rejects a mapping onto `sub`.

The Claims Mapper table sits under Scope Handling on the domain Token Exchange settings, and under the inheritance toggle on the application OAuth settings. Each row names the source token, the source claim and the target claim. 

Application Level Claims Mapper
image

Domain Level Claims Mapper
image

The docs need the new EL address, the omit-on-absent behaviour and the delegation detector above. The AM docs live outside this repo, so they are not in this PR.

The generated OpenAPI spec lists the enum values in upper case while the API serializes them in lower case. `scopeHandling` behaves the same way on master, so this is the existing pattern rather than something this branch introduces. The Console sends lower case, and the API accepts either.

@gravitee-io/am


gravitee-am-service/src/main/java/io/gravitee/am/service/validators/tokenexchange/TokenExchangeClaimsMapperValidator.java
```diff
@@ -0,0 +1,71 @@
...
+            Claims.SUB,
+            Claims.ISS,
+            Claims.AUD,
+            Claims.EXP,
+            Claims.IAT,
+            Claims.NBF,
+            Claims.JTI,
+            Claims.ACT,
+            Claims.CLIENT_ID,
+            Claims.SCOPE);
```
rafalpodles: **Blocker.** This set does not cover several claims AM writes *before* the mapper's `forEach(jwt::put)`, so a mapping can silently clobber them.

In `TokenServiceImpl.createAccessTokenJWT` the insertion point is line 502, and these are all set earlier:

- **`cnf`** (`:481-492`) — replacing it breaks the DPoP/mTLS sender constraint. `OAuth2AuthHandlerImpl:131` gates DPoP enforcement on `getDPoPConfirmationThumbprint() != null`, so the issued token silently degrades to a usable bearer token. A non-map value also hits the `(Map)` cast at `TokenServiceImpl:436` and 500s the token endpoint.
- **`permissions`** (`:621`) — UMA 2.0. `UmaStrategy:304` reads it back to decide resource access, and it survives refresh because `Claims.getAllClaims()` does not strip it.
- **`authorization_details`** (`:627`) — RFC 9396 RAR grant contents.
- **`domain`** (`:610`) — this one bypasses revocation. `DomainTokenValidator:59-71` skips the `findAccessTokenByJti` lookup entirely when `token.getDomain()` does not match the current domain, so a token with a rewritten `domain` claim, later replayed as a `subject_token`, is accepted after being revoked.

And in `IDTokenServiceImpl`: **`acr`** (`:221`), **`auth_time`** (`:204`), **`nonce`** (`:210`). `acr` is the sharpest of these — `acr -> acr` from a trusted IdP looks like an obviously reasonable mapping to configure, and it hands a remote issuer the ability to assert any authentication assurance level.

Also `sub_profile` / `client_profile`: the emission guards at `TokenServiceImpl:530,539` are `jwt.get(...) == null`, so a mapped value not only survives but *suppresses* AM's own authoritative value.

None of this is remotely exploitable without an administrator configuring the mapping — but that is exactly what this deny-list exists to prevent, and the javadoc above already makes the argument ("a mapping can read from an external trusted-issuer subject token"). It just stops short of its own logic.

Suggestion: base the set on the existing `Claims.getAllClaims()` (`ISS, SUB, AUD, EXP, NBF, IAT, AUTH_TIME, UPDATED_AT, JTI, DOMAIN, CLAIMS, IP_ADDRESS, USER_AGENT, SCOPE, CNF, CLIENT_ID`) — already used as the "AM-owned claims" list when building refresh tokens at `TokenServiceImpl:586` — plus `GIO_INTERNAL_SUB, ACT, SUB_PROFILE, CLIENT_PROFILE, permissions, authorization_details, nonce, acr`.

Worth considering the stronger fix too: move the `forEach(jwt::put)` *above* AM's own writes so the deny-list becomes defence in depth rather than the only control. Custom claims would still win over the mapper, and AM would win over both.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
```diff
@@ -815,6 +819,11 @@ public Single<Application> patch(Domain domain, String id, PatchApplication patc
                         return Single.error(new InvalidParameterException("Invalid token claims: " + claimValidation.invalidClaims()));
                     }
 
+                    var mapperValidation = tokenExchangeClaimsMapperValidator.validate(tokenExchangeClaimsMapperOf(toPatch));
```
rafalpodles: This is the only call site of the new validator, so `create` and `update` are not covered.

`create(Domain, Application)` (`:465`) and `update(Application)` (`:750`) both go through `create0`/`innerUpdate` to `validateApplicationMetadata` (`:1136`), which does not run it. Today that is unreachable only by accident — the REST create endpoint takes `NewApplication` (no oauth settings), and the DCR path happens to drop `tokenExchangeOAuthSettings` entirely. Fix the DCR omission mentioned in the review summary and this immediately becomes a self-service route to map an external issuer's claim onto a reserved name.

Moving the call into `validateApplicationMetadata` covers `create`, `update`, `updateType` and DCR in one place. Note `customClaimsValidator` on line 817 has the same gap, so this would fix both.

Domain side is correctly wired — `DomainServiceImpl:955` inside `validateDomain` covers create and patch.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/TokenExchangeServiceImpl.java
```diff
@@ -310,6 +313,41 @@ private ActorTokenInfo extractActorInfo(ValidatedToken actorToken, ValidatedToke
...
+                                                    ValidatedToken subjectToken, ValidatedToken actorToken) {
+        List<TokenExchangeClaimMapping> mappings = TokenExchangeOAuthSettings.getInstance(domain, client).getClaimsMapper();
+        if (mappings == null || mappings.isEmpty()) {
+            return Map.of();
+        }
+
+        Map<String, Object> mappedClaims = new HashMap<>();
+        for (TokenExchangeClaimMapping mapping : mappings) {
+            ValidatedToken source = mapping.getSource() == TokenExchangeClaimSource.ACTOR_TOKEN ? actorToken : subjectToken;
+            if (source == null || mapping.getSourceClaim() == null || mapping.getTokenClaim() == null) {
```
rafalpodles: The mapper accepts input the sibling `trustedIssuers` validation in this same feature explicitly rejects:

- **Blank names.** Only `!= null` is checked here, and `TokenExchangeClaimsMapperValidator` only checks the reserved list — `""` is not reserved, so it passes and produces `mappedClaims.put("", value)`, i.e. a JWT claim with an empty name. The console trims (`grant-flows.component.ts:259`) but the REST API does not.
- **Duplicate `tokenClaim`.** Two mappings onto the same name silently collapse via `HashMap.put` at line 345; the winner depends on list order. `TokenExchangeSettingsValidatorImpl:111-113` rejects duplicate trusted issuers.
- **Unbounded list size.** `trustedIssuers` is capped by `${domain.tokenExchange.trustedIssuers.maxCount:5}` in the same validator class; `claimsMapper` has no cap, so N mappings carrying arbitrarily large external claim values inflate every issued token.

Suggest trimming and rejecting blanks, rejecting duplicate `tokenClaim`, and adding a `domain.tokenExchange.claimsMapper.maxCount` bound to mirror the neighbour.

Minor related point: the validator trims `tokenClaim` before the reserved check but the untrimmed value is what gets stored, so `"  sub  "` is rejected with the message `[sub]`. Normalising on write would be the cleaner fix.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
```diff
@@ -497,6 +497,12 @@ private JWT createAccessTokenJWT(OAuth2Request request, Client client, User user
             jwt.setClaimsRequestParameter(requestParameters.getFirst(Parameters.CLAIMS));
         }
 
+        // Token Exchange (RFC 8693) - copy the claims resolved by the declarative mapper.
+        // Applied before the custom claims so a tokenCustomClaims entry of the same name wins.
+        if (request.getTokenExchangeMappedClaims() != null && !request.getTokenExchangeMappedClaims().isEmpty()) {
+            request.getTokenExchangeMappedClaims().forEach(jwt::put);
```
rafalpodles, line 503: This block and its `IDTokenServiceImpl` twin have no unit test — `TokenServiceImplTest` and `IDTokenServiceTest` both exist and are untouched, and `getTokenExchangeMappedClaims` appears nowhere in test sources.

The end-to-end coverage is genuinely good — the new jest specs decode the real issued JWT, and the choice of `jti` as the source claim is a neat way to prove the value came from the subject token rather than the issued one. But the ordering guarantee stated in the comment above ("a tokenCustomClaims entry of the same name wins") is only asserted in the integration layer, so it depends on the gateway lane staying green for unrelated reasons.

Two small tests would close this cheaply — set `tokenExchangeMappedClaims` plus a `tokenCustomClaims` entry of the same name and assert the custom claim wins, once here and once for the ID token path.

While in the area: `StrategyGranterAdapterTest` was updated for arity only, so the `data.mappedClaims() == null ? Map.of() : ...` fallback in `StrategyGranterAdapter:326` is never exercised — all four call sites pass `Map.of()`. The tests already capture the `OAuth2Request` with an `ArgumentCaptor`, so asserting propagation there is nearly free.

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.ts
```diff
@@ -246,6 +251,38 @@ export class GrantFlowsComponent implements OnInit {
...
+  get tokenExchangeClaimsMapper(): TokenExchangeClaimMapping[] {
+    return this.oauthSettings.tokenExchangeOAuthSettings?.claimsMapper ?? [];
+  }
+
+  isNewClaimMappingValid(): boolean {
+    return !!this.newClaimMapping.sourceClaim?.trim() && !!this.newClaimMapping.tokenClaim?.trim();
+  }
+
+  addClaimMapping() {
+    this.oauthSettings.tokenExchangeOAuthSettings = {
```
rafalpodles, line 291: Turning off "Inherit configuration" silently drops the domain's claim mappings, with no signal to the user.

`ngOnInit:104` initialises the override as `{inherited: true, scopeHandling}` with no `claimsMapper`, and nothing seeds it from the domain. Because the backend resolves settings wholesale (`TokenExchangeOAuthSettings.getInstance`), an application that overrides never consults the domain mapper.

Scenario: the domain defines three mappings; an admin opens the application and un-inherits purely to switch Scope Handling to Permissive. The Claims Mapper table renders empty, which reads as "no mappings configured yet" rather than "you just dropped three". They save, and every token issued for that application quietly loses those claims.

This already applies to `scopeHandling`, but the consequence there is visible in one field, whereas here it is an empty table that looks like a legitimate starting state. Suggest seeding `claimsMapper` (and `scopeHandling`) from the domain when inheritance is switched off, and/or a hint under the toggle: "An override replaces the domain configuration entirely, including its claim mappings."

Two smaller things in the same feature: there is no client-side check for duplicate or reserved `tokenClaim` names, although `tokens.component.ts:63-80` does exactly that for custom claims a couple of hundred lines away; and `grant-flows.component.spec.ts` already has a `describe('Token Exchange scope handling')` block covering the sibling feature, so the absence of tests for `addClaimMapping`/`removeClaimMapping`/`isNewClaimMappingValid` is a gap rather than the local convention.

rafalpodles (changes requested):
Reviewed the full diff plus the surrounding runtime, persistence, validation and console code.

The mechanism is well built: the ordering rationale (custom claims win over the mapper), the wholesale inheritance semantics, the `Application -> Client` deep copy and the "only the token exchange strategy fills this context" isolation all hold up under inspection. I also confirmed both caveats in the description — nothing in the repo branches on the presence of `token_exchange`, and the OpenAPI enum-case difference is the existing `scopeHandling` pattern (lowercase on the wire is correct thanks to the global enum serializer in `ObjectMapperResolver`).

One blocking item: `RESERVED_TOKEN_CLAIMS` does not cover several claims that AM writes *before* the mapper runs, so the deny-list does not yet enforce the threat model its own javadoc describes. Details inline. The remaining four comments are non-blocking.

Separately, worth its own ticket (pre-existing, but this PR widens the blast radius): `ClientServiceImpl.getSettings(Client)` in the gateway DCR path rebuilds `ApplicationOAuthSettings` field by field and never copies `tokenExchangeOAuthSettings` back, so a routine `PUT /oidc/register/{client_id}` silently wipes `inherited`, `scopeHandling` and now `claimsMapper` too.

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.html
```diff
@@ -104,6 +104,78 @@ <h5>Token Exchange</h5>
             </mat-select>
             <mat-hint>Controls how scopes are narrowed during token exchange.</mat-hint>
           </mat-form-field>
+
+          <div class="gv-form-section-title" style="margin-top: 16px">
+            <h5>Claims Mapper</h5>
```
wajb: thought: the title describes a component rather than the functionality it serves. Token exchange's trusted issuers use "Scope Mappings" for a similar concept. Would "Claim Mappings" fit better?

(Same suggestion for the UI in domain-level settings.)

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.html
```diff
@@ -104,6 +104,78 @@ <h5>Token Exchange</h5>
...
+              mat-stroked-button
+              type="button"
+              [disabled]="readonly || !isNewClaimMappingValid()"
+              (click)="addClaimMapping()"
+            >
+              <mat-icon svgIcon="gio:plus"></mat-icon>ADD
+            </button>
+          </div>
+          <mat-hint style="font-size: 75%">
+            A claim that is absent from its source token is skipped. A custom token claim of the same name overrides the mapping.
```
wajb, line 181: question: any reason this is split from the label in the application level but combined in one hint in the domain level settings?

gravitee-am-model/src/main/java/io/gravitee/am/model/application/TokenExchangeOAuthSettings.java
```diff
@@ -33,12 +36,19 @@ public class TokenExchangeOAuthSettings {
             "issued token to a subset of the original scopes.",
             defaultValue = "DOWNSCOPING")
     private TokenExchangeScopeHandling scopeHandling = TokenExchangeScopeHandling.DOWNSCOPING;
+    @Schema(description = "Claims copied from the validated subject or actor token onto the issued token. " +
+            "A claim that is absent from its source token is skipped.")
+    private List<TokenExchangeClaimMapping> claimsMapper;
```
wajb: suggestion: like the UI name, this feels clunky. Match TrustedIssuer's naming in the APIs too?

```suggestion
    private List<TokenExchangeClaimMapping> claimMappings;
```

sclarkamlab:
Good catch on the deny-list. Fixed in `2c9e1bc938`, `RESERVED_TOKEN_CLAIMS` is now built from `Claims.getAllClaims()` plus `gis`, `act`, `nonce`, `acr`, `amr`, `azp`, `permissions`, `authorization_details`, `client_profile` and `sub_profile`. The other four are done too, details in the threads.

I'll raise a technical task for the `ClientServiceImpl.getSettings` DCR omission, it's pre-existing and wants its own fix.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/tokenexchange/TokenExchangeClaimsMapperValidator.java
sclarkamlab: Done in `2c9e1bc938`. Deny-list is `Claims.getAllClaims()` plus the extras you listed, and the javadoc now names the `cnf` and `domain` stakes instead of dancing round them.

Won't do the reordering. Mapped claims land at `IDTokenServiceImpl:257`, after the user profile claims get processed, so moving the `forEach` above AM's writes would make a mapping onto something like `email` silently lose to the user's own value. The deny-list stays the control.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/ApplicationServiceImpl.java
sclarkamlab: Done in `dc093da8a8`. Both this and `customClaimsValidator` moved into `validateApplicationMetadata`, so create, update, updateType and DCR all get it.

Put them at the front of the chain rather than the end, otherwise a bad redirect_uri masks the claim error. Added a test that `update` rejects a mapping onto `cnf`.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/impl/TokenExchangeServiceImpl.java
sclarkamlab: Blanks and duplicates done in `ebab7b8722`, cap in `dc093da8a8` via `domain.tokenExchange.claimsMapper.maxCount` defaulting to 20. Put the cap in the claims mapper validator rather than the settings one so the app-level path gets it too.

Skipping normalise on write. Anything already stored stays unnormalised, so the gateway needs to trim regardless, and it does now.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/TokenServiceImpl.java
sclarkamlab: Done in `b2fa98ae06`. Two tests each side, mapped claims land on the token and a `tokenCustomClaims` entry of the same name wins. `StrategyGranterAdapterTest` now asserts propagation onto the `OAuth2Request` and covers the null fallback.

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.ts
sclarkamlab: Done in `45360b8a4f`. Un-inheriting now seeds `claimsMapper` and `scopeHandling` from the domain, once, keyed off `claimsMapper` being undefined so later edits survive a re-toggle. Hint under the toggle says the override replaces the domain config entirely.

Also added the reserved and duplicate check on the target claim, and specs for the add/remove/validate methods.

sclarkamlab:
Actually fixed it here, `c35a5cd0dd`. `Client` already carries `tokenExchangeOAuthSettings`, `getSettings` just never copied it back, so it was a one liner. Worth noting it wasn't only the register PUT, `renewSecret` goes through the same `convert` so a secret renewal wiped it too.

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.html
sclarkamlab: Yeah, "Claim Mappings" is better. Renamed in `a2cd723de9`, both the application and the domain settings UI.

gravitee-am-ui/src/app/domain/components/oauth2-settings/grant-flows/grant-flows.component.html
sclarkamlab: No reason, just how they got written. Aligned the domain level to the application one in `a2cd723de9`, so both have the "copy a claim..." line as the section description and the "absent claim is skipped" bit as a hint by the add row.

gravitee-am-model/src/main/java/io/gravitee/am/model/application/TokenExchangeOAuthSettings.java
sclarkamlab: Good catch. Renamed to `claimMappings` in `11b2781d24`, model, mongo, validator, openapi and the UI types. Nothing to migrate since the field only exists on this branch.

Renamed `TokenExchangeClaimsMapperValidator` to `TokenExchangeClaimMappingsValidator` while I was in there.

wajb (approved):

podlesrafal (approved):

PR #8418, opened 2026-08-18
feat: AM-7386 CSP directives validation

wajb:
[AM-7386](https://gravitee.atlassian.net/browse/AM-7386)

Enhance validation and UX of domain-level CSP configuration:
- Replace chip input with a structured Name/Value list editor for domain-level CSP directives.
- Validate format of directive names and values in UI and mAPI/aAPI (known directives are used for autocompletion in the UI only).
- Fix runtime gateway issues:
  - HTTP 500 Vert.x error when reportOnly is enabled with no `report-uri`/`report-to` directive.
  - Mixed-case script-src defeating the nonce lookup.
  - Valueless directives (like `upgrade-insecure-requests`) being silently dropped.

## :memo: Test scenarios 
See [AM-7368](https://gravitee.atlassian.net/browse/AM-7368)

### Before
image

### After
image

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy

@gravitee-io/am 

[AM-7386]: https://gravitee.atlassian.net/browse/AM-7386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-7368]: https://gravitee.atlassian.net/browse/AM-7368?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-model/src/main/java/io/gravitee/am/model/webprotection/CspDirective.java
+++ b/gravitee-am-model/src/main/java/io/gravitee/am/model/webprotection/CspDirective.java
@@ -1,0 +1,129 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.model.webprotection;
+
+import java.util.Locale;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/**
+ * A single Content Security Policy directive.
+ * <p>
+ * A trailing {@code ;} is optional and is discarded on parse. Directive names are case-insensitive
+ * per the CSP specification, so comparisons should use {@link #canonicalName()}; the name as typed
+ * by the operator is preserved in {@link #name()}.
+ * <p>
+ * Source-list values are not interpreted: everything after the first whitespace run is kept verbatim.
+ *
+ * @author GraviteeSource Team
+ */
+public record CspDirective(String name, String value) {
+
+    private static final Pattern VALID_NAME = Pattern.compile("^[a-z0-9-]+$");
+
+    /**
+     * Directives that carry no value at all, e.g. {@code upgrade-insecure-requests}.
+     */
+    private static final Set<String> NO_VALUE_DIRECTIVES = Set.of(
+            "upgrade-insecure-requests",
+            "block-all-mixed-content"
+    );
+
+    /**
+     * Directives whose value is optional, e.g. a bare {@code sandbox} is the most restrictive form.
+     */
+    private static final Set<String> OPTIONAL_VALUE_DIRECTIVES = Set.of(
+            "sandbox",
+            "trusted-types"
+    );
+
+    public static final String REPORT_URI = "report-uri";
+    public static final String REPORT_TO = "report-to";
+
+    /**
+     * Parses a stored directive entry.
+     *
+     * @return the parsed directive, or {@code null} if the entry is null, blank, or only a semicolon
+     */
+    public static CspDirective parse(String entry) {
+        if (entry == null) {
+            return null;
+        }
+        final String stripped = stripTrailingSemicolon(entry);
+        if (stripped.isEmpty()) {
+            return null;
+        }
+        final String[] parts = stripped.split("[ \t]+", 2);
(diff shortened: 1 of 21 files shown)


gravitee-am-model/src/main/java/io/gravitee/am/model/webprotection/CspDirective.java
```diff
@@ -0,0 +1,122 @@
...
+            "block-all-mixed-content"
+    );
+
+    /**
+     * Directives whose value is optional, e.g. a bare {@code sandbox} is the most restrictive form.
+     */
+    private static final Set<String> OPTIONAL_VALUE_DIRECTIVES = Set.of(
+            "sandbox",
+            "trusted-types"
+    );
```
sclarkamlab, line 51: Are these ok to be hardcoded? Is it possible this would need a product update in the future to provide additional values? Maybe these could be configured in gravitee.yml?

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/domain/DomainValidatorImpl.java
```diff
@@ -141,4 +146,11 @@ private Completable validateCertificateBasedAuthSettings(Domain domain) {
         return Completable.complete();
     }
 
+    private Completable validateCspSettings(Domain domain) {
```
sclarkamlab, line 149: This runs on every domain patch, not just web protection ones, because `DomainServiceImpl.patch` merges into the whole domain and then validates the result. The old chip editor let people save duplicates and whole policies packed into one entry, so those domains will now fail on something unrelated like a rename, and get a CSP error back. Either gate this on the patch actually touching `webProtectionSettings`, or normalise the existing lists first.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/webprotection/CspSettingsValidatorImpl.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * @return a failed {@link Completable} describing the problem, or {@code null} when the entry is valid
+     */
+    private Completable validateEntry(String entry, Set<String> seenNames) {
+        final CspDirective directive = CspDirective.parse(entry);
+        if (directive == null) {
+            return error("CSP directives must not be blank");
+        }
+
+        if (CspDirective.stripTrailingSemicolon(entry).contains(";")) {
```
sclarkamlab, line 86: We reject `;` but nothing rejects control characters. I ran `CspDirective.parse` on `default-src 'self'\r\nSet-Cookie: x` and it comes back as a valid name with that value and passes every check, then Vert.x `putHeader` throws on anything below 32 apart from tab and the login page 500s on every request. Same failure this PR is fixing for report only, so I'd reject control chars here too.

sclarkamlab (commented):
Looks good, tests are thorough. I checked the Vert.x source and the 500 claim in the javadoc is right, `CSPHandlerImpl.handle` does `ctx.fail` when the policy map has no `report-uri` or `report-to`.

Two inline comments on things I think need a look. Smaller stuff not worth its own thread:

Falling back to `NoOpCspHandler` for report only with no target is the right call I think, but it does mean a gravitee.yml policy can lose CSP entirely with only an error log to say so.

`csp.jest.spec.ts` copies the inline `beforeAll`/`afterAll` from the cors spec rather than the fixture pattern `GUIDELINES.md` asks for on new specs, and `toBeDefined()` on required fields is called out in there too.

`web-protection.component.html:375` flips customise to customize, which is the wrong way round for us and isn't related to the feature.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/domain/DomainValidatorImpl.java
wajb: I think this is fine since domain-level web protection settings is new to 4.13.

gravitee-am-service/src/main/java/io/gravitee/am/service/validators/webprotection/CspSettingsValidatorImpl.java
wajb: Thanks. I'll expand the 500 runtime fix with a character blacklist.

gravitee-am-model/src/main/java/io/gravitee/am/model/webprotection/CspDirective.java
wajb: My understanding is these valueless types are supported for historical reasons but newer specifications favour expressive parameters rather than switches - so it would be unlikely we'd have to support more. I think if we can avoid the configuration complexity we should here. Wdyt?

sclarkamlab: Yeah sounds good just wanted to run it by you first to get your thoughts on it

sclarkamlab (approved):

PR #8420, opened 2026-08-18
test: cover mongo audit reporter index lifecycle (AM-7537)

sagarchauhan-qa:
[AM-7537](https://gravitee.atlassian.net/browse/AM-7537)

Adds MongoAuditReporterIndexTest — 8 tests over the Mongo audit reporter's index lifecycle.

Driven by the AM-7454 / AM-7457 P1 (Zendesk 17891): the retired r1a1ta1t_1 index survived an upgrade and left security domains failing to sync. Nothing asserted that the. retired-index sweep ran, or that its list was complete.

N/A

## :memo: Test scenarios 
N/A

N/A

N/A

[AM-7537]: https://gravitee.atlassian.net/browse/AM-7537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
+++ b/gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
@@ -27,10 +27,12 @@ import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.ApplicationContext;
 import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.test.context.support.AnnotationConfigContextLoader;
 import org.springframework.test.util.ReflectionTestUtils;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -78,6 +80,13 @@ import static org.junit.Assert.fail;
  */
 @RunWith(SpringRunner.class)
 @ContextConfiguration(classes = {MongoReporterJUnitConfiguration.class}, loader = AnnotationConfigContextLoader.class)
+// Matches MongoAuditReporterTest so both classes share one cached Spring context, and so the
+// module starts a single Mongo testcontainer. startReporter() sets purgeEnabled and
+// retentionDays explicitly, so these values never reach an assertion.
+@TestPropertySource(properties = {
+    "services.purge.enabled=true",
+    "services.purge.audits.retention.days=90"
+})
 public class MongoAuditReporterIndexTest {
 
     /** Indexes the reporter declares unconditionally. */
@@ -141,7 +150,7 @@ public class MongoAuditReporterIndexTest {
 
         Map<String, Document> byName = indexesByName();
         expectedKeys().forEach((name, key) ->
-                assertEquals("wrong key for index " + name, key, byName.get(name).get("key")));
+                assertKeyEquals(name, key, byName.get(name).get("key", Document.class)));
     }
 
     @Test
@@ -168,7 +177,7 @@ public class MongoAuditReporterIndexTest {
                 "the retention index was not created");
 
         Document key = indexesByName().get(INDEX_TIMESTAMP_ID_NAME).get("key", Document.class);
-        assertEquals(new Document(FIELD_TIMESTAMP, 1).append(FIELD_ID, 1), key);
+        assertKeyEquals(INDEX_TIMESTAMP_ID_NAME, new Document(FIELD_TIMESTAMP, 1).append(FIELD_ID, 1), key);
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


podlesrafal (commented):
Thanks for adding this — the retired-index sweep genuinely had no coverage before, and pinning it to constants rather than literals is the right call. Three notes below, only the first one affects what the test actually proves.

Minor stuff I'm deliberately not commenting on inline: the two 3s settle sleeps (they can only fail slow, never flake red), and the fact that `startReporter`/`indexNames` are near-copies of the private helpers in `MongoAuditReporterTest` — worth folding into a shared base class at some point, but not blocking here.

Separately: you spotted that `OLD_INDICES` lists `ref_1_time_-1` twice and worked around it with `.distinct()`. That's a wart in `MongoAuditReporterConstants`, not in this PR — probably worth a follow-up to make it a `Set` (`::contains` semantics are unchanged).

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
```diff
@@ -0,0 +1,302 @@
...
+        seedIndex(OLD_INDEX_REFERENCE_ACTOR_TARGET_TIMESTAMP_NAME_SHORT_NAME, historicalActorTargetKey());
+
+        startReporter(reporter -> {});
+
+        awaitIndexes(names -> names.containsAll(CURRENT_INDEXES),
+                "the current index set did not land");
+
+        Map<String, Document> byName = indexesByName();
+        expectedKeys().forEach((name, key) ->
+                assertEquals("wrong key for index " + name, key, byName.get(name).get("key")));
```
podlesrafal: This assertion is weaker than it looks: `org.bson.Document.equals()` delegates to `Map.equals()`, which is order-insensitive. For a compound index the field order *is* the semantics, so a refactor that swapped e.g. `referenceType` and `referenceId` would still pass here.

Suggest comparing the key order explicitly:

```java
assertEquals("wrong key for index " + name,
        new ArrayList<>(key.keySet()),
        new ArrayList<>(byName.get(name).get("key", Document.class).keySet()));
```

(or compare `toJson()` on both sides, which covers order and values in one go). Same applies to the retention-index assertion on line 171.

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
```diff
@@ -0,0 +1,302 @@
...
+ * Index lifecycle of the Mongo audit reporter.
+ *
+ * Covers the retired-index sweep, the current index set and its definitions, and the two
+ * independent ensureIndexOnStart switches. Driven by AM-7454 / AM-7457, where the retired
+ * index r1a1ta1t_1 survived an upgrade and left security domains failing to sync.
+ *
+ * The management-node gate is covered by MongoAuditReporterTest and is not repeated here.
+ */
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = {MongoReporterJUnitConfiguration.class}, loader = AnnotationConfigContextLoader.class)
```
podlesrafal, line 82: `MongoAuditReporterTest` carries `@TestPropertySource(properties = {"services.purge.enabled=true", "services.purge.audits.retention.days=90"})`; this class doesn't. That makes the two `MergedContextConfiguration` cache keys differ, so Spring builds a second context — and since `MongoDBContainer` is a context-scoped bean, the module now boots a second Mongo testcontainer per build.

Adding the identical `@TestPropertySource` here shares the cached context. It's safe: `startReporter()` overrides `purgeEnabled` and `retentionDays` via `ReflectionTestUtils` anyway, so the property values never reach the assertions.

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+        MongoAuditReporter reporter = new MongoAuditReporter();
+        context.getAutowireCapableBeanFactory().autowireBean(reporter);
+        ReflectionTestUtils.setField(reporter, "configuration", configuration);
+        ReflectionTestUtils.setField(reporter, "globalEnsureIndexOnStart", true);
+        ReflectionTestUtils.setField(reporter, "reporterEnsureIndexOnStart", true);
+        ReflectionTestUtils.setField(reporter, "purgeEnabled", false);
+        ReflectionTestUtils.setField(reporter, "retentionDays", 0);
+        overrides.accept(reporter);
+        reporter.afterPropertiesSet();
```
podlesrafal, line 244: Not a change request — a warning worth recording, because it isn't visible from this file.

The eight reporters started here are never stopped, so each leaks its `bulkProcessor.buffer(1s, …)` subscription for the lifetime of the class. The obvious cleanup (`@After { reporter.stop(); }`) would **break the suite**: `doStop()` calls `clientWrapper.releaseClient()`, and `MongoClientWrapper.releaseClient()` shuts down the client once its reference count hits zero — that's the shared testcontainer client, so every subsequent test would fail.

The leak is the lesser evil as things stand. A proper fix belongs in `TestMongoConnectionProvider` (hand out a per-reporter wrapper). Could you drop a one-line comment here so nobody "tidies this up" later and spends an afternoon on it?

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
sagarchauhan-qa: Sorted. I swapped referenceType/referenceId in r1t_1 and all 8 tests still passed.

Now comparing ordered entrySet() lists, which covers order and values together. Same mutation now fails with:

expected: [referenceType=1, referenceId=1, timestamp=-1]
but was:  [referenceId=1, referenceType=1, timestamp=-1]

Applied to the retention index on line 171 too.

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
sagarchauhan-qa: Sorted. The module printed the container startup line twice. Added the matching @TestPropertySource; it prints once now.
 
Your safety point holds: startReporter() sets purgeEnabled and retentionDays explicitly, so the property values never reach an assertion.

gravitee-am-reporter/gravitee-am-reporter-mongodb/src/test/java/io/gravitee/am/reporter/mongodb/audit/MongoAuditReporterIndexTest.java
sagarchauhan-qa: Sorted. TestMongoConnectionProvider returns the same wrapper from every accessor, and releaseClient() shuts down at refcount zero.

Comment added in startReporter() explaining why it's deliberate and that the fix belongs in TestMongoConnectionProvider.

podlesrafal (approved):

PR #8422, opened 2026-08-19
feat: store access_token and refresh_token in one batch

lgw-gravitee:
to test: 
1. mongo and jdbc 
2. ensure querry/command by enabling driver's logs
3. create an app with refresh_token grant_flow and obtain user's token (password or auth_code grant_flow) - two tokens should appear in tokens collection/table

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/oauth2/api/JdbcTokenRepository.java
+++ b/gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/oauth2/api/JdbcTokenRepository.java
@@ -31,8 +31,10 @@ import io.reactivex.rxjava3.core.Maybe;
 import io.reactivex.rxjava3.core.Observable;
 import io.reactivex.rxjava3.core.Single;
 import io.reactivex.rxjava3.schedulers.Schedulers;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.relational.core.query.Query;
+import org.springframework.data.relational.core.sql.SqlIdentifier;
 import org.springframework.r2dbc.core.DatabaseClient;
 import org.springframework.stereotype.Repository;
 
@@ -40,6 +42,7 @@ import java.time.LocalDateTime;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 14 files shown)


gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/oauth2/api/JdbcTokenRepository.java
```diff
@@ -59,6 +61,13 @@ public class JdbcTokenRepository extends AbstractJdbcRepository implements Token
     @Autowired
     private RetryOnConcurrencyFailureConfiguration retryOnConcurrencyFailureConfiguration;
 
+    private static final String ACCESS_TOKEN_PREFIX = "at_";
+    private static final String REFRESH_TOKEN_PREFIX = "rt_";
+
+    private static final List<String> TOKEN_COLUMNS = List.of(
```
sclarkamlab: `TOKEN_COLUMNS`, `bindMarkers` and `bindToken` are three hand maintained copies of the `JdbcToken` mapping. Correct today, but add a `@Column` to `JdbcToken` next year and `getTemplate().insert()` picks it up while this path silently drops it. `AbstractJdbcRepository` already has `FieldSpec` + `addQuotedFields` which keeps column name, getter and type in one place, any reason not to use that here?

gravitee-am-repository/gravitee-am-repository-tests/src/test/java/io/gravitee/am/repository/oauth2/api/TokenRepositoryTest.java
```diff
@@ -77,6 +79,74 @@ public void shouldCreateAndFindTokensByJti() {
...
+
+        AccessToken accessToken = newAccessToken("batch-at-" + suffix, "domain-id", "client-id", "user-id", "parent-jti-" + suffix);
+        accessToken.setAuthorizationCode("batch-code-" + suffix);
+        accessToken.setRefreshToken("batch-rt-" + suffix);
+        accessToken.setCreatedAt(createdAt);
+        accessToken.setExpireAt(expireAt);
+
+        RefreshToken refreshToken = newRefreshToken("batch-rt-" + suffix, "domain-id", "client-id", "user-id", "parent-jti-" + suffix);
+        refreshToken.setCreatedAt(createdAt);
+        refreshToken.setExpireAt(expireAt);
```
sclarkamlab, line 96: These are all happy path. Nothing fails the second row and asserts neither token landed, which is the bit the batch actually buys us. As it stands you could swap this back to two sequential inserts and the suite would still be green.

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/oauth2/MongoTokenRepository.java
```diff
@@ -95,6 +95,20 @@ public Single<AccessToken> create(AccessToken accessToken) {
...
+    public Completable create(AccessToken accessToken, RefreshToken refreshToken) {
+        if (refreshToken == null) {
+            return create(accessToken).ignoreElement();
+        }
+        if (refreshToken.getId() == null) {
+            refreshToken.setId(RandomString.generate());
+        }
+
+        return Completable
+                .fromPublisher(tokenCollection.insertMany(List.of(convert(accessToken), convert(refreshToken))))
```
sclarkamlab, line 108: `insertMany` is ordered but it isn't atomic without a transaction, so mongo can still end up with an access token and no refresh token. JDBC gets a single statement so it's all or nothing there. Not a regression, but the javadoc on `storeTokens` says single batch write and I'd read that as a guarantee on both backends.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/oauth2/api/JdbcTokenRepository.java
```diff
@@ -97,6 +106,54 @@ public Single<AccessToken> create(AccessToken accessToken) {
...
+    @Override
+    public Completable create(AccessToken accessToken, RefreshToken refreshToken) {
+        if (refreshToken == null) {
+            return create(accessToken).ignoreElement();
+        }
+        accessToken.setId(accessToken.getId() == null ? RandomString.generate() : accessToken.getId());
+        refreshToken.setId(refreshToken.getId() == null ? RandomString.generate() : refreshToken.getId());
+        LOGGER.debug("Create accessToken with id {} and refreshToken with id {}", accessToken.getId(), refreshToken.getId());
+
+        DatabaseClient.GenericExecuteSpec spec = getTemplate().getDatabaseClient().sql(batchInsertStatement());
```
sclarkamlab: We rebuild this string on every token issue, two streams plus the identifier quoting per request. `JdbcGroupRepository` does it once in `afterPropertiesSet`, same thing here would be nice given the PR is about making this path cheaper.

sclarkamlab (commented):
Had a proper look through this, green across the full testcontainer matrix so the multi row insert is fine on postgres/mysql/mariadb/mssql, and the gateway side is a clean swap. The win is real, one round trip and one commit instead of two, and the no refresh token path still falls back to the old single insert so nothing regresses there.

One perf note on the inline comment about rebuilding the statement. The old `getTemplate().insert()` path never went near `NamedParameterExpander`, the raw `sql()` call does, and it looks it up twice per execution keyed on the SQL string. Building a fresh string each time means it hashes the whole thing both times instead of reusing a cached hash. Small, but free to fix.

Also minor: `TokenRepository` in the `-api` module gained an abstract method, so anything implementing it out of tree breaks on upgrade. A `default` delegating to the two single creates would avoid that if we care.

gravitee-am-repository/gravitee-am-repository-tests/src/test/java/io/gravitee/am/repository/oauth2/api/TokenRepositoryTest.java
lgw-gravitee: Fair, added the failing-row coverage.

`TokenRepositoryBatchCreateTest` on the JDBC side collides the second row's PK and asserts neither token lands (plus the same in the other direction). I checked it actually discriminates: patching `create(at, rt)` back to two sequential inserts fails it with `expected null, but was:<AccessToken@...>`. Green on postgres, mysql, mariadb and mssql.

Worth being explicit that this can't live in the shared suite, which ties into your `insertMany` comment: the batch is only all-or-nothing on JDBC, where it's one multi-values INSERT. On Mongo an ordered `insertMany` is not a transaction, and testcontainers/local-stack run standalone, so there's no session to make it atomic. I've pinned that real behaviour in a mongo-only `TokenRepositoryBatchCreateTest` — it asserts the access token *does* survive a failing refresh-token doc — so nobody later banks on a rollback that never happens. Behaviour alone can't catch a revert there, so there's also a `MongoTokenRepositoryBatchTest` asserting a single `insertMany` with both docs and no `insertOne`; that one does go red on a revert. It mocks the driver, which is a new pattern here — say the word and I'll drop it.

Also added `TokenServiceTest#shouldNotReturnTokenWhenBatchStoreFails`: the store-fails path was stubbed to `Completable.complete()` everywhere before.

I'll tighten the `storeTokens` javadoc on your other thread so "single batch write" doesn't read as an atomicity guarantee on both backends.

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/oauth2/MongoTokenRepository.java
lgw-gravitee: I updated javadocs

sclarkamlab (approved):

PR #8425, opened 2026-08-19
fix(console): keep the factor screens usable without DOMAIN_RESOURCE

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7478

Follow-up to #8424. That PR moved the resources listing endpoint onto `DOMAIN_RESOURCE[LIST]` and merged before these two parts went up.

The Console reaches that endpoint from the factor screens, not only from the resources screen. `factors/new` and `factors/{factorId}` resolve the resource list, and a failed resolver stops the navigation. So a custom role that grants the factor permissions without `domain_resource_list` cannot open the factor screens at all on master today.

`ResourceService.findByDomainWhenPermitted` maps a 403 to an empty list and suppresses the error snackbar. Any other status still propagates. A new `FactorResourcesResolver` calls it, and the two factor routes use that resolver. Both factor components already disable the resource field when the list is empty, so the form works and the picker is simply unavailable.

The domain settings resources screen keeps `ResourcesResolver`. A 403 there means something is genuinely wrong, and it should still fail loudly.

The permission sweep from AM-7448 excluded this route while the guard was wrong. The exclusion comes out here, so the sufficiency test asserts that `domain_resource_list` alone opens the endpoint.

### Automated

From the repo root, with the local stack up:

```bash
npm --prefix gravitee-am-test run test -- specs/management/permissions/permission-sufficiency.jest.spec.ts specs/management/permissions/endpoint-permission-sweep.jest.spec.ts
```

99 tests, all green. The case to look for is `should admit domain_resource_list alone to /organizations/{organizationId}/environments/{environmentId}/domains/{domain}/resources`. This branch removes the exclusion that skipped it, so it fails against the old guard.

```bash
cd gravitee-am-ui && yarn test --test-path-pattern 'services/resource.service.spec.ts'
```

3 tests, all green: permitted, 403 gives an empty list, other errors propagate. Note the flag is kebab-case. `--testPathPattern` is rejected by the builder, and bare `npx jest` cannot run a spec at all because the jest preset comes from the Angular builder.

### Console, by hand

Walked end to end on a local stack. Build a role that can manage factors but cannot list
resources, then use the Console as a user holding it.

1. Get an admin token.

```bash
MGMT=http://localhost:8093/management
TOKEN=$(curl -s -u admin:adminadmin -X POST "$MGMT/auth/token" | jq -r .access_token)
```

2. Create the role, then set its permissions. Creation rejects a `permissions` property, so it is two calls.

```bash
ROLE=$(curl -s -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -X POST "$MGMT/organizations/DEFAULT/roles" \
  -d '{"name":"factor-admin-no-resources","assignableType":"ORGANIZATION"}' | jq -r .id)

curl -s -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -X PUT "$MGMT/organizations/DEFAULT/roles/$ROLE" \
  -d '{"name":"factor-admin-no-resources","permissions":["environment_list","environment_read","organization_tag_list","domain_list","domain_read","domain_settings_read","domain_factor_list","domain_factor_read","domain_factor_create"]}'
```

Every permission there earns its place. Without `environment_list` the Console renders an empty
shell and never reaches a domain. Without `organization_tag_list` the domain overview raises its
own "Access denied" toast, which muddies the check in step 6. `domain_resource_list` is the one
permission deliberately left out, and that omission is the whole test.

3. Create the user and give it the role.

```bash
USER=$(curl -s -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -X POST "$MGMT/organizations/DEFAULT/users" \
  -d '{"username":"factoradmin","password":"Password123!","firstName":"Factor","lastName":"Admin","email":"factoradmin@example.com","preRegistration":false}' | jq -r .id)

curl -s -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -X POST "$MGMT/organizations/DEFAULT/members" \
  -d "{\"memberId\":\"$USER\",\"memberType\":\"USER\",\"role\":\"$ROLE\"}"
```

4. Log in to the Console at `http://localhost:4200` as `factoradmin` / `Password123!`.

5. Open a domain, then Settings, then Security, then Multifactor Auth. Open an existing factor,
and separately click New.

6. Open the browser console. This is the check, and it is unambiguous:

```
GET .../domains/{domain}/resources  ->  403 (Forbidden)
```

The request fails, and the factor screen renders anyway, with no "Access denied" toast. That is
the fix: the 403 becomes an empty list, and the route proceeds. On master the same 403 rejects
the resolver and the router never leaves the page you were on.

7. Optional, to see the picker itself: the resource field only exists for factor plugins whose
schema has a `graviteeResource` widget, such as SMS and call factors. The MOCK factor has no
such field, so it proves the route but shows no picker. For the populated case, use an admin,
create a service resource whose plugin category matches the factor plugin, and reopen the
screen. The picker then lists it by name, which shows the read-only state above is the
permission and not an empty domain.

Screenshot to follow: the factor form for a role without `domain_resource_list`, showing the disabled resource field.

The upgrade note for AM-7478 says a missing `domain_resource_list` breaks the factor screens. Once this merges the factor screens degrade instead, so the note should say the resource picker is unavailable.

@gravitee-io/am

### DIFF
--- a/gravitee-am-ui/src/app/services/resource.service.spec.ts
+++ b/gravitee-am-ui/src/app/services/resource.service.spec.ts
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { TestBed } from '@angular/core/testing';
+import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
+import { HttpErrorResponse, provideHttpClient, withInterceptorsFromDi } from '@angular/common/http';
+
+import { AppConfig } from '../../config/app.config';
+
+import { ResourceService } from './resource.service';
+
+describe('ResourceService', () => {
+  let httpTestingController: HttpTestingController;
+  let resourceService: ResourceService;
+
+  const domainId = 'domain-1234';
+  const resourcesUrl = `${AppConfig.settings.domainBaseURL}${domainId}/resources`;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      teardown: { destroyAfterEach: false },
+      imports: [],
+      providers: [ResourceService, provideHttpClient(withInterceptorsFromDi()), provideHttpClientTesting()],
+    });
+
+    httpTestingController = TestBed.get(HttpTestingController);
+    resourceService = TestBed.get(ResourceService);
+  });
+
+  afterEach(() => {
+    httpTestingController.verify();
+  });
+
+  describe('findByDomainWhenPermitted', () => {
+    it('returns the resources when the user is permitted', (done) => {
+      const resources = [{ id: 'resource-1', name: 'my-smtp', type: 'smtp-am-resource' }];
+
+      resourceService.findByDomainWhenPermitted(domainId).subscribe((result) => {
+        expect(result).toEqual(resources);
+        done();
+      });
+
+      httpTestingController.expectOne({ method: 'GET', url: resourcesUrl }).flush(resources);
+    });
+
+    it('returns an empty list when the user lacks DOMAIN_RESOURCE[LIST]', (done) => {
+      resourceService.findByDomainWhenPermitted(domainId).subscribe((result) => {
+        expect(result).toEqual([]);
+        done();
+      });
+
+      httpTestingController
+        .expectOne({ method: 'GET', url: resourcesUrl })
+        .flush({ message: 'Permission denied' }, { status: 403, statusText: 'Forbidden' });
+    });
+
+    it('propagates errors other than 403', (done) => {
+      resourceService.findByDomainWhenPermitted(domainId).subscribe({
+        next: () => done.fail('expected the error to propagate'),
+        error: (error: unknown) => {
+          expect(error).toBeInstanceOf(HttpErrorResponse);
+          expect((error as HttpErrorResponse).status).toEqual(500);
+          done();
+        },
+      });
+
+      httpTestingController
+        .expectOne({ method: 'GET', url: resourcesUrl })
+        .flush({ message: 'Server error' }, { status: 500, statusText: 'Internal Server Error' });
+    });
+  });
+});
--- a/gravitee-am-ui/src/app/services/resource.service.spec.ts
+++ b/gravitee-am-ui/src/app/services/resource.service.spec.ts
@@ -15,9 +15,10 @@
  */
 import { TestBed } from '@angular/core/testing';
 import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
-import { HttpErrorResponse, provideHttpClient, withInterceptorsFromDi } from '@angular/common/http';
+import { provideHttpClient, withInterceptorsFromDi } from '@angular/common/http';
 
 import { AppConfig } from '../../config/app.config';
+import { SKIP_ERROR_SNACKBAR } from '../interceptors/http-request.interceptor';
 
 import { ResourceService } from './resource.service';
 
@@ -61,19 +62,18 @@ describe('ResourceService', () => {
         done();
       });
 
-      httpTestingController
-        .expectOne({ method: 'GET', url: resourcesUrl })
-        .flush({ message: 'Permission denied' }, { status: 403, statusText: 'Forbidden' });
+      const req = httpTestingController.expectOne({ method: 'GET', url: resourcesUrl });
+      expect(req.request.context.get(SKIP_ERROR_SNACKBAR)).toBe(true);
+      req.flush({ message: 'Permission denied' }, { status: 403, statusText: 'Forbidden' });
     });
 
-    it('propagates errors other than 403', (done) => {
+    it('returns an empty list when the request fails for any other reason', (done) => {
       resourceService.findByDomainWhenPermitted(domainId).subscribe({
-        next: () => done.fail('expected the error to propagate'),
-        error: (error: unknown) => {
-          expect(error).toBeInstanceOf(HttpErrorResponse);
-          expect((error as HttpErrorResponse).status).toEqual(500);
+        next: (result) => {
+          expect(result).toEqual([]);
           done();
         },
+        error: (error: unknown) => done(error),
       });
 
       httpTestingController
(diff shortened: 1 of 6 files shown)


gravitee-am-ui/src/app/services/resource.service.ts
```diff
@@ -29,6 +31,18 @@ export class ResourceService {
     return this.http.get<any>(this.resourcesURL + domainId + '/resources');
   }
 
+  // The factor screens render without the resource list, so a user who lacks
+  // DOMAIN_RESOURCE[LIST] gets an empty list rather than a failed route.
+  findByDomainWhenPermitted(domainId): Observable<any[]> {
+    return this.http
+      .get<any[]>(this.resourcesURL + domainId + '/resources', {
+        context: new HttpContext().set(SKIP_ERROR_SNACKBAR, true),
```
podlesrafal, line 40: `SKIP_ERROR_SNACKBAR` is set on the request unconditionally, but `catchError` recovers only 403. A 5xx (or `status: 0`) is therefore both silenced **and** rethrown: the resolver rejects, the router cancels the navigation, and nothing surfaces it — there is no `NavigationError` subscriber and no custom `ErrorHandler` in `src/app`, and `RouterModule.forRoot` configures none.

On master `findByDomain` carried no context, so the same 500 at least produced a `Server error` snackbar. After this change, clicking a factor silently does nothing — for every user, including one holding `domain_resource_list`. (401 still snackbars and 404 still redirects, since `SKIP_404_REDIRECT` isn't set.)

Worth making the suppression and the handling the same width. The simplest version is `catchError(() => of([]))`: the comment above already says the screens render without the list, and `factor.component.ts:139/169` marks the field readonly on an empty list either way. `memberships.resolver.ts:36` already does exactly this. That does mean dropping the `propagates errors other than 403` case, so it's a deliberate call rather than a free fix.

gravitee-am-ui/src/app/services/resource.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      });
+
+      httpTestingController
+        .expectOne({ method: 'GET', url: resourcesUrl })
+        .flush({ message: 'Permission denied' }, { status: 403, statusText: 'Forbidden' });
+    });
+
+    it('propagates errors other than 403', (done) => {
+      resourceService.findByDomainWhenPermitted(domainId).subscribe({
+        next: () => done.fail('expected the error to propagate'),
```
podlesrafal: `done.fail` is a jasmine API that jest-circus does not provide, and the runner here is unambiguously jest-circus: `jest` is 29.7.0, `jest.config.js` sets no `testRunner`, and `yarn.lock` has `jest-circus@npm:29.7.0` (:11977) with no `jest-jasmine2`. `@types/jest` still declares `DoneCallback.fail`, so this compiles and lints clean.

The line is unreachable today because the 500 does take the `error` branch — but it is the guard for exactly the regression this test exists to catch. Widen the `catchError` and `next` fires instead: `TypeError: done.fail is not a function` thrown inside an RxJS `next` handler goes through `reportUnhandledError` rather than failing synchronously, `done` is never called, and you get a 5s timeout instead of the intended message.

It is also the only `done.fail` in the repo. Suggest:

```ts
next: () => done(new Error('expected the error to propagate')),
```

gravitee-am-ui/src/app/services/resource.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    it('returns an empty list when the user lacks DOMAIN_RESOURCE[LIST]', (done) => {
+      resourceService.findByDomainWhenPermitted(domainId).subscribe((result) => {
+        expect(result).toEqual([]);
+        done();
+      });
+
+      httpTestingController
+        .expectOne({ method: 'GET', url: resourcesUrl })
```
podlesrafal: The `context:` option isn't covered anywhere. This TestBed registers no interceptor (`imports: []`, and `withInterceptorsFromDi()` only activates `HTTP_INTERCEPTORS`, which are provided in `app.module.ts`), and this test never inspects the outgoing request.

So the spec pins the 403 → `[]` mapping but not the snackbar suppression — and the interceptor's `tap` runs upstream of the service's `catchError`, so dropping the context option would bring `Access denied` back on every visit to both factor screens with all three tests still green.

One line closes it:

```ts
const req = httpTestingController.expectOne({ method: 'GET', url: resourcesUrl });
expect(req.request.context.get(SKIP_ERROR_SNACKBAR)).toBe(true);
req.flush({ message: 'Permission denied' }, { status: 403, statusText: 'Forbidden' });
```

Reasonable to decline — no other spec here asserts an `HttpContext` token, and the existing `SKIP_*` call sites in `domain.service.ts` / `protected-resource.service.ts` have no coverage at all.

podlesrafal (commented):
Reviewed the diff against the surrounding code. The shape is right: a separate `FactorResourcesResolver` for the two factor routes, the strict `ResourcesResolver` left in place on the domain settings screen, and a narrowly scoped new service method rather than loosening the existing one. I verified the claims in the description rather than taking them on trust — `paramsInheritanceStrategy: 'always'` makes the `route.parent.data` lookup sound, both factor components really do fall back to `property['readonly'] = true` on an empty list (`factor.component.ts:169`, `step2.component.ts:113`), and the backend still enforces the permission (`ServiceResourcesResource#list` → `checkAnyPermission(..., DOMAIN_RESOURCE, Acl.LIST)`, payload filtered to id/name/type), so this is a UX fix and not an authorization change.

One behavioural gap worth addressing: `SKIP_ERROR_SNACKBAR` is set for every status while `catchError` recovers only 403, so a 5xx on the resources endpoint now cancels the factor navigation with no message at all, where master showed `Server error`. Two smaller test points below. None of these block the fix, which does what the description says.

wajb (commented):
Is there value in displaying most factors if the user cannot access resources? It feels like a trap since most types require resource data and blocking access might be the clearer feedback that permissions are lacking.

gravitee-am-ui/src/app/services/resource.service.ts
sclarkamlab: Good catch. Done in `5dc9063c89`, it's `catchError(() => of([]))` now so the suppression and the handling are the same width. Taking the deliberate call: the point of the resolver is that the factor route never fails on this lookup, and the screen already renders readonly on an empty list, so swallowing the 5xx is consistent with that. Dropped the propagate case and the test with it.

gravitee-am-ui/src/app/services/resource.service.spec.ts
sclarkamlab: Done in `5dc9063c89`. The wider `catchError` took that test with it, it now asserts the empty list on a 500 and there's no `done.fail` left in the repo.

gravitee-am-ui/src/app/services/resource.service.spec.ts
sclarkamlab: Yeah, fair. Added in `5dc9063c89`, it's the two lines you suggested on the 403 test. Checked it bites too, dropping the `context` option fails that test and only that one.

sclarkamlab:
Only factors with a `graviteeResource` widget in their schema need the list, so OTP and recovery code are unaffected. Blocking the screen would break those for someone who legitimately has no `DOMAIN_RESOURCE`, which felt like the worse trap. For the SMS/email ones the field goes readonly rather than empty-and-clickable, so it does read as "you can't set this", just without stopping the rest of the page.

wajb (approved):

podlesrafal (approved):

PR #8426, opened 2026-08-19
fix: AM-7515 reconcile and report mongo index creation

wajb:
[AM-7515](https://gravitee.atlassian.net/browse/AM-7515)

Add retry resilience, reconciliation and logging to mongo index creation. This hardens replacing indexes that can be replaced and describes those that cannot - chiefly to reconcile unexpected state during upgrades.

- Index creation mirrors `MongoUtils.dropIndexes` retry pattern, handling concurrent builds and transient failures.
- Only indexes defined in AM are updated.
- Logging:
  - DEBUG = nothing to do.
  - WARN = collection diverged, or an index was left/replaced on purpose.
  - ERROR = a declared guarantee is missing.

## :memo: Test scenarios 
See Jira.

n/a

n/a

@gravitee-io/am 

[AM-7515]: https://gravitee.atlassian.net/browse/AM-7515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoIndexManager.java
+++ b/gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoIndexManager.java
@@ -1,0 +1,366 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.repository.mongodb.common;
+
+import com.mongodb.ErrorCategory;
+import com.mongodb.MongoClientSettings;
+import com.mongodb.MongoCommandException;
+import com.mongodb.client.model.IndexModel;
+import com.mongodb.client.model.IndexOptions;
+import com.mongodb.reactivestreams.client.MongoCollection;
+import io.reactivex.rxjava3.core.Completable;
+import io.reactivex.rxjava3.core.Flowable;
+import io.reactivex.rxjava3.core.Single;
+import lombok.CustomLog;
+import lombok.experimental.UtilityClass;
+import org.bson.BsonDocument;
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+/**
+ * Applies a collection's declared indexes to MongoDB, and reconciles the collection with the
+ * declaration when they have drifted apart.
+ * <p>
+ * The rule this applies is that <em>AM owns its index names, not its index keys</em>. An index
+ * sitting on a name AM declares is AM's to replace, so it is dropped and rebuilt to match the
+ * declaration - this is what makes an upgrade that changes a definition converge. An index sitting
+ * on a key AM declares but under somebody else's name is left alone: its coverage is already there,
+ * and it may well have been created deliberately by an operator. What AM does not do is stay quiet
+ * about either case; see {@link MongoIndexReport}.
+ *
+ * @author GraviteeSource Team
+ */
+@UtilityClass
+@CustomLog
+public final class MongoIndexManager {
+
+    private static final int MAX_ATTEMPTS = 3;
+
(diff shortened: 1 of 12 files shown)


gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoIndexManager.java
```diff
@@ -0,0 +1,352 @@
...
+            MongoIndexReport.failed(collectionName, name, "the id index cannot be replaced");
+            return Completable.complete();
+        }
+        final String detail = "replaced " + describe(existing) + " with " + describe(model);
+        log.warn("Index {} of collection {} no longer matches its declaration - dropping and rebuilding it ({}).",
+                name, collectionName, detail);
+        return dropIndex(collection, name)
+                .andThen(createOne(collection, model))
+                .doOnComplete(() -> MongoIndexReport.rebuilt(collectionName, name, detail))
+                .onErrorResumeNext(error -> reportFailure(collectionName, model, error));
```
sclarkamlab: This sends every post-drop failure to `reportFailure`, including 85, but `reconcileOne` just above routes 85 to `reportShadowed`. I checked on mongo 7 and TTL-vs-TTL on the same key under a different name does give you 85. So if the rebuild hits a foreign index with the same TTL, you get an ERROR saying expired records will never be removed when retention is fine. `e1` is the TTL name across all the oauth2 collections, and `MongoEventRepository`'s value is operator-configurable, so this isn't theoretical. Same branch as `reconcileOne` would do it, though you'd need to pass the snapshot list into `rebuild` to reuse `reportShadowed`.

gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoIndexManager.java
```diff
@@ -0,0 +1,352 @@
...
+     * but if the declaration asked for a guarantee the incumbent cannot provide, something really is lost
+     * and this is a failure, not a note.
+     */
+    private static void reportShadowed(String collectionName, IndexModel model, List<Document> existing) {
+        final String name = indexName(model);
+        final String key = keyOf(model).toJson();
+        final Document incumbent = firstMatching(existing, index -> sameKey(index.get("key", Document.class), model));
+        // an incumbent created earlier in this same pass is not in the snapshot we read
+        final String incumbentName = incumbent != null ? incumbent.getString("name") : "another index";
+        final String missing = incumbent != null ? guaranteeMissingFrom(incumbent, model) : null;
```
sclarkamlab: `incumbent == null` falls through to `missing == null` and reports SHADOWED, which reads as nothing lost. But we only read the snapshot once, before the pass, so anything created during the pass isn't in it. I tried it. A batch with two declarations on one key gets rejected wholesale. Reconcile then creates the first and takes 85 on the second with `incumbent` null, so a TTL that never got created comes out as benign. I'd treat not-found as failed rather than fine, or re-read the list when the incumbent isn't there.

gravitee-am-repository/gravitee-am-repository-mongodb-api/src/main/java/io/gravitee/am/repository/mongodb/common/MongoIndexReport.java
```diff
@@ -0,0 +1,96 @@
...
+    static void shadowed(String collection, String index, String detail) {
+        record(new IndexOutcome(collection, index, Status.SHADOWED, detail));
+    }
+
+    static void failed(String collection, String index, String detail) {
+        record(new IndexOutcome(collection, index, Status.FAILED, detail));
+    }
+
+    private static void record(IndexOutcome outcome) {
+        OUTCOMES.put(outcome.collection() + '.' + outcome.index(), outcome);
```
sclarkamlab, line 71: This keys on collection name with no database. `DataPlaneRegistryImpl` holds a provider per data plane, and each one has a `MongoUserRepository` on a collection called `users`. So they overwrite each other's outcomes, and an ENSURED from one plane hides a FAILED from another. `action_lease` clashes the same way across the management and gateway packages. Passing `getNamespace().getFullName()` as the collection instead of `getCollectionName()` would sort it.

sclarkamlab (commented):
Nice catch on the retry. The old `zipWith(Flowable.range(1, 3))` completed successfully once the range ran out, so an index that never got created reported as fine. `Attempt.spent()` re-throwing is the real fix here. Tests are good too, especially asserting on report status rather than just index presence.

Couple of things outside the diff that are the same shape as what you're fixing. `MongoAuditReporter.java:413-415` still does `deleteOldIndexes.andThen(createNewIndexes).subscribe()`, so a failed sweep skips index creation and the error goes to the global handler. That's what you fixed in `MongoUserRepository:812-816` with `onErrorComplete`. And `MongoOrganizationUserRepository:652-665` is still on the old try/catch pattern, so `organization_users` gets none of the reconciliation despite declaring the same `rt1ri1u1s1_unique` index.

One minor thing. Nothing outside tests reads `MongoIndexReport`, and nothing evicts entries, so `reporter_audits_<domainId>` builds up per domain for the life of the JVM. Fine if that's deliberate groundwork, worth a line in the description either way.

sclarkamlab (approved):

PR #8427, opened 2026-08-20
feat: trusted domains become the single source of trust (AM-7372)

lgw-gravitee:
A trusted domain now declares which kind of trust it represents — SPIFFE or token exchange — and describes where its signing keys come from in one shape regardless of kind: a JWKS URL, an inline JWK set, or a PEM certificate. Names are unique per (kind, name) within a security domain, so a SPIFFE and a token-exchange trusted domain may share a name; the constraint is enforced by a unique index on both storage backends — partial on MongoDB, plain on PostgreSQL, MySQL and MariaDB, filtered on SQL Server — with the issuer column bounded at 512 characters so the key stays inside InnoDB's 3072-byte limit at utf8mb4.

Trusted-issuer keys are resolved through the same hardened, cached fetcher SPIFFE bundles already use, instead of a separate Nimbus JWKS source with fixed timeouts and no limits. A JWKS URL on an unsecured scheme or a private address is refused unless the security domain permits it, the advertised response-size limit is enforced by aborting the read rather than buffering, and the per-issuer JWTProcessor cache is retired in favour of the shared bundle cache, so keys survive a transient fetch failure and refresh on a key-id miss.

The limits governing retrieval — unsecured scheme, private address, fetch timeout, response size, cache TTL and cache size — leave the SPIFFE settings block for oidc.keyRetrievalSettings, which honestly describes what they govern. The JWT validation policy that is genuinely SPIFFE-specific stays where it is. Values written before the move are not lost: the old fields stay readable, deprecated and nullable, OIDCSettings falls back to them when the new block is absent, and an idempotent upgrader relocates them verbatim — a limit deliberately set to zero included.

Trusted domains become resident in the gateway. The gateway's event-type resolution had no case for TRUST_DOMAIN, so every trust-domain event resolved to no type and was silently discarded, and the gateway compensated with a database read on every token request. A trusted-domain event and its resolution case close that gap, and a manager holds a security domain's trusted domains in memory with two kind-scoped lookups — by name for SPIFFE, by issuer for token exchange. Readiness is reported per trusted domain behind a marker registered before the preload starts, so a domain reports as synchronised only once its trusted domains are loaded.

Every trusted issuer declared inline migrates to a trusted domain of its own, named from the slugified full issuer URL so the same configuration migrates to the same names in every environment. Where two issuers would slugify alike, both — never just the one declared second — carry a digest of their issuer, keeping the outcome free of declaration order. Entities are written straight to the repository, since TrustDomainService applies an SSRF guard the inline shape never had and an issuer resolving to a private address, legal when it was configured, would fail the upgrade instead of migrating. The migration runs behind a system task rather than the plain async upgrader because it creates rows under a unique index, and re-running neither duplicates nor overwrites.

The inline list then stops being persisted and becomes a projection over token-exchange trusted domains: reads assemble it, writes translate into entity creates, updates and deletes, so an issuer omitted from a written list stops being trusted. The management API keeps the permission it accepts today; the trusted-domains API requires the trusted-domain permission.

Console gains one Trusted Domains page under Security. The list shows each entry's kind and tolerates a SPIFFE and a token-exchange trusted domain sharing a name; creation asks for the kind first; and key material is edited through one control for both kinds. The trusted-issuers tab is gone from Token Exchange, which links to Trusted Domains instead, and its settings page stops sending trustedIssuers altogether — a settings save carrying a stale list would have replaced every token-exchange trusted domain. The cross-links are hidden from principals without the trusted-domain permission, who would otherwise be led to a page AuthGuard refuses.


lgw-gravitee:
Created Stacked PRs

PR #8429, opened 2026-08-20
feat: pin the storage of an identity provider that reuses the system cluster

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7264

On a shared cloud data plane a domain admin can point a mongo idp at the AM database and pick a collection name that collides with a platform one. This adds a mode that takes the choice away at creation time.

Two settings in `gravitee.yml`, per the ticket:

```yaml
repositories:
  system-cluster-idp:
    pin-database: true
    prefix-users-collection: true
```

`pin-database` takes the database from the repository layer settings. `prefix-users-collection` forces the users collection to `idp_<id>`. Each defaults to the installation type, so managed cloud gets both and everywhere else gets neither, and an operator can set either explicitly. They are independent, so I kept one flag on the model rather than two columns: `systemClusterRestricted` is set when either rule fires, and `MongoAbstractProvider` checks `isPinDatabase()` alongside it so a provider flagged only for the collection keeps the database it was given.

Once a provider carries the flag, `useSystemCluster`, `database`, `usersCollection` and `datasourceId` are all frozen. An update that tries to turn `useSystemCluster` on for a provider created without it is rejected too, since pinning on update would move the collection and hide the users already in the old one.

A create request cannot choose the id while the collection rule is on. `resolveId` in `IdentityProviderServiceImpl` drops a supplied id and generates one, keeping it only for a system create or an `AutomationNewIdentityProvider`. The id becomes the collection name, so a caller that picked one could point a new provider at someone else's storage.

The flag marks a provider created under the mode. Anything that exists today carries `useSystemCluster` with the flag unset and the policy leaves it alone whatever the settings become later. No back-fill.

`resolvePlatformDatabase` mirrors `MongoConnectionProvider#getDatabaseName`, which is private to the repository plugin. That means honouring `use-management-settings` and falling back to `gravitee-am`. Get this wrong and the platform pins one database while the node reads another, which is exactly what happens on a stock install where `system-cluster` is `gateway` and the gateway scope reuses the management block.

The policy only pins a database when the system cluster scope names a mongo repository. The packaged `gravitee.yml` carries a mongodb block on a jdbc platform too, so without that check it would pin a database with no store behind it. The collection rule still applies there.

The database key stays in the configuration rather than being removed. The plugin schema requires it unless a datasource is named, so dropping it would leave the stored config invalid and block every later update.

The automation API is declarative, so a re-applied manifest carries the storage the operator wrote rather than what the platform pinned. `carryPinnedStorage` copies the pinned values on before the guard sees it.

Console side, the edit screen marks `database` and `usersCollection` readonly. The checkbox widget only binds a form control when `readonly` is falsy, so the toggle reads the same flag and ends up disabled. Angular drops a disabled control from the form value, so the component sends the stored storage settings back with the update. The creation screen keeps the fields editable and adds a hint to whichever ones the active rules will overwrite, because the schema makes `usersCollection` mandatory and the id that drives it does not exist yet.

`GET /platform/configuration/installation` now returns a typed `InstallationConfiguration` with both rules, so the console reads them instead of guessing. The SDK returns it properly now instead of `void`.

## :memo: Test scenarios

Unit tests cover each rule on its own and both together, the scope resolution against `use-management-settings`, and the packaged database default. Cloud jest suite is green on mongo, 16 tests. Repository round trip runs on mongo and jdbc.

Edit screen of a pinned provider shows the mongo storage fields readonly and the system cluster toggle disabled. Everything else on the form stays editable. No new screen, no layout change.

`identities` gets a `system_cluster_restricted` column defaulting to false. The changelog has a `columnExists` precondition with `MARK_RAN` so a rerun is safe.

`MongoAbstractProvider` resolves the database from its own node's client wrapper. A management API and a gateway can name different scopes in `repositories.system-cluster` and reach different databases for the same provider. That is existing `useSystemCluster` behaviour and I have not changed it.

@gravitee-io/am


gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/MongoAbstractProvider.java
```diff
@@ -83,6 +83,14 @@ public void afterPropertiesSet() {
...
 
         this.clientWrapper = this.buildClientWrapper(systemScope);
+        // The platform owns the database of a restricted provider, so take it from the client
+        // wrapper of this node the way configureDatasourceClient() does for the datasource path.
+        if (this.identityProviderEntity != null
+                && this.identityProviderEntity.isSystemClusterRestricted()
+                && this.configuration.isUseSystemCluster()
+                && !shouldUseDatasource()) {
+            this.configuration.setDatabase(this.clientWrapper.getDatabaseName());
+        }
```
sclarkamlab: refactor this if statement into a readable function

docs/mapi/openapi.yaml
```diff
@@ -15469,8 +15472,6 @@ components:
             type: string
         external:
           type: boolean
-        id:
```
sclarkamlab: why has this changed?

gravitee-am-service/src/main/java/io/gravitee/am/service/model/NewIdentityProvider.java
```diff
@@ -29,6 +30,11 @@
 @Setter
 public class NewIdentityProvider {
 
+    /**
+     * Set by the internal default-provider path only. The id becomes the users collection name, so a
+     * request must not choose it.
+     */
+    @JsonIgnore
```
sclarkamlab: Is this right? Surely this could be considered a breaking change? Why not ignore the Id silently if the system is in cloud mode

gravitee-am-ui/src/app/services/cloud-mode.service.ts
sclarkamlab: Double check the changes here

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/IdentityProviderServiceImpl.java
```diff
@@ -61,6 +61,7 @@
 import io.gravitee.am.service.model.UpdateIdentityProvider;
 import io.gravitee.am.service.reporter.builder.AuditBuilder;
 import io.gravitee.am.service.reporter.builder.management.IdentityProviderAuditBuilder;
+import io.gravitee.am.service.idp.SystemClusterIdpPolicy;
```
podlesrafal: Nit: import order. `io.gravitee.am.service.idp.SystemClusterIdpPolicy` lands after `io.gravitee.am.service.reporter.*`, so the block is no longer alphabetical. Same in two other files touched here:

- `gravitee-am-service/src/test/java/io/gravitee/am/service/IdentityProviderServiceTest.java:60` (after `service.model.UpdateIdentityProvider`)
- `gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/test/java/io/gravitee/am/management/handlers/automation/AutomationJerseySpringTest.java:36` (before `service.ReporterService`)

The build does not enforce import order, so this is purely cosmetic — just an IDE optimise-imports away.

gravitee-am-ui/src/app/domain/settings/providers/provider/settings/settings.component.ts
```diff
@@ -183,10 +191,27 @@ export class ProviderSettingsComponent implements OnInit {
...
+
+  /**
+   * Angular leaves a disabled control out of the form value, so the system cluster flag of a pinned
+   * provider would reach the server as undefined and read as a change. Send back what is stored.
+   */
+  private withPinnedStorage(configuration: any): any {
+    if (!this.provider.systemClusterRestricted || typeof this.provider.configuration !== 'string') {
+      return configuration;
+    }
+    const stored = JSON.parse(this.provider.configuration);
```
podlesrafal, line 209: Nit: `JSON.parse` here has no try/catch. Risk is low in practice — a restricted provider's configuration was written by the platform, so it is always valid JSON, and the `typeof === 'string'` guard above already covers the null case.

Worth noting this file already parses the same string unguarded at lines 109, 143 and 169, so the new code is consistent with what is there. If it is ever hardened, it should probably be all four call sites at once rather than this one.

gravitee-am-test/playwright/pages/identity-provider-settings.page.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /** Picks a provider type in step 1 of the wizard and moves to the configuration step. */
+  async chooseType(label: string): Promise<void> {
+    await this.page.getByText(label, { exact: true }).click();
+    await this.page.locator('button:has-text("Next")').click();
+  }
+
+  get databaseField(): Locator {
+    return this.page.locator('input[placeholder="The database used to run query"]');
```
podlesrafal: Nit: `databaseField` and `usersCollectionField` (line 40) match on placeholder copy (`"The database used to run query"` / `"The collection used to run query"`). Formally brittle — a wording change in the plugin schema silently breaks the spec with a timeout rather than a clear failure.

In practice the text comes from the mongo IdP plugin schema and is stable, so this is acceptable as-is. Flagging it mainly so it is a conscious choice; a `formControlName`/`data-testid` hook would be sturdier if one is available on those inputs.

podlesrafal (commented):
A few nits from a second pass. All optional — nothing here blocks the merge.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/IdentityProviderServiceImpl.java
sclarkamlab: Good catch, fixed in `0b7ad08787`. All three files, import sits between `service.exception.*` and `service.model.*` now.

gravitee-am-ui/src/app/domain/settings/providers/provider/settings/settings.component.ts
sclarkamlab: Agreed, leaving it. Same reasoning as yours: the config comes back from the server so it's always valid JSON, and the three existing parses at 109, 143 and 169 are unguarded too. If we harden it I'd rather do all four in one go than leave the file half and half.

gravitee-am-test/playwright/pages/identity-provider-settings.page.ts
sclarkamlab: Fair point, swapped both to `input[name="database"]` and `input[name="usersCollection"]` in `d1538d9caa`. ajsf's material input binds `[attr.name]="controlName"` so the name is the schema key, and that's what the other page objects use. Haven't run playwright against it yet, will do before merge.

sclarkamlab: Ran it, green. 3 passed, the edit test covers both selectors.

Worth flagging while I was in there: the spec skips on any normal local stack. It gates on `readsPinStorage`, and nothing we ship turns those two rules on outside managed cloud. `--cloud` doesn't help either, the fixture can't create a domain in DEFAULT there ("Data Plane [default] is not linked to this environment"). I had to set `repositories.system-cluster-idp.pin-database` and `prefix-users-collection` by hand on a standalone stack. Probably wants a documented way to run it, separate ticket.

podlesrafal (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/IdentityProviderServiceImpl.java
```diff
@@ -219,9 +229,25 @@ private Single<IdentityProvider> innerCreate(IdentityProvider identityProvider)
                 });
     }
 
-    private static IdentityProvider prepareIdp(NewIdentityProvider newIdentityProvider, ReferenceType domain, String domain1, boolean system) {
+    /**
+     * A caller-supplied id would point a new provider at another provider's collection, so it is
+     * honoured only on an internal path: a system provider, or the automation API's payload type.
```
wajb, line 234: nitpick: use US English ("honour" spelling in a few places).

sclarkamlab: Spoke with William, ok to merge as is

PR #8430, opened 2026-08-20
feat: AM-7265 allow default mongo IDPs to use data plane

wajb:
[AM-7265](https://gravitee.atlassian.net/browse/AM-7265)

When `repositories.system-cluster` is set to "gateway", sync domains' default mongo IDPs to their associated data planes. Historically default mongo IDPs always synced to the management repository settings.

New default mongo IDPs persist `useSystemCluster=true`, like non-system instances, when nodes configure `repositories.system-cluster` as "gateway"; existing IDPs are unaffected since they do not persist `useSystemCluster` at all and fall back to the existing management scope sync.

In the "use system cluster" mode, default IDPs use the database name configured for the data plane - so this is synced for default IDPs only. It's a latent inconsistency that non-system IDPs use the data plane's client but the database still defined on the form - but this behaviour is preserved.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7265]: https://gravitee.atlassian.net/browse/AM-7265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
+++ b/gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
@@ -396,7 +396,7 @@ repositories:
   # specify which scope is used as reference
   # to initialize the IdentityProviders with the "use system cluster"
   # option enabled (only management and gateway scopes are allowed as value)
-  system-cluster: management
+  system-cluster: gateway
   # Management repository is used to store global configuration such as domains, clients, ...
   # This is the default configuration using MongoDB (single server)
   # For more information about MongoDB configuration, please have a look to:
--- a/gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/MongoAbstractProvider.java
+++ b/gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/MongoAbstractProvider.java
@@ -97,7 +97,14 @@ public abstract class MongoAbstractProvider implements InitializingBean {
             final var provider = this.dataPlaneRegistry.getProviderById(this.identityProviderEntity.getDataPlaneId());
             // make sure the DataPlane plugin is a Mongo one
             if (provider.canHandle(ConnectionProvider.BACKEND_TYPE_MONGO)) {
-                return provider.getClientWrapper();
+                final ClientWrapper<MongoClient> clientWrapper = provider.getClientWrapper();
+                // Only a domain's default identity provider follows the data plane's database: it has no
+                // connection settings of its own to honor. A user-configured provider keeps the database
+                // from its own form to preserve historical behavior.
+                if (this.identityProviderEntity.isSystem()) {
+                    this.configuration.setDatabase(clientWrapper.getDatabaseName());
+                }
+                return clientWrapper;
             }
         }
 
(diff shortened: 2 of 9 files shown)


sclarkamlab (commented):
Nice change and the tests are good. One thing I'd want sorted before this merges though, see the comment on the gateway yml.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -396,7 +396,7 @@ repositories:
   # specify which scope is used as reference
   # to initialize the IdentityProviders with the "use system cluster"
   # option enabled (only management and gateway scopes are allowed as value)
-  system-cluster: management
+  system-cluster: gateway
```
sclarkamlab, line 399: This flips it for user configured mongo IdPs too, not just the default ones. Anything with `useSystemCluster` ticked was resolving to `commonConnectionProvider.getClientWrapper("management")` before, now it hits the data plane branch at `MongoAbstractProvider:96` and gets the gateway client instead. The `database` is kept but the cluster isn't, so on a setup with `repositories.gateway.use-management-settings: false` those IdPs move cluster on upgrade and logins break. Was that intentional? Feels like it needs an upgrade note at minimum, or a guard so only system IdPs follow the new scope.

sclarkamlab: To be clear, AM-7265 covers the default IdP case so the flip itself is fine by me. It's the user configured IdPs with the flag ticked I'm asking about, the ticket doesn't mention those.

wajb: This is only the defaults of the templates so this changes the OOTB behaviour for new installations and upgrades that preserve the installation settings won't be affected.

Having said that, the `use-management-settings` should be flipped too to be consistent and for logins to be functional OOTB. This is something I've outlined in the testing steps (jira). I can look into whether we can safely align helm defaults too. Wdyt? Ok to flip the default as long as it's self-consistent?

wajb: In fact we can keep `use-management-settings: true` OOTB because it still assumes the simple setup of sharing a single database. That way we don't need to add boilerplate for dataplanes here.

sclarkamlab: Yeah that sounds good to me, consistency is key!

sclarkamlab (approved):

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
wajb: I decided not to add the settings to the chart in this change. It's the one deployment path where an upgrade silently changes the value for existing installations. Leaving Helm alone means Helm installs keep resolving management - the current, safe behaviour. The distribution templates carry the new default for fresh standalone and Docker installs only.

PR #8431, opened 2026-08-21
fix(management-api): check permissions before parsing query parameters

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7476

Several management API GETs read their query parameters before they check permissions. A caller with no rights on the domain gets a server error instead of a refusal, and a caller who leaves a parameter out gets one too.

The permission check now runs first on all of them, so an unauthorised caller gets 403 rather than a server error.

One deliberate exception. The protected-resource `type` is a typed enum parameter, so Jersey converts it before the resource method runs and an unrecognised value 400s ahead of the permission check. I kept it that way. A rejected enum value tells the caller nothing the OpenAPI spec doesn't already publish, and taking `type` as a String to defer the parse would cost the typed signature and the generated schema. There's a test pinning it.

Also in here:

- The `type` parameter is optional on both protected-resource GETs. The list returns every type when it's absent, and read-by-id works on the id alone.
- An unrecognised `type` gives 400. Jersey reports a failed parameter conversion as 404 by default, so a converter maps it.
- `/applications/search/_cursor` gives 400 for a missing or malformed cursor, rather than an NPE out of the base64 decode.

DELETE keeps `type` required. It's a guard on a destructive call, so I left it as it is.

`ProtectedResource.Type.fromString` is gone. It's the method in the ticket's stack trace, nothing called it any more, and the converter does the same job without swallowing a null into an unchecked throw.

The generated TS SDK still marks `type` required on both operations. Regenerating it needs a running management API, and every caller supplies `type` today, so the SDK is just stricter than the API. The console always sends it too.

## :memo: Test scenarios

Unit coverage in the branch. Full reactor suite is green. The shared repository suite covers the untyped query on Mongo and JDBC both.

The jest permission sweeps drove these routes for the first time, so their exclusions are gone. The three analytics and cursor routes moved to the sufficiency-only list, because a granted caller still needs a real time window or cursor.

No UI changes.

The list now accepts no `type`, which widens the contract rather than breaking it. The repositories drop the type filter when it's null, on Mongo and JDBC both.

Analytics 400s come back as JSON now. They were a bare string under an `application/json` content type before, so nothing could parse them. Same status and same message, just a body that parses.

@gravitee-io/am


gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ProtectedResourceResource.java
```diff
@@ -84,13 +84,15 @@ public ProtectedResourceFlowsResource getFlowsResource() {
...
+            @ApiResponse(responseCode = "400", description = "Unrecognised type"),
             @ApiResponse(responseCode = "500", description = "Internal server error")})
     public void get(
             @PathParam("organizationId") String organizationId,
             @PathParam("environmentId") String environmentId,
             @PathParam("domain") String domainId,
             @PathParam("protected-resource") String protectedResourceId,
-            @NotNull @QueryParam("type") ProtectedResource.Type type,
+            @Parameter(description = "Asserts the resource has this type. The resource is read on its id alone when it is absent.")
+            @QueryParam("type") ProtectedResource.Type type,
```
wajb, line 95: question: I don't understand the asymmetry between GET and DELETE tbh. Recall the redundant `type` param tripped me up in the recent review [here](https://github.com/gravitee-io/gravitee-access-management/pull/8406#discussion_r3802910912) too. Can we remove required from both, or remove the parameter altogether?

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/AnalyticsParam.java
```diff
@@ -107,54 +106,33 @@ public void setField(String field) {
...
 
-    public void validate() throws WebApplicationException {
-        if (type.getValue() == null) {
-            throw new WebApplicationException(Response
-                    .status(Response.Status.BAD_REQUEST)
-                    .entity("Query parameter 'type' is not valid")
-                    .build());
+    public void validate() {
+        if (type == null || type.getValue() == null) {
+            throw new BadRequestException("Query parameter 'type' is not valid");
```
wajb, line 111: thought: should we isolate layers (model/resource/etc.) by throwing agnostic validation errors rather than assuming the data is request driven? In some places we wrap exceptions to provide such context - but worth checking what's consistent. The usage is unchanged here though, so not a blocker.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/ProtectedResourceTypeParamConverterProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.Locale;
+
+/**
+ * Without this, Jersey reports an unrecognised {@code type} query parameter as a 404, because a
+ * failed parameter conversion is a {@code NotFoundException} by default.
+ */
+public class ProtectedResourceTypeParamConverterProvider implements ParamConverterProvider {
```
wajb: thought: should we have special handling for this single property? Strong typing is better for the OAS and we could think about updating other fixed types (where sets are well defined) - where a generic mapping could drive the response mapping - or loosening this parameter to match surrounding convention. I'm not sure the complexity is worth it for the fix here.

gravitee-am-test/specs/management/permissions/fixtures/permission-sweep-tables.ts
```diff
@@ -75,6 +65,14 @@ export const SUFFICIENCY_ONLY_EXCLUDED: Record<string, string> = {
   // INSTALLATION is only relevant to the PLATFORM tier, so it cannot be granted by an
   // organization-assignable role and this technique cannot reach it.
   '/platform/installation': 'PLATFORM-tier permission, not grantable at organization level',
+
+  // Authorisation runs first on these (AM-7476), so the negative sweep drives them, but a granted
+  // caller still gets 400 without a genuine time window, analytics type, or cursor to supply.
```
wajb: nitpick: remove story-related comment.

wajb (commented):
> The generated TS SDK still marks type required on both operations. Regenerating it needs a running management API...

Can you regenerate this to keep it in sync?

gravitee-am-test/specs/management/permissions/fixtures/permission-sweep-tables.ts
sclarkamlab: Man this gets past me every time!!

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/resources/organizations/environments/domains/ProtectedResourceResource.java
sclarkamlab: Good catch, and it's worse than the asymmetry. `shouldReturnBadRequest_deleteWithoutType` granted no permission and still asserted 400, because bean validation runs `@NotNull` before the method body. DELETE was leaking the same thing this PR fixes on GET.

Dropped it in `d95ecc0e8f`. `ProtectedResourceServiceImpl.delete` already guarded on `expectedType != null` so the service needed nothing. That test is now a 403, plus a 204 for delete-by-id.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/AnalyticsParam.java
sclarkamlab: Disagree for this class. `AnalyticsParam` is a JAX-RS binding bean, `@QueryParam` and `@Parameter` on every field, and it lives in the rest module. It isn't a domain model, so a JAX-RS exception is the right layer.

It threw `WebApplicationException` before this PR, same package. I swapped it for the subclass so the error body comes back as JSON.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/ProtectedResourceTypeParamConverterProvider.java
sclarkamlab: The provider is the fix. Jersey turns a failed `@QueryParam` enum conversion into a `NotFoundException`, so an unrecognised type was 404ing. `shouldReturnBadRequest_whenTypeUnknownAndPermissionDenied` covers it.

A `String` param would 400 too, but the enum drops out of the OAS and the SDK. Agree on not generalising it, there's one enum query param to serve.

gravitee-am-test/specs/management/permissions/fixtures/permission-sweep-tables.ts
sclarkamlab: Man this gets past me every time!!

Gone in `65bc659989`. I'd left two more in `protected-resource-tier.jest.spec.ts` and `ProtectedResourceResourceTest`, so I took all three.

sclarkamlab:
Regenerated in `4f3a4cce5d`, all three ops now take `type?`. Spec regenerated in `77e6d743c8` for the delete change.

Didn't need a running instance in the end. `scripts/regen-oas.sh` does a build-time scan, so I served the generated `docs/mapi/openapi.yaml` on a local port and pointed the SDK generator at that. Kept the diff to `DomainApi.ts` and `ProtectedResourceApi.ts` and reverted the rest. There's stale drift in `CspSettings` and a property reorder in `NewMcpTool`/`UpdateMcpTool` sitting on master, worth a separate pass.

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/model/AnalyticsParam.java
wajb: You're right, I misread its package. Thanks!

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/ProtectedResourceTypeParamConverterProvider.java
wajb: So the complexity is justified because it's a one-off? Not sure I follow. It sets an awkward precedent for new enums IMO.

wajb (approved):

gravitee-am-management-api/gravitee-am-management-api-rest/src/main/java/io/gravitee/am/management/handlers/management/api/provider/ProtectedResourceTypeParamConverterProvider.java
sclarkamlab: You're right and I was wrong, sorry. I answered a question you hadn't asked.

I also told you there was one enum query param to serve. There are three. `ReferenceType` on the org roles list, and `Template` on five form and email endpoints, both have the same bug. Measured on the stack before I changed anything:

```
GET /organizations/DEFAULT/roles?type=NOT_A_TYPE          -> 404, empty body
GET /organizations/DEFAULT/forms?template=NOT_A_TEMPLATE  -> 404, empty body
```

So the one-off left two params both called `type` answering the same bad input differently.

Made it generic in `b2d464e2ef`. `EnumParamConverterProvider` claims `rawType.isEnum()` instead of one class, and it came out shorter than the version you reviewed. Roles and forms now 400 with the value and the allowed set. No param enum defines `fromString`, so nothing gets overridden. New tests in `FormsResourceTest`, full rest module suite green at 605.

The converter also uppercases, which I should have mentioned. That part matters. We serialise `type` as `mcp_server` and the console passes it straight back into the query string, `overview.component.ts:45` and `tools.component.ts:104`. `?template=login` is the same shape and 404s today, so forms had that one too.

PR #8432, opened 2026-08-21
feat: trusted domains gain a SPIFFE matcher, shared key material and guarded retrieval (AM-7372)

lgw-gravitee:
The SPIFFE trust domain an authority vouches for moves out of the name into a
matcher of its own, spiffeTrustDomain, leaving the name a plain label. The name
stays unique within a security domain and is now free-form and renameable; the
DNS-label rule that gave it meaning moves to the matcher, where it still means
something. A trusted domain also describes where its signing keys come from in
one shape — a JWKS URL, an inline JWK set, or a PEM certificate — rather than
the SPIFFE-only bundle source it carried before.

Trusted domains stored before this change are trusted for SPIFFE under their
name, so the matcher is backfilled from it and read back from it on any row the
backfill has not reached; their legacy bundle source maps onto the shared shape.
No data migration is required, and a create that declares no matcher still means
the name, so payloads written against the SPIFFE-only API keep working.

The deprecated bundleSource/jwksUrl fields stay accepted on write and present on
read, keeping existing automation and the current Console page
working until the merged page lands. The gateway resolves all three key sources;
a PEM certificate holds a single key and is matched without a kid.

Trusted-issuer keys are now resolved through the same hardened, cached fetcher
SPIFFE bundles already use, instead of a separate Nimbus JWKS source with fixed
timeouts and no limits. A JWKS URL on an unsecured scheme or a private address
is refused unless the security domain permits it, and the refusal is logged with
the trusted domain's name. The response-size limit the domain already advertised
is now enforced: the fetcher aborts the read once the body passes the cap rather
than buffering it. The per-issuer JWTProcessor cache is retired, so cached keys
survive a transient fetch failure and refresh on a key-id miss, matching SPIFFE.

The limits that govern retrieval — unsecured scheme, private address, fetch
timeout, response size, cache TTL and cache size — leave the SPIFFE settings
block for oidc.keyRetrievalSettings, which honestly describes what they govern:
retrieval for every trusted domain, token exchange included. The JWT validation
policy that is genuinely SPIFFE-specific — max SVID lifetime, clock skew,
algorithm allowlist — stays where it is. Values written before the move are not
lost: the limits stay readable on the SPIFFE block, deprecated and nullable, and
OIDCSettings falls back to them when the new block is absent, so a domain reads
the same limits before and after the upgrader has run — gateways included. The
upgrader only touches domains that still carry a legacy limit, so it is
idempotent, and copies values verbatim including a limit deliberately set to
zero. The deprecated fields stay accepted on write and are relocated onto the
new block, because the management API rejects unknown properties. The new block
also reaches the automation API, whose domain apply replaces settings wholesale
and would otherwise silently drop the limits. Console binds its existing
controls to the new block and no longer hides them behind the SPIFFE toggle,
since they now apply whether or not SPIFFE is enabled.

Two behaviour changes worth noting on upgrade: a trusted issuer whose JWKS URL
is http:// or private now needs the domain's allowances turned on, and an
expired subject token from a trusted issuer now reports "has expired" rather
than the generic invalid-token description. Both keep the same invalid_request
status.

Trusted issuers are still declared in the security domain's token-exchange
settings; only retrieval and the location of the limits changed.
---


wajb (commented):
Couple of thoughts:
1. The deprecation approach is something new to the AM MAPI and, while the strangler approach works for reviewing of this PR stack, I'm worried about leaving the two pathways as supported in the next minor release. We can declare the new schema as the only supported form. Since it's a version boundary, can we remove and declare the breaking change? (Our AAPI terraform provider is not GA though we should probably send a note to the devex team.)
2. The `TrustDomainKind` embeds the two shapes into the same resource with no way to share key configuration between the two usages. Was the option of flattening the trust domain definition into a single and reusable shape, with no `type`, explored? e.g. requiring, separately, both SPIFFE `name` and token exchange `url`/iss`. Since it's defined as a first-class resource, having a more agnostic shape would make the move from having two separate schemas to one more worthwhile than storing and presenting them together but as incompatible flavours.

lgw-gravitee:
@wajb 

Thanks both — (2) is implemented, (1) I've left out of this stack.

**2. Flattening.** Done, and the stack is re-authored rather than patched on top, so `TrustDomainKind` never appears in any commit of it.

A trusted domain is now a single agnostic shape:

```
name                 // label, unique per security domain, no matching semantics
spiffeTrustDomain?   // matcher: sub = spiffe://<this>/...
issuer?              // matcher: iss claim
keyMaterial, refreshIntervalSeconds, allowedAlgorithms
scopeMappings, userBindingEnabled, userBindingCriteria
```

The two matchers are siblings, declared separately as you suggested, and at least one must be present. Declaring both lets one authority serve JWT-SVID client assertions *and* token exchange over the same key material — the sharing the previous shape made impossible.

The root of the problem was `name`: it doubled as the SPIFFE matcher while being a mere derived label for token exchange, and that asymmetry is what forced a discriminator. Separating label from matchers dissolves it.

Falling out of that:

- The cross-kind guards are gone — there is no longer a wrong kind to reject.
- Uniqueness is per matcher: partial unique indexes on `spiffeTrustDomain` and on `issuer`, alongside the label's own.
- Rows written before this read back with `spiffeTrustDomain` defaulted from `name`; rows the migration wrote carry an issuer and are correctly not treated as SPIFFE.
- The Console offers the two usages as checkboxes — "SPIFFE" and "OIDC - Trusted Issuer" — rather than a kind to pick up front.

gravitee-am-model/src/main/java/io/gravitee/am/model/oidc/OIDCSettings.java
```diff
@@ -55,6 +56,9 @@ public class OIDCSettings {
 
     private SpiffeDomainSettings workloadIdentitySettings;
 
+    @Schema(description = "Fetch, SSRF and cache limits applied to every trusted domain in the security domain.")
+    private KeyRetrievalSettings keyRetrievalSettings;
```
wajb: thought: `oidc` is becoming a bucket and isn't the right protocol for the generic fetch/cache settings or token exchange to look up. Do you think this could be promoted to `Domain`, alongside `tokenExchangeSettings`?

gravitee-am-model/src/main/java/io/gravitee/am/model/oidc/KeyRetrievalSettings.java
```diff
@@ -0,0 +1,105 @@
...
+ * Domain-level limits applied whenever key material is retrieved for a trusted domain, whatever
+ * kind of trust it represents.
+ *
+ * @author GraviteeSource Team
+ */
+@Getter
+@Setter
+@Schema(title = "Trusted domain key retrieval settings",
+        description = "Fetch, SSRF and cache limits applied to every trusted domain in the security domain, " +
+                "regardless of its kind.")
```
wajb: nitpick: "regardless of its kind" doesn't chime well now the model has been flattened. IMO we can remove this from the docs and schema description.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -64,10 +70,12 @@ public class TrustDomainServiceImpl implements TrustDomainService {
...
 
     /**
-     * SPIFFE trust-domain names are case-insensitive DNS-style host labels per the
+     * SPIFFE trust domains are case-insensitive DNS-style host labels per the
      * <a href="https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md#21-trust-domain">SPIFFE ID spec</a>.
      */
-    private static final Pattern NAME_PATTERN = Pattern.compile("^[a-z0-9]([a-z0-9.\\-]*[a-z0-9])?$");
+    private static final Pattern SPIFFE_TRUST_DOMAIN_PATTERN = Pattern.compile("^[a-z0-9]([a-z0-9.\\-]*[a-z0-9])?$");
+
+    private static final int MAX_NAME_LENGTH = 255;
```
wajb: suggestion: can we add a max length annotation to the schema to advertise this limit? This looks like something previously missed, since the limit is the same.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -238,32 +251,95 @@ private Completable publish(Domain domain, TrustDomain trustDomain, Action actio
...
+        if (td.getName() == null) {
+            return Completable.error(new InvalidTrustDomainException("name is required"));
         }
-        if (td.getBundleSource() == null) {
-            return Completable.error(new InvalidTrustDomainException("bundleSource is required"));
+        if (td.getName().length() > MAX_NAME_LENGTH) {
+            return Completable.error(new InvalidTrustDomainException("name must be at most " + MAX_NAME_LENGTH + " characters"));
         }
-        if (td.getBundleSource() != SpiffeBundleSource.JWKS_URL) {
+        if (td.getSpiffeTrustDomain() == null || !SPIFFE_TRUST_DOMAIN_PATTERN.matcher(td.getSpiffeTrustDomain()).matches()) {
```
wajb, line 333: issue: IMO we should length-check the SPIFFE trust domain too (also limited to 255 characters).

gravitee-am-service/src/main/java/io/gravitee/am/service/jwk/impl/WebClientJWKSetFetcher.java
```diff
@@ -71,11 +82,37 @@ private Maybe<JWKSetFetchResponse> fetch(String jwksUri) {
         }
     }
 
-    private Maybe<JWKSetFetchResponse> toResponse(HttpResponse response) {
-        String raw = response.bodyAsString();
+    private Single<FetchedResponse> send(String url, long maxResponseSizeBytes) {
```
wajb, line 97: thought: it might be worth considering hardening this to include timeout and disabling follow-redirects too. Mostly for the shared jwks client though, because the callers wrap it I believe, but they shouldn't have to ideally.

wajb (approved):

PR #8433, opened 2026-08-21
feat: trusted domains gain an issuer matcher, gateway residency and a migration (AM-7372)

lgw-gravitee:
A trusted domain gains a second matcher alongside the SPIFFE one: the issuer it
vouches for, matched against the "iss" claim of an external token during an RFC
8693 exchange. It sits flat beside spiffeTrustDomain rather than in a block of
its own, together with the scope mappings and user binding that govern what an
accepted token becomes. A trusted domain must declare at least one matcher, and
declaring both lets one authority serve both usages over the same key material —
there is no kind to choose between them. Scope mappings and user binding are the
only settings tied to a usage: they need an issuer, because nothing else
consumes them.

Trusted domains are created, read, listed, amended and removed through the
trusted-domains API under the trusted-domain permission, each with its own audit
record.

An issuer is vouched for by at most one trusted domain per security domain. The
service rejects a duplicate before writing, and every backend carries the
constraint: a partial unique index on MongoDB, since a sparse one would admit
every issuerless SPIFFE document and collide; a plain unique index on
PostgreSQL, MySQL and MariaDB, which treat nulls as distinct; and a filtered one
on SQL Server, which does not. The issuer column is bounded at 512 characters so
the index key stays inside InnoDB's 3072-byte limit at utf8mb4.

Registering, amending or removing a trusted domain now takes effect at the
gateway promptly and without restarting the security domain's reactor, and
validating a SPIFFE client assertion no longer costs a database round trip. The
gateway's event-type resolution had no case for TRUST_DOMAIN, so every
trust-domain event the management API published resolved to no type and was
silently discarded; the gateway compensated by reading from the database on
every token request. A trusted-domain event and its resolution case close that
gap.

A trusted-domain manager holds all of a security domain's trusted domains in
memory, following the established manager pattern rather than a lazy cache:
trusted domains are bounded configuration, so full residency is cheap and the
database leaves the token request path. It keeps one lookup per matcher — by
SPIFFE trust domain and by issuer — so a trusted domain declaring both is
reachable through either, and one declaring a single matcher is simply absent
from the other lookup. Amending re-indexes under the new matchers and evicts
cached key material, so a corrected JWKS URL takes effect promptly; removal
takes effect immediately rather than after a cache lifetime.

Readiness is reported per trusted domain, plus one marker registered before the
preload starts, so a security domain reports as synchronised only once its
trusted domains are loaded rather than while the load is still in flight. A load
failure is reported as an initialisation failure and leaves the domain unstable
rather than passing silently. Removal now advances the domain's last-sync
timestamp as a load does: the readiness service could only report a plugin
unloaded, which drops it from matching without recording that anything happened,
so a deletion was invisible to the sync barrier integration tests wait on.
Every applied change is therefore waitable, and the tests wait on it instead of
polling for side effects. An event referencing an entity that no longer exists
drops it from residency; one belonging to another security domain, or to another
reference type, is ignored.

On upgrade, every trusted issuer declared inside a security domain's
token-exchange settings gains a trusted domain of its own, carrying its issuer,
key material, scope mappings and user-binding configuration. Migrated entities
are named from the slugified full issuer URL, so the same configuration migrates
to the same names in every environment and two issuers sharing a host but
differing in path stay distinct. Where two issuers would slugify alike, both —
never just the one declared second — carry a digest of their issuer, keeping the
outcome free of declaration order; a derived name an existing trusted domain
already holds is disambiguated the same way rather than dropped, since a skipped
issuer would silently stop being trusted at cutover. Entities are written
straight to the repository rather than through TrustDomainService, because that
service applies the SSRF guard the inline shape never had, so an issuer
resolving to a private address — legal when it was configured — would fail the
upgrade instead of migrating. The migration creates rows under a unique index,
so it runs behind a system task like the password-policy migration rather than
the plain async upgrader: two management-api instances booting together would
otherwise race. Re-running neither duplicates nor overwrites.

The inline declarations are deliberately left in place and stay authoritative;
nothing reads the new entities for token exchange yet, so both the entities and
the migration can be proven before anything depends on them. The token-exchange
path still reads its inline trusted issuers; moving it onto the manager is the
cutover that follows.
---

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/trustdomain/impl/TrustDomainManagerImpl.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/trustdomain/impl/TrustDomainManagerImpl.java
@@ -1,0 +1,173 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.oidc.service.trustdomain.impl;
+
+import io.gravitee.am.common.event.EventManager;
+import io.gravitee.am.common.event.TrustDomainEvent;
+import io.gravitee.am.common.event.Type;
+import io.gravitee.am.gateway.handler.oidc.service.trustdomain.TrustDomainKeyService;
+import io.gravitee.am.gateway.handler.oidc.service.trustdomain.TrustDomainManager;
+import io.gravitee.am.model.Domain;
+import io.gravitee.am.model.ReferenceType;
+import io.gravitee.am.model.common.event.Payload;
+import io.gravitee.am.model.oidc.TrustDomain;
+import io.gravitee.am.monitoring.DomainReadinessService;
+import io.gravitee.am.repository.management.api.TrustDomainRepository;
+import io.gravitee.common.event.Event;
+import io.gravitee.common.event.EventListener;
+import io.gravitee.common.service.AbstractService;
+import io.reactivex.rxjava3.schedulers.Schedulers;
+import lombok.CustomLog;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.annotation.Autowired;
+
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+@CustomLog
+public class TrustDomainManagerImpl extends AbstractService implements TrustDomainManager, InitializingBean, EventListener<TrustDomainEvent, Payload> {
+
+    private static final String PRELOAD_PLUGIN_ID = Type.TRUST_DOMAIN.name();
+
+    @Autowired
+    private Domain domain;
+
+    @Autowired
+    private EventManager eventManager;
+
+    @Autowired
+    private TrustDomainRepository trustDomainRepository;
+
+    @Autowired
+    private TrustDomainKeyService trustDomainKeyService;
+
+    @Autowired
+    private DomainReadinessService domainReadinessService;
+
+    private final ConcurrentMap<String, TrustDomain> byId = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, TrustDomain> bySpiffeTrustDomain = new ConcurrentHashMap<>();
+    private final ConcurrentMap<String, TrustDomain> byIssuer = new ConcurrentHashMap<>();
+
+    @Override
+    public void afterPropertiesSet() {
+        log.info("Initializing trusted domains for domain {}", domain.getName());
+        domainReadinessService.initPluginSync(domain.getId(), PRELOAD_PLUGIN_ID, Type.TRUST_DOMAIN.name());
+        trustDomainRepository.findByReference(ReferenceType.DOMAIN, domain.getId())
+                .subscribeOn(Schedulers.io())
+                .subscribe(
+                        trustDomain -> {
+                            domainReadinessService.initPluginSync(domain.getId(), trustDomain.getId(), Type.TRUST_DOMAIN.name());
+                            index(trustDomain);
+                            log.info("Trusted domain {} loaded for domain {}", trustDomain.getName(), domain.getName());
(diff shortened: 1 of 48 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/trustdomain/impl/TrustDomainManagerImpl.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Override
+    public Optional<TrustDomain> findByIssuer(String issuer) {
+        return issuer == null ? Optional.empty() : Optional.ofNullable(byIssuer.get(issuer));
+    }
+
+
+    private void reload(String trustDomainId) {
+        log.info("Loading trusted domain {} for domain {}", trustDomainId, domain.getName());
+        domainReadinessService.initPluginSync(domain.getId(), trustDomainId, Type.TRUST_DOMAIN.name());
```
wajb, line 126: question: I don't think any other manager flags a general type marker like this. What is the benefit of tracking the trust domain loading state in the service when we already have logs? 🤔

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-trust-domains-token-exchange.yml
```diff
@@ -0,0 +1,65 @@
...
+        - onFail: MARK_RAN
+        - not:
+            - columnExists:
+                tableName: trust_domains
+                columnName: issuer
+      changes:
+        - addColumn:
+            tableName: trust_domains
+            columns:
+              - column: { name: issuer, type: nvarchar(512), constraints: { nullable: true } }
```
wajb, line 18: question: how will we handle issuer URLs that are longer than 512 at cutover? I don't think there is a limit currently.

gravitee-am-test/specs/gateway/blueprint-agents/spiffe-prefix-assertion.jest.spec.ts
```diff
@@ -16,10 +16,11 @@
 import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
 import { performPost } from '@gateway-commands/oauth-oidc-commands';
 import { decodeJwt } from '@utils-commands/jwt';
-import { setup } from '../../test-fixture';
+import { retryImmediatelyForThisFile, setup } from '../../test-fixture';
 import { SpiffePrefixFixture, setupSpiffePrefixFixture } from './fixtures/spiffe-prefix-fixture';
 
 setup(120000);
+retryImmediatelyForThisFile();
```
wajb: suggestion: can we avoid linear tests? The guidelines spell them out as not recommended. Repeating setup is better so that tests can be run individually.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/trustdomain/impl/TrustDomainManagerImpl.java
lgw-gravitee: it was for integration tests to avoid having sleep(5000).

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-trust-domains-token-exchange.yml
lgw-gravitee: I dont think we would ever have issuer that long. would you lift that from database column or put 512 on API?

wajb: I think you're right that it's unlikely, but maybe an edge case worth checking that the migrator won't fail in jdbc (next PR).

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oidc/service/trustdomain/impl/TrustDomainManagerImpl.java
wajb: Ah that's a valid use. After all the readiness service was primarily to support testing 👍 .

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/resources/liquibase/changelogs/v4_13_0/4.13.0-trust-domains-token-exchange.yml
lgw-gravitee: migrateIssuer now warns and leaves such an issuer inline, same as the existing derived-name-taken case. Two tests: the skip, and 512 exactly still migrating.

Related bug found and fixed in the cutover PR: the purge used Flowable.all(), which cancels on the first false, so one unmigratable issuer aborted every issuer after it. Now a reduce.

wajb (approved):

PR #8434, opened 2026-08-21
feat: trusted domains become the only source of trust (AM-7372)

lgw-gravitee:
The security domain stops persisting its inline trusted-issuer list on both
storage backends. The deprecated domain-level field becomes a projection over
the trusted domains carrying an issuer: reads assemble it, writes translate into
entity creates, updates and deletes, so an issuer omitted from a written list
stops being trusted. A write preserves the SPIFFE matcher of any domain it
touches, so a trusted domain serving both usages keeps both. The management API
keeps the permission it accepts today; the trusted-domains API requires the
trusted-domain permission.

The gateway resolves external subject and actor tokens by matching the issuer
against the resident trusted domains, and the derived issuer lookup on the
token-exchange settings is gone. Matchers are independent, so a single trusted
domain can serve JWT-SVID client assertions and token exchange over the same key
material.

The upgrader drops the inline declarations once every issuer it found is
represented as an entity, so no stored domain keeps a vestigial list.
---

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/ValidatedToken.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/ValidatedToken.java
@@ -15,8 +15,9 @@
  */
 package io.gravitee.am.gateway.handler.oauth2.service.token.tokenexchange;
 
-import io.gravitee.am.model.TrustedIssuer;
+import io.gravitee.am.model.oidc.TrustDomain;
 import lombok.Builder;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.Singular;
 
@@ -33,6 +34,7 @@ import java.util.*;
  */
 @Builder
 @Getter
+@EqualsAndHashCode
 public class ValidatedToken {
 
     /**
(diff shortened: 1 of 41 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/tokenexchange/ValidatedToken.java
```diff
@@ -144,11 +145,11 @@ public ValidatedToken withParents(Set<String> parentJtis) {
...
+        return Objects.equals(subject, that.subject) && Objects.equals(issuer, that.issuer) && Objects.equals(claims, that.claims) && Objects.equals(scopes, that.scopes) && Objects.equals(expiration, that.expiration) && Objects.equals(issuedAt, that.issuedAt) && Objects.equals(notBefore, that.notBefore) && Objects.equals(tokenId, that.tokenId) && Objects.equals(audience, that.audience) && Objects.equals(clientId, that.clientId) && Objects.equals(tokenType, that.tokenType) && Objects.equals(domain, that.domain) && Objects.equals(domainParentJtis, that.domainParentJtis) && Objects.equals(trustedDomain, that.trustedDomain);
```
wajb: suggestion: use `lombok.EqualsAndHashCode`.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/upgrades/DomainTrustedIssuerUpgrader.java
```diff
@@ -134,22 +126,35 @@ private Completable migrateIssuers(Domain domain, List<TrustedIssuer> issuers, L
...
-        return trustDomainRepository.create(asTrustDomain(domain, issuer, name)).ignoreElement();
+        return trustDomainRepository.create(asTrustDomain(domain, issuer, name)).map(created -> true);
+    }
+
+    /**
+     * Drops the inline declarations now that the entities are authoritative. The storage layer no
+     * longer persists them, so re-saving the security domain is what removes them.
+     */
+    private Completable purgeInlineIssuers(Domain domain) {
+        domain.getTokenExchangeSettings().setTrustedIssuers(null);
```
wajb: minor: I don't think clearing this actually has any effect since the storage layer ignores it when writing.

gravitee-am-test/specs/gateway/token-exchange/trusted-issuers.jest.spec.ts
```diff
@@ -17,17 +17,16 @@ import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
...
+import { getDomain, waitForOidcReady } from '@management-commands/domain-management-commands';
+import { listTrustDomains } from '@management-commands/trust-domain-management-commands';
+import { TrustDomain } from '@management-models/TrustDomain';
+import { retryImmediatelyForThisFile, setup } from '../../test-fixture';
+import { setupTrustedIssuerFixture, TrustedIssuerFixture, patchDomainRaw } from './fixtures/trusted-issuer-fixture';
 import { signJwtForTrustedIssuer } from './fixtures/trusted-issuer-jwt-helper';
 import { TOKEN_EXCHANGE_TEST } from './fixtures/token-exchange-fixture';
 
 setup();
+retryImmediatelyForThisFile();
```
wajb: suggestion: avoid sequential tests if possible.

wajb (approved):

PR #8435, opened 2026-08-21
feat: the Console presents a single Trusted Domains page (AM-7372)

lgw-gravitee:
Everything a security domain trusts is now managed from one entry under
Security. A trusted domain carries a label and, independently, a SPIFFE trust
domain and an issuer, so the list shows which usages each entry serves and one
entry can serve both. Creation and editing offer the two usages as checkboxes
labelled SPIFFE and OIDC - Trusted Issuer, revealing each one's settings as it
is enabled, and key material — JWKS URL, inline JWK set or PEM — is edited
through one control shared by both. Detail pages are addressed by a stable
identifier rather than a position in a list, so a page can be linked to
directly.

The trusted-issuers tab is gone from Token Exchange, which keeps its settings
tab and, like SPIFFE Settings, links to Trusted Domains so operators looking for
issuer configuration where it used to live are led to it. Both cross-links are
hidden from principals without the trusted-domain permission, who would
otherwise be led to a page AuthGuard refuses; the prose still names where
trusted domains live and that the permission is needed.

Because the inline list is now a projection, the Token Exchange settings page
stops sending trustedIssuers altogether: a settings save that carried a stale
list would have replaced every trusted domain serving an issuer.

The trusted-domain update call enumerates the fields it sends, so it had kept
sending the removed nested block; it now sends the flat shape the management API
accepts, which rejects unknown properties.

The key-material control no longer clobbers itself. The parent stored the
emitted object back on the bound input, so Angular re-ran ngOnChanges on every
keystroke and re-seeded the control's local state; typing a JWK set was
therefore impossible, since the first keystroke that left the JSON incomplete
wiped the textarea and the error with it. The control now ignores an input it
just emitted, pinned by a unit test — the E2E suite missed this because fill()
delivers valid JSON in one event.

An empty allowedAlgorithms list is what clears a SPIFFE override: the gateway
reads an empty list as "inherit the domain default", while a null one means
"leave alone" and silently reverted the removal.

The Playwright trusted-issuer suite and its page objects are rewritten against
the merged page rather than duplicated, with a positive anchor before its
absence assertion, retrying count assertions, and a deterministic scope-mapping
helper now that the two domain-scope inputs carry distinct test identifiers. The
generated client picks up the deprecation of the inline list.
---


gravitee-am-ui/src/app/app-routing.module.ts
```diff
@@ -3024,18 +3022,6 @@ export const routes: Routes = [
                         children: [
                           { path: '', redirectTo: 'settings', pathMatch: 'full' },
                           { path: 'settings', component: TokenExchangeSettingsComponent },
```
wajb: suggestion: can we remove the Settings layer so we aren't left with a huge tab?

Image

gravitee-am-ui/src/app/domain/settings/oauth/token-exchange/token-exchange-settings/token-exchange-settings.component.html
```diff
@@ -25,6 +25,21 @@
         <mat-hint style="font-size: 75%">Enable RFC 8693 OAuth 2.0 Token Exchange</mat-hint>
       </div>
 
+      <div class="gv-form-section" fxLayout="column">
```
wajb: question: do we need to display this even when token exchange is disabled? I would have expected it to be nested within the block below.

Also, note the styling changes in chrome when the feature is toggled, which makes the hint text jump.

gravitee-am-ui/src/app/domain/settings/openid/spiffe/spiffe-settings.component.html
```diff
@@ -26,8 +26,13 @@ <h1>SPIFFE Workload Identity</h1>
           </mat-slide-toggle>
           <mat-hint style="font-size: 75%">
             Allow clients in this domain to authenticate using the <code>spiffe_jwt</code> token-endpoint method. Requires at least one
-            Trust Domain configured under Workload Identity → Trust Domains.
+            trusted domain of the SPIFFE kind.
           </mat-hint>
+          <div style="margin-top: 12px" *hasPermission="['domain_trust_domain_list']">
```
wajb: question: like with token exchange, does this need to be visible when SPIFFE is disabled?

gravitee-am-ui/src/app/domain/settings/openid/spiffe/spiffe-settings.component.html
```diff
@@ -26,8 +26,13 @@ <h1>SPIFFE Workload Identity</h1>
...
+            <a mat-stroked-button data-testid="trustedDomainsLink" [routerLink]="['../trust-domains']">
+              <mat-icon>verified_user</mat-icon> Trusted Domains
+            </a>
+          </div>
         </div>
 
         <div class="gv-form-section" fxLayout="column">
```
wajb: suggestion: keeping key retrieval settings under SPIFFE is probably non-obvious now, particularly when only using trusted domains for token exchange. Instead, should we consider moving the settings to Trusted Domains, perhaps as a gear icon?

The SPIFFE page then becomes quite lean, but that's a better tradeoff I believe.

lgw-gravitee: yes, I moved it to Trust Domains

gravitee-am-test/playwright/pages/domain-token-exchange.page.ts
```diff
@@ -16,22 +16,20 @@
...
-  /* ------------------------------------------------------------------ */
-  /*  Tab navigation                                                     */
-  /* ------------------------------------------------------------------ */
-
-  get settingsTab(): Locator {
-    return this.page.locator('a[mat-tab-link]').filter({ hasText: /settings/i });
+  /** Tab bar the page no longer carries now that settings are its only content. */
+  get tabLinks(): Locator {
+    return this.page.locator('a[mat-tab-link]');
   }
```
wajb: nitpick: there's no need to read or assert this anymore.

wajb (approved):

PR #8440, opened 2026-08-24
feat: async mode for token storage

lgw-gravitee:
fixes AM-7504
ref gravitee-io/issues#11729

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
@@ -1,0 +1,74 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.oauth2.service.token.impl;
+
+import io.gravitee.am.gateway.handler.oauth2.service.token.TokenManager;
+import io.gravitee.am.repository.oauth2.api.TokenRepository;
+import io.gravitee.am.repository.oauth2.model.AccessToken;
+import io.gravitee.am.repository.oauth2.model.RefreshToken;
+import io.gravitee.common.service.AbstractService;
+import io.reactivex.rxjava3.core.Completable;
+import io.reactivex.rxjava3.processors.FlowableProcessor;
+import io.reactivex.rxjava3.processors.PublishProcessor;
+import io.reactivex.rxjava3.subjects.CompletableSubject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * @author GraviteeSource Team
+ */
+public class AsyncTokenManager extends AbstractService implements TokenManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(AsyncTokenManager.class);
+
+    private static final long DRAIN_TIMEOUT_MS = 5000;
+
+    private final TokenRepository tokenRepository;
+    private final FlowableProcessor<Completable> pendingWrites = PublishProcessor.<Completable>create().toSerialized();
+    private final CompletableSubject drained = CompletableSubject.create();
+
+    public AsyncTokenManager(TokenRepository tokenRepository) {
+        this.tokenRepository = tokenRepository;
+        this.pendingWrites.flatMapCompletable(write -> write)
+                .subscribe(drained::onComplete, drained::onError);
+    }
+
+    @Override
+    protected void doStart() throws Exception {
+        super.doStart();
+    }
+
+    @Override
+    protected void doStop() throws Exception {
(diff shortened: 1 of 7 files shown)


sclarkamlab (commented):
Nice, and the flag defaulting to off makes it safe to land. The thing that bothers me is that this only makes creates async while every delete stays sync, so any read then delete path can race the queued write. Details inline.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public Completable storeTokens(AccessToken accessToken, RefreshToken refreshToken) {
+        pendingWrites.onNext(tokenRepository.create(accessToken, refreshToken)
+                .doOnError(error -> logger.error("Unable to store tokens asynchronously, domain={} client={} accessTokenJti={} refreshTokenJti={}",
+                        accessToken.getDomain(), accessToken.getClient(), accessToken.getToken(),
+                        refreshToken != null ? refreshToken.getToken() : null, error))
+                .onErrorComplete());
+        return Completable.complete();
```
sclarkamlab, line 72: This only makes creates async. Every delete is still sync, so anything that reads then deletes can run against a row that is still sitting in `pendingWrites`, and the queued write puts it back afterwards. Three places I can see:

`TokenServiceImpl:312` rotates the refresh token. `BaseIntrospectionTokenService:90` skips the store while `now < iat + offlineVerificationTimerSeconds` (10s default), so a refresh inside that window passes on the signature alone, the delete hits nothing, and the queued write reinserts the refresh token. Single use token is replayable for the rest of its lifetime.

`revokeAccessToken` does the same introspect then delete. Row comes back, token stays live, and we still return 200 and write a `revoked()` audit so nothing looks wrong to the operator.

`AuthorizationCodeServiceImpl:85` uses `findAccessTokenByAuthorizationCode` to kill the tokens issued from a reused code. If the replay beats the write there is nothing to find, so we skip the revoke RFC 6749 4.1.2 asks for, and log nothing either.

Pushing the deletes onto the same serialized stream would cover all three since ordering then holds per token, but that's your call.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
```diff
@@ -0,0 +1,74 @@
...
+        super.doStop();
+    }
+
+    @Override
+    public Completable storeTokens(AccessToken accessToken, RefreshToken refreshToken) {
+        pendingWrites.onNext(tokenRepository.create(accessToken, refreshToken)
+                .doOnError(error -> logger.error("Unable to store tokens asynchronously, domain={} client={} accessTokenJti={} refreshTokenJti={}",
+                        accessToken.getDomain(), accessToken.getClient(), accessToken.getToken(),
+                        refreshToken != null ? refreshToken.getToken() : null, error))
+                .onErrorComplete());
```
sclarkamlab, line 71: `onErrorComplete()` flips this from fail closed to fail open. Today the token request fails if the write fails. With this on we log a line and hand the token out anyway, and nothing can revoke or purge it afterwards.

There are no custom meters anywhere in the gateway today so I'm not going to pretend there's a pattern to follow, but a store outage being visible only in the log feels thin. `services.metrics.prometheus` is already wired up if you wanted a counter on the dropped writes.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final long DRAIN_TIMEOUT_MS = 5000;
+
+    private final TokenRepository tokenRepository;
+    private final FlowableProcessor<Completable> pendingWrites = PublishProcessor.<Completable>create().toSerialized();
+    private final CompletableSubject drained = CompletableSubject.create();
+
+    public AsyncTokenManager(TokenRepository tokenRepository) {
+        this.tokenRepository = tokenRepository;
+        this.pendingWrites.flatMapCompletable(write -> write)
```
sclarkamlab, line 47: No concurrency limit here, so a burst puts every write in flight at once and they all hold repo connections the read paths also need. Worth `flatMapCompletable(w -> w, false, maxConcurrency)` with a configurable cap, and a gauge on the pending count if we do end up adding meters.

gravitee-am-gateway/gravitee-am-gateway-standalone/gravitee-am-gateway-standalone-distribution/src/main/resources/config/gravitee.yml
```diff
@@ -366,6 +366,13 @@ handlers:
 
     response:
       strict: false
+#    tokens:
+#      # Store the access and refresh tokens in a fire-and-forget mode: the token response is sent
+#      # back to the client without waiting for the database write to complete.
+#      # Caveats: a token may be returned but not stored yet, so introspection (see
```
sclarkamlab, line 372: The caveat wording undersells it a bit. "May not see the token immediately" reads like a delayed read, but what actually happens is the delete is a no-op and then the queued write resurrects the token. Might be worth saying that outright.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-oidc/src/main/java/io/gravitee/am/gateway/handler/oauth2/service/token/impl/AsyncTokenManager.java
leleueri: also, is there any backpressure manage for this operator if so how to manage a buffer filled by token creation event?

lgw-gravitee:
Team decided to not provide it

PR #8448, opened 2026-08-25
test: sign in to the Console as an admin defined in gravitee.yml (AM-2192)

sagarchauhan-qa:
## Summary

The AM-2192 Console sign-in test never ran. It skipped unless `AM_ALT_ADMIN_USERNAME` and `AM_ALT_ADMIN_PASSWORD` were set, and nothing in the local stack or CI set them — `AM_ALT_ADMIN` appeared nowhere else in the repo. It reported as skipped, which looks green without having tested anything.

This provisions the administrator the test needs, removes the guard, and covers the three scenarios from the ticket.

## Changes

- Second administrator as `security.providers[0].users[1]` on the management service
- Credentials moved into `test-constants.ts`; skip guard deleted, so the test runs unconditionally
- Three scenarios: the configured admin signs in · the default admin still works alongside it · credentials absent from the configuration are refused
- Unit test for a second user in the configuration file — existing coverage only ever exercised `users[0]`

## On the shared compose change

`docker-compose.yml` is used by every suite, so worth stating explicitly: enabling the inline provider is *additive*. `loadIdentityProviders()` merges it with the built-in Gravitee provider rather than replacing it, so `admin:adminadmin` and the compose healthcheck are unaffected, and the provider is not persisted — `GET /organizations/DEFAULT/identities` is unchanged.

The BCrypt hash is written with `$$` in the compose file so Compose does not interpolate it away.

## Testing

All run locally against a Mongo local stack:

| Suite | Result |
|---|---|
| `login-flows.spec.ts` | 8 passed, **0 skipped** |
| `playwright/tests/auth/` | 39 passed |
| `IdentityProviderManagerTest` | 7 passed |
| Jest management (88 suites) | 1198 passed / 1215 |

The 17 Jest failures are confined to `ciba.jest.spec.ts` and `openfga.jest.spec.ts`, both of which need services absent from the local stack profile used here (OpenFGA refuses connection on `:8090`; no CIBA container). Neither touches security providers or admin authentication. The 1198 that passed all authenticate through `requestAdminAccessToken()`, which is direct confirmation at scale that the default admin is unaffected by the compose change.

Each new assertion was deliberately broken once to confirm it was live, then reverted. The sign-in scenarios assert the account shown in the navbar, so they distinguish "this admin is signed in" from "a console rendered" — the previous assertion would have passed for the default admin too.

### DIFF
--- a/gravitee-am-test/playwright/utils/test-constants.ts
+++ b/gravitee-am-test/playwright/utils/test-constants.ts
@@ -36,9 +36,6 @@ export const ADMIN_PASSWORD = 'adminadmin';
 export const ALT_ADMIN_USERNAME = 'altadmin';
 export const ALT_ADMIN_PASSWORD = 'altadminaltadmin';
 
-/** A username that is deliberately absent from the configuration file (AM-2192 negative case) */
-export const UNCONFIGURED_ADMIN_USERNAME = 'ghostadmin';
-
 /** Regex matching a valid JWT (three Base64url segments separated by dots) */
 export const JWT_FORMAT = /^[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+\.[A-Za-z0-9_-]+$/;
 
--- a/gravitee-am-test/playwright/utils/test-constants.ts
+++ b/gravitee-am-test/playwright/utils/test-constants.ts
@@ -26,6 +26,19 @@ export const ADMIN_USERNAME = 'admin';
 /** Default admin password */
 export const ADMIN_PASSWORD = 'adminadmin';
 
+/**
+ * AM-2192: a second administrator declared in the configuration file rather than the database.
+ * Provisioned as security.providers[0].users[1] by the management service in
+ * docker/local-stack/dev/docker-compose.yml — these must stay in step with that file.
+ * Deliberately a different password from ADMIN_PASSWORD, so a test signing in as this
+ * account cannot pass by accidentally authenticating as the default admin.
+ */
+export const ALT_ADMIN_USERNAME = 'altadmin';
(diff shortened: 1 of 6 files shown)


gravitee-am-test/playwright/utils/test-constants.ts
```diff
@@ -26,6 +26,19 @@ export const ADMIN_USERNAME = 'admin';
...
+ * docker/local-stack/dev/docker-compose.yml — these must stay in step with that file.
+ * Deliberately a different password from ADMIN_PASSWORD, so a test signing in as this
+ * account cannot pass by accidentally authenticating as the default admin.
+ */
+export const ALT_ADMIN_USERNAME = 'altadmin';
+export const ALT_ADMIN_PASSWORD = 'altadminaltadmin';
+
+/** A username that is deliberately absent from the configuration file (AM-2192 negative case) */
+export const UNCONFIGURED_ADMIN_USERNAME = 'ghostadmin';
+
```
lgw-gravitee: I think its unnecessary to create a constant value for not existing username. I would just use a literal string in test. wdyt?

sagarchauhan-qa: Agreed — removed the constant and used the literal in the test (980057981d).

Kept `ALT_ADMIN_USERNAME`/`ALT_ADMIN_PASSWORD` as constants since those have to stay in step with the credentials configured in `docker-compose.yml`, but `ghostadmin` has no such constraint, so the literal reads better at the call site. Spec still passes 8/8.

lgw-gravitee (approved):

PR #8451, opened 2026-08-25
test: role mappers on an organization identity provider (AM-2219)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-2219

## Summary

Role mapping was covered on a security domain, where the mapped role lands in the user's `dynamicRoles`. The same thing at **organization** level, applied when signing in to the Console, was not covered anywhere.

At that level the mechanism is different: `AuthenticationServiceImpl` records the mapped role as an organization **membership** flagged `fromRoleMapper`, so that is what these tests assert.

## Scenarios

The three from AM-2219:

- a user whose claims match the mapper rule receives the mapped role, `fromRoleMapper: true`
- a user whose claims do not match falls back to `ORGANIZATION_USER`, `fromRoleMapper: false`
- two organization providers each apply their own mapper

## Why Jest, and why only these three

The decision logic — mapped roles win, an empty set falls back to the default role — is **already unit-tested** in `AuthenticationServiceTest` (3 tests), so it is not re-tested here. Those tests stub `userDetailsMock.getRoles()`, i.e. they assume the principal already carries mapped roles. What only end-to-end reaches, and what these tests are for, is that a mapper persisted on an organization provider is loaded and its `claim=value` rule evaluated against the claims the external provider actually returns.

## Notes for review

**Nothing existing was modified** — two new files only. The UC-AM3 fixture and `management-auth-helper.ts` are imported, not edited; `extractSocialUrlFromManagementLoginHtml` takes `.first()` and cannot distinguish two providers, so a name-selective variant lives in the new fixture rather than changing a helper four files depend on.

**How to run — `npm run test` will not work.** `dev.setup.js` sets `AM_INTERNAL_GATEWAY_URL=http://localhost:8092`, but that URL is fetched by the management API from *inside its container*, where `localhost` is the management container itself. The CI config sets `http://gateway:8092`. AM-2207's specs have the same constraint.

```
npx jest --config=api/config/ci.config.js specs/management/identity-provider/org-idp-role-mapper.jest.spec.ts
```

**Parallel-run caveat, flagged for a second opinion.** The fixture patches organization settings, which is global state, and `ci:management:parallel` runs `--maxWorkers=2`. Three existing files already patch the same settings; this makes four. Cleanup here re-reads and removes only the two ids it added, rather than writing back a stale snapshot, so it will not clobber another file's providers — but the reverse is not something this PR can prevent without editing a fixture it does not own. CI on this PR is the real evidence.

## Testing

Run locally against a Mongo local stack:

| Suite | Result |
|---|---|
| `org-idp-role-mapper.jest.spec.ts` | 3 passed |
| `specs/management/identity-provider` + `specs/management/api-management` | 63 passed, 7 suites |

The neighbouring run includes AM-2207's own spec, which shares the organization settings this fixture touches.

**Negative control:** with the mapper rule changed to match no one, scenarios 1 and 3 fail with `Expected <mapped role> / Received <ORGANIZATION_USER>`, and scenario 2 still passes — confirming the rule, not the sign-in, drives the outcome.


gravitee-am-test/specs/management/identity-provider/fixtures/org-idp-role-mapper-fixture.ts
```diff
@@ -0,0 +1,555 @@
...
+    accessToken,
+    providerBName,
+    domainB.domain.hrid!,
+    domainB.clientId,
+    domainB.clientSecret,
+    internalGatewayUrl,
+    { [roleB.id!]: [`preferred_username=${matchingB}`] },
+  );
+
+  await defaultApi.patchOrganizationSettings({
```
sclarkamlab: this whole setup has no try/catch, unlike the other fixtures in this batch. if anything after this point throws, say waitForProvidersOnLoginPage on line 479 times out, cleanUp is never returned and both org providers stay registered on the shared DEFAULT organization for the rest of the run. worth wrapping setup and tearing down what has been created so far on failure.

gravitee-am-test/specs/management/identity-provider/fixtures/org-idp-role-mapper-fixture.ts
```diff
@@ -0,0 +1,555 @@
...
+  defaultRoleId: string;
+  defaultRoleName: string;
+  providerA: OrgProvider;
+  providerB: OrgProvider;
+}
+
+/**
+ * Pick a specific provider's sign-in link from the Console login page.
+ *
+ * The shared `extractSocialUrlFromManagementLoginHtml` takes `.first()`, which is fine when a
```
sclarkamlab: you have spotted the .first() here but i think it bites us anyway. ci:management:parallel runs specs/management with maxWorkers=2, so api-management-login-social.jest.spec.ts can be running while these two extra providers are registered on the shared organization, and its .first() then picks one of yours. same read-modify-write race on the identities list in setup, which cleanUp is careful about but setup is not. can we scope this to its own organization, or at least fix the shared helper to select by name?

gravitee-am-test/specs/management/identity-provider/fixtures/org-idp-role-mapper-fixture.ts
sagarchauhan-qa: Fixed. Setup is wrapped in a try/catch now — it keeps track of what it has created and, on any failure, deletes the providers, roles and domains and puts the organization's identity list back before rethrowing.

Checked it by making `waitForProvidersOnLoginPage` throw on purpose. With the catch in place nothing is left behind. With the catch taken out, the same failure left exactly what you described: both providers still registered on the organization, plus the two roles and two domains. (Cleaned those up afterwards.)

gravitee-am-test/specs/management/identity-provider/fixtures/org-idp-role-mapper-fixture.ts
sagarchauhan-qa: Good catch. There's no endpoint to delete an organization, so giving these their own would leak one on every CI run — so I went with your second suggestion.

`extractSocialUrlFromManagementLoginHtml` now takes an optional provider name, and `api-management-login-social-fixture` passes its own. The parameter is optional, so the other callers are unchanged. I also deleted the copy of this helper I'd written in my own fixture and use the shared one instead.

One thing worth knowing: `data-testid` is the provider **type** (`social-provider-oauth2-generic-am-idp`), which any two generic providers share — so matching on that wouldn't have fixed it. `login.html` puts the name in the button text, so that's what it matches on, exactly.

Checked by pointing the lookup at a name that isn't registered:
`Could not find social provider "NotThisProvider" in login form. Present: Social (social-provider-oauth2-generic-am-idp)`

Both spec folders run together now, since they share the helper: 63 passed, 7 suites.

gravitee-am-test/specs/management/identity-provider/fixtures/org-idp-role-mapper-fixture.ts
```diff
@@ -0,0 +1,580 @@
...
+      domainB.clientId,
+      domainB.clientSecret,
+      internalGatewayUrl,
+      { [roleB.id!]: [`preferred_username=${matchingB}`] },
+    );
+    createdOrgIdpIds.push(providerBId);
+
+    await defaultApi.patchOrganizationSettings({
+      organizationId: ORG_ID,
+      patchOrganization: { identities: [...existingIdentities, providerAId, providerBId] },
```
sclarkamlab: the .first() half is sorted, thanks. this half is still open though. existingIdentities is read on line 445 and written back here, and between the two you create two domains and two org providers, so the window is wide. if api-management-login-social registers Social in that gap this patch drops it, which is the exact thing your cleanUp comment warns about. re-reading the settings immediately before this patch would close it.

sagarchauhan-qa: Fixed — setup re-reads the settings immediately before the patch now, the same as `cleanUp` does.

Checked it by registering a decoy provider inside that window, as a parallel spec would:

| | decoy survives the patch |
|---|---|
| with the re-read | true |
| with the old snapshot | false |

Worth noting this spec passed either way — the damage only shows up in the other one, which is why it was invisible. Same as the `.first()` half.

63 pass across `specs/management/identity-provider` and `specs/management/api-management`.

sclarkamlab (approved):

PR #8452, opened 2026-08-25
test: authenticate through an application after a context path change (AM-2224)

sagarchauhan-qa:
## Summary

Changing a domain's context path is already well covered — validation rules, overlap detection, and the OIDC discovery document advertising endpoints under the new path. But advertising an endpoint and being able to *use* it are different things, and the second half of AM-2224 was untested: nothing ever created an application and signed in through it afterwards.

## Scenarios

The three from the ticket:

- a user signs in through an application on the new context path, and a token is issued
- that token's issuer matches the new context path
- the previous context path is no longer served

## Why Jest, and nothing at the other layers

Path validation is already unit-tested in `DomainValidatorTest` (empty, null, multiple slashes, non-leading slash), so none of that is duplicated here. What was missing is emergent across gateway routing → OIDC discovery → token issuance, which a unit test cannot reach. Playwright would add a browser without adding an assertion: the ticket asks about authenticating through an application, not about changing the path in the Console.

## Notes for review

**A shared helper carries an assumption this ticket breaks.** `signInUser` asserts the login redirect points at `${AM_GATEWAY_URL}/${domain.hrid}/login`, i.e. that a domain is served on its hrid — exactly what changing the context path stops being true. It is given a domain-shaped object carrying the new path instead, which turns its internal assertion into a useful one for this suite: it confirms the login page moved along with the other endpoints. The helper itself is untouched.

**The old-path check needed a discriminator.** Asserting "the previous path returns 4xx" would pass identically if the domain had never been routed there at all. The fixture therefore probes the original path *before* the change and the test asserts it returned 200 first, so the pair genuinely demonstrates the move.

**New file, not an extension of `path-mode.jest.spec.ts`.** That suite is order-dependent shared state — its `beforeAll` creates the domains and a later test depends on an earlier one having applied the path change. This work needs an application, a user and a full gateway sign-in, and bolting that on would make an already fragile file worse. The two are also different concerns: path *validation* versus authenticating *through* a moved path.

**How to run — `npm run test` will not work.** These drive the gateway, and the dev config points `AM_INTERNAL_GATEWAY_URL` at `localhost`, which is not the gateway from inside the management container.

```
npx jest --config=api/config/ci.config.js specs/management/domain/context-path-auth.jest.spec.ts
```

## Testing

Run locally against a Mongo local stack:

| Suite | Result |
|---|---|
| `context-path-auth.jest.spec.ts` | 3 passed |
| `specs/management/domain` (whole suite) | 126 passed, 11 suites |

The wider run includes the existing `path-mode.jest.spec.ts`, which covers the same feature.

**Negative controls.** Each new assertion was deliberately broken and confirmed to fail, then reverted:

- sign-in — `Expected "…/ctxpath-auth-generic-insect-477/login"`, `Received "…/ctxpath/login"`, so the login page really does move
- issuer — `Expected "/ctxpath-auth-awesome-cat-9035"`, `Received "http://localhost:8092/ctxpath/oidc"`, so the issuer really does follow the new path

### DIFF
--- a/gravitee-am-test/specs/management/domain/context-path-auth.jest.spec.ts
+++ b/gravitee-am-test/specs/management/domain/context-path-auth.jest.spec.ts
@@ -15,7 +15,8 @@
  */
 import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
 import { jira } from '@specs-utils/jira';
-import { performGet, requestToken, signInUser } from '@gateway-commands/oauth-oidc-commands';
+import { introspectToken, performGet, performPost, requestToken, signInUser } from '@gateway-commands/oauth-oidc-commands';
+import { applicationBase64Token } from '@gateway-commands/utils';
 import { setup } from '../../test-fixture';
 import { ContextPathAuthFixture, setupContextPathAuthFixture } from './fixtures/context-path-auth-fixture';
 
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/management/domain/context-path-auth.jest.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  it(jira`the previous context path is no longer served ${'AM-2224'}`, async () => {
+    // The domain was reachable there before the change — otherwise a refusal below would prove
+    // nothing, since an unrouted path is refused whether or not anything ever moved.
+    expect(fixture.originalPathStatusBeforeChange).toEqual(200);
+
+    const response = await performGet(process.env.AM_GATEWAY_URL, `${fixture.originalPath}/oidc/.well-known/openid-configuration`);
+
+    expect(response.status).toBeGreaterThanOrEqual(400);
+    expect(response.status).toBeLessThan(500);
+  });
```
lgw-gravitee, line 83: question: should we introspect the token minted before the path is changed? (/token, updatepath, introspect -> fail) ?

sagarchauhan-qa: Good question — tried it, and it doesn't fail. Added a test in 846dee3511.

A token minted on the old path is still active when introspected after the change, and keeps its original iss, which no longer resolves:

 {"active":true,"iss":"http://localhost:8092/ctxpath-auth-.../oidc", ...}

Moving a context path is a routing change rather than a revocation event, so tokens already issued keep working at the endpoints in their new location. The introspection endpoint on 
the old path is itself gone (404).

The test asserts all three: still active, old issuer retained, old endpoint unrouted.

lgw-gravitee (approved):

PR #8454, opened 2026-08-25
test: remove custom metadata from an application (AM-2173)

sagarchauhan-qa:
[AM-2173](https://gravitee.atlassian.net/browse/AM-2173)

Cover removing custom metadata from an application. Removing the entry in the Console is already covered; whether the removal takes effect for a claim that referred to it was not.

- The entry disappears from the application, a freshly issued token no longer carries it, and the remaining metadata is untouched and still readable
- The removal is confirmed by polling a freshly issued token rather than `waitForSyncAfter`, since an application change does not reliably advance the domain's `lastSync`
- Adds the same optional `applicationMetadata` to the token identity fixture as #8453 (AM-2175). Whichever of the two merges second will need the duplicated lines removed

## :memo: Test scenarios
See jira.

n/a

n/a

@gravitee-io/am

[AM-2173]: https://gravitee.atlassian.net/browse/AM-2173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+});
+
+/** A fresh token's claims, so the gateway's in-memory view is what gets inspected. */
+const freshClaims = async (): Promise<Record<string, any>> => {
+  const tokens = await fixture.passwordGrant('openid');
+  return decodeToken(tokens.access_token).payload;
+};
+
+describe('Application metadata removal (AM-2173)', () => {
```
wajb: minor: IMO we don't need to add the jira reference to the describe block names. It duplicates to every test path and can't be parsed either. e.g. in [junit-gateway.xml](https://output.circle-artifacts.com/output/job/8ad59ea7-5a93-40f3-8360-4b26ce0ae75d/artifacts/0/jest-reports/junit/junit-gateway.xml):
```xml
<testcase classname="Application metadata removal (AM-2173) > after the entry is removed" name="the entry no longer appears in the application metadata [AM-2173]" time="0.013" file="specs/gateway/token-claims/application-metadata-removal.jest.spec.ts">
```

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+      // observable — a freshly issued token — is both accurate and what this test cares about.
+      const deadline = Date.now() + 30_000;
+      for (;;) {
+        if ((await freshClaims()).app_tier === undefined) {
+          return;
+        }
+        if (Date.now() > deadline) {
+          throw new Error('gateway still issued the removed metadata value 30s after it was deleted');
+        }
+      }
```
wajb: suggestion: use `retryUntil` helper.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { jira } from '@specs-utils/jira';
+import { getApplication, patchApplication } from '@management-commands/application-management-commands';
+import { retryImmediatelyForThisFile, setup } from '../../test-fixture';
+import { decodeToken, setupTokenIdentityFixture, TokenIdentityFixture } from './fixtures/token-identity-fixture';
+
+setup(200000);
+// The removal happens once, between the two describes below, so the tests are order-dependent.
+retryImmediatelyForThisFile();
```
wajb: thought: rather than making this suite sequential (which should be avoided - see GUIDELINES.md) could we rewrite this file as a more general `application-metadata.jest.spec.ts` that tests, separately, both adding and removing the metadata? This would be more balanced coverage and likely more resilient, even if the adding part is repeated.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * AM-2173 / UC-AM22 — removing custom metadata from an application.
+ *
+ * Removing the row in the Console is covered by Playwright. What that cannot show is whether the
+ * removal takes effect for a claim that referred to it: the application is held in memory by the
+ * gateway, so a stale value would surface in a token requested after the change rather than in
+ * the saved settings.
+ */
```
wajb: minor: can we remove this comment? The `jira` tagged template should be enough to link to the test case and the description is prone to become stale in the codebase.

gravitee-am-test/specs/gateway/token-claims/fixtures/token-identity-fixture.ts
```diff
@@ -215,6 +221,8 @@ export const setupTokenIdentityFixture = async (options: TokenIdentityOptions =
         },
         advanced: { skipConsent: true },
       },
+      // Application metadata is a top-level field, not part of the oauth settings.
```
wajb: nitpick: comment is noisy.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
sagarchauhan-qa: Done — describes are now `Application metadata - <behaviour>`, matching the other specs in the folder. The `jira` template still tags each test, so the key appears once instead of in both `classname` and `name`.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
sagarchauhan-qa: Done. Worth flagging that this was more than style: my loop had no delay between attempts, so it hit the token endpoint as fast as the network allowed for up to 30s. `retryUntil` gives it a 500ms interval.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
sagarchauhan-qa: Agreed, done — renamed to `application-metadata.jest.spec.ts` and split into two self-contained tests, one for reading an entry back as a claim and one for removing it. Each arranges its own state, so `retryImmediatelyForThisFile()` is gone. Both now pass individually via `-t` as well as together.

Restructuring surfaced a second problem worth mentioning: the old `beforeAll` polled until `app_tier` was `undefined` and the test then asserted it was absent, which was true by construction. The wait now keys on a `revision` entry rewritten on every change, so it is independent of the entry under test. Confirmed with a negative control — forcing the removal not to happen fails cleanly at the assertion (`Received value: "platinum"`) rather than timing out in the poll.

gravitee-am-test/specs/gateway/token-claims/application-metadata-removal.jest.spec.ts
sagarchauhan-qa: Done, removed.

gravitee-am-test/specs/gateway/token-claims/fixtures/token-identity-fixture.ts
sagarchauhan-qa: Done, removed.

sagarchauhan-qa:
Thanks @wajb — all five addressed in a7f5464.

Verified locally against a Mongo local stack:

| Check | Result |
|---|---|
| `application-metadata.jest.spec.ts` | 2 passed |
| each test run alone via `-t` | both pass independently |
| `specs/gateway/token-claims` (whole folder) | 81 passed, 12 suites |

One note on the earlier run in this PR: the local failures I saw were the gateway having OOM’d (every Vert.x event-loop thread threw `OutOfMemoryError`, port 8092 timing out), not the spec. Restarted it before re-testing.

wajb (approved):

PR #8457, opened 2026-08-26
test: identity provider selection rule edge cases (AM-2171)

sagarchauhan-qa:
[AM-2171](https://gravitee.atlassian.net/browse/AM-2171)

Cover two gaps in identity provider selection rules. Routing a matching user to the right provider, and a user matching no rule at all, are already covered in `login-flow-inline.jest.spec.ts` and are not duplicated here.

- **A rule that cannot be evaluated.** `selectionRuleMatches` catches the failure, logs a warning and returns false, so the provider is skipped rather than the sign-in failing outright — a typo in a rule silently excludes a provider. Covered from both sides: with a second provider present the sign-in succeeds and returns *that* provider's profile, and when the broken rule is the only provider the sign-in is refused as `invalid_user` rather than surfacing an internal error.
- **A provider carrying no rule.** The existing fallback test asserts only that a code came back, and the fixture user exists in both providers, so it would pass with selection rules ignored entirely. The test here reads the profile back, and the ruled provider holds the higher priority, so a pass can only mean it was excluded rather than beaten to it.

No existing files changed. The weaker fallback test is left as it is, with this one alongside.

## :memo: Test scenarios
See jira.

n/a

n/a

@gravitee-io/am

[AM-2171]: https://gravitee.atlassian.net/browse/AM-2171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/login-flow/idp-selection-rules.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/login-flow/idp-selection-rules.jest.spec.ts
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { jira } from '@specs-utils/jira';
+import { getHeaderLocation, login, loginUserNameAndPassword } from '@gateway-commands/login-commands';
+import { logoutUser, performGet, requestToken } from '@gateway-commands/oauth-oidc-commands';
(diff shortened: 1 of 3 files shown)


gravitee-am-test/specs/gateway/login-flow/idp-selection-rules.jest.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+let appMalformedWithFallback: any;
+let appMalformedOnly: any;
+let appUnruledAlongsideRuled: any;
+
+/**
+ * Both inline providers hold the same username with a different profile, so signing in and
+ * reading `given_name` back says which provider actually answered. In every application below
+ * the ruled provider carries the higher priority, so it would answer if its rule were ignored.
+ */
+const signInAndReadGivenName = async (app: any): Promise<string> => {
```
sclarkamlab: this is almost identical to signInAndReadProfile from https://github.com/gravitee-io/gravitee-access-management/pull/8456/changes can we reuse the existing function

sagarchauhan-qa: Good spot — moved it into the shared `login-flow-inline-fixture` as `signInAndReadProfile`, taking the application and OIDC configuration as arguments so it isn't bound to one spec. Both now use the single copy (d5c81fbde3).

That does mean a small change to `idp-priority-order.jest.spec.ts` from #8456, since that's where the original lived — happy to pull it back to just this spec if you'd rather keep the PR narrower.

login-flow suite passes: 5 suites, 29 tests.

sclarkamlab (approved):

PR #8458, opened 2026-08-26
test: make the hidden login form test able to fail (AM-2169)

sagarchauhan-qa:
[AM-2169](https://gravitee.atlassian.net/browse/AM-2169)

**This modifies an existing test** (`login-flows-external-oidc.spec.ts`, originally by @Fu Yu) rather than adding one, because the existing test could not fail on the behaviour it covers.

It fell back to doing the thing the feature is supposed to prevent:

```js
try {
  await page.waitForURL(providerLoginRe);         // redirect worked
} catch {
  await page.waitForURL(clientLoginRe);           // redirect did NOT work
  await clickClientOauth2GenericAmSocial(page);   // ...click the provider button by hand
  await page.waitForURL(providerLoginRe);
}
expect(page.url()).not.toMatch(clientLoginRe);    // passes either way
```

With "hide login form" broken, the catch branch ran, clicked through manually, and the final assertion still passed — indistinguishable from the ordinary external-provider test (AM-2207).

Changes:

- **Fallback removed.** If the redirect does not happen the browser stays on a rendered login form and the test fails.
- **Assert on the response, not the URL.** `LoginHideFormHandler` is a handler on `/login` itself and answers with a 302 instead of rendering the form, so the browser legitimately reaches that URL. Asserting "never visited `/login`" would fail against working software; the test records the response statuses for that URL and asserts they are redirects.
- **Added the setting-off comparison**, where the same domain serves its login form. Same single external provider either way, so only the setting differs.

`AM-2207`'s test in the same file is untouched.

## :memo: Test scenarios
See jira.

n/a

Not covered: more than one external provider. `LoginHideFormHandler` only redirects when `isHideFormEnabled && hasExactlyOneSocialProvider`, and this fixture deletes the default provider to get down to one, so that case needs a different setup.

Verified the fix does what it claims by running the corrected test with `hideLoginForm: false` — it fails on the redirect that never comes (`TimeoutError: page.waitForURL`), where the original passed. Suite is green: 4 passed.

@gravitee-io/am

[AM-2169]: https://gravitee.atlassian.net/browse/AM-2169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/playwright/tests/auth/login/login-flows-external-oidc.spec.ts
+++ b/gravitee-am-test/playwright/tests/auth/login/login-flows-external-oidc.spec.ts
@@ -67,15 +67,19 @@ test.describe('Hide login form with external IdP (AM-2169)', () => {
     linkJira(testInfo, 'AM-2169');
 
     const clientId = externalOidcBundle.clientApp.settings.oauth.clientId;
-    const providerLoginRe = new RegExp(`${externalOidcBundle.providerDomain.hrid}.*/login`, 'i');
-    const clientLoginRe = new RegExp(`${externalOidcBundle.clientDomain.hrid}/login`, 'i');
+
+    // Matched on the path alone rather than anywhere in the URL. The provider's authorize URL
+    // carries the client's own `/login/callback` inside its redirect_uri parameter, so a
+    // whole-URL match counts hops that never went near the client's login page.
+    const clientLoginPath = `/${externalOidcBundle.clientDomain.hrid}/login`;
+    const providerLoginPath = `/${externalOidcBundle.providerDomain.hrid}/login`;
 
     // LoginHideFormHandler sits on /login itself, so the browser does reach that URL — it is
-    // answered with a 302 rather than a rendered form. Recording the statuses is therefore how
+    // answered with a 302 rather than a rendered form. Recording the status is therefore how
     // "the form was hidden" is proved; the URL alone cannot show it.
     const clientLoginStatuses: number[] = [];
     page.on('response', (response) => {
-      if (clientLoginRe.test(response.url())) {
+      if (new URL(response.url()).pathname === clientLoginPath) {
         clientLoginStatuses.push(response.status());
       }
     });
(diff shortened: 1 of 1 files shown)


gravitee-am-test/playwright/tests/auth/login/login-flows-external-oidc.spec.ts
```diff
@@ -67,20 +67,29 @@ test.describe('Hide login form with external IdP (AM-2169)', () => {
...
+    });
+
+    await page.goto(buildAuthorizeUrl(externalOidcBundle.clientGatewayUrl, clientId));
+
+    // No fallback to clicking the provider button by hand: if the redirect does not happen, the
+    // browser stays on a rendered login form and this fails, which is the point of the feature.
+    await page.waitForURL(providerLoginRe, { timeout: BRIEF_TIMEOUT * 6 });
+
+    expect(clientLoginStatuses.length).toBeGreaterThan(0);
+    expect(clientLoginStatuses.every((status) => status === 302)).toBe(true);
```
sclarkamlab: nit, this fails with true !== false and hides what the statuses actually were. expect(clientLoginStatuses).toEqual([302]) tells you straight away, and it covers the length check on the line above too.

sclarkamlab (approved):

gravitee-am-test/playwright/tests/auth/login/login-flows-external-oidc.spec.ts
sagarchauhan-qa: Done — though your line needed one change underneath to work.

The array was actually `[302, 302]`. The second one wasn't a request to the client's login page at all: it was the provider's `/oauth/authorize`, which matched because the client's `/login/callback` sits inside its `redirect_uri` parameter. So the old assertion was counting an unrelated hop.

Matching on the path rather than anywhere in the URL fixes that, and then it really is `[302]` and your version works. `providerLoginRe` had the same problem, so that's a path check now too.

4 passed in the file.

sclarkamlab (approved):

PR #8459, opened 2026-08-26
test: a policy in the login flow that refuses a sign-in (AM-2199)

sagarchauhan-qa:
[AM-2199](https://gravitee.atlassian.net/browse/AM-2199)

Every policy covered so far adds or sets something, and every one of those tests ends with the sign-in succeeding. Nothing showed that a policy can turn a user away, so a flow engine that stopped honouring a refusal would go unnoticed.

- A Validate Request policy with a rule that cannot be satisfied refuses the sign-in, and no authorization code is issued
- The same user gets in once the policy is removed — the discriminator, since same user and credentials with only the policy differing means a refusal cannot be mistaken for bad credentials or a broken domain
- A Groovy policy that fails at runtime also stops the sign-in

The policies are attached to the Login **post** step, so credentials are accepted and the policy then refuses. On the pre step the refusal lands on the login page request itself, before any credentials are entered.

Two new files; nothing existing changed.

## :memo: Test scenarios
See jira.

n/a

The ticket left the runtime-failure outcome open ("the outcome is the one the product intends, rather than left to chance"). Determined by running it rather than assumed: **a policy that errors mid-flow stops the sign-in**. The redirect carries the Groovy error itself (`Cannot invoke method ... on null object`), which shows the script ran and threw rather than being rejected before execution.

Not done: the ticket also asks to retag, since `flow-policy.jest.spec.ts` carries this ticket's key but only checks that sign-in still works. Left alone as tagging cleanup is out of scope here — tracked as **[AM-7590](https://gravitee.atlassian.net/browse/AM-7590)**, to be actioned once this PR and #8460 have merged.

@gravitee-io/am

[AM-2199]: https://gravitee.atlassian.net/browse/AM-2199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AM-7590]: https://gravitee.atlassian.net/browse/AM-7590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
+++ b/gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect } from '@jest/globals';
+import { Domain } from '@management-models/Domain';
+import {
+  createDomain,
+  DomainOidcConfig,
+  getDomainFlows,
+  safeDeleteDomain,
+  startDomain,
+  updateDomainFlows,
+  waitForDomainStart,
+} from '@management-commands/domain-management-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { getAllIdps } from '@management-commands/idp-management-commands';
+import { createUser } from '@management-commands/user-management-commands';
+import { createApplication, updateApplication } from '@management-commands/application-management-commands';
+import { lookupFlowAndResetPolicies } from '@management-commands/flow-management-commands';
+import { waitForSyncAfter } from '@gateway-commands/monitoring-commands';
+import { uniqueName } from '@utils-commands/misc';
+import { Fixture } from '../../../test-fixture';
+import { FlowEntityTypeEnum } from '../../../../api/management/models';
+
+export const REDIRECT_URI = 'https://callback';
+
+export const REFUSING_USER = {
+  username: 'refusing-policy-user',
+  password: 'SomeP@ssw0rd',
+  firstName: 'Refusing',
+  lastName: 'Policy',
+  email: 'refusing-policy@example.com',
+  preRegistration: false,
+};
+
+/**
+ * A Validate Request rule that cannot be satisfied: it requires a request parameter that the
+ * sign-in never sends, so the policy refuses every request reaching this flow step.
+ */
+export const refusingPolicy = (status = '401') => ({
+  name: 'Validate Request',
+  policy: 'policy-request-validation',
+  description: '',
+  condition: '',
+  enabled: true,
+  configuration: JSON.stringify({
+    status,
+    rules: [
+      {
+        input: "{#request.params['a-parameter-never-sent']}",
+        isRequired: true,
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
```diff
@@ -0,0 +1,152 @@
...
+  DomainOidcConfig,
+  getDomainFlows,
+  safeDeleteDomain,
+  startDomain,
+  updateDomainFlows,
+  waitForDomainStart,
+} from '@management-commands/domain-management-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { getAllIdps } from '@management-commands/idp-management-commands';
+import { createUser } from '@management-commands/user-management-commands';
```
sclarkamlab, line 30: heads up, #8460 adds this same new file, and its copy is a superset of this one (extra enrichProfilePolicy and readUserProfile). whichever lands second is an add/add conflict. worth stacking one branch on the other so the shared fixture only lands once.

gravitee-am-test/specs/gateway/flows/refusing-policy.jest.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+  it(jira`a policy that fails at runtime also stops the sign-in ${'AM-2199'}`, async () => {
+    await fixture.setLoginPolicies('post', [throwingPolicy()]);
+
+    const location = await signInAndReadRedirect();
+
+    // Pinning the behaviour down rather than assuming it: a policy erroring mid-flow turns the
+    // user away, it does not let them through. The message is the Groovy failure itself, which
+    // shows the script ran and threw rather than being rejected before execution.
+    expect(location).toContain('error=login_failed');
+    expect(location).toContain('Cannot+invoke+method');
```
sclarkamlab: this pins a groovy runtime message into a user facing redirect as expected behaviour. two things i would rather not lock in: the wording breaks whenever groovy changes it, and an internal error string reaching the browser is arguably a leak we should be fixing rather than asserting. can we assert the failure without the jvm message, and raise the leak separately?

sagarchauhan-qa: Done. It now asserts `error=login_failed` and that the user is back on the domain's own login page with no code issued — no JVM wording.

On the leak: the raw message does reach the redirect as `error_description`, so it ends up in browser history and access logs. Raising that separately.

gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
sagarchauhan-qa: Stacked — #8460 now sits on top of this branch, so the fixture only lands once. It shows there as +30/-1 rather than a new file.

This one merges first.

gravitee-am-test/specs/gateway/flows/refusing-policy.jest.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    // Pinning the behaviour down rather than assuming it: a policy erroring mid-flow turns the
+    // user away, it does not let them through.
+    expect(url.searchParams.get('error')).toEqual('login_failed');
+    expect(url.pathname).toEqual(`/${fixture.domain.hrid}/login`);
+    expect(location).not.toMatch(/[?&]code=/);
+    expect(location).not.toContain(REDIRECT_URI);
+
+    // The gateway currently puts the script's own runtime message into error_description. That is
+    // deliberately not asserted here: the wording belongs to the JVM rather than to AM, and
+    // whether an internal message should reach the browser at all is tracked separately.
```
sclarkamlab: assertions look right now. same ask as 8461 though, can you name the ticket here rather than tracked separately? you gave AM-7591 on the other one and this reads as the same kind of note.

sagarchauhan-qa: Named — it's AM-7593. Added a note that if the description is made generic, this test keeps passing.

sclarkamlab (approved):

PR #8460, opened 2026-08-26
test: several policies in one flow where one refuses or fails (AM-2198)

sagarchauhan-qa:
[AM-2198](https://gravitee.atlassian.net/browse/AM-2198)

Policies running together, in order, and domain and application flows combining are covered already. Every one of those tests ends with the sign-in succeeding, so nothing showed what a chain does when a member of it turns the user away or breaks.

- A chain where every policy runs — the anchor for the two below
- A refusing policy among others stops the sign-in, and what the earlier policy wrote is kept
- A policy failing partway stops the ones after it from running

Enrich User Profile is the observable step, since it writes to the user's stored profile rather than only to the token — so whether it ran can still be established after a sign-in that issued nothing.

>⚠️ **Stacked on #8459 — that one merges first.**

This branch is based on `test/AM-2199-policy-refuses-signin`, not on master, so the shared fixture is added once by #8459 and only extended here (+30/-1, adding `enrichProfilePolicy` and `readUserProfile`). Merging this before #8459 would land a spec whose fixture is not on master.

One new file of its own; nothing existing changed.

## :memo: Test scenarios
See jira.

n/a

The ticket left two outcomes open. Both were determined by running them rather than assumed:

- a refused sign-in **still leaves** what earlier policies wrote against the user — a refusal is not a rollback
- a policy failing partway **stops** the policies after it from running

Not done, and tracked separately: `flow-policy.jest.spec.ts` carries the AM-2198 and AM-2199 keys but cannot exercise flows as written. It signs in with `grant_type=password`, which never reaches the login page, so Login-flow policies do not run for it — its fixture even defines a `groovy_claim` that never resolves. Confirmed by testing: adding the missing Enrich Auth Flow policy and moving the policies to the Root step both still gave `undefined`. Fixing it means moving that spec onto the authorization-code flow, which rewrites two tests that are not this ticket's, so it is left for a separate PR, tracked as **[AM-7590](https://gravitee.atlassian.net/browse/AM-7590)**. That deletion has to wait until this PR and #8459 have merged, since until then those two tests are the only ones on master carrying the AM-2198 and AM-2199 keys.

@gravitee-io/am

[AM-2198]: https://gravitee.atlassian.net/browse/AM-2198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AM-7590]: https://gravitee.atlassian.net/browse/AM-7590?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
+++ b/gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
@@ -27,7 +27,7 @@ import {
 } from '@management-commands/domain-management-commands';
 import { requestAdminAccessToken } from '@management-commands/token-management-commands';
 import { getAllIdps } from '@management-commands/idp-management-commands';
-import { createUser } from '@management-commands/user-management-commands';
+import { createUser, listUsers } from '@management-commands/user-management-commands';
 import { createApplication, updateApplication } from '@management-commands/application-management-commands';
 import { lookupFlowAndResetPolicies } from '@management-commands/flow-management-commands';
 import { waitForSyncAfter } from '@gateway-commands/monitoring-commands';
@@ -68,6 +68,24 @@ export const refusingPolicy = (status = '401') => ({
   }),
 });
 
+/**
+ * An Enrich User Profile policy writing a fixed value onto the signing-in user.
+ *
+ * Its effect is stored against the user rather than only reaching the token, so whether it ran
+ * can still be established after a sign-in that was refused and issued no token.
+ */
+export const enrichProfilePolicy = (key: string, value: string) => ({
+  name: 'Enrich User Profile',
+  policy: 'policy-am-enrich-profile',
+  description: '',
+  condition: '',
+  enabled: true,
+  configuration: JSON.stringify({
+    exitOnError: false,
+    properties: [{ claim: key, claimValue: value }],
+  }),
+});
+
 /** A Groovy policy whose script throws, for the runtime-failure case. */
 export const throwingPolicy = () => ({
   name: 'Groovy',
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/flows/fixture/refusing-policy-fixture.ts
```diff
@@ -0,0 +1,181 @@
...
+  DomainOidcConfig,
+  getDomainFlows,
+  safeDeleteDomain,
+  startDomain,
+  updateDomainFlows,
+  waitForDomainStart,
+} from '@management-commands/domain-management-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { getAllIdps } from '@management-commands/idp-management-commands';
+import { createUser, listUsers } from '@management-commands/user-management-commands';
```
sclarkamlab, line 30: heads up, #8459 adds this same new file with a different body. yours is a superset, it also has enrichProfilePolicy and readUserProfile. whichever lands second is an add/add conflict, and if that gets resolved towards the 8459 version this spec breaks on master. worth stacking one on the other so the fixture only lands once.

sagarchauhan-qa: Stacked — this PR is now based on #8459's branch, so the fixture only lands once. It shows here as +30/-1 rather than a new file, and the conflict is resolved to this superset version.

#8459 merges first. All 14 tests in specs/gateway/flows/ pass on the rebased branch.

sclarkamlab (approved):

PR #8461, opened 2026-08-26
test: an application with no client authentication method set (AM-2232)

sagarchauhan-qa:
[AM-2232](https://gravitee.atlassian.net/browse/AM-2232)

Every client authentication method is covered elsewhere with that method fixed on the application. The arbitrary option itself was untested: one application, no method set, accepting whichever method the incoming request uses.

- An unset application accepts credentials in the authorization header, and in the request body
- A fixed application accepts its own method and refuses a different one

Both applications live in one domain, so the comparison differs only in that setting.

Two new files; nothing existing changed.

## :memo: Test scenarios
See jira.

n/a

The ticket asked for the assertion behaviour to be confirmed against a running gateway before expectations were agreed. It was, and it looks like a defect: **an unset application refuses `client_secret_jwt`** even though `ClientAssertionAuthProvider.canHandle` carries the same empty-method branch as the basic and post providers, which both work.

Isolated to a single variable — the identical assertion request is accepted by a second application in the same domain whose method is set to `client_secret_jwt`, and refused when the method is empty. Raised as **AM-7591**. That case is deliberately left out of this suite rather than asserting the current behaviour as expected; it should be added once the bug is resolved.

One trap worth knowing, which is why the first test exists: **omitting `tokenEndpointAuthMethod` does not create an unset application** — the management API stores `client_secret_basic`. The first probe of this work did exactly that, and every method behaved as though fixed. The application has to be sent an explicit empty string. That first test asserts the stored method is `""`, so the rest of the file cannot silently run against the wrong subject.

Not covered: `private_key_jwt` against an unset application, which needs a registered key pair; and mTLS, whose existing tests need the `gateway-mtls` container.

@gravitee-io/am

[AM-2232]: https://gravitee.atlassian.net/browse/AM-2232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/token-auth-methods/arbitrary-auth-method.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/token-auth-methods/arbitrary-auth-method.jest.spec.ts
@@ -36,8 +36,9 @@ setup(200000);
  * Assertion-based methods are deliberately not covered here. An unset application refuses
  * client_secret_jwt even though `ClientAssertionAuthProvider.canHandle` carries the same
  * empty-method branch as basic and post — the identical assertion is accepted once the method is
- * set explicitly. That looks like a defect rather than intended behaviour, so it is raised
- * separately instead of being written down here as expected.
+ * set explicitly. That looks like a defect rather than intended behaviour, so it is tracked as
+ * AM-7591 instead of being written down here as expected. Add the assertion cases to this file
+ * once that is resolved.
  */
 let fixture: ArbitraryAuthMethodFixture;
 
--- a/gravitee-am-test/specs/gateway/token-auth-methods/arbitrary-auth-method.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/token-auth-methods/arbitrary-auth-method.jest.spec.ts
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
-import jwt from 'jsonwebtoken';
-import crypto from 'crypto';
 import { jira } from '@specs-utils/jira';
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/token-auth-methods/arbitrary-auth-method.jest.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+ * is the arbitrary option itself: one application, no method set, accepting whichever the incoming
+ * request uses. `ClientBasicAuthProvider.canHandle` allows it when the stored method is null or
+ * empty and the matching credentials are present.
+ *
+ * The two applications live in one domain, so the comparison differs only in that setting.
+ *
+ * Assertion-based methods are deliberately not covered here. An unset application refuses
+ * client_secret_jwt even though `ClientAssertionAuthProvider.canHandle` carries the same
+ * empty-method branch as basic and post — the identical assertion is accepted once the method is
+ * set explicitly. That looks like a defect rather than intended behaviour, so it is raised
```
sclarkamlab: can you put the ticket key in here? this is the most interesting thing in the PR and right now raised separately is not traceable to anything, so it gets lost the moment this file stops being read.

sagarchauhan-qa: Added. It's AM-7591, now named in the file with a note to add the assertion cases once it's fixed.

It was in the PR description, but you're right that that's the wrong place for it — the file outlives the PR.

sclarkamlab (approved):

PR #8462, opened 2026-08-26
test: a translation used on the sign-in form (AM-2184)

sagarchauhan-qa:
[AM-2184](https://gravitee.atlassian.net/browse/AM-2184)

Creating a translation and saving its entries was already covered, but nothing rendered a page — a domain that stored translations and ignored them would have passed.

- The custom wording appears when the sign-in page is asked for in the translated language
- Another language falls back to the default wording
- An entry left out of the translation keeps its default
- A key of the domain's own is shown once the form's template asks for it

Two new files in a new `specs/gateway/i18n/` directory; nothing existing changed.

## :memo: Test scenarios
See jira.

n/a

**An untranslated entry does not fall back to English.** AM ships its own bundle per language, so the page stays in the language that was asked for and only the translated entries are replaced — with a French translation covering just the title, the username label renders as `Identifiant`, not `Username`. The test asserts the French default and that the English wording is absent.

**Overriding a shipped key and adding a new one are different things.** A key of your own renders only if the template references it (`<label th:text="#{login.extra.text}"/>`), which is why that test replaces the domain's login form. Worth knowing: `login.button` is not a key — the sign-in button uses `login.button.submit`, and writing the wrong key fails silently.

**Sync.** The gateway is polled until it serves each write rather than waiting on the domain's `lastSync`, which a dictionary update does not reliably advance — waiting on it hung for the full 90 seconds.

Not covered: translations on other forms (registration, forgotten password, MFA), and translated wording in emails, which use a different templating mechanism and belong to their own ticket.

@gravitee-io/am

[AM-2184]: https://gravitee.atlassian.net/browse/AM-2184?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
+++ b/gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
@@ -1,0 +1,155 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Domain } from '@management-models/Domain';
+import {
+  createDomain,
+  DomainOidcConfig,
+  safeDeleteDomain,
+  startDomain,
+  waitForDomainStart,
+} from '@management-commands/domain-management-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { getAllIdps } from '@management-commands/idp-management-commands';
+import { createApplication, updateApplication } from '@management-commands/application-management-commands';
+import { createDictionary, updateDictionaryEntries } from '@management-commands/dictionary-management-commands';
+import { getDomainApi } from '@management-commands/service/utils';
+import { readFileSync } from 'fs';
+import { join } from 'path';
+import { uniqueName } from '@utils-commands/misc';
+import { Fixture } from '../../../test-fixture';
+
+export const REDIRECT_URI = 'https://callback';
+
+/** The language the translation is written for. Not English, so the fallback case is distinguishable. */
+export const TRANSLATED_LANGUAGE = 'fr';
+
+/** Recognisable wording, so a match cannot come from the shipped bundle by coincidence. */
+export const TRANSLATED_TITLE = 'AM-2184 translated sign-in title';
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
```diff
@@ -0,0 +1,155 @@
...
+
+      await getDomainApi(accessToken).createForm({
+        organizationId: process.env.AM_DEF_ORG_ID!,
+        environmentId: process.env.AM_DEF_ENV_ID!,
+        domain: domain.id,
+        newForm: { template: 'LOGIN' as any, content: injected, enabled: true },
+      });
+      templateReferencesCustomKey = true;
+    } catch (e) {
+      console.warn('Could not override the login form; the custom-key test will be skipped:', e);
```
sclarkamlab: the warn says the custom-key test will be skipped, but the test asserts templateReferencesCustomKey is true, so it fails instead. and it fails in the spec with no sign of the real cause, which is buried in this warn. i would let this throw out of the fixture.

gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
```diff
@@ -0,0 +1,155 @@
...
+    // replaced with one that does. Overriding a shipped key needs no template change; adding a
+    // new one does — that difference is what the custom-key test exists to show.
+    let templateReferencesCustomKey = false;
+    try {
+      const loginHtmlPath = join(
+        __dirname,
+        '../../../../../gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/resources/webroot/views/login.html',
+      );
+      const original = readFileSync(loginHtmlPath, 'utf-8');
+      const injected = original.replace('<body>', `<body><label id="am2184-extra" th:text="#{${CUSTOM_KEY}}"></label>`);
```
sclarkamlab: if login.html ever stops having a literal <body> this replace is a no-op, and we still set templateReferencesCustomKey to true on line 117. the custom-key test then fails looking like a gateway bug. worth asserting the injected content actually changed.

gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
sagarchauhan-qa: Agreed — it throws out of the fixture now. The try/catch is gone, and so is `templateReferencesCustomKey`; it only existed for that assertion, which turned a setup problem into what looked like a gateway failure.

gravitee-am-test/specs/gateway/i18n/fixtures/translation-fixture.ts
sagarchauhan-qa: Fixed — it compares against the original and throws if nothing changed:

```
No <body> found in .../views/login.html, so the custom key was never added to the template
```

Checked by pointing the replace at a tag that isn't there: you get that message instead of a test failing as though the wording had not rendered. 4/4 pass otherwise.

sclarkamlab (approved):

PR #8463, opened 2026-08-26
test: healthy and expired certificates on the certificates page (AM-2220)

sagarchauhan-qa:
[AM-2220](https://gravitee.atlassian.net/browse/AM-2220)

A near-expiry certificate being flagged was already covered, but nothing showed a healthy one going unflagged — a Console that put a warning on every row would have passed.

- A near-expiry certificate is flagged, and its warning carries the tooltip saying how long is left
- A healthy certificate in the same domain is not flagged
- An already-expired certificate is refused on upload

Both certificates are created in the same domain and read from the same page, so only the expiry date differs. Two new files; nothing existing changed, including the original test in `certificate-expiry-warning.spec.ts`.

## :memo: Test scenarios
See jira.

n/a

**The ticket's third scenario cannot happen as written.** It assumed an expired certificate could be uploaded and its display inspected. AM refuses it: `400 — "The certificate you uploaded has already expired. Please select a different certificate to upload."` Only a certificate that expires while installed reaches that state, which a test cannot wait for, so the refusal is what is pinned down instead. A valid certificate uploads to the same domain in the test above, so the refusal is about the expiry rather than the upload.

**Certificates are built with node-forge** rather than by shelling out to OpenSSL. `openssl req` refuses a non-positive `-days`, and `-not_before`/`-not_after` only arrived in OpenSSL 3.5 — newer than the image CI runs on. Building them in JavaScript keeps every validity window available on any machine.

**Aliases are derived per certificate.** AM rejects a second certificate reusing an alias already in the domain (`409 — "A certificate with alias [test] already exists"`), and these tests deliberately put several in one domain.

Not covered, and worth its own ticket: the expiry notification shown against the bell in the Console, and the expiry email. The email notifier ships disabled, runs on a daily cron, and uses thresholds of 20, 15, 10, 5 and 1 days — testing it means enabling the notifier and overriding the schedule rather than asserting on a page.

@gravitee-io/am

[AM-2220]: https://gravitee.atlassian.net/browse/AM-2220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/playwright/tests/security/certificates/certificate-expiry-states.spec.ts
+++ b/gravitee-am-test/playwright/tests/security/certificates/certificate-expiry-states.spec.ts
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { createCertificate } from '@management-commands/certificate-management-commands';
+import { waitForDomainSync } from '@management-commands/domain-management-commands';
+import { expect, test } from '../../../fixtures/base.fixture';
+import { DomainCertificatesPage } from '../../../pages/domain-certificates.page';
+import { linkJira } from '../../../utils/jira';
+import { uniqueTestName } from '../../../utils/fixture-helpers';
+import { alreadyExpired, validForDays } from '../../../utils/pkcs12-certificate-with-validity';
+
+/**
+ * AM-2220 / UC-AM69 — how the certificates page distinguishes certificates by expiry.
+ *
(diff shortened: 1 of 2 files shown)


gravitee-am-test/playwright/tests/security/certificates/certificate-expiry-states.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    await certPage.navigateTo(testDomain.id);
+
+    const nearExpiryWarning = certPage.certificateRow(nearExpiryName).locator('mat-icon.warning');
+    await expect(nearExpiryWarning).toBeVisible();
+    // The tooltip carries the remaining life, so the warning says how urgent it is rather than
+    // only that something is wrong.
+    await expect(nearExpiryWarning).toHaveAttribute('aria-describedby', /.+/);
+
+    // The half that was missing: without it, a page that flagged every row would still pass.
+    await expect(certPage.certificateRow(healthyName).locator('mat-icon.warning')).toHaveCount(0);
```
sclarkamlab: this passes if the healthy row has not rendered yet, which is the case it is meant to rule out. worth asserting the row is visible first, then that it has no warning icon in it.

sagarchauhan-qa: Fixed — the row is asserted visible first, then that it has no warning icon.

Checked by pointing the healthy row at a certificate that was never created: it now fails on `toBeVisible`, where the old line passed. 3/3 pass otherwise.

sclarkamlab (approved):

PR #8466, opened 2026-08-27
test: recovery codes are spent when used (AM-2216)

sagarchauhan-qa:
[AM-2216](https://gravitee.atlassian.net/browse/AM-2216)

Being issued recovery codes and signing in with one was already covered, here and in Playwright. Nothing covered what happens next — a code that stayed valid after use would be a lasting way past MFA, and no test would have noticed.

- A code signs the user in, and the same code is refused afterwards
- A code that was never issued is refused
- Spending one code leaves the others usable
- Spending every code leaves the user with none
- Regenerating the codes stops the old ones working

Two new files; nothing existing changed, including `mfa-factor.spec.ts` and the other `mfa-factors/` specs.

## :memo: Test scenarios
See jira.

n/a

**A reused code and a code that was never issued return the same error** (`error=mfa_challenge_failed`). Every refusal is therefore paired with that same code having worked first — otherwise a test would pass equally well if the code had never worked at all.

**The MFA rate limit shapes the fixture.** It allows 2 attempts per minute keyed on user + factor + client, and one full recovery-code sign-in spends that bucket, so a second attempt on the same application is answered with `request_limit_error=mfa_request_limit_exceed` before the code is ever checked — the test would have been measuring the rate limiter. Retries go through a **second application in the same domain** instead, which has its own bucket. No rate-limit configuration is touched. It also makes the assertion stronger: a spent code works nowhere, rather than merely being rejected inside the original app's session.

**Consumption is shown directly, not only inferred.** The domain enables self-service account management so the tests can read the user's remaining codes from `/{domain}/account/api/auth/recovery_code`, and confirm a spent code has actually left the set. The token for that call comes from the authorization code the MFA sign-in produces, so it is earned by passing the challenge rather than by a password grant that would skip MFA entirely.

**The number of codes is configurable** (`{"digit":5,"count":2}`), so exhausting a user's set costs two sign-ins rather than the ten a default set would need.

Not covered: recovery codes through the Console UI, which `playwright/tests/auth/login/login-flows-recovery.spec.ts` already exercises.

@gravitee-io/am

[AM-2216]: https://gravitee.atlassian.net/browse/AM-2216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/mfa-factors/fixtures/recovery-code-fixture.ts
+++ b/gravitee-am-test/specs/gateway/mfa-factors/fixtures/recovery-code-fixture.ts
@@ -1,0 +1,356 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { expect } from '@jest/globals';
+import * as cheerio from 'cheerio';
+import {
+  createDomain,
+  patchDomain,
+  safeDeleteDomain,
+  startDomain,
+  waitForDomainStart,
+} from '@management-commands/domain-management-commands';
+import { requestAdminAccessToken } from '@management-commands/token-management-commands';
+import { getAllIdps } from '@management-commands/idp-management-commands';
+import { createApplication, updateApplication } from '@management-commands/application-management-commands';
+import { createFactor } from '@management-commands/factor-management-commands';
+import { buildCreateAndTestUser, deleteUser } from '@management-commands/user-management-commands';
+import {
+  extractXsrfTokenAndActionResponse,
+  performGet,
+  performPost,
+  requestToken,
+} from '@gateway-commands/oauth-oidc-commands';
+import { loginUserNameAndPassword } from '@gateway-commands/login-commands';
+import { uniqueName } from '@utils-commands/misc';
+import { Fixture } from '../../../test-fixture';
+
+const REDIRECT_URI = 'https://auth-nightly.gravitee.io/myApp/callback';
+
+/** The code the mock factor always accepts, so the usual method never needs a real device. */
+const MOCK_FACTOR_CODE = '123456';
+
+/**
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/mfa-factors/fixtures/recovery-code-fixture.ts
```diff
@@ -0,0 +1,356 @@
...
+      },
+    };
+  } catch (error) {
+    if (domain?.id) {
+      try {
+        await safeDeleteDomain(domain.id, accessToken);
+      } catch (e) {
+        console.error('Cleanup failed:', e);
+      }
+    }
```
wajb: suggestion: this can be simplified since inputs are validated and errors are handled in the util:

```suggestion
    await safeDeleteDomain(domain?.id, accessToken);
```

gravitee-am-test/specs/gateway/mfa-factors/recovery-code.jest.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+setup(300000);
+
+/**
+ * AM-2216 / UC-AM46 — recovery codes during sign-in.
+ *
+ * Being issued recovery codes and signing in with one is already covered, both here and in
+ * Playwright. What nothing covered is that a code is spent when it is used: a code that stayed
+ * valid afterwards would be a lasting way past multi-factor authentication, and no test would
+ * have noticed.
```
wajb: nitpick: IMO it's better not to describe wider coverage or distant code in comments because over time it becomes wrong, then noise. (The commit itself links to jira and to this PR for context, including point in time.)

gravitee-am-test/specs/gateway/mfa-factors/fixtures/recovery-code-fixture.ts
sagarchauhan-qa: Applied. I checked `safeDeleteDomain` and you're right — it takes `null | undefined` and never throws, so the guard and the inner try/catch were duplication.

Confirmed by forcing a failure mid-setup: the domain is still deleted and the original error still propagates.

gravitee-am-test/specs/gateway/mfa-factors/recovery-code.jest.spec.ts
sagarchauhan-qa: Fair point, and taken generally rather than just here — that paragraph described what other specs cover, which is exactly the sort of thing that goes stale. Removed.

I've kept the paragraph about why every refusal is paired with the same code having worked first: a reused code and a code that was never issued return the same error, so that shapes how the tests in this file are built. Happy to drop it too if you'd rather.

sagarchauhan-qa:
Comments addressed — ready for re-review please.

wajb (approved):

PR #8470, opened 2026-08-28
test: a remembered device stops being trusted (AM-2218)

sagarchauhan-qa:
[AM-2218](https://gravitee.atlassian.net/browse/AM-2218)

A remembered device skipping the challenge is well covered. Nothing covered the trust ever ending — a device remembered forever, or trusted on a machine it was never granted on, would have gone unnoticed.

- The device is challenged again once the remembered period passes
- Another device is still challenged
- Deleting the remembered device challenges the user again

One new file. Nothing existing is modified — the shared MFA fixtures are imported only, and `mfa-challange-required.spec.ts` was run alongside these to confirm it still passes.

## :memo: Test scenarios
See jira.

n/a

**The tests read the device off the user's profile, not just the redirect.** `GET .../users/{user}/devices` is what backs the Devices list in the Console, and no test had used it before. That turns "were they challenged?" into something checkable at the source:

- the entry's `expiresAt - createdAt` equals the configured period exactly, so the window is asserted rather than assumed
- once it lapses the entry is **removed**, not left behind unused
- being challenged on a second device does not quietly add it to the trusted list
- after deletion the challenge page offers `name="rememberDeviceConsent"` again, so revoking a device is not a one-way door

**Every test remembers a device and confirms the challenge is skipped first.** Without that anchor, a "challenged" result would pass equally well if remember-device were broken outright.

**One scenario the ticket lists is already covered.** It names "turning the setting off" as a gap, but `specs/gateway/mfa/mfa-challange-required.spec.ts` already has *"When Challenge REQUIRED and factor enrolled and no session issued using MFA and RememberDevice disabled → should Challenge"*. The ticket's analysis looked only at the Playwright tests. Left alone here rather than duplicated, and noted on the ticket.

**The expiry test waits ~13s.** `expirationTimeSeconds` is set to 10 for the test application; there is no way to fast-forward the window.

@gravitee-io/am

[AM-2218]: https://gravitee.atlassian.net/browse/AM-2218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/mfa/mfa-remember-device.spec.ts
+++ b/gravitee-am-test/specs/gateway/mfa/mfa-remember-device.spec.ts
@@ -1,0 +1,172 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { jira } from '@specs-utils/jira';
+import { performDelete, performGet, getWellKnownOpenIdConfiguration } from '@gateway-commands/oauth-oidc-commands';
+import { waitFor } from '@management-commands/domain-management-commands';
+import { getDomainManagerUrl } from '@management-commands/service/utils';
+import { Domain, initClient, initDomain, enableDomain, removeDomain, TestSuiteContext } from './fixture/mfa-setup-fixture';
+import { processLoginFromContext, processMfaEndToEnd } from './fixture/mfa-flow-fixture';
+import { setup } from '../../test-fixture';
+
+setup(300000);
+
+/**
+ * AM-2218 / UC-AM45 — a remembered device stops being trusted.
+ *
+ * Every test remembers a device and confirms the challenge is skipped before doing anything else,
+ * so a later challenge means the trust ended rather than remember-device never having worked.
+ */
+const EXPIRATION_SECONDS = 10;
+
+/** Long enough past `expiresAt` that a slow gateway cannot make the result ambiguous. */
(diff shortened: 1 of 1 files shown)


gravitee-am-test/specs/gateway/mfa/mfa-remember-device.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * AM-2218 / UC-AM45 — a remembered device stops being trusted.
+ *
+ * Every test remembers a device and confirms the challenge is skipped before doing anything else,
+ * so a later challenge means the trust ended rather than remember-device never having worked.
+ */
+const EXPIRATION_SECONDS = 10;
+
+/** Long enough past `expiresAt` that a slow gateway cannot make the result ambiguous. */
+const WAIT_PAST_EXPIRY_MS = (EXPIRATION_SECONDS + 3) * 1000;
```
wajb: thought: is it possible to reduce this wait and still have reliable positive and negative tests? In practice the first test is taking 13.5 seconds, which is slow, plus there's a hardcoded 3 second wait in the `beforeAll`.

gravitee-am-test/specs/gateway/mfa/mfa-remember-device.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  return { ctx, userId };
+};
+
+beforeAll(async () => {
+  await initDomain(domain, 3);
+  client = await initClient(domain, 'remember-device', rememberDeviceSettings(domain));
+  await enableDomain(domain);
+  await waitFor(3000);
+  const oidc = await getWellKnownOpenIdConfiguration(domain.domain.domainHrid).expect(200);
```
wajb: suggestion: use `waitForOidcReady`.

sagarchauhan-qa: Done — replaced the waitFor(3000) + getWellKnownOpenIdConfiguration pair with waitForOidcReady, so the hardcoded 3s is gone.

gravitee-am-test/specs/gateway/mfa/mfa-remember-device.spec.ts
sagarchauhan-qa: Yes — the expiry doesn't wait on anything in the background, every device read filters on expiresAt, so trust ends as soon as that time passes. Dropped the expiry to 5s and now wait on the device's own expiresAt instead of a fixed period.

First test 13.5s → 5.9s, whole file 22s → 9s. The 3s in beforeAll is gone too.

wajb (approved):

PR #8471, opened 2026-08-28
test: step-up that does not apply leaves a session alone (AM-2209)

sagarchauhan-qa:
[AM-2209](https://gravitee.atlassian.net/browse/AM-2209)

Step-up not applying is covered for a fresh sign-in. Nothing covered a user who already holds a session and returns — which is the situation this use case describes, and the one where step-up would misfire in practice.

- An already signed-in user is not challenged when the rule does not apply
- The same is true when step-up is switched off
- A session one application challenges is let through where the rule does not apply

One new file. Nothing existing is modified — the shared MFA fixtures are imported only, and `mfa-general.spec.ts` was run alongside these to confirm it still passes.

## :memo: Test scenarios
See jira.

n/a

**"Not challenged" is only worth asserting if something is challenged.** The first two tests would pass equally well with multi-factor authentication switched off altogether, so the third carries that weight: one session is returned to the application whose rule applies — **challenged** — and then to the one where it does not — **let through**. Same session, opposite outcomes, so the rule is demonstrably doing the work.

**Already covered, so not repeated here.** The ticket's third scenario, a fresh sign-in not being challenged, already exists in `specs/gateway/mfa/mfa-general.spec.ts` as *"when stepUp=false"* and *"when stepUp is disabled"*. Those two are left alone rather than duplicated, and noted on the ticket.

**Tagging.** The ticket suggests linking those existing tests to AM-2209. That means editing a spec belonging to another area, so it has been left out of this PR and raised on the ticket instead.

@gravitee-io/am

[AM-2209]: https://gravitee.atlassian.net/browse/AM-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
+++ b/gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
+import { jira } from '@specs-utils/jira';
+import { getWellKnownOpenIdConfiguration } from '@gateway-commands/oauth-oidc-commands';
+import { waitFor } from '@management-commands/domain-management-commands';
+import { Domain, initClient, initDomain, enableDomain, removeDomain, TestSuiteContext } from './fixture/mfa-setup-fixture';
+import { get, processLoginFromContext, processMfaEndToEnd } from './fixture/mfa-flow-fixture';
+import { setup } from '../../test-fixture';
+
+setup(300000);
+
+/**
+ * AM-2209 / UC-AM43 — step-up authentication that does not apply, for a user who already has a
+ * session.
+ *
+ * "Not challenged" is only worth asserting if something here is challenged, otherwise these tests
+ * would pass just as well with multi-factor authentication switched off altogether. The last test
+ * carries that weight: one session, two applications, opposite outcomes.
+ */
+const domain = {
+  admin: { username: 'admin', password: 'adminadmin' },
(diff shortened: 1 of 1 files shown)


gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+beforeAll(async () => {
+  await initDomain(domain, 3);
+  const ruleDoesNotApply = await initClient(domain, 'step-up-rule-false', stepUpSettings(domain, '{{ false }}', true));
+  const stepUpOff = await initClient(domain, 'step-up-off', stepUpSettings(domain, '', false));
+  const ruleApplies = await initClient(domain, 'step-up-rule-true', stepUpSettings(domain, '{{ true }}', true));
+  await enableDomain(domain);
+  await waitFor(3000);
+
+  const oidc = await getWellKnownOpenIdConfiguration(domain.domain.domainHrid).expect(200);
```
wajb: suggestion: avoid the fixed wait
```suggestion
  const oidc = await waitForOidcReady(domain.domain.domainHrid);
```

gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  it(jira`an already signed-in user is not challenged when the rule does not apply ${'AM-2209'}`, async () => {
+    const session = await signInAndKeepSession(ruleDoesNotApplyCtx);
+
+    const again = await returnToApplication(ruleDoesNotApplyCtx, session);
+
+    expect(again.letThrough).toBe(true);
+    expect(again.location).not.toContain('/mfa/challenge');
+  });
+
+  it(jira`the same is true when step-up is switched off ${'AM-2209'}`, async () => {
```
wajb: suggestion: the test name implies an execution sequence. Name this as an independent test case within the block.

gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
sagarchauhan-qa: Done, taken as suggested. Also dropped the waitFor(3000) above it, since that was the fixed wait.

gravitee-am-test/specs/gateway/mfa/mfa-step-up-not-applied.spec.ts
sagarchauhan-qa: Renamed to an already signed-in user is not challenged when step-up is switched off so it mirrors the first test instead of referring back to it. Ran it on its own to check there was no real dependency behind the name. Also reworded a file comment that referred to "the last test" by position.

wajb (approved):

PR #8472, opened 2026-08-28
test: factor selection rules decide what a user is offered (AM-2819)

sagarchauhan-qa:
[AM-2819](https://gravitee.atlassian.net/browse/AM-2819)

Across the whole suite, only three places set a factor selection rule and **every one of them evaluates false** — so nothing showed a rule putting a factor in front of a user. A gateway that ignored rules and always offered the default would have passed everything we had.

- A user matching the rule is offered that factor as well as the unruled one
- A user not matching is offered only the unruled factor
- Two ruled factors, two users — each is offered only the one whose rule names them
- A user matching no rule falls back to the default factor
- A lone default factor is offered even when its rule does not pass
- A rule that cannot be evaluated leaves its factor out, and the sign-in continues

One new file. Nothing existing is modified — the shared MFA fixtures are imported only, and `mfa-general.spec.ts` was run alongside to confirm it still passes.

## :memo: Test scenarios
See jira.

**Rules that exclude are paired with rules that include.** A rule which excluded everything looks identical to one that was never evaluated, so the second and third tests reuse the *same application and the same rule* and vary only the user. That difference is what shows the rule is being applied to the user rather than ignored.

**Each test reads the factor ids actually rendered on the enrolment page**, so what is asserted is what a user would be shown, rather than what the configuration says.

**One assertion is deliberately weaker than it looks.** With a single factor ruled false, two separate mechanisms would each produce the same result — the default fallback in `getApplicableFactors`, and the single-factor exemption in `evaluateFactorRuleByFactorId`. The test pins the outcome a user sees and says so in a comment, rather than claiming which one delivered it.

**Already covered, so not repeated here:** one of two factors ruled false and therefore not offered, which `playwright/fixtures/mfa-default-factor-selection.fixture.ts` sets up with a rule matching an impossible username.

**Sources.** The scenarios come from the Xray manual cases AM-2774, AM-2775 and AM-2804 as well as the ticket; those describe combinations the ticket does not. The documented behaviour — *"This factor will be available for all users that does not meet any configured Selection Rule"* — is covered by the fallback test. The lone-default-factor case is **not documented anywhere**, which is part of why it is worth having.

@gravitee-io/am

[AM-2819]: https://gravitee.atlassian.net/browse/AM-2819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/mfa/mfa-factor-selection-rule.spec.ts
+++ b/gravitee-am-test/specs/gateway/mfa/mfa-factor-selection-rule.spec.ts
@@ -16,8 +16,7 @@
 import { afterAll, beforeAll, describe, expect, it } from '@jest/globals';
 import { jira } from '@specs-utils/jira';
 import * as cheerio from 'cheerio';
-import { getWellKnownOpenIdConfiguration } from '@gateway-commands/oauth-oidc-commands';
-import { waitFor } from '@management-commands/domain-management-commands';
+import { waitForOidcReady } from '@management-commands/domain-management-commands';
 import { Domain, initClient, initDomain, enableDomain, removeDomain, TestSuiteContext } from './fixture/mfa-setup-fixture';
 import { followUpGet, processLoginFromContext } from './fixture/mfa-flow-fixture';
 import { setup } from '../../test-fixture';
@@ -40,7 +39,7 @@ const domain = {
 const matchesUsername = (username: string) => `{#context.attributes['user'].username == '${username}'}`;
 
 /** Unparseable on purpose — an unterminated expression. */
-const MALFORMED_RULE = "{#context.attributes[";
+const MALFORMED_RULE = '{#context.attributes[';
 
 let factorA: string;
 let factorB: string;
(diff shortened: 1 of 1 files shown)


gravitee-am-test/specs/gateway/mfa/mfa-factor-selection-rule.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    applicationWith([
+      { id: factorA, selectionRule: '' },
+      { id: factorB, selectionRule: MALFORMED_RULE },
+    ]),
+  );
+
+  await enableDomain(domain);
+  await waitFor(3000);
+
+  const oidc = await getWellKnownOpenIdConfiguration(domain.domain.domainHrid).expect(200);
```
wajb: suggestion: avoid the fixed wait
```suggestion
  const oidc = await waitForOidcReady(domain.domain.domainHrid);
```

sagarchauhan-qa: Done, taken as suggested. Also dropped the waitFor(3000) above it, since that was the fixed wait.

wajb (approved):

PR #8473, opened 2026-08-28
test: the skip duration on optional enrolment (AM-2821)

sagarchauhan-qa:
[AM-2821](https://gravitee.atlassian.net/browse/AM-2821)

Three existing tests configure a skip duration of nothing, one hour and twenty-four hours, and all three end the same way — skip, reach the application. Nothing observed the duration, so they would pass unchanged if it were ignored.

- The user is left alone inside the skip window, and asked again once it passes
- A longer window still holds at the moment a shorter one has already lapsed

One new file. Nothing existing is modified — the matrix fixture and MFA helpers are imported only, and `mfa-enrollment-matrix.spec.ts` was run alongside these (27 tests together) to confirm it still passes.

## :memo: Test scenarios
See jira.

n/a

**The skip is stamped on the user; the window belongs to the application.** `MfaFilterContext.isEnrollSkipped()` is `skippedAt + skipTimeSeconds > now`, where `mfaEnrollmentSkippedAt` is a property of the **user** and `skipTimeSeconds` of the **application**. The second test uses that: one skip, then two applications asked at the same instant whether it still stands — the short one says no, the long one yes. No second wait and no timing race, so the duration is the only variable.

**The wait polls rather than sleeps.** `waitUntilMfaEnrollmentSkipWindowExpired` watches the Management API, so the test waits exactly as long as the window needs.

**Partly overlapping, deliberately.** The second half of the first test — enrolment returning once the window passes — is also covered by the AM-2190 test in `mfa-enrollment-matrix.spec.ts`. It is repeated here because it is the anchor that gives the first half meaning: "not asked" proves nothing unless the same flow does ask later. This version also asserts the **skip option is offered again**, which the existing test does not check and which the manual case specifies.

**Already covered, so not repeated:** a user being able to skip optional enrolment at all, which the existing AM-2821 test in the same file covers.

**Sources.** The scenarios come from the ticket and the Xray manual cases AM-2743 (within the window, application shows) and AM-2742 (window passed, enrolment shows *with the skip option*). Documented behaviour: *"You can specify the period of time during which enrollment can be skipped. Once the timer has ended, users will be asked to enroll."*

@gravitee-io/am

[AM-2821]: https://gravitee.atlassian.net/browse/AM-2821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-test/playwright/tests/mfa/mfa-enrollment-skip-duration.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+                active: true,
+                type: 'OPTIONAL',
+                forceEnrollment: false,
+                enrollmentSkipActive: true,
+                enrollmentSkipRule: '{{ true }}',
+                skipTimeSeconds: LONG_WINDOW_SECONDS,
+              },
+              challenge: { active: false },
+            },
+            // The consent page is not part of this behaviour; skipping it lets the sign-in run
```
wajb: nitpick: use US spelling "behavior".

wajb (approved):

gravitee-am-test/playwright/tests/mfa/mfa-enrollment-skip-duration.spec.ts
sagarchauhan-qa: Fixed.

PR #8474, opened 2026-08-28
test: the skip rule on conditional enrolment (AM-2825, AM-2826, AM-2827)

sagarchauhan-qa:
[AM-2825](https://gravitee.atlassian.net/browse/AM-2825), [AM-2826](https://gravitee.atlassian.net/browse/AM-2826), [AM-2827](https://gravitee.atlassian.net/browse/AM-2827)

Three tickets in one PR because they are three settings of the same mechanism and they discriminate each other — "skip rule holds, so the option appears" proves nothing without "skip rule does not hold, so it does not".

**AM-2825 — the skip rule holds**
- The user is offered the chance to skip, and skipping completes the sign-in
- Declining the offered skip takes them through enrolment instead
- A user who has already enrolled is let straight through
- They are left alone inside the skip window and asked again once it passes

**AM-2826 — the skip rule does not hold**
- No skip is offered and the user must enrol

**AM-2827 — altering the duration**
- A longer duration still holds when a shorter one has lapsed

One new file. Nothing existing is modified — the matrix fixture and MFA helpers are imported only. Run alongside `mfa-enrollment-matrix.spec.ts` and `mfa-enrollment-skip-duration.spec.ts` (31 passing together) and Lukasz's Jest `mfa-enrollment-conditional.spec.ts` (3 passing).

## :memo: Test scenarios
See jira.

n/a

**`forceEnrollment` is deliberately on, and that is what makes these tests mean anything.** The enrolment page renders its skip control from `!MfaUtils.isCanSkip(...)`, and

```java
isCanSkip = enrollmentActive && (isNotForcedEnrollment || isConditionalAndSkipped)
```

With `forceEnrollment` **off**, `isNotForcedEnrollment` short-circuits and the skip control appears **whatever the skip rule says**. My first run had it off and the AM-2826 assertion failed for exactly that reason. Only with force on does `isConditionalAndSkipped` — the session flag `MFAEnrollStep.conditional()` sets when the skip rule holds — decide the control. So these tests would have passed while proving nothing about the rule had that not been caught.

**The skip is stamped on the user; the window belongs to the application.** AM-2827 uses that: one skip, then two applications asked at the same instant whether it still stands. No second wait and no timing race.

**Enrolling a factor is always verified**, whatever the challenge setting says, so the tests that enrol answer a challenge before reaching the callback.

**Partly overlapping, deliberately.** The AM-2841/2842/2843 tests in `mfa-enrollment-matrix.spec.ts` already configure conditional enrolment with a skip rule that holds and then enrol — so "declines the skip, enrolment proceeds" is exercised there. What was missing is the assertion that the skip was offered at all, which is what makes it a decision rather than an outcome. That test is here to complete the case, not to cover new ground.

**Already covered, so not repeated:** conditional rule true bypassing MFA, and conditional rule false leading to enrolment, both in Lukasz's `specs/gateway/mfa/mfa-enrollment-conditional.spec.ts`.

**Sources.** Scenarios come from the three tickets and the Xray manual cases AM-2690 ("no skip option will display"), AM-2691 ("the skip option will display"), AM-2739 (offered and declined), AM-2742 and AM-2743 (outside and inside the window). Documented behaviour: *"To allow users to skip conditional enrollment, you enable the 'Allow users to skip Conditional Enrollment' toggle, which allows you to add an additional rule that allows some users to skip MFA enrollment for some time even if they matched the first conditional rule."*

@gravitee-io/am

[AM-2825]: https://gravitee.atlassian.net/browse/AM-2825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-2826]: https://gravitee.atlassian.net/browse/AM-2826?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-2827]: https://gravitee.atlassian.net/browse/AM-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-skip-rule.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+ * `forceEnrollment` is deliberately **on** throughout. With it off, `isCanSkip` short-circuits on
+ * `isNotForcedEnrollment` and the skip control appears whatever the rule says — so the rule would
+ * not be the thing under test. The presence and absence of that control are asserted against each
+ * other, since a page without a skip button looks the same whether the rule was evaluated or never
+ * consulted.
+ */
+const SHORT_WINDOW_SECONDS = 3;
+const LONG_WINDOW_SECONDS = 3600;
+
+const skipButton = (page) => page.locator('button[name="user_mfa_enrollment"][value="false"]');
```
wajb: suggestion: we could add this to `mfa-helpers.ts` with a way to assert its visibility so the selector is maintainable in one place (in fact it's declared there already but within the `skipMfaEnrollment` function itself, so the first visibility check below is redundant).

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-skip-rule.spec.ts
```diff
@@ -0,0 +1,294 @@
...
+    expect(after.askedToEnrol).toBe(true);
+    await expect(skipButton(page)).toBeVisible();
+  });
+});
+
+/* ------------------------------------------------------------------ */
+/*  AM-2826 — the skip rule does not hold                              */
+/* ------------------------------------------------------------------ */
+
+test.describe('Conditional enrolment where the skip rule does not hold (AM-2826)', () => {
```
wajb: nitpick: "enrollment" is US spelling used elsewhere, including jest tests (check file name too).

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-skip-rule.spec.ts
sagarchauhan-qa: Moved the selector into mfa-helpers.ts with an expectSkipEnrollmentOffered helper, and skipMfaEnrollment now uses it too, so it's declared in one place. Dropped the redundant check you spotted. Kept the other three, since those tests don't skip afterwards.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-skip-rule.spec.ts
sagarchauhan-qa: Fixed throughout, and renamed the file.

wajb (approved):

PR #8476, opened 2026-08-28
feat: AM-7553 evaluate reporter custom attribute mappings

wajb:
[AM-7553](https://gravitee.atlassian.net/browse/AM-7553)

Evaluate and produce `customAttributes` on audit events as they are built. The reporter types that support custom attributes mapping (Kafka, TCP and File) receive events containing the new property when mappings are configured and any mapping produces a value; data for existing reporters without mappings is unaffected.

Additionally, relax restriction on custom attribute mappings' `expression` fields - wrapping in `{}` is no longer required and any literal value as well as interpolated expressions is now supported and consistent with other usages of EL in AM.

Supported contexts are `user`, `client`, `request` and `audit` - example: `{#context.attributes['request'].ip}`. Availability of contextual data principally depends on the audit type. When an entire expression produces no value, the attribute is omitted from the audit data.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7553]: https://gravitee.atlassian.net/browse/AM-7553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-reporter/gravitee-am-reporter-api/src/main/java/io/gravitee/am/reporter/api/audit/model/AuditEnrichmentContext.java
+++ b/gravitee-am-reporter/gravitee-am-reporter-api/src/main/java/io/gravitee/am/reporter/api/audit/model/AuditEnrichmentContext.java
@@ -1,0 +1,48 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.reporter.api.audit.model;
+
+import io.gravitee.am.model.User;
+import io.gravitee.am.model.oidc.Client;
+import io.gravitee.am.model.safe.ClientProperties;
+import io.gravitee.am.model.safe.UserProperties;
+
+/**
+ * Carries the user and client behind an {@link Audit}, for reporters that resolve attribute mappings.
+ *
+ * <p>Exposes them only as the sanitized {@link UserProperties} and {@link ClientProperties}
+ * projections; the raw models must not be exposed.
+ *
+ * @author GraviteeSource Team
+ */
+public class AuditEnrichmentContext {
+
+    private final User user;
+    private final Client client;
+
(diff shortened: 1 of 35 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/attribute/ReporterAttributeResolver.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * @return the resolved attributes keyed by the operator's chosen export name, in declaration order
+     */
+    public Map<String, Object> resolve(List<ReporterAttributeMapping> mappings, Audit audit) {
+        if (mappings == null || mappings.isEmpty() || audit == null) {
+            return Map.of();
+        }
+
+        TemplateEngine engine = TemplateEngine.templateEngine();
+        engine.getTemplateContext().setVariable(CONTEXT_VARIABLE, new EvaluableAuditContext(attributesOf(audit)));
```
sclarkamlab, line 58: This sits outside the try below, so anything that fails while building the attribute map drops the event for this reporter. `new UserProperties(user, false)` throws a ClassCastException on a user with a non string `locale` claim and no preferredLanguage, and the mapping doesn't even need to reference the user for that to happen. Ran it through the event bus, the plain reporter got the event and the mapped one got nothing.

Something like this in `AttributeMappingReporter.report` would cover it:
```java
Map<String, Object> resolved;
try {
    resolved = resolver.resolve(mappings, audit);
} catch (Exception ex) {
    log.warn("Attribute mappings failed for audit {}, delivering it unenriched", audit.getId(), ex);
    delegate.report(reportable);
    return;
}
```

gravitee-am-reporter/gravitee-am-reporter-api/src/main/java/io/gravitee/am/reporter/api/audit/model/AuditEnrichmentContext.java
```diff
@@ -0,0 +1,62 @@
...
+ * passwords, client secrets and provider tokens are unreachable by construction.
+ *
+ * @author GraviteeSource Team
+ */
+public class AuditEnrichmentContext {
+
+    private final User user;
+    private final Client client;
+
+    // Benign race: concurrent builds yield equivalent projections.
```
sclarkamlab: Not benign I'm afraid. `UserProperties` has 24 non final fields and this isn't volatile, so a second thread can see the reference before the fields and the attribute just goes missing. It is reachable, the same `Audit` instance lands on every reporter for the domain and I got 4 wrappers handling one publish on 4 different event loop threads. Can't make the reorder happen on demand but the JMM allows it. `volatile` on both fields fixes it.

wajb: Quite right and not necessary - removed since `resolve()` ends up using the context objects once now.

sclarkamlab (approved):

PR #8477, opened 2026-08-28
test: a required challenge after conditional enrolment with a skip rule (AM-2842)

sagarchauhan-qa:
[AM-2842](https://gravitee.atlassian.net/browse/AM-2842)

The existing test sets conditional enrolment, a skip rule that holds and a required challenge — then enrols and answers the verification that always follows enrolment. Neither named setting influences the result: the skip is never used, and the required challenge is never reached on its own.

- A user who enrolled is asked for a code on a later sign-in
- A user who skipped has no method, so the required challenge never applies

One new file. Nothing existing is modified — the matrix fixture and MFA helpers are imported only, and `mfa-enrollment-matrix.spec.ts` was run alongside these (27 passing together) to confirm it still passes.

## :memo: Test scenarios
See jira.

n/a

**The two tests are each other's control.** Same application, same required challenge, both users signing in twice — the only difference is whether the first visit ended in enrolling or skipping. One is then challenged and the other is not. Either test alone would show very little: "challenged" could be the post-enrolment verification rather than the configured challenge, and "not challenged" could mean the challenge was never configured at all.

**Two different routes through `MFAEnrollStep`.** A user holding a method reaches the challenge through `conditional() -> userHasFactor -> continueFlow`. A user who skipped goes through `MFAEnrollPostEndpoint`, which marks the challenge complete for that authorisation leg at the same moment it records the skip:

```java
// Skipping enrollment also skips MFA challenge for this authorisation leg
routingContext.session().put(ConstantKeys.MFA_ENROLLMENT_COMPLETED_KEY, true);
routingContext.session().put(ConstantKeys.MFA_CHALLENGE_COMPLETED_KEY, true);
```

That answers the third scenario, which the ticket records as unestablished: a user who skips is asked for neither, and it is deliberate rather than a side effect.

**`forceEnrollment` is on.** With it off, `MfaUtils.isCanSkip` short-circuits on `isNotForcedEnrollment` and the skip control appears whatever the skip rule says — so the rule would not be the thing under test. Same reasoning as PR #8474.

**Already covered elsewhere:** the skip control being offered under conditional enrolment is also asserted in #8474, with the challenge switched off. It is asserted here as well because this ticket's combination includes the required challenge, and that is the configuration its first scenario describes.

**Sources.** Scenarios come from the ticket and Xray manual case AM-2731 — *conditional = false, challenge Required, user enrolled → "the MFA challenge page will display"* — which nothing automated covered. The nearest existing test, AM-2829, reaches a required challenge with enrolment switched **off**, a different branch.

@gravitee-io/am

[AM-2842]: https://gravitee.atlassian.net/browse/AM-2842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
+++ b/gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { expect } from '@playwright/test';
+import { test } from '../../fixtures/mfa-enrollment-matrix.fixture';
+import { linkJira } from '../../utils/jira';
+import { clearSessionOnly } from '../../utils/webauthn-helpers';
+import {
+  buildAuthorizeUrl,
+  completeMfaChallenge,
+  enrollMockFactor,
+  reachOAuthAuthorizationCallback,
+  skipMfaEnrollment,
+  submitLogin,
+} from '../../utils/mfa-helpers';
+import { API_USER_PASSWORD, AUTH_CODE_FORMAT, MULTI_PHASE_TEST_TIMEOUT } from '../../utils/test-constants';
+
+/**
+ * AM-2842 — conditional enrolment with a skip rule, alongside a required challenge.
+ *
+ * The two tests are each other's control. Both applications are configured identically and both
+ * users sign in twice; the only difference is whether the first visit ended in enrolling or in
+ * skipping. One is then challenged and the other is not, so the challenge setting is shown to be
+ * live and the skip is shown to be what suppresses it.
+ *
+ * `MFAEnrollStep.conditional()` reaches the challenge through `userHasFactor -> continueFlow` when
+ * the user already holds a method. Skipping instead goes through `MFAEnrollPostEndpoint`, which
+ * marks the challenge complete for that authorisation leg at the same time as recording the skip.
+ */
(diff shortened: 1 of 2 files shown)


gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    enrollSkipActive: true,
+    enrollSkipRule: '{{ true }}',
+    enrollSkipTimeSeconds: 3600,
+    challengeActive: true,
+    challengeType: 'REQUIRED',
+    challengeRule: '',
+  });
+  test.setTimeout(MULTI_PHASE_TEST_TIMEOUT);
+
+  const signIn = async (page, gatewayUrl: string, clientId: string, username: string) => {
```
wajb: suggestion: this looks common to the setup in https://github.com/gravitee-io/gravitee-access-management/pull/8474. Worth a `waitFor` helper?

gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await page.goto(buildAuthorizeUrl(gatewayUrl, clientId));
+    await page.waitForURL(/.*login.*/i);
+    await submitLogin(page, username, API_USER_PASSWORD);
+    await page.waitForURL(
+      (url) => /\/mfa\/(enroll|challenge)/i.test(url.href) || url.searchParams.has('code'),
+    );
+    return {
+      askedToEnrol: /\/mfa\/enroll/i.test(page.url()),
```
wajb: nitpick: spelling enrol/enroll is inconsistent (use US enroll).

gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
sagarchauhan-qa: Agreed — moved it to mfa-helpers.ts as signInAndReportMfaStop. It was in three files; #8474 and #8493 use it too, so I'll switch those over as they come up.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-skip-required-challenge.spec.ts
sagarchauhan-qa: Fixed throughout.

wajb (approved):

PR #8492, opened 2026-09-01
feat: configure Cross App Access on a trusted domain (AM-7520)

lgw-gravitee:
A trusted domain now carries what AM issues towards it: the resource servers it exposes, the aud_sub expression and the outbound scope mappings. The block is written through the management API and from the Console, next to the inbound Token exchange settings.

fixes AM-7520
image

image


gravitee-am-model/src/main/java/io/gravitee/am/model/oidc/CrossAppAccessSettings.java
```diff
@@ -0,0 +1,72 @@
...
+    private String audSubMapping;
+
+    @Schema(description = "One-to-one mapping from domain scope to the name this authority knows it by. "
+            + "Unmapped domain scopes are dropped (fail-closed).")
+    private Map<String, String> scopeMappings;
+
+    public CrossAppAccessSettings(CrossAppAccessSettings other) {
+        this.enabled = other.enabled;
+        this.resourceServers = other.resourceServers != null
+                ? other.resourceServers.stream().map(CrossAppAccessResourceServer::new).collect(toCollection(ArrayList::new))
```
sclarkamlab, line 76: `resourceServers: [null]` NPEs here. `CrossAppAccessResourceServer::new` reads `other.id` on the null element, so you get a 500 instead of the 400.

`copyWithGeneratedIds` runs before `validate()` at `TrustDomainServiceImpl:134` and `:184`, which is why the null check at `:511` never fires. Moving `validate()` first sorts both.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -259,23 +275,71 @@ public Completable delete(Domain domain, String id, User principal) {
 
     /**
      * Sets the matchers a trusted domain is recognised by. A payload that declares neither is written
-     * against the SPIFFE-only API that preceded the issuer matcher, and means the name.
+     * against the SPIFFE-only API that preceded the issuer matcher, and means the name. Cross App
+     * Access is a usage in its own right, so a payload that declares it is not falling back.
      */
     private static void applyMatchers(TrustDomain td, String spiffeTrustDomain, String issuer) {
-        String spiffe = spiffeTrustDomain == null && issuer == null
+        String spiffe = spiffeTrustDomain == null && issuer == null && !td.trustsCrossAppAccess()
```
sclarkamlab: `trustsCrossAppAccess()` is `crossAppAccess != null && enabled`, so a create with `"crossAppAccess": {"enabled": false}` and no matcher still falls back and AM stamps `spiffeTrustDomain = name`. The javadoc above says the opposite.

Did you mean `td.getCrossAppAccess() == null`?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -416,6 +486,97 @@ private Optional<String> validateUserBinding(TrustDomain td) {
                 : Optional.empty();
     }
 
+    private Optional<String> validateCrossAppAccess(CrossAppAccessSettings settings) {
```
sclarkamlab, line 502: `enabled: true` with no resource servers saves fine. No matcher, no key material, nothing to mint towards. The form lets it through too, `crossAppAccessErrors()` only walks the rows that exist.

Worth making enabled imply at least one?

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -416,6 +486,97 @@ private Optional<String> validateUserBinding(TrustDomain td) {
...
+    private Optional<String> validateAudSubMapping(String audSubMapping) {
+        if (audSubMapping == null || audSubMapping.isBlank()) {
+            return Optional.empty();
+        }
+        if (audSubMapping.length() > CrossAppAccessSettings.AUD_SUB_MAPPING_MAX_LENGTH) {
+            return Optional.of("crossAppAccess.audSubMapping must be at most "
+                    + CrossAppAccessSettings.AUD_SUB_MAPPING_MAX_LENGTH + " characters");
+        }
+        try {
+            EXPRESSION_PARSER.parseExpression(audSubMapping);
```
sclarkamlab: Not blocking. We parse `audSubMapping` here but `userBindingCriteria[].expression` only gets a blank check in `validateUserBinding`. Two EL fields, two rules.

gravitee-am-test/specs/management/domain/cross-app-access-validation.jest.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+  });
+
+  it('should reject a trusted domain that declares no usage at all', async () => {
+    await expect(
+      fixture.createTrustDomain({
+        name: uniqueName('xaa-nothing'),
+        issuer: '  ',
+        crossAppAccess: { ...fixture.crossAppAccess('https://calendar.example.com/xaa-nothing'), enabled: false },
+      }),
+    ).rejects.toThrow();
```
sclarkamlab: These eight only assert something threw, so the 500 above would pass all of them. We do this in `protected-resources.jest.spec.ts:258`:

```ts
).rejects.toMatchObject({ response: { status: 400 } });
```

sclarkamlab (dismissed):
Shape is good. The model, both repos, the migration, the SDK and the UI all line up and the tests are broad.

Main one is `resourceServers: [null]`, that NPEs before validation and comes back as a 500. Rest is two validation gaps and the jest assertions.

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
```diff
@@ -155,11 +165,28 @@ <h5>Signing algorithms</h5>
...
-  <div *ngIf="issuerEnabled" class="gv-form-section" fxLayout="column">
+  <div *ngIf="tokenExchangeEnabled" class="gv-form-section" fxLayout="column">
     <div class="gv-form-section-title">
-      <h5>Scope mappings</h5>
+      <h5>Token exchange</h5>
       <mat-divider></mat-divider>
     </div>
+    <mat-hint class="section-hint">What AM accepts from this issuer during an RFC 8693 token exchange.</mat-hint>
+
+    <mat-form-field appearance="outline" class="trust-domain-field">
```
wajb, line 175: question: isn't this assertion also required for XAA? 🤔

gravitee-am-model/src/main/java/io/gravitee/am/model/oidc/CrossAppAccessSettings.java
lgw-gravitee: Took the ordering fix rather than making the copy null-tolerant — tolerating it just carries a null further into the object.

`create` and `update` now hold the written block as-is until `validate()` passes, and `copyWithGeneratedIds` runs in the `Single.defer` right before the repository write. Nothing between the two needs ids: `rejectDuplicates` works off name, spiffe, issuer and audience.

Confirmed both directions — reverting the ordering makes the new `shouldRejectANullResourceServerEntry` fail with `NullPointerException: Cannot read field "id" because "other" is null`; with it, 400 and the message from `validateResourceServers`.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
lgw-gravitee: Yes, `getCrossAppAccess() == null` is what the javadoc promises. Fixed.

`{"crossAppAccess":{"enabled":false}}` with no matcher is now a 400 from `validateMatchers` instead of a silent `spiffeTrustDomain = name`. Covered by `shouldNotFallBackToTheNameWhenTheCrossAppAccessBlockIsDisabled` and a jest case. Update is unaffected — it only calls `applyMatchers` when a matcher is actually written.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
lgw-gravitee: Agreed. `enabled` now implies at least one resource server, server-side and in `crossAppAccessErrors()`.

The "no key material, nothing to mint towards" half got a second answer from wajb's thread: the block also requires an `audience` now, so an enabled block can no longer be contentless.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
lgw-gravitee: Done. Parsing moved into a shared `parses(String)` that both `validateUserBinding` and `validateAudSubMapping` call, so it's one rule now.

Two things worth flagging, neither blocking. Both parse with `SpelExpressionParser` while the gateway evaluates with `TemplateEngine` (`TrustedIssuerUserResolver.buildCriteria`) — so in practice it's a balanced-brace check, not real validation. Making both use `TemplateEngine` is the correct fix but changes shipped `audSubMapping` behaviour, so it's a separate ticket.

And: the deprecated inline `trustedIssuers` list projects through `trustDomainService.create/update`, so a domain already storing a malformed expression will 400 on its next `patchDomain`, including one that doesn't touch user binding. That config is already broken at exchange time, but it surfaces at an unexpected moment.

gravitee-am-test/specs/management/domain/cross-app-access-validation.jest.spec.ts
lgw-gravitee: All of them are `.rejects.toMatchObject({ response: { status: 400 } })` now, and there are more of them since the suite grew. Checked every exception reachable on those paths — `InvalidTrustDomainException` and the four `TrustDomain*AlreadyExistsException` — returns `BAD_REQUEST_400`, so none of them pass on a 5xx any more.

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
lgw-gravitee: You were right, and it went further than the form.

draft-ietf-oauth-identity-assertion-authz-grant-04: the ID-JAG's `aud` "MUST contain the issuer identifier of the Resource Authorization Server", and `resource` (RFC 8707) names the resource server behind it. So `audience` was at the wrong level — it identified the AS but lived on the row. Two resource servers behind one authorization server were unmodellable: both rows would need the same `audience`, and the uniqueness rule rejected the second.

`audience` moved up to the block, rows carry `resource`. Uniqueness moved with it — `audience` unique per security domain, `resource` unique within a trusted domain. The Console gained an "Authorization server" field above the table.

Reusing `TrustDomain.issuer` for it wasn't viable: a non-null issuer also indexes the trusted domain into the gateway's `byIssuer` map and surfaces it in `TrustedIssuerProjection`, so setting it for an outbound purpose would have made AM start trusting inbound tokens from that issuer — and pulled the key-material requirement back in.

One deliberate deviation: `resource` is OPTIONAL in the draft, required here, otherwise two rows under one `aud` are indistinguishable.

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
```diff
@@ -60,21 +60,31 @@ <h5>Usage</h5>
     >
       OIDC - Trusted Issuer
```
wajb, line 61: suggestion: can we drop this outer checkbox? The sub-checkboxes are both compatible with SPIFFE so it could simply be
- [ ] Token Exchange
- [ ] XAA
- [ ] SPIFFE

This saves one click for most combinations.

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
wajb, line 126: issue: this doesn't show on first render. When you change key source to something else and back to JWKS_URL then it appears.

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.ts
wajb, line 323: issue: missing third option (XAA).

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
wajb: 🤔 An alternative to consider might be keeping the two top categories as separate "enable" sections that own their specialised properties. Only general properties live outside these (like Signing algorithms). Token exchange and XAA could be sub sections of OIDC.

This is closer to what you have but ties the checkbox actions directly to the availability of options in a clearly additive way.

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -411,11 +488,143 @@ private Optional<String> validateUserBinding(TrustDomain td) {
...
+     * rather than at mint time. A value carrying no expression is a valid literal.
+     */
+    private static boolean parses(String expression) {
+        try {
+            EXPRESSION_PARSER.parseExpression(expression);
+            return true;
+        } catch (ParseException e) {
+            return false;
+        }
+    }
```
leleueri, line 626: suggestion: what if we extract this in a Utility Class so we can apply validation on EL everywhere EL can be assigned. (ex on custom token claim)

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
lgw-gravitee: I flatten the checkboxes. tell me what you think. I would rather opt for current selection mode as the API model have distinct configuraiton blocks

gravitee-am-ui/src/app/domain/settings/trust-domains/form/trust-domain-form.component.html
lgw-gravitee: thanks for that

gravitee-am-ui/src/app/domain/settings/trust-domains/trust-domains.component.ts
```diff
@@ -53,6 +53,11 @@ export class DomainSettingsTrustDomainsComponent implements OnInit {
     return trustDomainUsagesLabel(trustDomain);
   }
 
+  usagesBadgeLabel(trustDomain: TrustDomain): string {
+    const usages = trustDomainUsages(trustDomain);
+    return usages.length > 1 ? `${usages.length} usages` : trustDomainUsagesLabel(trustDomain);
```
wajb, line 58: minor: won't the fallback always be empty? If so we can remove it.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
```diff
@@ -106,11 +106,21 @@ private Completable replace(Domain domain, List<TrustedDomain> existing, List<Tr
...
+    }
+
+    /**
+     * Drops an issuer the written list no longer declares, keeping the row when the trusted domain
+     * also issues towards the authority: the deprecated settings must not delete what they cannot see.
+     */
+    private Completable withdraw(Domain domain, TrustedDomain trustDomain, User principal) {
+        return trustDomain.trustsCrossAppAccess()
+                ? trustDomainService.update(domain, trustDomain.getId(), asTokenExchangeCleared(trustDomain), principal).ignoreElement()
+                : trustDomainService.delete(domain, trustDomain.getId(), principal);
```
sclarkamlab, line 130: issue: this deletes SPIFFE trust through the deprecated field. The javadoc right above says the deprecated settings must not delete what they cannot see, but the guard only covers XAA. A trusted domain with SPIFFE plus token exchange gets deleted outright when someone PATCHes the domain with a `trustedIssuers` list that omits it, and the SPIFFE trust domain and key material go with it.

```java
return trustDomain.trustsCrossAppAccess() || trustDomain.trustsSpiffe()
        ? trustDomainService.update(domain, trustDomain.getId(), asTokenExchangeCleared(trustDomain), principal).ignoreElement()
        : trustDomainService.delete(domain, trustDomain.getId(), principal);
```

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
```diff
@@ -463,11 +538,111 @@ private Optional<String> validateUserBinding(TrustedDomain td) {
         boolean incomplete = criteria.stream().anyMatch(c -> c == null
                 || c.getAttribute() == null || c.getAttribute().isBlank()
                 || c.getExpression() == null || c.getExpression().isBlank());
+        if (incomplete) {
+            return Optional.of("userBindingCriteria entries must have a non-blank attribute and expression");
+        }
+        return criteria.stream()
+                .map(c -> parses(c.getExpression())
```
wajb, line 545: question: we never want to allow literal values, only expressions? I can't find any other examples where we validate expressions at configuration time.

wajb (approved):

gravitee-am-repository/gravitee-am-repository-mongodb/src/main/java/io/gravitee/am/repository/mongodb/management/LegacyTrustDomainMigration.java
```diff
@@ -0,0 +1,115 @@
...
+                        .append("on", FIELD_ID)
+                        .append("whenMatched", "keepExisting")
+                        .append("whenNotMatched", "insert")));
+
+        return Completable.fromPublisher(database.getCollection(LEGACY_COLLECTION_NAME).aggregate(pipeline))
+                .doOnError(error -> log.warn("Unable to migrate the {} collection into {}",
+                        LEGACY_COLLECTION_NAME, targetCollectionName, error))
+                .onErrorComplete();
+    }
+
```
lgw-gravitee, line 87:
```
  { "$cond": {
      "if":   { "$eq": [ { "$type": "$keyMaterial" }, "object" ] },
      "then": "$keyMaterial",
      "else": { "$switch": {
          "branches": [
            { "case": { "$eq": ["$bundleSource", "STATIC_JWKS"] },
              "then": { "source": "JWK_SET",
                        "refreshIntervalSeconds": "$refreshIntervalSeconds" } },
            { "case": { "$eq": [ { "$type": "$bundleSource" }, "string" ] },
              "then": { "source": "JWKS_URL",
                        "jwksUrl": "$jwksUrl",
                        "refreshIntervalSeconds": "$refreshIntervalSeconds" } }
          ],
          "default": "$$REMOVE" } } } }
```

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
lgw-gravitee: now only tokenExchange is cleared
the row survives if it has SPIFFE or XAA

gravitee-am-ui/src/app/domain/settings/trust-domains/trust-domains.component.ts
lgw-gravitee: when length == 1 we display a badge with name

gravitee-am-service/src/main/java/io/gravitee/am/service/impl/TrustDomainServiceImpl.java
lgw-gravitee: parses() uses TemplateParserContext("{#", "}") -> so a value with no {#…} block is a literal and passes

lgw-gravitee: I didnt want to update that for "Custom Token Claims" for example but  I guess its worth to add in seperate ticket

sclarkamlab (approved):

PR #8493, opened 2026-09-01
test: a conditional challenge after conditional enrolment with a skip rule (AM-2843)

sagarchauhan-qa:
[AM-2843](https://gravitee.atlassian.net/browse/AM-2843)

The existing test for this combination stops after the first visit: the user enrols and answers the verification that always follows enrolment, so the conditional challenge is never evaluated for a user who already holds a method. Across the whole suite, every conditional-challenge test sets `challengeRule: '{{ true }}'`. The rule-does-not-hold side appears exactly once — AM-2831, with enrolment switched **off**.

- An enrolled user is not challenged on a later sign-in while the rule holds
- The same user is challenged on a later sign-in once the rule no longer holds
- A user offered the skip is not challenged, though the rule would have challenged them

One new file. Nothing existing is modified — the matrix fixture and MFA helpers are imported only, and the whole `playwright/tests/mfa/` folder was run together (38 passing) to confirm the neighbouring specs still pass.

## :memo: Test scenarios
See jira.

n/a

**The tests are each other's controls.** The enrolment settings are identical throughout and only the challenge rule differs. The first two users both enrol and sign in again, and one is challenged where the other is not — that is what shows the rule is live. The third meets the same rule as the second but takes the skip instead of enrolling, so the enrolment, not the challenge setting, is what separates them.

**Clearing the session is load-bearing.** `MFAChallengeStep.conditional()` skips the challenge on three separate grounds:

```java
if (challengeConditionSatisfied(client, context, ruleEngine)) {
    continueFlow(context, flow);                       // the rule holds
} else if (context.getRememberDeviceSettings().isSkipChallengeWhenRememberDevice()) {
    required(context, flow);
} else if (context.isUserStronglyAuth()) {
    continueFlow(context, flow);                       // already strongly authenticated
} else {
    challenge(context, flow);
}
```

Carrying the first visit's session into the second reaches the third branch, so the challenge is skipped whatever the rule says. Verified by running it: with the session kept and the rule at `{{ false }}`, the second `/authorize` landed on `…/callback?code=…` with no challenge page at all. Both enrolling tests therefore call `clearSessionOnly` between visits; without it neither would consult the rule.

**Observation on existing tests, not changed here.** AM-2835, AM-2837, AM-2828 and AM-2830 reuse the session across their second `/authorize` via `secondAuthorizeExpectCallbackWithoutMfa`, so they exit through that same `isUserStronglyAuth()` branch and would pass with their rules inverted. AM-2831 is the only one that clears cookies first. Flagging it for whoever owns those rather than touching them in this PR.

**`forceEnrollment` is on**, so the skip control is decided by the skip rule rather than short-circuited by `MfaUtils.isCanSkip`. Same reasoning as #8474 and #8477.

**Already covered elsewhere:** the skip control being offered under conditional enrolment is asserted in #8474 (challenge off) and #8477 (required challenge). The third test here asserts it again because what follows the skip differs — under a *conditional* challenge whose rule would challenge the user, `MFAEnrollPostEndpoint` marks the challenge complete alongside the skip, and that combination is not covered anywhere.

**Sources.** Scenarios come from the ticket and Xray manual cases AM-2729 (rule true, enrolled user → application displays) and AM-2727 (rule false, enrolled user → challenge displays), neither of which anything automated covered. The [MFA policies documentation](https://documentation.gravitee.io/am/guides/multi-factor-authentication/mfa-policies) confirms both sides of the rule are intended behaviour.

@gravitee-io/am

[AM-2843]: https://gravitee.atlassian.net/browse/AM-2843?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrollment-conditional-challenge.spec.ts
wajb, line 1: minor: rename enrol -> enroll.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-conditional-challenge.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  }, testInfo) => {
+    linkJira(testInfo, 'AM-2843');
+
+    const clientId = matrixApp.settings.oauth.clientId;
+
+    const first = await signIn(page, gatewayUrl, clientId, matrixUser.username);
+    expect(first.askedToEnrol).toBe(true);
+
+    // The enrolment rule did not hold, but the skip rule did, so a way past enrolment is offered.
+    await expect(page.locator('button[name="user_mfa_enrollment"][value="false"]')).toBeVisible();
```
wajb: suggestion: reuse `expectSkipEnrollmentOffered` helper once merged.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-conditional-challenge.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * `MFAChallengeStep.conditional()` skips the challenge on three separate grounds: the rule holding,
+ * a remembered device, or the user already being strongly authenticated. The last of those is the
+ * reason the two enrolling tests clear the session cookie between visits — carrying the first
+ * visit's session into the second would skip the challenge whatever the rule said, and both would
+ * pass without the rule ever being consulted.
+ */
+
+/** One sign-in, reporting where the gateway stopped rather than assuming which page comes next. */
+const signIn = async (page: Page, gatewayUrl: string, clientId: string, username: string) => {
```
wajb: suggestion: reuse `signInAndReportMfaStop` helper once merged.

wajb (approved):

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrollment-conditional-challenge.spec.ts
sagarchauhan-qa: Done throughout, and renamed the file.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-conditional-challenge.spec.ts
sagarchauhan-qa: Done, now that #8474 has merged.

gravitee-am-test/playwright/tests/mfa/mfa-conditional-enrolment-conditional-challenge.spec.ts
sagarchauhan-qa: Done, now that #8477 has merged.

PR #8496, opened 2026-09-01
fix: pin a system cluster identity provider to the database it actually reads

leleueri:
## Problem

A mongo identity provider created with **use system cluster** had its `database` written from the management API's own repository settings. On a multi data plane installation that names the wrong store: the gateway serves such a provider from the data plane of its domain, so the users live there while the console and the stored configuration point at the management database.

Reproducing AM-7594 — a domain bound to `dp1`, an IDP created with `use system cluster` — the stored configuration read `database: management-db` while the users were written to `dp1-db.<collection>`. An operator looking for those users was sent to a database that does not hold them.

## What changed

**The database is resolved from the node that actually serves the provider.** New `SystemClusterDatabaseResolver` follows `repositories.system-cluster`:

| `repositories.system-cluster` | Database |
|---|---|
| `management` | `repositories.management.mongodb.uri` / `.dbname`, as before |
| `gateway` | the data plane bound to the identity provider, read from `<propertiesBase>.mongodb.*` |

A provider naming no data plane falls back to `default`; a data plane that is unknown or not a mongo one falls back to the scope settings rather than naming a store at random. The same resolver serves both creation and the upgrader, so a provider created between two restarts is never wrong.

**The two storage rules become one.** `repositories.system-cluster-idp.pin-database` and `repositories.system-cluster-idp.prefix-users-collection` are replaced by **`repositories.system-cluster-restricted`**, which owns both the database and the users collection. They were never meaningfully independent — a collection the platform names inside a database the form chooses is a half-owned location — and the state they created had to be carried through the runtime, where `MongoAbstractProvider` re-read the setting to tell the two regimes apart. It no longer does: the provider's own `systemClusterRestricted` flag is enough. That also settles the ticket's second point — the gateway has no opinion on this setting, so the two `gravitee.yml` templates no longer have to agree on a key only one of them declared.

**A system upgrader realigns the stored database on every management API start.** `SystemClusterIdpDatabaseUpgrader` (order 2) walks the providers carrying the flag and rewrites only those whose stored value has actually drifted, so a start with nothing to repair emits no identity provider event and leaves the gateways alone. It writes through a new `IdentityProviderService.updatePinnedStorage`, modelled on `assignDataPlane`: the regular `update` path goes through `SystemClusterIdpPolicy.applyOnUpdate`, which exists to *reject* a storage change — precisely what the platform is doing here — while still emitting the `IDENTITY_PROVIDER/UPDATE` event the gateways need.

It is keyed on the provider's flag rather than on the current setting: a provider created under the regime stays in it whatever the setting becomes later, so its stored database must keep following the platform even once the setting is turned off.

## API and console

`GET /platform/configuration/installation` returns one boolean instead of the two-rule object:

```diff
-{ "type": "managed", "identityProviderStorage": { "pinDatabase": true, "prefixUsersCollection": true } }
+{ "type": "managed", "systemClusterRestricted": true }
```

The `IdentityProviderStorage` schema is dropped. This is a clean replacement with no fallback on the old keys — the feature is unreleased, and `4.13.0` has no patch release yet, so the compatibility check exempts it. `docs/mapi/openapi.yaml` was regenerated with `scripts/regen-oas.sh`; the staleness check passes.

The console reads the single flag and hints both pinned fields on the creation form.

## Testing

- New `SystemClusterDatabaseResolverTest` (8) covering each scope, the data plane path, the default plane, the uri/dbname fallback, an unknown and a non-mongo plane, and the jdbc case.
- New `SystemClusterIdpDatabaseUpgraderTest` (4): drift repaired, no drift left alone, unflagged provider skipped, failed update propagated so the node stops.
- `SystemClusterIdpPolicyTest` and `MongoAbstractProviderTest` lose the "one rule on, the other off" cases, which no longer exist.
- `gravitee-am-service`, `gravitee-am-identityprovider-mongo` and `gravitee-am-management-api-service` build green; console `provider.form.enricher` jest suite passes.

## Not addressed

The ticket's last point — `usersCollection` is `required` in the mongo IDP `schema-form.json` while always overwritten, same for `database` — is left out. Making it conditional trips the schema compatibility checker and changes the console form; it deserves its own change.

## Still to verify

The end-to-end reproduction from the ticket (multi data plane stack, IDP created on `dp1`, a database forced wrong in the store and repaired on the next management API start) has not been run.


gravitee-am-identityprovider/gravitee-am-identityprovider-mongo/src/main/java/io/gravitee/am/identityprovider/mongo/MongoAbstractProvider.java
```diff
@@ -121,14 +120,15 @@ private ClientWrapper<MongoClient> buildClientWrapper(Scope scope) {
                 : getClientWrapperBasedOnConfig(scope);
     }
 
-    // The flag marks a provider created under the regime, which may have pinned the collection
-    // without pinning the database. Only the database rule decides this override.
+    // The flag marks a provider whose storage the platform owns: the database it stores is the one
+    // the platform named, so this node's own client wrapper is the authority on it. The flag is
+    // read rather than the setting, so a node that never enabled the regime still serves such a
+    // provider from the right store.
```
wajb: nitpick: this is historical and noisy to keep in code docs IMO.

gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterDatabaseResolver.java
```diff
@@ -0,0 +1,145 @@
...
+ * runtime. The node that serves the provider is the authority on that name, so this resolver
+ * follows the same settings it does:
+ *
+ * <ul>
+ *     <li>{@code repositories.system-cluster: management} — the provider shares the management
+ *     store, whose database comes from the {@code repositories.management.mongodb} block;</li>
+ *     <li>{@code repositories.system-cluster: gateway} — the provider is served by the data plane
+ *     of its domain, so the database comes from that data plane's own block. This is the case a
+ *     scope-only resolution gets wrong as soon as several data planes are declared: the management
+ *     block names a store the users are not in.</li>
```
wajb: nitpick: can remove this bit since it's historical noise IMO.

gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterDatabaseResolver.java
```diff
@@ -0,0 +1,145 @@
...
+        this.dataPlaneRegistry = dataPlaneRegistry;
+    }
+
+    /**
+     * @return the database the identity provider reads at runtime, or {@code null} when the
+     * platform cannot name one — the caller then leaves the stored value alone rather than
+     * replacing it with a guess.
+     */
+    public String resolve(IdentityProvider identityProvider) {
+        final String scope = environment.getProperty(SYSTEM_CLUSTER, String.class, DEFAULT_SYSTEM_CLUSTER);
```
wajb, line 68: thought: this lookup, the keys and the default are duplicated in a few places, usually in the form of `Scope.fromName(...)`. Can we abstract the lookup to the repository package? 🤔

gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterDatabaseResolver.java
```diff
@@ -0,0 +1,145 @@
...
+            return null;
+        }
+        if (!MONGODB_TYPE.equalsIgnoreCase(description.get().type())) {
+            return null;
+        }
+        return databaseAt(description.get().propertiesBase());
+    }
+
+    private String dataPlaneId(IdentityProvider identityProvider) {
+        // Mirrors DataPlaneRegistryImpl, which reads a domain without a data plane as the default one.
```
wajb: suggestion: rather than mirroring, could this live somewhere in `gravitee-am-dataplane-api` where both can use it?

gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterDatabaseResolver.java
```diff
@@ -0,0 +1,145 @@
...
+    private Optional<DataPlaneDescription> findDataPlane(String dataPlaneId) {
+        // getDataPlanes rather than getProviderById: naming a database must not start a connection.
+        return dataPlaneRegistry.getDataPlanes().stream()
+                .filter(description -> dataPlaneId.equals(description.id()))
+                .findFirst();
+    }
+
+    /**
+     * Mirrors {@code MongoConnectionProvider#getDatabaseName}, which is private to the repository
+     * plugin. A node that resolves the database differently from this method reads one value and
```
wajb, line 101: thought: again, can we consider extracting the common logic?

wajb (approved):

PR #8497, opened 2026-09-01
feat(mfa): let users return to the enrollment page from the verify page

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7609

If you land on the OTP verify page before you scanned the QR code you're stuck, since the code was on the enroll page you just left and there's no way back to it. So this adds a link back to the enroll page.

The catch is that going back didn't actually work. `MFAEnrollEndpoint` calls `provider.enroll()` on every render and TOTP generates a fresh `SharedSecret` each time, so you'd land back on the enroll page holding a different QR code and a different secret to the one you already scanned. Post that form and every code your authenticator gives you is wrong.

So the enroll page now reuses the shared secret already in the session when there is one, and rebuilds the bar code from it with the provider's existing `generateQrCode`. Same secret, same QR, rescanning is optional. This fixes the browser back button too, which had the same problem.

The link only renders while the enrolled factor is `PENDING_ACTIVATION`. Once it's activated I don't render it, otherwise anyone who got past the password on its own could walk back to the enroll page and read the TOTP secret.

`MFAEnrollEndpoint` also clears `mfaEnrollmentCompleted` when it renders. Rendering that page means you haven't finished enrolling, however you got there.

Both copies of `mfa_challenge.html` are updated, the gateway one and the management api preview one. No CSS, the link reuses the styles from the "try other methods" link below it.

## :memo: Test scenarios

`MFAEnrollEndpointTest`:

- secret in session for this factor, reused, `enroll` never called
- nothing in session, fresh `enroll`
- secret in session under a different factor id, fresh `enroll`
- `mfaEnrollmentCompleted` is gone by the time the page renders

`MFAChallengeGetEndpointTest`:

- pending factor, back action is in the template data
- activated factor, absent

```
mvn install -pl gravitee-am-common,gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core
```

538 tests, green.

New jest spec `specs/gateway/mfa/mfa-enroll-back-navigation.spec.ts` walks the whole thing against a real gateway: enroll, challenge, follow the back link, assert the second enroll page carries the same secret, then verify with a TOTP code generated from the first one and land on the callback.

```
npm --prefix gravitee-am-test run test -- specs/gateway/mfa/mfa-enroll-back-navigation.spec.ts
```

Green. The rest of `specs/gateway/mfa` fails on my stack with `Plugin type mock-am-factor not deployed`, nothing to do with this.

Falls back to a fresh `enroll()` if `generateQrCode` comes back empty, so a provider that keeps a key but doesn't do bar codes still works.

Domains with a custom `mfa_challenge` form keep their own markup and won't get the link, same as any other template change here.

@gravitee-io/am

### DIFF
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
@@ -106,11 +106,11 @@ public interface ConstantKeys {
 
     String ENROLLED_FACTOR_KEY = "enrolledFactor";
 
-    String MFA_CHALLENGE_QR_CODE_KEY = "qrCode";
-
     String MFA_ALTERNATIVES_ACTION_KEY = "mfaAlternativesAction";
 
     String MFA_ALTERNATIVES_ENABLE_KEY = "mfaAlternativesEnabled";
+
+    String MFA_ENROLL_BACK_ACTION_KEY = "mfaEnrollBackAction";
 
     String USER_MFA_ENROLLMENT = "user_mfa_enrollment";
     String MFA_FORCE_ENROLLMENT = "mfa_force_enrollment";
--- a/gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
+++ b/gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
@@ -106,6 +106,8 @@ public interface ConstantKeys {
 
     String ENROLLED_FACTOR_KEY = "enrolledFactor";
 
+    String MFA_CHALLENGE_QR_CODE_KEY = "qrCode";
+
     String MFA_ALTERNATIVES_ACTION_KEY = "mfaAlternativesAction";
 
     String MFA_ALTERNATIVES_ENABLE_KEY = "mfaAlternativesEnabled";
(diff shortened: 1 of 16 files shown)


wajb (commented):
> French bundle has no mfa_challenge page keys in it at all.

This isn't true. Perhaps you misread the encoding? 🤔

sclarkamlab:
> > French bundle has no mfa_challenge page keys in it at all.
> 
> This isn't true. Perhaps you misread the encoding? 🤔

It's just claude being a turkey with the PR description, will remove

lgw-gravitee:
The idea is to generate another spare QR code (the first one during enrollment and the second one during the challenge)?

Would it be cleaner to add a "Back" button that effectively removes the selected (but not yet enrolled) factor, so the whole process can start over?

wajb:
> > > French bundle has no mfa_challenge page keys in it at all.
> > 
> > 
> > This isn't true. Perhaps you misread the encoding? 🤔
> 
> It's just claude being a turkey with the PR description, will remove

Not just the description - you need to add strings to `messages_fr.properties`.

sclarkamlab:
> The idea is to generate another spare QR code (the first one during enrollment and the second one during the challenge)?
> 
> Would it be cleaner to add a "Back" button that effectively removes the selected (but not yet enrolled) factor, so the whole process can start over?

This issue can occur when the user logs in they are automatically redirected to the verify page, so I guess in that situation, back would not take them where they need to go.

lgw-gravitee:
> This issue can occur when the user logs in they are automatically redirected to the verify page, so I guess in that situation, back would not take them where they need to go.

It's because the factor is enrolled, but not verified. If the Back button removes the factor, could the authorization flow redirect you to the enrollment step again?

lgw-gravitee (approved):

PR #8498, opened 2026-09-02
fix: AM-7601 decouple entrypoint sync events from data planes

wajb:
[AM-7601](https://gravitee.atlassian.net/browse/AM-7601)

- Fan entrypoint events out to all gateways to avoid the category error of filtering by data plane.
- Warm entrypoints cache per ennvironment on gateway startup to self-heal domain startup.
- Reconcile entrypoints on cloud ENVIRONMENT commands to avoid deletion/recreation churn.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7601]: https://gravitee.atlassian.net/browse/AM-7601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-services/gravitee-am-gateway-services-sync/src/main/java/io/gravitee/am/gateway/services/sync/SyncManager.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-services/gravitee-am-gateway-services-sync/src/main/java/io/gravitee/am/gateway/services/sync/SyncManager.java
@@ -30,6 +30,7 @@ import io.gravitee.am.monitoring.DomainState;
 import io.gravitee.am.repository.management.api.DomainRepository;
 import io.gravitee.am.repository.management.api.EventRepository;
 import io.gravitee.am.monitoring.DomainReadinessService;
+import io.gravitee.am.service.EntryPointManager;
 import io.gravitee.am.common.event.EventManager;
 import io.gravitee.node.api.Node;
 import io.reactivex.rxjava3.core.Completable;
@@ -117,6 +118,10 @@ public class SyncManager implements InitializingBean, DisposableBean {
 
     @Autowired
     private DomainReadinessService domainReadinessService;
+
+    @Lazy
+    @Autowired
+    private EntryPointManager entryPointManager;
 
     private Optional<List<String>> shardingTags;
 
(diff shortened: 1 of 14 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/impl/AbstractEntryPointManager.java
```diff
@@ -112,6 +113,25 @@ public List<Entrypoint> findAllByEnvironmentId(String environmentId) {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public Completable ensureEnvironmentLoaded(String environmentId) {
+        if (environmentId == null) {
+            return Completable.complete();
+        }
+        return Completable.defer(() -> findAllByEnvironmentId(environmentId).isEmpty()
+                ? environmentLoads.computeIfAbsent(environmentId, id -> loadEnvironment(id).cache())
```
sclarkamlab, line 122: `environmentLoads` only drops the entry on error, so once a load succeeds it sticks around forever. If that first read found nothing, the entry is a completed `Completable` and this environment can never be loaded on demand again. Line 121 already checks the cache is empty, so I think the map only needs to dedupe an in-flight read, `doFinally(() -> environmentLoads.remove(id))` would give you the sharing and keep the retry.

gravitee-am-gateway/gravitee-am-gateway-services/gravitee-am-gateway-services-sync/src/main/java/io/gravitee/am/gateway/services/sync/SyncManager.java
```diff
@@ -266,7 +271,15 @@ private Completable deployDomains() {
     }
 
     private Completable deployOne(Domain domain) {
-        return securityDomainManager.deployReactive(domain);
+        return warmEntrypointCache(domain).andThen(securityDomainManager.deployReactive(domain));
+    }
+
+    private Completable warmEntrypointCache(Domain domain) {
+        Completable warm = entryPointManager.ensureEnvironmentLoaded(domain.getReferenceId());
+        if (eventsTimeOut > 0) {
```
sclarkamlab, line 279: When this times out nothing gets logged. The `doOnError` in `loadEnvironment` is upstream of the `.cache()`, so the `TimeoutException` raised here never reaches it and `onErrorComplete()` just eats it. You end up with domains showing DEPLOYED, an empty entrypoint cache and no log line saying why, worth a `doOnError(e -> log.warn(...))` before the `onErrorComplete`.

sclarkamlab (commented):
Approach looks right to me. The null `dataPlaneId` works because both `MongoEventRepository` and `SpringEventRepository` already match `data_plane_id is null`, and the reconcile kills a lot of real churn. Main things are the cache memoisation never expiring on success, and one test that doesn't test what it says.

gravitee-am-service/src/test/java/io/gravitee/am/service/impl/AbstractEntryPointManagerTest.java
```diff
@@ -243,6 +248,79 @@ public void shouldKeepCachedEntrypointWhenReloadFails() {
...
+        cache(entrypoint);
+
+        cut.ensureEnvironmentLoaded("env#1").blockingAwait();
+
+        verifyNoInteractions(environmentRepository);
+        assertEquals(List.of(entrypoint), cut.findAllByEnvironmentId("env#1"));
+    }
+
+    @Test
+    public void shouldShareOneReadBetweenConcurrentCallers() {
```
sclarkamlab, line 310: This one isn't testing concurrency. `ensureEnvironmentLoaded` is a `defer` so `computeIfAbsent` only runs on subscribe, and `first.blockingAwait()` completes and fills the cache before `second` ever subscribes, so `second` short circuits on the `isEmpty` check. It's the same test as `shouldNotReadTheStoreAgainOnceAnEnvironmentIsLoaded`, you'd need an async source to actually overlap the two subscriptions.

sclarkamlab (approved):

PR #8515, opened 2026-09-03
test: attestation root certificates keep a valid issue (AM-7617)

sagarchauhan-qa:
[AM-7617](https://gravitee.atlassian.net/browse/AM-7617)

Six attestation root certificates are compiled into `GraviteeWebAuthnOptions`, and nothing looks at their expiry dates. An expired one used to fail the `webAuthn` bean at start-up and stop security domains deploying; AM-7055 made that non-fatal, so it is now dropped from the trust set silently instead. Neither the expiry nor the drop is reported.

- Every attestation root identity keeps a valid certificate
- No identity's last certificate expires within 90 days

One new file, `GraviteeWebAuthnOptionsTest`. No production code touched. Pure JUnit, so it runs in `junit_tests` on every PR.

## :memo: Test scenarios
See jira.

n/a

**Why identities and not certificates.** Asserting all six load goes red immediately — only five do:

| Constant | Expires | Subject |
|---|---|---|
| `ANDROID_KEYSTORE_ROOT` | 2036-01-06 | `CN=Android Keystore Software Attestation Root` |
| `ANDROID_KEYSTORE_ROOT_1` | **2026-05-24 — expired** | `serialNumber=f92009e853b6b045` |
| `ANDROID_KEYSTORE_ROOT_2` | 2034-11-18 | `serialNumber=f92009e853b6b045` |
| `ANDROID_KEYSTORE_ROOT_3` | 2036-11-13 | `serialNumber=f92009e853b6b045` |
| `ANDROID_KEYSTORE_ROOT_4` | 2042-03-15 | `serialNumber=f92009e853b6b045` |
| `ANDROID_KEYSTORE_ROOT_5` | 2035-07-15 | `CN=Key Attestation CA1` |

`ROOT_1` through `ROOT_4` are rotated issues of one Google hardware root, so `ROOT_1` lapsing costs nothing while three successors remain valid. Counting identities catches an identity losing its **last** certificate — the point attestation silently stops validating — and ignores retired issues by design. `ROOT_1` is deliberately left in place.

Verification, sources and the coverage survey are on [AM-7617](https://gravitee.atlassian.net/browse/AM-7617). The other half of this P1 — that a domain still deploys when a root has expired — needs a running gateway and is [AM-7618](https://gravitee.atlassian.net/browse/AM-7618).

@gravitee-io/am

[AM-7617]: https://gravitee.atlassian.net/browse/AM-7617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-7617]: https://gravitee.atlassian.net/browse/AM-7617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/GraviteeWebAuthnOptionsTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/GraviteeWebAuthnOptionsTest.java
@@ -1,0 +1,114 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.vertx.auth.webauthn;
+
+import org.junit.jupiter.api.Test;
+
+import java.security.cert.X509Certificate;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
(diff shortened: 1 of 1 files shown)


wajb (commented):
I'm not sure we need to do anything more to cover expiry tbh. We already log when certs are dropped, including by expiry [here](https://github.com/gravitee-io/gravitee-access-management/blob/master/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/main/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/GraviteeWebAuthnOptions.java#L244-L259). It can actually be too noisy which is why I think the stacktrace was removed by default.

Also, testing for expiry is just a time bomb that doesn't help with anything already in production. (The root certs vertx provides aren't covered here either.) I think the logs are probably a better signal, where we could potentially include the exception message if we want the cause to be more visible.

sagarchauhan-qa:
Thanks @wajb. Agreed - closing this. The startup log already tells the operator, and a build-time warning doesn't help anything already released. AM-5812 is the right fix: loading the certs from a file means a replacement doesn't need a new AM build, which removes the failure mode rather than warning about it.

PR #8516, opened 2026-09-03
fix: AM-7612 enabling system mode is irreversible

wajb:
[AM-7612](https://gravitee.atlassian.net/browse/AM-7612)

Under `repositories.system-cluster-resticted` mode, prevent disabling of Mongo IDP "use system cluster" option once enabled. Previously this was only guarded for an IDP that was _created_ under restricted mode, leaving a gap for existing IDPs where inadvertent disabling of the mode could never be undone.

Bring attention to this fact by adding disclaimers to the console UI:
- A hint during IDP creation that "use system cluster" cannot be updated once set (restricted mode only)
- A recommendation onscreen to use datasource instead

Additionally, to draw more attention to the guidance, this markup is now shown onscreen during IDP creation.

## :memo: Test scenarios 
See jira.

#### Mongo IDP data plane guidance
image

#### Augmented "use system cluster" warning hint
image

n/a

@gravitee-io/am 

[AM-7612]: https://gravitee.atlassian.net/browse/AM-7612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterIdpPolicy.java
+++ b/gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterIdpPolicy.java
@@ -32,6 +32,9 @@ import java.util.Optional;
  * provider id. Only providers created under this regime are affected, so an existing provider keeps
  * working whatever the setting becomes later.
  *
+ * <p>Under that regime a provider's use of the system cluster is settled at creation, whether or not
+ * the platform pinned it.
+ *
  * @author GraviteeSource Team
  */
 @Component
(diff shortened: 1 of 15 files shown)


gravitee-am-service/src/main/java/io/gravitee/am/service/idp/SystemClusterIdpPolicy.java
```diff
@@ -66,14 +71,28 @@ public void applyOnUpdate(IdentityProvider stored, IdentityProvider identityToUp
...
+            rejectSystemClusterDisable(identityToUpdate);
             return;
         }
         rejectSystemClusterSwitch(identityToUpdate);
     }
 
+    // Leaving the regime puts the provider back on the connection settings of the form, while the
+    // users it authenticates stay behind in the system cluster.
+    private void rejectSystemClusterDisable(IdentityProvider identityToUpdate) {
+        if (!isEligible(identityToUpdate)) {
```
leleueri: question: should we prevent the update if the "system-cluster-restricted" is disabled ? I know it may bring some confusion but in the same time, this is the current behaviour... 🤔

wajb: Limiting the change to restricted mode actually simplifies the logic a lot and removes the need for the confirmation prompt (the "use system cluster" option is symmetrical and immutable only in restricted mode). Yes, let's preserve existing behaviour otherwise.

leleueri (approved):

PR #8517, opened 2026-09-03
fix: skip mongo uri setting for default idp relying on system cluster

leleueri:
This commit remove the mongo connection settings from the default idp config.
Is introduce a new setting to disable this behaviour (domains.identities.default.useSystemCluster)
If disable the default config is initialized using the management scope settings.

In cloud mode (installation.type: managed), the system cluster usage is enforced.

Important Reminder: the setting "repositories.system-cluster" has to be configure in a consistent manner across Management API and Gateway services. Only two values are possible
* management: idp collections are in the same cluster/database as the configuration entities (domain, app, factor configuration...)
* gateway: idp are manage by the DataPlane storage. That means in the MAPI, the connection rely on the dataplane definition linked to the domain and on Gateway side, the connection rely on the gateway scope settings. It is required that the gateway repository scope settings on the GW gravitee.yaml match the same cluster and db as the dataplane definition on MAPI

Fix AM-7606

### DIFF
--- a/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DefaultIdentityProviderServiceImpl.java
+++ b/gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DefaultIdentityProviderServiceImpl.java
@@ -18,6 +18,7 @@ package io.gravitee.am.management.service.impl;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.BaseEncoding;
+import io.gravitee.am.common.env.CloudProperties;
 import io.gravitee.am.common.env.RepositoriesEnvironment;
 import io.gravitee.am.identityprovider.api.User;
 import io.gravitee.am.management.service.DefaultIdentityProviderService;
@@ -43,6 +44,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
+import static io.gravitee.am.repository.BackendConfigurationUtils.DEFAULT_SYSTEM_CLUSTER;
 import static io.gravitee.am.repository.BackendConfigurationUtils.getMongoDatabaseName;
 import static io.gravitee.am.repository.BackendConfigurationUtils.SYSTEM_CLUSTER;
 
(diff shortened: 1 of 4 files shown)


lgw-gravitee (approved):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DefaultIdentityProviderServiceImpl.java
```diff
@@ -116,7 +124,7 @@ private <T extends NewIdentityProvider> Single<T> populateDefault(Domain domain,
     public Map<String, Object> createProviderConfiguration(String referenceId, NewIdentityProvider identityProvider) {
         // A new default identity provider always reuses the system cluster; the Mongo provider decides at
```
wajb: minor: the new option means this comment is no longer accurate.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DefaultIdentityProviderServiceImpl.java
```diff
@@ -172,9 +180,13 @@ private Map<String, Object> buildProviderConfiguration(String referenceId, NewId
...
 
-            configMap.put("uri", mongoUri);
-            configMap.put("host", (mongoHost != null) ? mongoHost : "");
-            configMap.put("port", (mongoPort != null) ? Integer.parseInt(mongoPort): null);
+            if (!boundToSystemCluster) {
+                configMap.put("uri", mongoUri);
+                configMap.put("host", (mongoHost != null) ? mongoHost : "");
+                configMap.put("port", (mongoPort != null) ? Integer.parseInt(mongoPort) : null);
+            } else {
+                configMap.put("uri", String.format("from-system-cluster:%s", environment.getProperty(SYSTEM_CLUSTER, DEFAULT_SYSTEM_CLUSTER)));
```
wajb, line 186: thought: do we want to state which scope is being used? This can be changed on the gw side, mapi and gw could disagree or it could be a connection taken from the data plane 🤔

leleueri: I think is can be beneficial for support to troubleshoot.

With useSystemCluster, the configuration across GW and MAPI has to be consistent, if MAPI define management scope, the GW config also have to define management scope. (maybe something we can improve and persist the scope so the GW always use the same as the MAPI... but it is an improvement to manage in another ticket)

there is only 2 scopes :management and gateway
* management: idp collections are in the same cluster/db as the configuration entities (domain, app, ...)
* gateway: idp are manage by the DataPlane storage so in the MAPI, the connection rely on  the dataplane definition linked to the domain and on GW side, the connection rely on the gateway scope settings. It is required that the gateway scope settings on the GW gravitee.yaml match the same cluster and db as the dataplane definition on MAPI (this is documented)

wajb (approved):

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/impl/DefaultIdentityProviderServiceImpl.java
wajb: If these configurations are the only supported and it is already documented then I agree it's not an issue.

PR #8519, opened 2026-09-04
test: an unloadable attestation root degrades the trust set (AM-7618)

sagarchauhan-qa:
[AM-7618](https://gravitee.atlassian.net/browse/AM-7618)

AM-7055 made an unloadable attestation root non-fatal at start-up, after an expired one stopped security domains deploying. That fix has no test. It swallows the exception at two points and neither is exercised: `pushAdditionalRootCertificate` for the roots compiled into the product, and `putRootCertificate` for the ones a domain supplies through its WebAuthn settings.

- Constructing the options does not propagate a certificate failure
- A rejected root leaves the certificates that did load intact
- A malformed certificate in a domain's WebAuthn settings does not fail the `webAuthn` bean
- One bad domain-supplied certificate does not cost the roots shipped with the product

One new file, `WebAuthnFactoryResilienceTest`. No production code touched. Pure JUnit, so it runs in `junit_tests` on every PR.

## :memo: Test scenarios
See jira.

n/a

**The domain-supplied path is the one worth having.** `WebAuthnFactory` is the `FactoryBean` named in the original stack trace — a throw from `getObject()` surfaced as an unsatisfied `webAuthn` dependency on `rootProvider` and stopped `SyncManager` entirely. A malformed certificate in a domain's own settings reaches that path, and before AM-7055 it failed the bean every security domain on the gateway depends on. Nothing covered it.

**The assertions are not only "does not throw".** Each pairs that with the trust set surviving. Degrading is acceptable; emptying is not. Without the second half these would pass against a version that caught the exception and returned an empty root set — which boots cleanly and then validates no attestation at all.

Verification and scope are on [AM-7618](https://gravitee.atlassian.net/browse/AM-7618). The other half of this P1, noticing a certificate is near expiry before it lapses, is [AM-7617](https://gravitee.atlassian.net/browse/AM-7617) / #8515.

@gravitee-io/am

[AM-7618]: https://gravitee.atlassian.net/browse/AM-7618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-7618]: https://gravitee.atlassian.net/browse/AM-7618?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/WebAuthnFactoryResilienceTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/WebAuthnFactoryResilienceTest.java
@@ -1,0 +1,139 @@
+/**
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.gravitee.am.gateway.handler.vertx.auth.webauthn;
+
+import io.gravitee.am.gateway.handler.vertx.auth.webauthn.store.RepositoryCredentialStore;
+import io.gravitee.am.model.Domain;
+import io.gravitee.am.model.login.WebAuthnSettings;
+import io.vertx.ext.auth.webauthn.WebAuthn;
+import io.vertx.rxjava3.core.Vertx;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+import java.security.cert.X509Certificate;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.Mockito.when;
+
+/**
+ * Pins the behaviour AM-7055 introduced: an attestation root certificate that cannot be loaded must
+ * degrade the trust set, not stop the gateway starting.
+ *
+ * <p>Before that fix an expired certificate threw out of the {@link GraviteeWebAuthnOptions}
+ * constructor. {@link WebAuthnFactory} is a {@code FactoryBean}, so the throw surfaced as
+ * "FactoryBean threw exception on object creation" for the {@code webAuthn} bean, left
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/WebAuthnFactoryResilienceTest.java
```diff
@@ -0,0 +1,139 @@
...
+                .as("An unloadable additional root must be skipped, not propagated.")
+                .doesNotThrowAnyException();
+
+        assertThat(options.getAdditionalRootCertificate(ANDROID_KEY))
+                .as("The certificates that did load must be left intact when a later one is rejected.")
+                .containsExactlyElementsOf(before);
+    }
+
+    @Test
+    void shouldBuildWebAuthnWhenTheDomainSuppliesACertificateThatCannotBeLoaded() {
```
wajb: thought: this is the same functional area but only this test covers the actual `WebAuthnFactory`. That's at integration level whereas the others are unit level tests of `GraviteeWebAuthnOptions`.

Maybe a @Nested block of tests would be enough to group those while keeping them under this topic?

sagarchauhan-qa: Good call — done in fddd2630e6.

Two `@Nested` groups, split by level as you suggested:

- **Loading certificates into `GraviteeWebAuthnOptions`** — the three unit-level ones (constructor, `pushAdditionalRootCertificate`, `putRootCertificate`)
- **Building the `webAuthn` bean through `WebAuthnFactory`** — the one that goes through the factory

Keeping them in one file felt right since they're two halves of the same property: the factory is where the failure actually surfaced in the incident (`FactoryBean threw exception on object creation` → `rootProvider` unsatisfied → no domain deployed), but the swallow it depends on lives in the options. Splitting them would have hidden that.

One wrinkle worth knowing: with `@Nested`, Surefire reports the outer class as `Tests run: 0` and records the nested ones separately in the same XML. All four still execute and the module total is unchanged at 617 — I checked the XML rather than trusting the summary line. But running the class alone now needs `-Dtest='WebAuthnFactoryResilienceTest$*'`; the plain name matches nothing.

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-core/src/test/java/io/gravitee/am/gateway/handler/vertx/auth/webauthn/WebAuthnFactoryResilienceTest.java
```diff
@@ -0,0 +1,158 @@
...
+ * {@code pushAdditionalRootCertificate} for the roots compiled into the product, and
+ * {@code putRootCertificate} for the ones a domain supplies through its WebAuthn settings. The
+ * second is reachable by any customer with a malformed certificate in their configuration.
+ *
+ * <p>The two nested groups separate the levels: loading certificates into the options object, and
+ * building the {@code webAuthn} bean through the factory that failed in the incident.
+ *
+ * <p>AM-7617 covers the other half of this P1 — noticing that a certificate is close to expiry
+ * before it lapses.
+ */
```
wajb: minor: we should probably remove this. It exposes some internal facts.

sagarchauhan-qa: Agreed — removed the whole block in 050d6cea32. Shouldn't have been in a public repo at all.

wajb (approved):

PR #8520, opened 2026-09-04
test: map a provider duplicate to a SCIM conflict (AM-7620)

sagarchauhan-qa:
[AM-7620](https://gravitee.atlassian.net/browse/AM-7620)

`ProvisioningUserServiceImpl` checks for an existing username before persisting, but that check is not atomic. Two concurrent `POST /scim/Users` for the same username both pass it and race to the identity provider; the loser hits the database's unique constraint. The provider reports that as `UserAlreadyExistsException`, and `ProvisioningUserServiceImpl:342` maps it to `UniquenessException` so `ScimErrorMapper` can return a 409. Unmapped it surfaces as a 500, which is what was reported on AM-7285.

That mapping had no test. `ProvisioningUserServiceTest` had zero references to `UserAlreadyExistsException`.

- A duplicate reported by the provider is returned to the caller as a conflict, not an internal error

One method appended to `ProvisioningUserServiceTest`, plus its import. No production code touched, no existing test changed.

## :memo: Test scenarios
See jira.

n/a

**The concurrency itself is already covered**, at the layer below: `JdbcUserProvider_Test#shouldReturnUserAlreadyExists_onConcurrentCreate` runs genuinely concurrent creates and asserts every loser surfaces `UserAlreadyExistsException` rather than a raw database error. It shipped with the AM-7285 fix. What was missing was only the hop after it — the one that decides whether the customer sees 409 or 500.

**It sits with the two tests it completes.** `shouldNotCreateUserWhenUsernameAlreadyUsed` and `shouldNotCreateUserWhenExternalIdAlreadyUsed` cover the pre-check path, which was never the broken one. This covers the race that gets past them.

Verification is on [AM-7620](https://gravitee.atlassian.net/browse/AM-7620).

@gravitee-io/am

[AM-7620]: https://gravitee.atlassian.net/browse/AM-7620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-7620]: https://gravitee.atlassian.net/browse/AM-7620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
@@ -367,25 +367,12 @@ public class ProvisioningUserServiceTest {
         verify(userRepository, never()).create(any());
     }
 
-    /**
-     * The two checks above run before the user is persisted, so they are not atomic. Two concurrent
-     * requests for the same username both pass them and race to the identity provider; the loser hits
-     * the database's unique constraint, which the provider reports as
-     * {@link UserAlreadyExistsException}. That much is covered at the provider layer by
-     * {@code JdbcUserProvider_Test#shouldReturnUserAlreadyExists_onConcurrentCreate}.
-     *
-     * <p>This pins the hop after it. {@code ScimErrorMapper} turns {@link UniquenessException} into
-     * the 409 the caller sees, so leaving the provider's exception unmapped surfaces a 500 instead —
-     * which is what was reported for {@code POST /scim/Users} on AM-7285, fixed in 4.9.27, 4.10.18,
-     * 4.11.12 and 4.12.2.
-     */
     @Test
     public void shouldNotCreateUserWhenTheProviderReportsADuplicate() {
         UserProvider userProvider = mock(UserProvider.class);
         when(userProvider.create(any())).thenReturn(Single.error(new UserAlreadyExistsException("username-1")));
 
-        // nothing exists yet, so both checks pass and the request reaches the provider — the window
-        // two concurrent creates get through together
+        // nothing exists yet, so the pre-checks pass and the request reaches the provider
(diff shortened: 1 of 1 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
```diff
@@ -366,6 +367,41 @@ public void shouldNotCreateUserWhenExternalIdAlreadyUsed() {
...
+     * requests for the same username both pass them and race to the identity provider; the loser hits
+     * the database's unique constraint, which the provider reports as
+     * {@link UserAlreadyExistsException}. That much is covered at the provider layer by
+     * {@code JdbcUserProvider_Test#shouldReturnUserAlreadyExists_onConcurrentCreate}.
+     *
+     * <p>This pins the hop after it. {@code ScimErrorMapper} turns {@link UniquenessException} into
+     * the 409 the caller sees, so leaving the provider's exception unmapped surfaces a 500 instead —
+     * which is what was reported for {@code POST /scim/Users} on AM-7285, fixed in 4.9.27, 4.10.18,
+     * 4.11.12 and 4.12.2.
+     */
```
wajb: suggestion: these javadocs are noisy in several ways and can just be dropped IMO:
- describes _other tests_ and their scope (will diverge)
- references historical jira ticket (whose git history will not align with this change)
- pins versions unnecessarily
- describes pathways it does not exercise (unreachable 500 code)

sagarchauhan-qa: Agreed on all four — dropped in 0bc078665b. The context lives on AM-7620 instead.

wajb (approved):

PR #8541, opened 2026-09-08
test: valueless SCIM multi-valued attributes on write and patch (AM-7621)

sagarchauhan-qa:
[AM-7621](https://gravitee.atlassian.net/browse/AM-7621)

Follows the fix in #8521, which rejects SCIM writes where `emails`, `phoneNumbers`, `ims` or `photos` contain an entry without a value, and filters such entries when reading users already persisted in that state. The incident behind it had two symptoms — the write that created the state, and reads failing afterwards — and two paths were left unverified.

**The HTTP contract.** `scim-users.jest.spec.ts` asserted `400` and `scimType: invalidValue` for a malformed body, a missing `userName`, and invalid `userName`, `familyName` and `givenName`, but nothing covered a valueless entry in a multi-valued attribute. Five tests added in a new `Multi-valued attribute validation` block:

- `POST` with a valueless email, and with a valueless phone number, so the check is shown not to be email-only
- `PUT` and `PATCH` with a valueless email
- a positive control creating a user whose entries all have values, without which the rejections would pass against an implementation that rejected everything

Each rejection asserts the full error contract through a shared helper: status, `scimType`, and the exact `detail`.

**Patching a user whose stored record is already affected.** `shouldNotPatchUser_whenAnEmailHasNoValue` covers a bad *payload*. Nothing covered a *stored* record holding a valueless entry, which is the case that failed: `patch()` begins with `innerGet(userId, baseUrl)` and so runs the same conversion. One test added to `ProvisioningUserServiceTest` — it captures the SCIM user handed to the patch operation and asserts the valueless entry was dropped, so it pins the self-heal rather than only the absence of an exception.

No production code touched, no existing test changed.

## :memo: Test scenarios
See jira.

n/a

**Why one of these is a unit test rather than an integration one.** After #8521 the affected state can no longer be created through the API, so reproducing "already-persisted bad data, then patch it" over HTTP would need a direct database write. A mocked repository is the right level for it. The equivalent read case was already covered by `shouldListUsers_withStoredEmailAttributeWithoutValue` in #8521.

`ims` and `photos` are covered at the unit layer only. They share the single `requireAttributeValues` path that the `emails` and `phoneNumbers` cases already exercise over HTTP.

Verified against a local stack built from master: jest 30/30 on the full spec file, unit 46/46 on the full test class. Each new assertion was checked by inverting it and confirming the test fails.

@gravitee-io/am

[AM-7621]: https://gravitee.atlassian.net/browse/AM-7621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
@@ -1688,10 +1688,8 @@ public class ProvisioningUserServiceTest {
         testObserver.assertNoErrors();
         testObserver.assertComplete();
 
-        // The stored valueless entry must be dropped while reading, not carried into the patch
-        // and not raised as an error. Reading it is what returned 500 for the whole domain.
         List<Attribute> emailsHandedToThePatch = convertedScimUser.getValue().getEmails();
-        assertEquals(1, emailsHandedToThePatch.size());
+        assertEquals("the valueless entry should be dropped before the patch is applied", 1, emailsHandedToThePatch.size());
         assertEquals("real@example.com", emailsHandedToThePatch.get(0).getValue());
     }
 
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
@@ -1647,6 +1647,54 @@ public class ProvisioningUserServiceTest {
         verify(userRepository, never()).update(any(), any());
     }
 
+    @Test
+    public void shouldPatchUser_whenTheStoredUserHasAnEmailWithoutValue() throws Exception {
+        final String userId = "user-id";
+
+        io.gravitee.am.model.User existingUser = new io.gravitee.am.model.User();
+        existingUser.setId(userId);
+        existingUser.setUsername("nullmail2");
+        existingUser.setSource("user-idp");
+        existingUser.setEmail("real@example.com");
+        existingUser.setReferenceId(DOMAIN_ID);
+        existingUser.setReferenceType(ReferenceType.DOMAIN);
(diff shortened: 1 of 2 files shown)


lgw-gravitee (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
```diff
@@ -1647,6 +1647,54 @@ public void shouldNotPatchUser_whenAnEmailHasNoValue() throws Exception {
...
+        when(tokenService.deleteByUser(any())).thenReturn(Completable.complete());
+        when(userRepository.update(any(), any())).thenAnswer(invocation -> Single.just(invocation.getArgument(0)));
+        when(eventService.create(any(), any())).thenReturn(Single.just(new Event()));
+
+        TestObserver<User> testObserver = userService.patch(userId, patchOp, null, "/", null, null).test();
+        testObserver.assertNoErrors();
+        testObserver.assertComplete();
+
+        // The stored valueless entry must be dropped while reading, not carried into the patch
+        // and not raised as an error. Reading it is what returned 500 for the whole domain.
```
wajb: suggestion: remove this comment. It describes a historical scenario that needs more context and it doesn't stand by itself.

gravitee-am-test/specs/gateway/scim/scim-users.jest.spec.ts
wajb, line 1: thought: IMO these new tests test the same pathway that's already covered (more cheaply) at the unit level. Fine here, since they're quick (<20ms), but we should consider settling on a stable test pyramid rather than put too much granularity near the top.

wajb (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
sagarchauhan-qa: Done - removed. Moved the intent into the assertion message so it shows up on failure rather than sitting above it.

sagarchauhan-qa:
Hi @wajb - I have decided to keep them - POST, PUT and PATCH are separate handlers and separate service methods, so each could miss the validation independently and the unit tests would still pass. The other two are the phoneNumbers case and a positive control.

PR #8542, opened 2026-09-08
feat: AM-7554 restrict custom reporter attributes

wajb:
[AM-7554](https://gravitee.atlassian.net/browse/AM-7555)

- Define extensible list of sensitive attributes that are filtered from audit records.
  - This list can be extended with system setting.
- Expose opt-in list of event types where custom attributes are allowed.
- Expose 23 missing event types in `GET /platform/audits/events` endpoint and Kafka reporter configuration.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am

[AM-7554]: https://gravitee.atlassian.net/browse/AM-7554?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/attribute/SensitiveAttributeDenylist.java
```diff
@@ -0,0 +1,107 @@
...
+        return denied.contains(normalized)
+                || DENIED_SUFFIXES.stream().anyMatch(normalized::endsWith);
+    }
+
+    /**
+     * Removes the denied names from {@code attributes} in place.
+     */
+    public void scrub(Map<String, ?> attributes) {
+        if (attributes != null && !attributes.isEmpty()) {
+            attributes.keySet().removeIf(this::isDenied);
```
sclarkamlab: `removeIf` on the keySet only strips the top level, so anything nested one deeper still goes out. `additionalInformation` comes straight off a JSON body so a nested map is easy to make. I ran this against gravitee-el 4.3.0 and got the value back:

```
{#context.attributes['user']['claims']['idp']['access_token']}  ==>  SECRET-AT
```

Either recurse into nested maps here, or drop any claim whose value isn't itself exportable before the mappings run.

gravitee-am-test/specs/gateway/reporters/domain-reporter-attribute-mappings.jest.spec.ts
```diff
@@ -268,4 +268,56 @@ describe('Reporter attribute mappings - Domain Level Gateway', () => {
...
   });
+
+  describe('Sensitive attributes', () => {
+    it('should never export a denied attribute, however the expression spells it', async () => {
+      const topic = uniqueName('mapping-denied-attribute', true);
+      await fixture.addReporter(
+        topic,
+        [],
+        [
+          { expression: "{#context.attributes['user'].claims['op_access_token']}", exportedName: 'provider_value' },
```
sclarkamlab: This passes with the denylist deleted. The fixture user only has `employeeId` and `department`, so `op_access_token` was never in the claims to begin with. It's the wrong probe anyway, `UserProperties.removeSensitiveClaims` strips it before `scrub` ever runs. I'd put a name only the denylist covers on the fixture user, `azure_b2c_refresh_token` say, and assert `employeeId` still comes through.

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/attribute/AttributeMappingReporter.java
```diff
@@ -89,6 +88,10 @@ public void report(io.gravitee.reporter.api.Reportable reportable) {
         delegate.report(reportable);
     }
 
+    private boolean appliesTo(Audit audit) {
+        return eventTypes.isEmpty() || eventTypes.contains(audit.getType());
```
sclarkamlab, line 92: This runs outside the try that keeps the event alive, and `Set.copyOf(...).contains(null)` throws NPE (checked on 25). So a null audit type drops the event instead of passing it through unenriched. `audit.getType() != null && eventTypes.contains(audit.getType())` covers it.

gravitee-am-common/src/main/java/io/gravitee/am/common/audit/EventType.java
```diff
@@ -394,46 +398,30 @@ public interface EventType {
...
-                I18N_DICTIONARY_CREATED, I18N_DICTIONARY_UPDATED, I18N_DICTIONARY_DELETED,
-                THEME_UPDATED, THEME_DELETED, THEME_CREATED, MFA_VERIFICATION_LIMIT_EXCEED,
-                TOKEN_CREATED, TOKEN_REVOKED, TOKEN_REVOKE_SCHEDULED, CLIENT_AUTHENTICATION,
-                MFA_CHALLENGE, MFA_ENROLLMENT, MFA_CHALLENGE_SENT, MFA_MAX_ATTEMPT_REACHED,
-                MFA_RATE_LIMIT_REACHED, MFA_REMEMBER_DEVICE,
-                PASSWORD_POLICY_CREATED, PASSWORD_POLICY_UPDATED, PASSWORD_POLICY_DELETED,
-                PROTECTED_RESOURCE_CREATED, PROTECTED_RESOURCE_UPDATED, PROTECTED_RESOURCE_DELETED,
-                PERMISSION_EVALUATED
-        ));
+        return new TreeSet<>(DeclaredTypes.VALUES);
```
sclarkamlab, line 401: This takes `types()` from 135 to 158. Looks like a genuine fix to me, every constant in there is a real event type, but it also widens the console audit filter dropdown and what `AnalyticsParam` will accept as `field`. Worth a line in the PR description.

wajb: Mentioned as last bullet but I did not call out analytics. It's an incidental fix, yeah, and should avoid drift in future.

sclarkamlab (approved):
Nested scrub looks right, and the jest spec actually fails without the denylist now. Good from me.

PR #8561, opened 2026-09-10
test: cover self-account consent, access control, username and factor QR (AM-7652)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7652

Adds 36 Jest tests across 4 new spec files, taking `specs/gateway/self-account/` from 34 to 70. The self-account API has 17 routes; these cover the four areas with no coverage at any layer.

- `self-account-consent` (9) — list, read and revoke consent records, each driven through the authorization-code flow to create one
- `self-account-access-control` (12) — the `selfServiceAccountManagement.enabled` switch on and off, plus missing, malformed and foreign tokens
- `self-account-username` (11) — the username validation rules from [AM-3686](https://gravitee.atlassian.net/browse/AM-3686), asserting the error messages rather than just the status
- `self-account-factor-qr` (4) — the QR code for an enrolled factor, including the three ways it refuses before reaching the provider

**Two consent tests are committed as skipped** against [AM-7654](https://gravitee.atlassian.net/browse/AM-7654): a signed-in user can currently read *and delete* another user's consent record, because neither endpoint scopes the lookup to the caller. They un-skip when it is fixed.

### :warning: One existing file is modified — `fixture/self-account-fixture.ts` (+6/-4)

`setupFixture` and `setupFactorFixture` gain an **optional** `domainNamePrefix`, defaulting to the exact literal each already used. Nothing is removed and no signature is broken, so **the four existing specs produce the same domain names and behave identically**.

Why it is needed: the fixtures hardcode the domain name they create, so every caller gets the same one. Jest runs spec files in parallel workers, and the random part of the name is not unique across processes that start in the same millisecond — two files asking for the same domain means the second fails to start. Adding four more callers made that fire on most runs; with a prefix per spec file the suite is stable.

That underlying weakness in `uniqueName()` affects ~700 call sites and has an identical twin in the Playwright helpers. Raised as [AM-7655](https://gravitee.atlassian.net/browse/AM-7655) rather than fixed here.

## :memo: Test scenarios 
Ran the full self-account folder 11 times consecutively: 70 tests, 68 passed, 2 skipped, no flakes.

NA

NA

[AM-3686]: https://gravitee.atlassian.net/browse/AM-3686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AM-7654]: https://gravitee.atlassian.net/browse/AM-7654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
+++ b/gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
@@ -70,9 +70,11 @@ export async function getEndUserAccessToken(fixture: SelfAccountFixture, passwor
   return response.body.access_token;
 }
 
-export const setupFixture = async (domainSettings: any): Promise<SelfAccountFixture> => {
+// Each spec file passes its own domain name prefix: two files sharing one can generate the same
+// name in parallel workers, and the second domain then fails to start.
+export const setupFixture = async (domainSettings: any, domainNamePrefix = 'self-account-change-password'): Promise<SelfAccountFixture> => {
   const accessToken = await requestAdminAccessToken();
-  const domain = await createDomain(accessToken, uniqueName('self-account-change-password', true), 'Description');
+  const domain = await createDomain(accessToken, uniqueName(domainNamePrefix, true), 'Description');
   await patchDomain(domain.id, accessToken, domainSettings);
   const idpSet = await getAllIdps(domain.id, accessToken);
 
@@ -154,9 +156,9 @@ const SELF_ACCOUNT_ENABLED_SETTINGS = {
   },
 };
 
-export const setupFactorFixture = async (): Promise<SelfAccountFactorFixture> => {
+export const setupFactorFixture = async (domainNamePrefix = 'self-account-factors'): Promise<SelfAccountFactorFixture> => {
   const accessToken = await requestAdminAccessToken();
-  const domain = await createDomain(accessToken, uniqueName('self-account-factors', true), 'Description');
+  const domain = await createDomain(accessToken, uniqueName(domainNamePrefix, true), 'Description');
   await patchDomain(domain.id, accessToken, SELF_ACCOUNT_ENABLED_SETTINGS);
   const idpSet = await getAllIdps(domain.id, accessToken);
 
--- a/gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
+++ b/gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
@@ -70,8 +70,47 @@ export async function getEndUserAccessToken(fixture: SelfAccountFixture, passwor
   return response.body.access_token;
 }
 
-// Each spec file passes its own domain name prefix: two files sharing one can generate the same
-// name in parallel workers, and the second domain then fails to start.
+/** Base URL of the self-account API for a fixture's domain. */
+export const accountApiUrl = (fixture: SelfAccountFixture, path: string) =>
+  `${process.env.AM_GATEWAY_URL}/${fixture.domain.hrid}/account/api${path}`;
+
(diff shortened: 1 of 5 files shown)


gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
```diff
@@ -70,9 +70,11 @@ export async function getEndUserAccessToken(fixture: SelfAccountFixture, passwor
   return response.body.access_token;
 }
 
-export const setupFixture = async (domainSettings: any): Promise<SelfAccountFixture> => {
+// Each spec file passes its own domain name prefix: two files sharing one can generate the same
+// name in parallel workers, and the second domain then fails to start.
```
wajb: nitpick: this'll likely be one of those comments that becomes out of date (you already have a ticket to resolve it). Better to remove it now IMO.

gravitee-am-test/specs/gateway/self-account/self-account-access-control.jest.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  userToken = await getEndUserAccessToken(enabled);
+});
+
+afterAll(async () => {
+  if (enabled) await enabled.cleanUp();
+  if (disabled) await disabled.cleanUp();
+});
+
+describe('SelfAccount - the feature switch', () => {
+  it.each(REPRESENTATIVE_PATHS)(jira`%s is not served when self-account management is disabled ${'AM-7652'}`, async (path) => {
```
wajb: issue: `AM-7652` is a subtask not a manual test case as expected by the `jira` tagged template.

gravitee-am-test/specs/gateway/self-account/self-account-username.jest.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+const SETTINGS = {
+  selfServiceAccountManagementSettings: {
+    enabled: true,
+    resetPassword: { oldPasswordRequired: false, tokenAge: 0 },
+  },
+};
+
+let fixture: SelfAccountFixture;
+
+const accountUrl = (path: string) => `${process.env.AM_GATEWAY_URL}/${fixture.domain.hrid}/account/api${path}`;
```
wajb: suggestion: this function at least is common to several suites so could be moved to the fixture.

gravitee-am-test/specs/gateway/self-account/fixture/self-account-fixture.ts
sagarchauhan-qa: Removed.

gravitee-am-test/specs/gateway/self-account/self-account-access-control.jest.spec.ts
sagarchauhan-qa: Tagged the username tests with the real manual cases from AM-3686 instead — AM-3936, AM-3937, AM-3938, AM-3939, AM-3940, AM-3941, AM-3942, AM-3946, AM-3947.

Consent, access control and factor QR have no manual cases, so I removed the tag there rather than invent one.

gravitee-am-test/specs/gateway/self-account/self-account-username.jest.spec.ts
sagarchauhan-qa: Done. Moved accountApiUrl, tokenFor and createUserWithToken into the fixture. That also removed the URL builder that was copied into all four files.

wajb (approved):

PR #8562, opened 2026-09-10
test: cover SCIM bulk validation, multi-operation and pre-registration (AM-7653)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-7653

Adds 34 Jest tests across 4 new spec files and 4 JUnit tests. SCIM bulk goes from 71 to 105 tests, `ProvisioningUserServiceTest` from 46 to 50.

Xray holds 65 manual cases for SCIM bulk; all 65 were read and mapped against the existing 13 tests, and about 9 were covered. These close the four areas that had none.

- `scim-bulk-preregistration` (4) — all `preRegistration: true`, all `false`, **mixed**, and each user completing registration from their own email
- `scim-bulk-validation` (10) — empty body, empty `Operations`, auth, duplicate `bulkId`; then per-operation `userName`, email format, `bulkId`, `schemas` and path
- `scim-bulk-multi-operation` (11) — multi-user PUT/PATCH with and without bulkIds, partial failure, idempotency, multi-delete, mixed delete, and a missing `userName` on the update path
- `scim-bulk-fail-on-errors` (9) — DELETE, PUT and PATCH, which were untested; the two existing tests are POST only

**JUnit** adds bulk in *synchronous* mode, bulk-level staging (N users → N staged emails), the retry policy and the dropped-email metric. The two existing async tests set `bulkEnabled=true` but create a single user.

### Integration tests now run with async email

`GRAVITEE_EMAIL_BULK_ENABLED=true` on the gateway, so SCIM pre-registration emails go through the staging processor rather than being sent inline — the more complex path, and the one worth exercising end to end. JUnit keeps the synchronous case.

`GRAVITEE_EMAIL_BULK_PERIOD=1` (default 5s) is the processor's tick, so emails land in 1–2s.

The flag is gateway-wide, so this was checked rather than assumed: it is read in only two places and its only behavioural branch is the SCIM pre-registration path. All 20 email-dependent specs were run with it enabled — **157 passed**, including the non-SCIM `confirm-pre-registration` spec.

### :warning: Two existing files are modified

**`api/commands/utils/email-commands.ts` (+57/-0)** — purely additive. `waitForEmail`, `waitForEmails` and `countEmailsFor` poll the mailbox instead of `getLastEmail`'s fixed 1-second read. Nothing is removed and `getLastEmail` is untouched, so existing callers are unaffected.

**`specs/gateway/scim/fixture/scim-fixture.ts` (+6/-4)** — `setupFixture` gains an **optional** `domainNamePrefix` defaulting to `'scim'`, and `extractConfirmRegistrationLink` polls rather than sleeping. The four existing SCIM specs produce the same domain names and behave identically; the polling change is what makes pre-registration assertions work in async mode.

### Deliberately not duplicated

`checkBulkRequest` runs once per request and is method-agnostic — the bulkId uniqueness check spans all operations regardless of method. The manual cases that repeat "empty request", "empty operations", "invalid token" and "duplicate bulkId" per method exercise the same code, so they are covered once rather than four times. The update path validates separately from create, so that one does have its own test.

## :memo: Test scenarios 
SCIM folder run repeatedly: 105 tests green. JUnit: 50 green. Each new assertion was checked with a negative control — breaking the input makes the test fail.

Two findings while writing them: `enabled` is `false` for a SCIM-created user whether or not it is pre-registered (neither gets a password), so asserting it would pass for the wrong reason — the tests assert `preRegistration` and the email instead. And the retry could only be observed by counting subscriptions, not mock invocations, since `retryWhen` resubscribes to the returned `Completable` rather than calling `push` again.

NA

NA

### DIFF
--- a/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
+++ b/gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
@@ -1418,44 +1418,13 @@ public class ProvisioningUserServiceTest {
         testObserver.assertNoErrors();
         testObserver.assertComplete();
 
-        // One initial attempt plus two retries. The eviction only happens once the retries are spent.
+        // One initial attempt plus two retries. Only once the retries are spent is the email
+        // dropped, which is both audited and counted.
         Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> assertEquals(3, attempts.get()));
-        Awaitility.await()
-                .atMost(10, TimeUnit.SECONDS)
-                .untilAsserted(() -> verify(emailService, times(1)).traceEmailEviction(any(), any(), any()));
-    }
-
-    /** A dropped email is counted, not only audited. */
-    @Test
-    public void shouldCountDroppedEmail_whenStagingKeepsFailing() {
-        ReflectionTestUtils.setField(userService, "bulkEnabled", true);
-        GraviteeUser newUser = mock(GraviteeUser.class);
-        when(newUser.getSource()).thenReturn("unknown-idp");
-        when(newUser.getUserName()).thenReturn("username");
-        when(newUser.getPassword()).thenReturn(null);
-        Map<String, Object> ai = new HashMap<>();
-        ai.put("preRegistration", true);
-        when(newUser.getAdditionalInformation()).thenReturn(ai);
-
-        when(userRepository.findByUsernameAndSource(any(), anyString(), anyString())).thenReturn(Maybe.empty());
-        when(identityProviderManager.getIdentityProvider(anyString())).thenReturn(new IdentityProvider());
-        when(identityProviderManager.getUserProvider(anyString())).thenReturn(Maybe.empty());
-
-        io.gravitee.am.model.User user = new io.gravitee.am.model.User();
-        user.setReferenceType(ReferenceType.DOMAIN);
-        user.setReferenceId(DOMAIN_ID);
-        user.setAdditionalInformation(ai);
-        user.setPreRegistration(true);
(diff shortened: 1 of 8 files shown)


gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
```diff
@@ -1295,6 +1296,168 @@ public void shouldCreateUser_with_preRegistration_bulkMode_trace_staging_failure
...
+        user.setEmail("user@acme.fr");
+        when(userRepository.create(any())).thenReturn(Single.just(user));
+
+        when(emailStagingService.push(any(), any())).thenReturn(Completable.error(new RuntimeException()));
+
+        TestObserver<User> testObserver = userService.create(newUser, null, "/", null, new Client()).test();
+        testObserver.assertNoErrors();
+        testObserver.assertComplete();
+
+        Awaitility.await().atMost(10, TimeUnit.SECONDS).untilAsserted(() -> verify(metricProvider, times(1)).incrementDroppedEmails());
```
leleueri: suggestion: I know it is better to have one test per assertion we want to highlight but I wondering if this assertion should be added in the previous test as finally the email is dropped and the increment should be done. If you move this assertion, we can remove this test.

leleueri (approved):

gravitee-am-gateway/gravitee-am-gateway-handler/gravitee-am-gateway-handler-scim/src/test/java/io/gravitee/am/gateway/handler/scim/service/ProvisioningUserServiceTest.java
sagarchauhan-qa: The dropped-email assertion now sits in the retry test alongside the eviction, since it's the same failing push. Removed the separate test, which drops about 30 lines of duplicated setup.

leleueri (approved):

PR #8564, opened 2026-09-11
feat: add dry-run support to Automation API domain create-or-update (AM-7507)

benoitgravitee:
## Ticket

[AM-7507 — GKO - Reconcile an AMSecurityDomain](https://gravitee.atlassian.net/browse/AM-7507)

Add dry-run support to the Automation API's domain create-or-update endpoint so the GKO admission webhook can validate a domain payload before reconciliation (AC7).

## Changes

Add `?dryRun=true` query parameter to `PUT /domains`. When set, the endpoint runs all server-side validation (data plane resolution, HRID uniqueness, domain settings validation) without persisting, creating events, or provisioning default resources. The endpoint always returns 200 with an `AutomationDomain` — on validation failure the domain's read-only `dryRunErrors` list is populated with errors (severity ERROR/WARNING + message); on success the field is absent.

To support this, `validateCreate()` and `validateUpdate()` are extracted from `DomainServiceImpl`. The existing `create()` and `update()` methods delegate to them — no behavior change on the non-dry-run path.

## Verification

- [ ] Existing automation and service tests pass (zero regression on non-dry-run path)
- [ ] OAS spec renders the new `dryRun` parameter and `dryRunErrors` field on `AutomationDomain`


docs/automation/openapi.yaml
```diff
@@ -489,29 +184,12 @@ paths:
...
-                    oidc:
-                      redirectUriStrictMatching: true
               schema:
-                $ref: "#/components/schemas/AutomationDomain"
-          description: The created or updated domain
+                type: array
+                items:
+                  $ref: "#/components/schemas/DryRunError"
+          description: "The created or updated domain, or when dryRun is true a list\
+            \ of validation errors"
```
leleueri: question: how the TF provider generator will behave if the regular response is not highlighted in the openApi ?

benoitgravitee: I don't know what happen here... that was not my intention. I need to rework that good call.

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/test/java/io/gravitee/am/management/handlers/automation/resource/DomainsResourceTest.java
```diff
@@ -287,4 +292,98 @@ void put_rejects_invalid_key_pattern() {
...
+
+        assertEquals(200, response.getStatus());
+        AutomationDomain body = readEntity(response, AutomationDomain.class);
+        assertEquals(1, body.getDryRunErrors().size());
+        assertEquals(DryRunError.Severity.ERROR, body.getDryRunErrors().get(0).severity());
+        assertEquals("path must start with /", body.getDryRunErrors().get(0).message());
+    }
+
+    @Test
+    void dryRun_does_not_persist() {
```
leleueri, line 330: suggestion: should we create a test explictly for the update as we are testing here the create case. So we are confident that an update do not persist data.

leleueri (approved):

gravitee-am-management-api/gravitee-am-management-api-automation/gravitee-am-management-api-automation-rest/src/test/java/io/gravitee/am/management/handlers/automation/resource/DomainsResourceTest.java
benoitgravitee: Good call — added `dryRun_update_does_not_persist` for the existing-domain path. Same `never().create` / `never().update` asserts as the create case.

PR #8565, opened 2026-09-11
test: application OAuth 2.0 / OIDC settings (AM-5418)

sagarchauhan-qa:
https://gravitee.atlassian.net/browse/AM-5418

Adds 6 Jest tests in a new `specs/gateway/application-oauth-settings/` folder for the application OAuth 2.0 / OIDC settings that had no integration coverage:

- consent duration on a scope (`scopeApproval`) — consent re-requested once it elapses, reused after it is raised
- scope added to / removed from the application — token issued, then `invalid_scope`
- refresh token validity — refresh works within it, `invalid_grant` after it

The rest of the ticket (refresh-token grant, access/ID token expiry) was already covered by existing specs and is not duplicated. No existing file is modified.

## :memo: Test scenarios
Green locally against MongoDB; both time-based tests fail when their wait is removed.

NA

NA

### DIFF
--- a/gravitee-am-test/specs/gateway/application-oauth-settings/app-oauth-settings.jest.spec.ts
+++ b/gravitee-am-test/specs/gateway/application-oauth-settings/app-oauth-settings.jest.spec.ts
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2015 The Gravitee team (http://gravitee.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


gravitee-am-test/specs/gateway/oauth2/oauth2-scopes-and-token-validity.jest.spec.ts
wajb, line 1: suggestion: an `application-oauth-settings` folder and suite don't fit well into the existing gateway tests because it's describing a MAPI/console concept. Could the coverage be folded into the `oauth-flows` or `oauth2` folders to align by feature category instead? We can create new fixtures and suites but existing ones can be enhanced or modified if they're closely related.

sagarchauhan-qa: Moved into oauth2/ and renamed by feature, so the folder's gone. Suites are now OAuth2 - consent duration on a scope, OAuth2 - adding and removing a scope, OAuth2 - refresh token validity.

Kept a separate fixture rather than folding into oauth2-fixture.ts — that one's built around grant types with 4 callers, mine provisions three apps with different consent and validity settings. Happy to merge if you'd rather.

wajb (approved):

PR #8566, opened 2026-09-11
fix: resolve CIMD clients on the social login and logout callbacks

leleueri:
## Problem

A CIMD client whose `client_id` is an HTTPS URL cannot complete an authorization-code flow when the domain authenticates users through an external (social) identity provider. After the user authenticates at the upstream IdP and the IdP returns a code to `/login/callback`, AM aborts with:

```
error=social_authentication_failed&error_description=Bad+client+credentials
```

The same client succeeds against a local IdP, and a pre-registered opaque `client_id` succeeds against the same external IdP.

## Root cause

A CIMD client is not a registered application: it is synthesized from its metadata document and a template application by `CimdAwareClientLookupServiceImpl`. Two lookup paths exist, and three handlers were still on the CIMD-blind one:

| Path | Lookup used before | CIMD? |
|---|---|---|
| `/oauth/authorize` → `AuthorizationRequestParseClientHandler` | `ClientLookupService` | yes |
| root routes (`/login`, `/mfa`, …) → `ClientRequestParseHandler` | `ClientLookupService` | yes |
| **`/login/callback` → `LoginCallbackParseHandler`** | `ClientSyncService` | **no** |
| **`/logout/callback` → `LogoutCallbackEndpoint`** | `ClientSyncService` | **no** |
| **`/logout` → `LogoutEndpoint`** (client restored from session) | `ClientSyncService` | **no** |

`RootProvider` already injected the CIMD-aware bean (`@Qualifier("regularClientLookupService")`, built as `CimdAwareClientLookupServiceImpl` in `CommonConfiguration`) for the other root routes — it simply kept passing `clientSyncService` to these three.

The logout case is worth calling out: after a social login, `SocialAuthenticationProvider` stores `client.getClientId()` on the user record (the local-IdP path stores `client.getId()` instead). So for a CIMD client that field holds the URL, `findById` misses, and the `findByClientId` fallback returned nothing — leaving `LogoutEndpoint` with no client at all, silently applying the **domain** `post_logout_redirect_uris` instead of the client's.

## Fix

Pass `clientLookupService` to `LoginCallbackParseHandler`, `LogoutEndpoint` and `LogoutCallbackEndpoint`.

Domains without CIMD are unaffected: the bean is a plain `DefaultClientLookupServiceImpl` delegating to `clientSyncService` there. Re-resolving on the callback is cheap — `CimdMetadataServiceImpl` caches the document, so the `/authorize` fetch is normally still warm.

## Tests

**Unit** — `LoginCallbackParseHandlerTest`: a CIMD client resolved on callback resumes the chain with the client in context; an unknown client still fails with `BadClientCredentialsException`. The two logout test classes were retargeted to `ClientLookupService`. Full `gravitee-am-gateway-handler-core` suite: 592 tests, 0 failures.

**E2E** — new `specs/gateway/cimd/`, on a shared fixture that stands up an upstream domain (inline IdP + web app redirecting to `{cimdDomain}/login/callback`) and binds an `oauth2-generic-am-idp` on the CIMD template:

- `cimd-social-login.jest.spec.ts` — authorize → external IdP login → callback issues a code to the metadata-declared `redirect_uri`.
- `cimd-logout.jest.spec.ts` — RP-initiated logout **without** a `client_id` parameter, so the client can only come from the session. Note `isMatchingRedirectUri` returns `true` on an empty registered list, so a null client is permissive, not failing — a plain "logout succeeds" assertion would not catch the regression. The fixture therefore registers a different `post_logout_redirect_uri` on the template app and at domain level; the app-level URI must be accepted and the domain-only one rejected, which only holds when the CIMD client was really re-synthesized.

## Out of scope

- URL-shaped `client_id` values land unescaped in login-redirect `Location` headers: `UriBuilderRequest.resolveProxyRequest(request, path, params)` resolves to the `encoded=false` overload. Repo-wide behaviour with a legacy toggle (`legacy.openid.sanitizeParametersEncoding`) behind it — deserves its own ticket and an audit of all callers rather than a flip here.
- Still on `ClientSyncService` and therefore CIMD-blind: `ErrorEndpoint` (error page loses client branding), `UserServiceImpl` (aud-based lookups), `AbstractUserConsentEndpointHandler`.


gravitee-am-test/specs/gateway/cimd/cimd-logout.jest.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+afterAll(async () => {
+  if (fixture) {
+    await fixture.cleanUp();
+  }
+});
+
+/**
+ * AM-7658: after a social login the user record stores the CIMD `client_id` (a URL), so RP-initiated
+ * logout has to re-synthesize the client from its metadata document. The lookup used to ignore CIMD,
+ * leaving the endpoint with no client at all and falling back to the domain settings.
```
wajb, line 44: nitpick: I'm not a fan of describing historical behaviour. Can the test suite stand by itself?

gravitee-am-test/specs/gateway/cimd/cimd-social-login.jest.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+afterAll(async () => {
+  if (fixture) {
+    await fixture.cleanUp();
+  }
+});
+
+describe('CIMD authorize - external (social) identity provider', () => {
+  // AM-7658: /login/callback used to re-resolve the client with a lookup that ignores CIMD,
+  // which aborted the flow with social_authentication_failed / Bad client credentials.
```
wajb, line 42: nitpick: I'm not a fan of describing historical behaviour. Can the test suite stand by itself?

sclarkamlab (approved):

wajb (approved):
Looks good. Though can we confirm that the three remaining direct usages of `ClientSyncService` are safe to be kept CIMD-blind?

lgw-gravitee (approved):

PR #8581, opened 2026-09-16
fix: AM-7668 preserve and mirror trusted domains for backwards compatibility

wajb:
[AM-7668](https://gravitee.atlassian.net/browse/AM-7668)

Preserve data for Token Exchange trusted issuers and SPIFFE workload identity settings during an upgrade to 4.13. This allows continuity of flows when the management API is upgraded before the gateways, and also after a rollback.

Additionally, mirror Trusted Domains that have token exchange usage back to the old domain token exchange settings so that they also have continuity during and after a rollback.

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am 

[AM-7668]: https://gravitee.atlassian.net/browse/AM-7668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
```diff
@@ -74,6 +85,53 @@ public Single<Domain> project(Domain domain) {
...
+     */
+    public Completable syncStored(String domainId) {
+        return Completable.defer(() -> domainRepository.findById(domainId)
+                .filter(stored -> stored.getTokenExchangeSettings() != null)
+                .flatMapCompletable(stored -> {
+                    List<TrustedIssuer> before = stored.getTokenExchangeSettings().getTrustedIssuers();
+                    return mirror(stored).flatMapCompletable(mirrored ->
+                            Objects.equals(before, mirrored.getTokenExchangeSettings().getTrustedIssuers())
+                                    ? Completable.complete()
+                                    : domainRepository.update(mirrored).ignoreElement());
```
sclarkamlab, line 120: A 4.12 gateway never sees this write. It only reloads the domain on a `DOMAIN` event, and `SyncManager` skips the reload unless `updatedAt` moved forward, which this doesn't touch. So in the mixed window a delete through the trusted domains API leaves the 4.12 gateways trusting the issuer until someone next edits the domain.

Think it needs the same thing `DomainServiceImpl.patch` does:

```java
mirrored.setUpdatedAt(new Date());
return domainRepository.update(mirrored)
        .flatMap(saved -> eventService.create(new Event(Type.DOMAIN, new Payload(saved.getId(), DOMAIN, saved.getId(), Action.UPDATE)), saved))
        .ignoreElement();
```

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
```diff
@@ -74,6 +85,53 @@ public Single<Domain> project(Domain domain) {
...
+                }))
+                .onErrorResumeNext(ex -> {
```
sclarkamlab, line 122: This swallows the failure after the trusted domain delete has already committed, so the caller gets a 204 and the stored list still has the issuer. Nothing runs the mirror again until someone edits the domain, and after a rollback that list is the only trust 4.12 reads.

I'd let the error propagate rather than log and complete. Could also move the mirror into `TrustDomainServiceImpl` so every caller gets it and we drop the four resource hooks.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
wajb: Yes, though this is a transitory state so I didn't think it was worth the plumbing. Having a DOMAIN event would also cause domains to be redeployed unnecessarily in 4.13. For the ongoing upgrade, live updates, however unlikely, can still be synced after a restart, which is ordinary practice.

gravitee-am-management-api/gravitee-am-management-api-service/src/main/java/io/gravitee/am/management/service/trustdomain/TrustedIssuerProjection.java
wajb: The error stays logged rather than raised because retries aren't idempotent after the commit. Though I can add a check at startup to sync the data.

Moving the mirroring could incur redundant reads and writes of the domain so I'm inclined to leave it for now.

sclarkamlab (approved):

PR #8583, opened 2026-09-17
fix: AM-7719 allow list and map types to be mapped in reporters

wajb:
[AM-7719](https://gravitee.atlassian.net/browse/AM-7719)

Allow mapping of list and map types in reporter custom attribute mapping.

Previously, `{#context.attributes['client'].metadata['tier']}` was mappable but not `{#context.attributes['client'].metadata}`.

- Instant and LocalDate values used to come out as "" and are now ISO strings.
- Max number of nodes is 100, where larger maps/lists are skipped
  - Max depth of node evaluated remains 10, where deeper nodes become truncated
- Set types are ordered, e,g, `{#context.attributes['user'].roles}`
- Output formatting depends on reporter type:
  - Kafka sends each of these as a JSON string, so the message format doesn't change.
  - File and TCP, in JSON or MessagePack format, write them as nested objects.
  - CSV and Elasticsearch formats still leave custom attributes out.
- Object types remain excluded, e.g. `{#context.attributes['user'].identities}`

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am 

[AM-7719]: https://gravitee.atlassian.net/browse/AM-7719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/attribute/ExportableValue.java
```diff
@@ -16,35 +16,122 @@
...
  *
  * @author GraviteeSource Team
  */
 public final class ExportableValue {
 
+    /** Levels of nesting below the top-level value. */
+    static final int MAX_DEPTH = 10;
+
+    /** Elements across the whole value, at any depth. */
+    static final int MAX_NODES = 100;
```
sclarkamlab, line 42: question: are these ok to be hardcoded, is a config option in gravitee.yml overkill?

lgw-gravitee (approved):

gravitee-am-service/src/main/java/io/gravitee/am/service/reporter/attribute/ExportableValue.java
wajb: I'd say overkill. These are really safeguards against recursion or huge objects that break the kafka message limit. The UI does say "prefer picking single attributes" but we could clarify in documentation how to avoid these limits 👍.

sclarkamlab (approved):

PR #8586, opened 2026-09-18
ci: resolve and publish Maven artifacts through the Azure Artifacts feeds

leleueri:
## Issue

https://gravitee.atlassian.net/browse/AM-7736

## Description

Maven now resolves from the Azure Artifacts feed, and publishes to the Azure feeds alongside Artifactory and Nexus. Same approach as APIM (gravitee-io/gravitee-api-management#19828 and gravitee-io/gravitee-api-management#19940). Only `.circleci/workflows.yml` changes.

**Resolution**
- `setup` reads `custom_field/xml_azure` (the settings shared with APIM) and writes it straight to its file with `ksm secret notation`. `keeper/env-export` cannot be used any more: it expands `${env.AZURE_ARTIFACTS_PAT}` inside an unquoted heredoc, which breaks the export and can print the file in the logs.
- `setup` then resolves `io.gravitee.canary:feed-canary` once through Maven, to check the settings wiring. It now runs on `cimg/openjdk` so Maven is available.
- A new `azure-artifacts-token` command gets an Entra token for the Gravitee.io Bot service principal (the same Keeper record as `purge-azure-registry`) and exports it as `AZURE_ARTIFACTS_PAT`. It then checks that the feed answers: 401/403/404 fail the job, anything else falls back to Artifactory.
- A service principal cannot hold a PAT: its token lives one hour. So each of the 19 jobs running Maven fetches its own token, which keeps "rerun from failed" and jobs waiting on an approval working.

**Publication** (nothing removed)
- Snapshots go to `gravitee-snapshots` from `build`, not from `deploy-snapshot-on-nexus`. Both deploy the same snapshot, and a feed is immutable, so the second would get a 409.
- Releases go to `gravitee` from `release`, only when not in dry-run. The step uses `gio-release`, never `gravitee-release`: the Central extension would take over the deploy phase.
- No `clean`: the `target/` just deployed to Artifactory is reused rather than rebuilt.
- Both `altDeploymentRepository` and `altSnapshotDeploymentRepository` (or `altReleaseDeploymentRepository` for releases) are passed, so the command line wins over any settings profile.
- A fresh token is fetched before each Azure deploy, because the tests before it may outlive the first one.
- Both deploys are fatal, like the Artifactory deploys.

The canary checks and the feed deploys go away when Artifactory is switched off.

## Verification

- `circleci config validate` passes.
- `circleci config process` with `gio_action=release, dry_run=false` expands the release deploy step.
- Not run on CircleCI yet: the pipeline on this PR exercises the token and the resolution. The snapshot deploy runs on the first push to master after merge. The release deploy is only proven by the first release.

**Security-sensitive**: this change touches CI credentials and publication.


.circleci/workflows.yml
```diff
@@ -219,6 +219,64 @@ commands:
...
+  azure-artifacts-token:
+    description: Get a short-lived Azure Artifacts token
+    # Exposes the token in AZURE_ARTIFACTS_PAT, which the Maven settings read as ${env.AZURE_ARTIFACTS_PAT}.
+    steps:
+      - keeper/install
+      - run:
+          name: Get an Azure Artifacts token
+          command: |
+            # Read into shell variables rather than exported: they stay inside this step.
+            CLIENT_ID=$(ksm secret notation keeper://UryantA7MvZe8fkWwcUt8g/field/login)
```
wajb, line 231: issue: looks like we need to bump the keeper orb version `0.6.3` -> `0.7.1` for this notation to work.

wajb (approved):

PR #8594, opened 2026-09-18
fix: remove leftover usage chip from trusted domain edit page

leleueri:
The usage is already selected in the form's Usage section; the chip duplicated it with outdated wording and did not reflect unsaved changes.

fix AM-7715


gravitee-am-test/playwright/tests/security/trusted-domains/crud.spec.ts
```diff
@@ -170,7 +170,7 @@ test.describe('Trusted Domains CRUD', () => {
 
     await detailPage.navigateToEdit(testDomain.id, trustDomainId);
     await expect(detailPage.issuerUrlInput).toHaveValue('https://linkable.example.com');
-    await expect(detailPage.usageBadge).toHaveText(/OIDC - Trusted Issuer/i);
```
wajb, line 173: suggestion: can we update the matching text used in `trust-domain.types.ts` too, since this old terminology is still used in the list page?

`OIDC - Trusted Issuer` -> `Token Exchange`.

wajb (approved):

PR #8604, opened 2026-09-21
fix: edit XAA resource servers in place in the Console

leleueri:
## Description

The resource-server table of a trusted domain only offered a remove action, so renaming a server or changing its resource URI meant deleting it and adding it again. The new row got a new `id`, which orphaned the Cross App Access mappings of every application referencing the old one.

Name and Resource URI cells are now editable inline, following the interaction already used on the IDP mappers and token claims pages: double-click a cell, then press Enter. The edit keeps the resource server's `id`, so the save is an update and application mappings still resolve.

- The edit only changes the row in the form; nothing is stored until the form's Save, as for adding or removing a row.
- Double-click does nothing outside edit mode, and an empty value is ignored.
- A URI duplicating another row still blocks the save through the existing validation.
- Removing a row closes any cell left open, since cells are keyed by row index.

No backend change: `copyWithGeneratedIds` in `TrustDomainServiceImpl` already keeps an incoming id that matches a stored resource server.

## Tests

Six specs drive the real table cells: name edit keeps the id, URI edit keeps the id, empty value ignored, duplicate URI rejected, nothing opens outside edit mode, removing a row closes open cells. The spec now loads the real `NgxDatatableModule` so cell templates render; the existing tests in that file still pass.

`yarn ng test --test-path-pattern='trust-domains'` — 98 passed. Prettier, eslint and stylelint clean on the changed files.

Not verified in a running Console, so the look and feel of the inline edit is untested.

## Follow-up

Deleting a resource server still leaves applications pointing at one that no longer exists. Out of scope here; could warn or block on delete while an application still maps it.

fix AM-7713


lgw-gravitee:
imho its a bad UX
1. a border is partially hidden by a layer above
image


2. the footer description can be easily omitted by an eye makes the functionality secret :) I think it would be better to add pencil icon

wdyt?

lgw-gravitee:
image

looks like its obvious or irrelevant :P

lgw-gravitee:
I can still click on Name/URI and enter edit mode. If I click and ammend both and save only one is updated

lgw-gravitee:
nitpick: 
image
icons are not vertically centerlized

lgw-gravitee (approved):

PR #8607, opened 2026-09-21
fix(xaa): expose the resolved ID-JAG target to TOKEN policies and claims

lgw-gravitee:
fixes AM-7714

Groovy policies and claims can read the resolved ID-JAG target

When a client omits resource on an ID-JAG token exchange, AM infers it. That inferred value was never visible to TOKEN flow policies or to ID-JAG custom claims, because request.params['resource'] only holds what the client actually sent.

ID-JAG requests now get a read-only idJag context attribute: {audience, resource, clientId}. It has the same value whether resource was sent or inferred.
- It's available in PRE_TOKEN and POST_TOKEN policies. RulesEngine gained a fire(...) overload that adds extra attributes to the context before any policy runs.
- It's available in custom claims and aud_sub expressions. It's set after the request's own attributes, so a PRE_TOKEN policy can't overwrite it.
- Groovy reads it as `context.attributes['idJag']['resource']`, EL as `{#context.attributes['idJag']['resource']}`
- Nothing is added to the request parameters, and other grants are unaffected.

How to test:

1. Set up a trusted domain with Cross App Access, one resource server (https://calendar.acme.com) and a scope mapping. Map an application to that one resource server.
2. On the application, add a TOKEN flow Groovy policy (pre):
     `def idJag = context.attributes['idJag']; if (idJag) { context.setAttribute('groovyResource', idJag['resource']) }`
3. Add two ID-JAG custom claims:
     - groovy_resource = {#context.attributes['groovyResource']}
     - resolved_resource = {#context.attributes['idJag']['resource']}
4. Get an ID token with the password grant. Exchange it with requested_token_type=urn:ietf:params:oauth:token-type:id-jag and audience=…, and leave out  both resource and scope.
5. Decode the ID-JAG: groovy_resource and resolved_resource should both be https://calendar.acme.com.
6. Repeat with resource=https://calendar.acme.com and expect the same values.
7. Run a normal password grant with the same Groovy policy. It should still succeed, because idJag is absent and the script skips it.


leleueri (approved):

gravitee-am-common/src/main/java/io/gravitee/am/common/utils/ConstantKeys.java
```diff
@@ -38,6 +38,7 @@ public interface ConstantKeys {
     String ALLOW_EMPTY_SCOPE_SELECTION = "allowEmptyScopeSelection";
     String AUTHORIZATION_REQUEST_CONTEXT_KEY = "authorization_request";
     String ID_TOKEN_CONTEXT_KEY = "idToken";
+    String ID_JAG_CONTEXT_KEY = "idJag";
```
leleueri, line 41: not sure about the name. we are naming it idJag when it is about the triple client, aud, resource (IdJagTarget) not the IdJag token... I don't know if there is a  better name but I flag it.

leleueri: as discussed internally, we will keep idJag.

PR #8609, opened 2026-09-21
fix(xaa): add ID-JAG lax mode to token exchange settings

lgw-gravitee:
fixes AM-7768

image

image

if enabled, you can use access_token as subject_token


gravitee-am-ui/src/app/domain/settings/oauth/token-exchange/token-exchange-settings/token-exchange-settings.component.html
```diff
@@ -88,6 +88,27 @@ <h5>Requested Token Configuration</h5>
             </mat-form-field>
           </div>
 
+          <!-- ID-JAG Settings Section -->
```
wajb: suggestion: would this be more intuitive placed in the Subject Token Configuration section?

lgw-gravitee: i think its more related o "Requested Token Configuration"
the box disappears when ID_JAG is not selected. 
I can put it inside the "Requested Token Configuration"
wdyt?

lgw-gravitee: I fixed as you suggested

gravitee-am-model/src/main/java/io/gravitee/am/model/TokenExchangeSettings.java
```diff
@@ -89,6 +89,11 @@ public class TokenExchangeSettings {
             "1–100.", defaultValue = "25")
     private int maxDelegationDepth = DEFAULT_MAX_DELEGATION_DEPTH;
 
+    @Schema(description = "Lax mode for ID-JAG issuance: also accept an access token issued to the requesting " +
+            "client as the subject token. By default only an ID token is accepted.",
+            defaultValue = "false")
+    private boolean idJagLaxMode = false;
```
leleueri: discussion: should we anticipate extra settings related to IDJag in that case a IDJagSettings object maybe better with a laxMode inside.

lgw-gravitee: good idea. fixed that too

wajb (approved):

PR #8610, opened 2026-09-21
fix: AM-7690 speed up PostgreSQL user search with pattern-ops indexes

wajb:
[AM-7690](https://gravitee.atlassian.net/browse/AM-7690)

`LIKE 'term%'` cannot use a plain btree under a non-C collation, so user search sequential-scanned the users table.

- Add pattern-ops indexes on the searched columns.
- Quote `additional_information->>'email'` so the JSON expression index can match.
- Use `COUNT(*)` instead of `COUNT(DISTINCT id)`.
- Zip the page and count queries so they run in parallel.

Indicative results with 430k users:

| | Before | After |
|--------|--------|--------|
| Search, zero-match term | 121.8 ms, 96,899 buffers, 430,000 rows filtered | 0.292 ms, 21 buffers |
| Count | 30.8 ms, 11,255 buffers | 0.013 ms, 18 buffers | 

## :memo: Test scenarios 
See jira.

n/a

n/a

@gravitee-io/am 

[AM-7690]: https://gravitee.atlassian.net/browse/AM-7690?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_9_0/4.9.33-dataplane-users-search-pattern-ops-indexes.yml
leleueri, line 1: question: should we remove the indexes created by the previous liquibase scripts as they are targeting the same fields ? Does it make sens to keep both 🤔 ? also is there any improvement on the other DB implementation.

gravitee-am-repository/gravitee-am-repository-jdbc/src/main/java/io/gravitee/am/repository/jdbc/management/api/JdbcOrganizationUserRepository.java
leleueri, line 1: question: I do not see liquibase script for organization _users table, should we create the same kind of indexes as for the domain users ?

gravitee-am-dataplane/gravitee-am-dataplane-jdbc/src/main/resources/liquibase/changelogs/v4_9_0/4.9.33-dataplane-users-search-pattern-ops-indexes.yml
wajb: I was looking at that...it turns out SCIM range comparisons and ordering by username still need the old indexes, so we have to keep them.

Regarding improvements on other DB types, only dropping `DISTINCT` and counting in parallel have impact.

Measurements for the first:
| Engine | Before `COUNT(DISTINCT id)` | After `COUNT(*)` | Improvement |
|---|---|---|---|
| **PostgreSQL 18.4** | 93.7 ms | **25.4 ms** | **3.7×** |
| **MySQL 8.4** | ~1365 ms | ~984 ms | ~28% |
| **MariaDB 11.6.2** | ~745 ms | ~658 ms | ~12% (noisy) |
| **MSSQL 2022** | 150 ms elapsed, **1153 ms CPU** | 98 ms elapsed, **275 ms CPU** | **4.2× less CPU** |

MySQL is slow because it refuses the `index_merge` and reads the whole domain - but this is unrelated.

leleueri (approved):

PR #8616, opened 2026-09-22
feat: xaa jest tests

lgw-gravitee:
still requries published XAA plugin


gravitee-am-test/specs/gateway/cross-app-access/id-jag-redemption.jest.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+  });
+
+  describe('the audit log answers who redeemed what', () => {
+    it('should audit a redemption in the token stream with the resource and the granted scope', async () => {
+      const assertion = await fixture.agent.assertion();
+
+      await fixture.agent.redeem(assertion, scopeParam(RESOURCE_SCOPE.write)).expect(200);
+
+      const audit = await fixture.awaitTokenAudit('SUCCESS', (detail) => JSON.stringify(detail).includes(fixture.mcpResource));
```
sclarkamlab, line 163: These can pass on an audit from an earlier test. Every redemption in this file has `mcpResource` and `mcp.write` in it, and every refusal has the jwt-bearer grant type, so `awaitTokenAudit` can hand back an older audit before this one lands.

The issuance spec matches on the jti instead, and the success audit records the access token jti so this would pin it:

```ts
const tokenId = decodeToken(response.body.access_token).jti;
const audit = await fixture.awaitTokenAudit('SUCCESS', (detail) => JSON.stringify(detail).includes(tokenId));
```

The failure audit has no id to match on, but `/audits` takes `from` (epoch millis) so a `Date.now()` taken before the request would do it. The `outcome.status` check on 183 can't fail either, the list is already filtered on `status=FAILURE`.

gravitee-am-test/specs/gateway/cross-app-access/id-jag-redemption-binding.jest.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+      expect(errorOf(response)).toBe('invalid_grant');
+    });
+
+    it('should create the local user on first redemption when the grant creates users', async () => {
+      const first = await fixture.creator.redeem(await fixture.creator.assertion({ user: 'stranger' })).expect(200);
+      const second = await fixture.creator.redeem(await fixture.creator.assertion({ user: 'stranger' })).expect(200);
+
+      expect(subjectOf(first)).toBe(subjectOf(second));
+    });
```
sclarkamlab, line 62: These two depend on order. The create test makes a local user for the stranger, and the refusal test above needs that user not to exist. CI runs with `JEST_RETRIES=2` and retries after the whole file, so if the refusal test flakes its retry gets a 200 and fails for good. GUIDELINES.md wants `retryImmediatelyForThisFile()` for ordered tests.

Also `first` and `second` would share a sub even if no user got created and the token just carried the assertion sub. Worth adding the `not.toBe(decodeToken(assertion).sub)` check the main spec has.

wajb: Ideally we'd make the tests run independently and not introduce the `retryImmediatelyForThisFile` exemption.

gravitee-am-test/specs/gateway/cross-app-access/id-jag-redemption.jest.spec.ts
lgw-gravitee: success audits matched by token jti, failure scoped with from, redundant status check dropped.

PR #8617, opened 2026-09-22
fix: isolate scheduled purge targets and keep valid purge exclusions

sclarkamlab:

https://gravitee.atlassian.net/browse/AM-7775
https://gravitee.atlassian.net/browse/AM-7776

Two fixes to the scheduled purge. Each one is its own commit with its own ticket ref.

- AM-7513: each purge target runs on its own now. When one target fails, the service logs the error and moves on to the next target. The targets still run one at a time. A sweeper that throws before it returns a Completable gets the same treatment. The "ended" debug line prints when the run really finishes.
- AM-7514: the service trims each name in `services.purge.exclude` and skips an unknown name with a warning. The valid exclusions still apply. Before, one bad name or a space after a comma made the service purge everything.

On JDBC the gateway purge is the only cleanup, so one failing table used to stop every table after it.

## :memo: Test scenarios 

Unit coverage in the branch, all green.

No UI changes.

For whoever fixes the Mergify PRs: from 4.11.x the test needs two lines changed, because the sweeper source type is different there.

```java
import io.gravitee.am.repository.common.ExpiredDataSweeperProvider;
private ExpiredDataSweeperProvider sweepers; // was ExpiredDataSweepers
```

I cherry-picked both commits onto 4.11.x and master with that edit. Both apply clean and the tests pass. Master needs JDK 25 to build.

@gravitee-io/am


gravitee-am-service/src/test/java/io/gravitee/am/service/purge/ScheduledPurgeServiceTargetsTest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Checks which targets a scheduled purge run purges: the exclusions it honours, and that
```
wajb: nitpick: use US spelling.

sclarkamlab:
moved to https://github.com/gravitee-io/gravitee-access-management/pull/8625 after the tickets changed to AM-7775 and AM-7776

